PR #14, opened 2021-05-21
Added testing help functions

ptesavol:
Added testing help functions runAndWaitForEvents() and runAndWaitForConditions() to make it easier to create tests that work consistently in different kinds of network environments (real network, simulator).

### DIFF
--- a/packages/test-utils/src/utils.ts
+++ b/packages/test-utils/src/utils.ts
@@ -57,7 +57,7 @@ export const waitForEvent = (emitter: EventEmitter, event: Event, timeout = 5000
  * @returns {Promise<unknown[]>} resolves with event arguments if event occurred
  * within timeout. Otherwise rejected.
  */
-export const runAndWaitForEvents = (
+export const runAndWaitForEvents = async (
     operations: () => void | (() => void)[], 
     waitedEvents: [emitter: EventEmitter, event: Event] | Array<[emitter: EventEmitter, event: Event]>, 
     timeout = 5000
--- a/packages/test-utils/src/utils.ts
+++ b/packages/test-utils/src/utils.ts
@@ -161,7 +161,7 @@ export const waitForCondition = async (
  * conditions evaluate to true on a retry attempt within timeout. If timeout
  * is reached with conditionFn never evaluating to true, rejects.
  */
-export const runAndWaitForConditions = (
+export const runAndWaitForConditions = async (
     operations: () => void | (() => void)[], 
     conditions: (() => (boolean | Promise<boolean>)) | (() => (boolean | Promise<boolean>)) [],
     timeout = 5000,
--- a/packages/test-utils/src/utils.ts
+++ b/packages/test-utils/src/utils.ts
@@ -57,8 +57,12 @@ export const waitForEvent = (emitter: EventEmitter, event: Event, timeout = 5000
  * @returns {Promise<unknown[]>} resolves with event arguments if event occurred
  * within timeout. Otherwise rejected.
  */
-export const runAndWaitForEvents = (operations: Function | Function[], waitedEvents: [emitter: EventEmitter, event: Event] | Array<[emitter: EventEmitter, event: Event]>, timeout = 5000): Promise<unknown[]> => {
-    let ops: Function[]
+export const runAndWaitForEvents = (
+    operations: () => void | (() => void)[], 
+    waitedEvents: [emitter: EventEmitter, event: Event] | Array<[emitter: EventEmitter, event: Event]>, 
+    timeout = 5000
+): Promise<unknown[]> => {
+    let ops: (() => void)[]
     if (Array.isArray(operations)) {
         ops = operations
     } 
@@ -75,16 +79,16 @@ export const runAndWaitForEvents = (operations: Function | Function[], waitedEve
     }
     
     const promise: Promise<unknown[]> = new Promise( (resolve, reject) => { 
-        Promise.all(evs.map(e => { waitForEvent(e[0], e[1], timeout) }))
-        .then((result) => {
-            resolve(result)
-        })
-        .catch( (e) => {
-            reject(e)
-        })
+        Promise.all(evs.map( (e) => { waitForEvent(e[0], e[1], timeout) }))
+            .then((result) => {
+                return resolve(result)
+            })
+            .catch( (e) => {
+                reject(e)
+            })
     })
 
-    ops.forEach(op=> { op() })
+    ops.forEach( (op) => { op() })
     
     return promise
 }
@@ -158,19 +162,19 @@ export const waitForCondition = async (
  * is reached with conditionFn never evaluating to true, rejects.
  */
 export const runAndWaitForConditions = (
-    operations: Function | Function[], 
+    operations: () => void | (() => void)[], 
     conditions: (() => (boolean | Promise<boolean>)) | (() => (boolean | Promise<boolean>)) [],
     timeout = 5000,
     retryInterval = 100,
     onTimeoutContext?: () => string
-    ): Promise<unknown[]> => {
+): Promise<unknown[]> => {
 
-    let ops: Function[]
+    let ops: (() => void)[]
     if (Array.isArray(operations)) {
         ops = operations
     } 
     else {
-        ops = [operations];
+        ops = [operations]
     }
     
     let conds: (() => (boolean | Promise<boolean>)) [] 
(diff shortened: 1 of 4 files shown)


packages/test-utils/src/utils.ts
```diff
@@ -47,6 +47,52 @@ export const waitForEvent = (emitter: EventEmitter, event: Event, timeout = 5000
...
+    let evs: Array<[emitter: EventEmitter, event: Event]>
+    if (Array.isArray(waitedEvents) && Array.isArray(waitedEvents[0]) ) {
+        evs = waitedEvents as Array<[emitter: EventEmitter, event: Event]>
+    } 
+    else {
+        evs = [waitedEvents as [emitter: EventEmitter, event: Event]]
+    }
+    
+    const promise: Promise<unknown[]> = new Promise( (resolve, reject) => { 
+        Promise.all(evs.map( (e) => { waitForEvent(e[0], e[1], timeout) }))
```
timoxley, line 82: Ah there's a critical bug here. The promise returned by `waitForEvent` isn't being returned due to the `{` in `(e) => { waitForEvent(…) }`, so this is essentially just creating an array of undefined and a bunch of unchained async. The `Promise.all` won't actually wait for anything here.

Can fix it like so, also explicitly named the arguments:

```suggestion
`const promise = Promise.all(evs.map(([emitter, event]) => waitForEvent(emitter, event, timeout)))`
```

packages/test-utils/src/utils.ts
```diff
@@ -47,6 +47,52 @@ export const waitForEvent = (emitter: EventEmitter, event: Event, timeout = 5000
...
+
+    let evs: Array<[emitter: EventEmitter, event: Event]>
+    if (Array.isArray(waitedEvents) && Array.isArray(waitedEvents[0]) ) {
+        evs = waitedEvents as Array<[emitter: EventEmitter, event: Event]>
+    } 
+    else {
+        evs = [waitedEvents as [emitter: EventEmitter, event: Event]]
+    }
+    
+    const promise: Promise<unknown[]> = new Promise( (resolve, reject) => {
```
timoxley, line 81: I don't think this `new Promise( (resolve, reject) => {` wrapper is necessary, you can just `const promise = Promise.all(…)` and remove the `.then` and `.catch` entirely

packages/test-utils/src/utils.ts
```diff
@@ -101,6 +147,59 @@ export const waitForCondition = async (
...
+ * @param conditions condition(s) to be evaluated; condition functions should return boolean or Promise<boolean> and have
+ * no side-effects.
+ * @param timeout amount of time in milliseconds to wait for
+ * @param retryInterval how often, in milliseconds, to re-evaluate condition
+ * @param onTimeoutContext evaluated only on timeout. Used to associate human-friendly textual context to error.
+ * @returns {Promise<unknown[]>} resolves immediately if
+ * conditions evaluate to true on a retry attempt within timeout. If timeout
+ * is reached with conditionFn never evaluating to true, rejects.
+ */
+export const runAndWaitForConditions = (
```
timoxley, line 164: Should be `async` for same reasons above.
```suggestion
export const runAndWaitForConditions = async (
```

packages/test-utils/src/utils.ts
```diff
@@ -101,6 +147,59 @@ export const waitForCondition = async (
...
+    let conds: (() => (boolean | Promise<boolean>)) [] 
+    if (Array.isArray(conditions)) {
+        conds = conditions
+    } 
+    else {
+        conds = [conditions]
+    }
+    
+    const promise: Promise<unknown[]> = new Promise( (resolve, reject) => { 
+        Promise.all(conds.map( (c) => { waitForCondition(c, timeout, retryInterval, onTimeoutContext) }))
```
timoxley, line 189: Same problems here: `new Promise` not needed, no promises are returned from `conds.map(…)`

timoxley (changes requested):
Neat but there's some bugs hidden in the usage of `Promise.all` here.

packages/test-utils/src/utils.ts
ptesavol: Thank you for the review! I made the suggested corrections and pushed again.

timoxley (approved):
ok issues addressed, nice one 👍 

Looking at this now though… I do feel like this utility shouldn't need to exist? Either:
1.  wrap the common case event waiting into the actual api interface (maybe as a new method), or 
2. let's get used to always registering listeners first, *then* doing the action that may trigger the events 

For example, using `Promise.all` instead of `runAndWaitForEvents` the code is almost identical but without any indirection, just using built-ins and existing utilities:

```js
// with wrapper
await runAndWaitForEvents([ () => { subscriberOne.start() }, () => { subscriberTwo.start() }],[
    [subscriberOne, NodeEvent.NODE_SUBSCRIBED],
    [subscriberTwo, NodeEvent.NODE_SUBSCRIBED],
    // @ts-expect-error private field
    [tracker.trackerServer, TrackerServerEvent.NODE_STATUS_RECEIVED]
])

// with no wrapper
await Promise.all([
    waitForEvent(subscriberOne, NodeEvent.NODE_SUBSCRIBED),
    waitForEvent(subscriberTwo, NodeEvent.NODE_SUBSCRIBED),
    // @ts-expect-error private field
    waitForEvent(tracker.trackerServer, TrackerServerEvent.NODE_STATUS_RECEIVED),
    subscriberOne.start(),
    subscriberTwo.start(),
])
```
or

```js
// with wrapper
await runAndWaitForEvents(() => { subscriberOne.unsubscribe('stream-2', 2) },[
    [subscriberTwo, NodeEvent.NODE_UNSUBSCRIBED],
    // @ts-expect-error private field
    [tracker.trackerServer, TrackerServerEvent.NODE_STATUS_RECEIVED]
])

// with no wrapper
await Promise.all([
    waitForEvent(subscriberTwo, NodeEvent.NODE_UNSUBSCRIBED),
    // @ts-expect-error private field
    waitForEvent(tracker.trackerServer, TrackerServerEvent.NODE_STATUS_RECEIVED),
    subscriberOne.unsubscribe('stream-2', 2),
])
```

This wrapper *does* enforce the pattern of event listeners getting registered first, which is definitely useful, but also note the wrapper does not support async "run" steps. 

IMO developing a habit of "always attach listeners first" is what we should be doing anyway, rather than relying on a utility to do it for us. i.e. if you're attaching the listeners afterward, the code is kinda just… wrong.

j9euu 0

That being said, I'm not strongly against this, just IMO better to use built-ins and less indirection where possible, rather than making our own dsl, especially if our dsl is pretty much on-par with built-ins with regard to reducing clutter and visual complexity, while also being a little less flexible.

ptesavol:
I agree that the syntax of calling the wrapper is a bit cumbersome and DSL-like. However, what I am afraid about the Promise.all() approach without the wrapper is that it does not force one to do the things in the right order.  It is possible for a new developer who is unaware of this issue to fall for the same mistake in the future (also for us, if we forget).

IMO we should even consider making waitForEvent() private in order to prevent it be used in the wrong way.

harbu:
> ok issues addressed, nice one +1
> 
> Looking at this now though… I do feel like this utility shouldn't need to exist? Either:
> 
>     1. wrap the common case event waiting into the actual api interface (maybe as a new method), or
> 
>     2. let's get used to always registering listeners first, _then_ doing the action that may trigger the events
> 
> 
> For example, using `Promise.all` instead of `runAndWaitForEvents` the code is almost identical but without any indirection, just using built-ins and existing utilities:
> 
> ```js
> // with wrapper
> await runAndWaitForEvents([ () => { subscriberOne.start() }, () => { subscriberTwo.start() }],[
>     [subscriberOne, NodeEvent.NODE_SUBSCRIBED],
>     [subscriberTwo, NodeEvent.NODE_SUBSCRIBED],
>     // @ts-expect-error private field
>     [tracker.trackerServer, TrackerServerEvent.NODE_STATUS_RECEIVED]
> ])
> 
> // with no wrapper
> await Promise.all([
>     waitForEvent(subscriberOne, NodeEvent.NODE_SUBSCRIBED),
>     waitForEvent(subscriberTwo, NodeEvent.NODE_SUBSCRIBED),
>     // @ts-expect-error private field
>     waitForEvent(tracker.trackerServer, TrackerServerEvent.NODE_STATUS_RECEIVED),
>     subscriberOne.start(),
>     subscriberTwo.start(),
> ])
> ```
> 
> or
> 
> ```js
> // with wrapper
> await runAndWaitForEvents(() => { subscriberOne.unsubscribe('stream-2', 2) },[
>     [subscriberTwo, NodeEvent.NODE_UNSUBSCRIBED],
>     // @ts-expect-error private field
>     [tracker.trackerServer, TrackerServerEvent.NODE_STATUS_RECEIVED]
> ])
> 
> // with no wrapper
> await Promise.all([
>     waitForEvent(subscriberTwo, NodeEvent.NODE_UNSUBSCRIBED),
>     // @ts-expect-error private field
>     waitForEvent(tracker.trackerServer, TrackerServerEvent.NODE_STATUS_RECEIVED),
>     subscriberOne.unsubscribe('stream-2', 2),
> ])
> ```
> 
> This wrapper _does_ enforce the pattern of event listeners getting registered first, which is definitely useful, but also note the wrapper does not support async "run" steps.
> 
> IMO developing a habit of "always attach listeners first" is what we should be doing anyway, rather than relying on a utility to do it for us. i.e. if you're attaching the listeners afterward, the code is kinda just… wrong.
> 
> j9euu 0
> 
> That being said, I'm not strongly against this, just IMO better to use built-ins and less indirection where possible, rather than making our own dsl, especially if our dsl is pretty much on-par with built-ins with regard to reducing clutter and visual complexity, while also being a little less flexible.

I'd say both ways are okay. I'd like to see some unit tests for these new methods.

ptesavol:
The changes have been integrated into https://github.com/streamr-dev/network-monorepo/pull/27
Closing this pull request as a duplicate

PR #16, opened 2021-05-21
NET-264: remove node-level buffering

harbu:
*Notice: Copied from network repository, original PR: https://github.com/streamr-dev/network/pull/547*

Remove buffering from `Node` (logic layer). 

The following changes to preexisting public API methods have been made.
- Method `publish(msg: StreamMessage): number` is now the designated "fire-and-forget" publish mechanism. It returns the number of neighbors the message was forwarded (attempted) to, which could server as a useful signal in certain cases.

The following new methods have been added to the public API.
- Method `asyncPublish(msg: StreamMessage, minNeighbors = 1, ...): Promise<number>` is a more "reliable" publish mechanism. It will wait for at least neighbors to be present before attempting to propagate. It too, signals the number of neighbors the message was forwaded (attempted) to. It will reject, if the minimum number of neighbors were not available within a given timeout.
- Method `getNumberOfNeighbors(streamId: string, streamPartition: number): number` returns the number of available neighbors for a stream at the present time.
- Method `waitForNeighbors(streamId: string, streamPartition: number, minNeighbors = 1, ...): Promise<number>` similar thinking to `asyncPublish` but we do not publish but just wait.

### Caveats
- Buffering will most likely need to be implemented on broker in order to buffer incoming messages from clients.
- Parameter `minNeighbors` is quite arbitrary and there is not a "correct" answer to what it should be currently (There are valid cases where there are no neighbors, however, setting it as zero means neighbors are not awaited for :shrug:.) The ideal would be not to have to supply such a parameter, however, that would require some significant changes on how the tracker sends instructions which I am hesitant to do right now.


packages/network/src/NetworkNode.ts
```diff
@@ -101,4 +135,52 @@ export class NetworkNode extends Node {
...
+
+    /**
+     * Wait until at least the given number of neighbors are available on a stream.
+     * @param streamId
+     * @param streamPartition
+     * @param minNeighbors minimum number of neighbors to a wait for
+     * @param timeoutInMs timeout (in milliseconds) after which to give up waiting
+     * @returns nubmer of neighbors available
+     */
+    public waitForNeighbors(
```
timoxley, line 156:
```suggestion
    public async waitForNeighbors(
```
Just for good measure. Everything is inside a new Promise so doesn't change anything in this case.

packages/network/src/logic/Node.ts
```diff
@@ -430,8 +404,12 @@ export class Node extends EventEmitter {
             }
         })
 
-        this.seenButNotPropagatedSet.delete(streamMessage)
+        if (subscribers.length === 0) {
+            this.logger.warn('no neighbors to propagate message %o to', streamMessage.getMessageID().toArray())
+        }
+
         this.emit(Event.MESSAGE_PROPAGATED, streamMessage)
```
timoxley, line 411: Previously this wouldn't fire for `subscribers.length === 0`, is it correct to still fire `Event.MESSAGE_PROPAGATED` if there were "no neighbors to propagate message to"?

packages/network/test/integration/message-duplication.test.ts
```diff
@@ -71,27 +66,22 @@ describe('duplicate message detection and avoidance', () => {
...
-        }))
         // Become subscribers (one-by-one, for well connected graph)
         otherNodes[0].subscribe('stream-id', 0)
         otherNodes[1].subscribe('stream-id', 0)
         otherNodes[2].subscribe('stream-id', 0)
         otherNodes[3].subscribe('stream-id', 0)
         otherNodes[4].subscribe('stream-id', 0)
 
-        await allNodesSubscribed
+        await Promise.all([
```
timoxley, line 78: Maybe `Promise.all(otherNodes.map((otherNode) => otherNode.waitForNeighbors('stream-id', 0, 2, 15 * 1000)))`?

packages/network/test/integration/message-propagation.test.ts
```diff
@@ -96,24 +95,26 @@ describe('message propagation in network', () => {
             payload: streamMessage.getParsedContent()
         }))
 
+        n1.subscribe('stream-1', 0)
```
timoxley, line 98: I wonder whether `subscribe` should return a `waitForNeighbors` promise, maybe as `asyncSubscribe`

packages/network/test/integration/storage-config.test.ts
```diff
@@ -93,7 +84,7 @@ describe('storage node', () => {
         const stream = createMockStream()
         config.addStream(stream)
         const message = createMockMessage(stream)
-        relayNode.publish(message)
+        await relayNode.asyncPublish(message)
```
timoxley, line 87: neat

packages/network/src/NetworkNode.ts
harbu: committed.

packages/network/src/logic/Node.ts
harbu: Looks like this event is currently not used anywhere. That being said, your suggestion makes sense. I'll fire the event only if there were subscribers. (Could also consider removing the event but maybe leave that for a separate time.)

packages/network/test/integration/message-duplication.test.ts
harbu: I think the repetition is low enough here for this to be a rather negligible change.

packages/network/test/integration/message-propagation.test.ts
harbu: I don't see this necessarily being a very common use case compared to e.g. publishing? Usually you'd just want to subscribe to a stream and then assign a handler for "when the messages start eventually appearing". For those situations where you want to wait, you can use `waitForNeighbors`?

harbu:
Closing this ticket. The cache was deemed necessary and thus this PR https://github.com/streamr-dev/network-monorepo/pull/211 replaced this one.

PR #17, opened 2021-05-21
NET-312: Storage plugin

teogeb:
Moved storage node logic to new `StoragePlugin`. Now there is no need to pass cassandra storage or storage configs to the network library. Divided the remaining functionality of `HttpPlugin` to `PublishHttpPlugin` and `MetricsPlugin`.

Also small improvements for resource cleanup in `Broker` and `StoragePlugin`.


packages/broker/src/storage/StoragePlugin.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    private async getCassandraStorage() {
+        const cassandraStorage = await startCassandraStorage({
+            contactPoints: [...this.pluginConfig.cassandra.hosts],
+            localDataCenter: this.pluginConfig.cassandra.datacenter,
+            keyspace: this.pluginConfig.cassandra.keyspace,
+            username: this.pluginConfig.cassandra.username,
+            password: this.pluginConfig.cassandra.password,
+            opts: {
+                useTtl: !this.brokerConfig.network.isStorageNode
```
harbu, line 67: Feel free to set this just as `false` for now to get rid of reliance on `this.brokerConfig.network.isStorageNode` (which should be removed)

harbu: Context: before we had the possibility to enable "storage" on non-storage nodes. This means that the node would **not** be treated as a storage node from the network's viewpoint, but it would still be configured with a Cassandra to be used as a short-term persistence solution. This Cassandra could be used as a caching solution of sorts, responding to resend requests that involved recent data. Hence the need for TTL.

harbu (approved):
Looking really good, nice :+1:

packages/broker/src/storage/StoragePlugin.ts
```diff
@@ -0,0 +1,91 @@
...
+export class StoragePlugin extends Plugin<StoragePluginConfig> {
+
+    private cassandra: Storage|undefined
+    private storageConfig: StorageConfig|undefined
+
+    constructor(options: PluginOptions) {
+        super(options)
+    }
+
+    async start() {
```
timoxley, line 34: I feel like this should have some sort of re-entry protection to prevent running `start` twice, or while either `start` or `stop` are still pending, etc. Might not be possible to do this currently, but should this ever happen it will produce some weird behaviour and might not be easy to spot what the problem is.

packages/broker/src/storage/StoragePlugin.ts
```diff
@@ -0,0 +1,91 @@
...
+    private async createStorageConfig() {
+        const brokerAddress = new Wallet(this.brokerConfig.ethereumPrivateKey).address
+        const storageConfig = await StorageConfig.createInstance(brokerAddress, this.brokerConfig.streamrUrl + '/api/v1', this.pluginConfig.storageConfig.refreshInterval)
+        storageConfig.startAssignmentEventListener(this.brokerConfig.streamrAddress, this.networkNode)
+        return storageConfig
+    }
+
+    async stop() {
+        return Promise.all([
+            this.cassandra!.close(),
```
timoxley, line 83: This should probably be `this.cassandra?.close()` otherwise this will crash with a `ReferenceError` due to `this.cassandra` being `undefined`:

```js
const s = new StoragePlugin(opts)
await s.stop()
```

or if `stop` is called before `start` finishes.

packages/broker/src/storage/StoragePlugin.ts
```diff
@@ -0,0 +1,91 @@
...
+            this.storageConfig!.cleanup()
```
timoxley, line 84: Same here.

in both cases the stop/start calls probably need to know about each other's current state or run through an async queue to make this robust.

packages/broker/src/storage/StoragePlugin.ts
```diff
@@ -0,0 +1,91 @@
...
+        this.cassandra = await this.getCassandraStorage()
+        this.storageConfig = await this.createStorageConfig()
+        this.storageConfig.getStreams().forEach((stream) => {
+            this.subscriptionManager.subscribe(stream.id, stream.partition)
+        })
+        this.storageConfig.addChangeListener({
+            onStreamAdded: (stream: StreamPart) => this.subscriptionManager.subscribe(stream.id, stream.partition),
+            onStreamRemoved: (stream: StreamPart) => this.subscriptionManager.unsubscribe(stream.id, stream.partition)
+        })
+        this.networkNode.addMessageListener((msg) => {
```
timoxley, line 44: If the plugin is stopped, does this still fire?

timoxley: Also note IIUC this messageListener function is never removed, nor can it be with the current public interface, there's no `networkNode.removeMessageListener`, thus this instance can never be garbage collected until the node is GC'ed.

packages/broker/src/storage/StoragePlugin.ts
harbu: One way to achieve what Tim described is to
- Make this class into an abstract base class
- Add a boolean field `this.started = false`
- Modify `start` method to work as follows
```ts
async start() {
    if (this.started) {
        throw new Error('started already')
    }
    this.started = true
    this.doStart()
}
```
- Re-name and change the visibility of method `async start()` to `abstract protected doStart()`, which the extending class would then implement.

packages/broker/src/storage/StoragePlugin.ts
teogeb: Fixed

packages/broker/src/storage/StoragePlugin.ts
teogeb: Added documentation to `Plugin.ts` about `start` and `stop` lifecycle methods. Plugins can therefore now assume that the methods are called in valid order, as the behaviour is documented. 

I'd say that there is no need to to add additional guards to Broker to prevent us from calling lifecycle methods in invalid order. If that happened, it would maybe even be better to catch the bug by crashing the app instead of silently ignoring the error. (Depends of how we want to handle error situations in general. I've understood that this approach is currently used in Broker as we are more in development mode than in maintenance mode).

packages/broker/src/storage/StoragePlugin.ts
teogeb: Good catch 👍 Added cleanup logic to `Broker` and `StoragePlugin`: 

- stops the HTTP server before the plugin stops
- removes the message listener (added `networkNode.removeMessageListener `to `network` package)
- unsubscribes from each storable stream
- unsubscribes from the `storage-node-assignments` stream

Added also` StoragePlugin.test.ts` to test the cleanup logic.

packages/broker/src/storage/StorageConfig.ts
```diff
@@ -144,8 +145,20 @@ export class StorageConfig {
...
+        }
+        networkNode.addMessageListener(messageListener)
         networkNode.subscribe(assignmentStreamId, 0)
+        return messageListener
+    }
+
+    stopAssignmentEventListener(messageListener: (msg: Protocol.StreamMessage) => void, streamrAddress: string, networkNode: NetworkNode) {
+        networkNode.removeMessageListener(messageListener)
+        const assignmentStreamId = this.getAssignmentStreamId(streamrAddress)
+        networkNode.unsubscribe(assignmentStreamId, 0)
```
harbu, line 157: Wondering actually if `subscriptionManager` should be used in this class? To guard from unsubscribing from streams that other plugins have subscribed to?

packages/broker/test/integration/storage/StoragePlugin.test.ts
```diff
@@ -0,0 +1,87 @@
...
+const STREAM_PARTS: StreamPart[] = [ 
+    { id: 'foo', partition: 0 },
+    { id: 'bar', partition: 0 }
+]
+
+const createMockPlugin = (networkNode: any, subscriptionManager: any) => {
+    const brokerConfig: any = {
+        ethereumPrivateKey: fastPrivateKey(),
+        network: {
+            isStorageNode: true
```
harbu, line 17: This should no longer have to be `true`?

packages/network/src/NetworkNode.ts
```diff
@@ -40,6 +40,10 @@ export class NetworkNode extends Node {
         this.on(NodeEvent.UNSEEN_MESSAGE_RECEIVED, cb)
     }
 
+    removeMessageListener(cb: (msg: MessageLayer.StreamMessage) => void): void {
```
harbu, line 43: :+1:

harbu (changes requested):
See comment on SubscriptionManager

packages/broker/src/storage/StorageConfig.ts
teogeb: Now it uses `subscriptionManager`

packages/broker/test/integration/storage/StoragePlugin.test.ts
teogeb: Fixed

harbu (approved):

PR #24, opened 2021-05-21
Add shared TypeScript config

timoxley:
Created a new internal package `@streamr/dev-config`.
This currently contains configs for typescript. 
This is now shared across the various packages e.g. 

```json
# packages/broker/tsconfig.json
{
  "extends": "@streamr/dev-config/ts/tsconfig.node.json",
  "compilerOptions": {
    "outDir": "dist",
    "rootDirs": ["src", "test"]
  },
  "include": [
    "src/**/*",
    "test/**/*"
  ]
}

```

Going forward babel & eslint configs could live in this package too.


harbu (commented):
LGTM; why were some of the github action workflows deleted though?

timoxley:
@harbu I tried cleaning out all the cruft, so unused dependencies and files. Workflow files in packages don't run in the monorepo, it only runs the top level `.github`

PR #36, opened 2021-05-25
Add StreamPartitionID class.

timoxley:
First pass at a shared structure to represent stream ids & partitions.

Likely will need revision or additional methods upon integration.

`StreamPartitionID` is a bit lengthy so I'm proposing perhaps `SPID` as the commonly used moniker. 
I admit this name does conjure up images of potatoes for me, but I think potatoes are cool.

If this looks reasonable to you at a high level then let's take a stab at actually integrating it and see how it looks/feels at that point.

We should do a pass over all the projects and convert as much as we can in one go, otherwise we end up creating https://xkcd.com/927/. PR(s) for integration will need to merge fast; this change will touch a lot of files across and thus will likely conflict a lot and be a PITA to maintain if the PR(s) are long-lived.

### DIFF
--- a/packages/protocol/src/utils/SPID.ts
+++ b/packages/protocol/src/utils/SPID.ts
@@ -32,7 +32,7 @@ export default class SPID {
      * @param partition - stream partition
      */
     constructor(id: string, partition = 0) {
-        this.id = typeof id === 'string' ? id.toLowerCase() : id
+        this.id = id
         this.partition = partition
         this.validate()
 
--- a/packages/protocol/src/utils/SPID.ts
+++ b/packages/protocol/src/utils/SPID.ts
@@ -1,3 +1,8 @@
+/**
+ * SPID - Stream + Partition ID
+ * See SPID constructor docs.
+ */
+
 import { format } from 'util'
 
 /**
@@ -38,7 +43,19 @@ class SPIDValidationError extends Error {
 }
 
 /**
- * SPID - Stream Partition ID
+ * SPID – Stream + Partition ID
+ * Simple Data container that represents a streamId + partition, and utilities
+ * to parse various formats into a SPID.
+ *
+ * Use this rather than inventing new data types to represenent or parse these.
+ *
+ * Usage:
+ * ```js
+ * const spid = new SPID(streamId, streamPartition)
+ * spid.id === streamId // true
+ * spid.partition === streamPartition // true
+ * ```
+ * See tests for more usage examples.
  */
 export default class SPID {
     /** stream id */
--- a/packages/protocol/src/utils/SPID.ts
+++ b/packages/protocol/src/utils/SPID.ts
@@ -1,3 +1,15 @@
+import { format } from 'util'
+
+class SPIDValidationError extends Error {
+    data: SPIDish
+    constructor(msg: string, data: SPIDish) {
+        super(format(msg, data))
+        this.data = data
+        if (Error.captureStackTrace) {
+            Error.captureStackTrace(this, this.constructor)
+        }
+    }
+}
 /**
  * SPID - Stream Partition ID
  */
@@ -26,9 +38,8 @@ export default class SPID {
         this.partition = partition
         this.validate()
 
-        // static values
-        const separator = (this.constructor as typeof SPID).SEPARATOR
-        this.key =`${this.id}${separator}${this.partition}`
+        // static cached values to prevent unnecessary runtime allocations
+        this.key = `${this.id}${SPID.SEPARATOR}${this.partition}`
         this.cachedSplit = [this.id, this.partition]
         // prevent all mutation
         Object.freeze(this)
@@ -40,11 +51,17 @@ export default class SPID {
      */
     private validate() {
         if (typeof this.id !== 'string' || this.id.length === 0) {
-            throw new Error(`SPID validation failed: id must be non-empty string: ${JSON.stringify({ id: this.id, partition: this.partition })}`)
+            throw new SPIDValidationError('SPID validation failed: id must be non-empty string: %o', {
+                id: this.id,
+                partition: this.partition
+            })
         }
 
         if (typeof this.partition !== 'number' || !Number.isSafeInteger(this.partition) || this.partition < 0) {
-            throw new Error(`SPID validation failed: paritition must be an integer >= 0: ${JSON.stringify({ id: this.id, partition: this.partition })}`)
+            throw new SPIDValidationError('SPID validation failed: partition must be a safe integer >= 0: %o', {
+                id: this.id,
+                partition: this.partition
+            })
         }
     }
 
@@ -56,17 +73,15 @@ export default class SPID {
         if (other === this) { return true }
         let otherSpid: SPID
         try {
-            otherSpid = (this.constructor as typeof SPID).from(other)
+            otherSpid = SPID.from(other)
         } catch (_err) {
+            // ignore error
             // not equal if not valid
             return false
         }
 
-        // check if matches both id + partition
-        return (
-            this.id === otherSpid.id
-            && this.partition === otherSpid.partition
-        )
+        // check key matches
+        return this.key === otherSpid.key
     }
 
     /**
@@ -85,11 +100,13 @@ export default class SPID {
             // @ts-expect-error object should have one of these, validated anyway
             const streamId = spidish.streamId || spidish.id
             // @ts-expect-error object should have one of these, validated anyway
-            const streamPartition = spidish.streamPartition || spidish.partition
-            return SPID.from(`${streamId}${this.SEPARATOR}${streamPartition}`)
+            const partition = spidish.streamPartition || spidish.partition
+            // try parse if a value was passed, but fall back to undefined i.e. default
+            const streamPartition = partition != null ? Number.parseFloat(partition) : undefined
+            return new SPID(streamId, streamPartition)
         } else {
             // TODO: add more conversions?
-            throw new Error(`SPID validation failed, input is malformed: ${spidish}`)
+            throw new SPIDValidationError('SPID validation failed, input is malformed: %o', spidish)
         }
     }
 
@@ -97,8 +114,10 @@ export default class SPID {
      * Convert String to SPID if possible.
      */
     static fromString(spidString: string): SPID {
-        const [id, parititionStr] = spidString.split(this.SEPARATOR)
-        return new SPID(id, Number.parseFloat(parititionStr))
+        const [id, partitionStr] = spidString.split(this.SEPARATOR)
+        // partition is optional, falls back to undefined i.e. default
+        const partition = partitionStr != null ? Number.parseFloat(partitionStr) : undefined
+        return new SPID(id, partition)
     }
 
     /**
@@ -112,7 +131,7 @@ export default class SPID {
      * Alias of toString.
      */
     toKey(): string {
-        return this.key
+        return this.toString()
     }
 
     /**
--- a/packages/protocol/src/utils/SPID.ts
+++ b/packages/protocol/src/utils/SPID.ts
@@ -1,8 +1,34 @@
 import { format } from 'util'
 
+/**
+ * Object version of SPID
+ */
+export type SPIDObject = {
+    streamId: string,
+    streamPartition: number
+}
+
+/**
+ * Represents partial SPIDObject e.g. for setting defaults
+ * Known keys
+ */
+export type SPIDObjectPartial = Partial<SPIDObject>
+
+/**
+ * SPID or String representing a SPID
+ * Object cases can be typechecked
+ * TODO: SPID string type safety
+ */
+export type SPIDLike = SPID | string | SPIDObject | { id: string, partition: number }
+
+/**
+ * Flexible input type
+ */
+export type SPIDLikePartial = SPIDLike | SPIDObjectPartial | Partial<{ id: string, partition: number }>
+
 class SPIDValidationError extends Error {
-    data: SPIDishPartial
-    constructor(msg: string, data: SPIDishPartial) {
+    data: SPIDLikePartial
+    constructor(msg: string, data: SPIDLikePartial) {
         super(format(msg, data))
         this.data = data
         if (Error.captureStackTrace) {
@@ -14,9 +40,6 @@ class SPIDValidationError extends Error {
 /**
  * SPID - Stream Partition ID
  */
-// if you're using JS numbers for
-// ids or keys going over MAX_SAFE_INTEGER has the potential for disasterous
-// effects e.g. serving private user information
 export default class SPID {
     /** stream id */
     public readonly id: string
@@ -65,7 +88,7 @@ export default class SPID {
     /**
      * True iff other value is equivalent.
      */
-    equals(other: SPIDish): boolean {
+    equals(other: SPIDLike): boolean {
         // check if same instance
         if (other === this) { return true }
         let otherSpid: SPID
@@ -81,18 +104,23 @@ export default class SPID {
         return this.key === otherSpid.key
     }
 
-    static toSPIDObjectPartial(spidish: SPIDishPartial): SPIDObjectPartial {
+    /**
+     * Convert SPIDLikePartial to SPIDObjectPartial
+     * i.e. normalizes various input types/shapes to { streamId?, streamPartition? }
+     * Note: does not throw on malformed input
+     */
+    static toSPIDObjectPartial(spidLike: SPIDLikePartial): SPIDObjectPartial {
         // convert from string
-        if (typeof spidish === 'string') {
-            const [streamId, partitionStr] = spidish.split(this.SEPARATOR)
+        if (typeof spidLike === 'string') {
+            const [streamId, partitionStr] = spidLike.split(this.SEPARATOR)
             // partition is optional, falls back to undefined i.e. default
             const streamPartition = partitionStr != null ? Number.parseFloat(partitionStr) : undefined
             return { streamId, streamPartition }
-        } else if (spidish && typeof spidish === 'object') {
+        } else if (spidLike && typeof spidLike === 'object') {
             // @ts-expect-error object should have one of these, validated anyway
-            const streamId = spidish.streamId || spidish.id
+            const streamId = spidLike.streamId || spidLike.id
             // @ts-expect-error object should have one of these, validated anyway
-            const partition = spidish.streamPartition || spidish.partition
+            const partition = spidLike.streamPartition || spidLike.partition
             // try parse if a value was passed, but fall back to undefined i.e. default
             const streamPartition = partition != null ? Number.parseFloat(partition) : undefined
             return { streamId, streamPartition }
@@ -108,32 +136,32 @@ export default class SPID {
      * fromDefaults(streamId, { partition: 0 })
      * ```
      */
-    static fromDefaults(spidish: SPIDish, defaultValues?: SPIDishPartial): SPID
-    static fromDefaults(spidish: SPIDishPartial, defaultValues: SPIDishPartial): SPID // requires id+partition if no defaults
-    static fromDefaults(spidish: SPIDishPartial, defaultValues?: SPIDishPartial): SPID {
+    static fromDefaults(spidLike: SPIDLike, defaultValues?: SPIDLikePartial): SPID
+    static fromDefaults(spidLike: SPIDLikePartial, defaultValues: SPIDLikePartial): SPID // requires id+partition if no defaults
+    static fromDefaults(spidLike: SPIDLikePartial, defaultValues?: SPIDLikePartial): SPID {
         // return spid if already spid
-        if (spidish instanceof SPID) {
-            return spidish
+        if (spidLike instanceof SPID) {
+            return spidLike
         }
 
         // defaults can be partial, e.g. { partition: 0 }
         // toSPIDObjectPartial can handle undefined input but we want external interface to check for it.
         const defaults = SPID.toSPIDObjectPartial(defaultValues!)
-        const { streamId = defaults?.streamId, streamPartition = defaults?.streamPartition } = SPID.toSPIDObjectPartial(spidish)
+        const { streamId = defaults?.streamId, streamPartition = defaults?.streamPartition } = SPID.toSPIDObjectPartial(spidLike)
         try {
             // constructor can handle partial input but we want external interface to check for it.
             return new SPID(streamId!, streamPartition!)
         } catch (err) {
             // TODO: add more conversions?
-            throw new SPIDValidationError(`SPID validation failed, input is malformed. ${err.message} %o`, spidish)
+            throw new SPIDValidationError(`SPID validation failed, input is malformed. ${err.message} %o`, spidLike)
         }
     }
 
     /**
      * Convert to SPID if possible.
      */
-    static from(spidish: SPIDish): SPID {
-        return SPID.fromDefaults(spidish)
+    static from(spidLike: SPIDLike): SPID {
+        return SPID.fromDefaults(spidLike)
     }
 
     /**
@@ -161,39 +189,14 @@ export default class SPID {
     }
 
     /**
-     * Returns a key for spidish
+     * Returns a key for spidLike
      * e.g.
      * ```js
      * const key = SPID.toKey({ streamId, streamPartition })
      * ```
      */
-    static toKey(spidish: SPIDish): string {
-        return SPID.from(spidish).key
+    static toKey(spidLike: SPIDLike): string {
+        return SPID.from(spidLike).key
     }
 }
 
-/**
- * Object version of SPID
- */
-export type SPIDObject = {
-    streamId: string,
-    streamPartition: number
-}
-
-/**
- * Represents partial SPIDObject e.g. for setting defaults
- * Known keys
- */
-export type SPIDObjectPartial = Partial<SPIDObject>
-
-/**
- * Flexible input type
- */
-export type SPIDishPartial = SPIDish | SPIDObjectPartial | Partial<{ id: string, partition: number }>
-
-/**
- * SPID or String representing a SPID
- * Object cases can be typechecked
- * TODO: SPID string type safety
- */
-export type SPIDish = SPID | string | SPIDObject | { id: string, partition: number }
--- a/packages/protocol/src/utils/SPID.ts
+++ b/packages/protocol/src/utils/SPID.ts
@@ -27,8 +27,6 @@ export default class SPID {
     /** string key representing SPID */
     public readonly key: string
 
-    protected readonly cachedSplit: [string, number] // cached split
-
     /**
      * @param id - stream id
      * @param partition - stream partition
@@ -40,10 +38,8 @@ export default class SPID {
 
         // static cached values to prevent unnecessary runtime allocations
         this.key = `${this.id}${SPID.SEPARATOR}${this.partition}`
-        this.cachedSplit = [this.id, this.partition]
         // prevent all mutation
         Object.freeze(this)
-        Object.freeze(this.cachedSplit)
     }
 
     /**
@@ -132,30 +128,6 @@ export default class SPID {
      */
     toKey(): string {
         return this.toString()
-    }
-
-    /**
-     * Returns an [id, partition] tuple.
-     * Useful for destructuring.
-     * e.g.
-     * ```js
-     * const [streamId, streamPartition] = spid.split()
-     * ```
-     */
-    split(): [string, number] {
-        return this.cachedSplit
-    }
-
-    /**
-     * Returns an [id, partition] tuple.
-     * Useful for destructuring.
-     * e.g.
-     * ```js
-     * const [streamId, streamPartition] = SPID.split(someString)
-     * ```
-     */
-    static split(spidish: SPIDish): [string, number] {
-        return SPID.from(spidish).split()
     }
 
     /**
--- a/packages/protocol/src/utils/SPID.ts
+++ b/packages/protocol/src/utils/SPID.ts
@@ -1,13 +1,16 @@
 /**
  * SPID - Stream Partition ID
  */
+// if you're using JS numbers for
+// ids or keys going over MAX_SAFE_INTEGER has the potential for disasterous
+// effects e.g. serving private user information
 export default class SPID {
     /** stream id */
     public readonly id: string
     /** stream partition */
     public readonly partition: number
     /** toString/fromString separator */
-    protected static readonly SEPARATOR = '|'
+    protected static readonly SEPARATOR = '#'
 
     /** string key representing SPID */
     public readonly key: string
@@ -37,11 +40,11 @@ export default class SPID {
      */
     private validate() {
         if (typeof this.id !== 'string' || this.id.length === 0) {
-            throw new Error('SPID validation failed: id must be non-empty string')
+            throw new Error(`SPID validation failed: id must be non-empty string: ${JSON.stringify({ id: this.id, partition: this.partition })}`)
         }
 
         if (typeof this.partition !== 'number' || !Number.isSafeInteger(this.partition) || this.partition < 0) {
-            throw new Error('SPID validation failed: paritition must be an integer >= 0 ')
+            throw new Error(`SPID validation failed: paritition must be an integer >= 0: ${JSON.stringify({ id: this.id, partition: this.partition })}`)
         }
     }
 
@@ -95,7 +98,7 @@ export default class SPID {
      */
     static fromString(spidString: string): SPID {
         const [id, parititionStr] = spidString.split(this.SEPARATOR)
-        return new SPID(id, Number.parseFloat(parititionStr) || 0)
+        return new SPID(id, Number.parseFloat(parititionStr))
     }
 
     /**
--- a/packages/protocol/src/utils/SPID.ts
+++ b/packages/protocol/src/utils/SPID.ts
@@ -1,8 +1,8 @@
 import { format } from 'util'
 
 class SPIDValidationError extends Error {
-    data: SPIDish
-    constructor(msg: string, data: SPIDish) {
+    data: SPIDishPartial
+    constructor(msg: string, data: SPIDishPartial) {
         super(format(msg, data))
         this.data = data
         if (Error.captureStackTrace) {
@@ -10,6 +10,7 @@ class SPIDValidationError extends Error {
         }
     }
 }
+
 /**
  * SPID - Stream Partition ID
  */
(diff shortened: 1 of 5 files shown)


juslesan (approved):

packages/protocol/src/utils/StreamPartitionID.ts
```diff
@@ -0,0 +1,154 @@
...
+
+        // convert from string
+        if (typeof spidish === 'string') {
+            return SPID.fromString(spidish)
+        } else if (spidish && typeof spidish === 'object') {
+            // @ts-expect-error object should have one of these, validated anyway
+            const streamId = spidish.streamId || spidish.id
+            // @ts-expect-error object should have one of these, validated anyway
+            const streamPartition = spidish.streamPartition || spidish.partition
+            return SPID.from(`${streamId}${this.SEPARATOR}${streamPartition}`)
```
harbu, line 86: More direct way of constructing instance `return new SPID(streamId, streamPartition)` to avoid redundant string parsing.

packages/protocol/src/utils/StreamPartitionID.ts
```diff
@@ -0,0 +1,154 @@
...
+            throw new Error(`SPID validation failed, input is malformed: ${spidish}`)
+        }
+    }
+
+    /**
+     * Convert String to SPID if possible.
+     */
+    static fromString(spidString: string): SPID {
+        const [id, parititionStr] = spidString.split(this.SEPARATOR)
+        return new SPID(id, Number.parseFloat(parititionStr) || 0)
```
harbu, line 98: I'm inclined to think it should be an error if the passed string does not conform to the pattern `stream-id|partition` or if the partition is non-numeric. In its current form, it seems like it will accept such a string but just default the partition to be zero?

teogeb:
Very nice implementation! 👍 

How about naming the class to `StreamPart`? It would describe that this object represents a part of a stream. The term would be short and precise.

Another alternative would be `Partition`. It would describe the content with a basic English word. I guess it it is quite obvious that in our project the Partition refers to a part of a stream, and not to some other partial object. E.g. Shopify uses same kind of semantic. The two main classes there are `Product` and `Variant`.

I agree that the class doesn't contain the actual "Stream" or part of it, and is just a reference to it. Therefore appending an "ID" suffix would be valid. But to shorten the term, it would maybe be ok to left that suffix out, too. 

If we shortened the term to SPID abbreviation, that would be one extra term that users need to know when  they start to use the Streamr project. Those kind of important but bit unclear key terms make the learning curve very steep. That's why I'd like to use some other, more descriptive term.

packages/protocol/src/utils/StreamPartitionID.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    /**
+     * Returns an [id, partition] tuple.
+     * Useful for destructuring.
+     * e.g.
+     * ```js
+     * const [streamId, streamPartition] = spid.split()
+     * ```
+     */
+    split(): [string, number] {
```
harbu, line 123: I'm skeptical whether caching the split provides more benefits vs. cost. How often is `split()`  actually used? Put another way: what proportion of `SPID` objects created would have their `split` method invoked more than once?

packages/protocol/src/utils/StreamPartitionID.ts
```diff
@@ -0,0 +1,154 @@
...
+    public readonly key: string
+
+    protected readonly cachedSplit: [string, number] // cached split
+
+    /**
+     * @param id - stream id
+     * @param partition - stream partition
+     */
+    constructor(id: string, partition = 0) {
+        this.id = typeof id === 'string' ? id.toLowerCase() : id
```
teogeb, line 22: Should not lowercase the string as the streamID's are not case senstive?

packages/protocol/src/utils/StreamPartitionID.ts
```diff
@@ -0,0 +1,154 @@
...
+     */
+    toString(): string {
+        return this.key
+    }
+
+    /**
+     * Alias of toString.
+     */
+    toKey(): string {
+        return this.key
```
harbu, line 112: Just to be safe (and faithful to the code comment) in the face of potential future modifications, I would have the implementation here call `this.toString()`.

packages/protocol/src/utils/StreamPartitionID.ts
```diff
@@ -0,0 +1,154 @@
...
+     */
+    static toKey(spidish: SPIDish): string {
+        return SPID.from(spidish).key
+    }
+}
+
+/**
+ * SPID or String representing a SPID
+ */
+export type SPIDish = SPID | string | { streamId: string, streamPartition: number } | { id: string, partition: number }
```
harbu, line 154: Nice one. I do think we want to probably move to use type `SPID` most of the time, but in the journey to get there, we can mark existing representations of streamid-partition as `SPIDish`. (Also there are some cases where you cannot avoid having to use a string representation of `SPID` such as object keys.)

packages/protocol/test/unit/utils/StreamPartitionID.test.ts
```diff
@@ -0,0 +1,188 @@
+import SPID  from '../../../src/utils/StreamPartitionID'
+
+describe('StreamPartitionID', () => {
```
harbu, line 3: Imo use `SPID` in describe

packages/protocol/src/utils/StreamPartitionID.ts
```diff
@@ -0,0 +1,154 @@
+/**
+ * SPID - Stream Partition ID
+ */
+export default class SPID {
```
harbu, line 4: Ah file name should probably match class name (as is convention)?

packages/protocol/src/utils/StreamPartitionID.ts
```diff
@@ -0,0 +1,154 @@
...
+    /** stream id */
+    public readonly id: string
+    /** stream partition */
+    public readonly partition: number
+    /** toString/fromString separator */
+    protected static readonly SEPARATOR = '|'
```
teogeb, line 10: Maybe we could use some char which is easily writable when using the command line. E.g. the CLI tool would then be able to use the same semantics. 

```
streamr stream subscribe 0x12345/foo/bar#123
```

The streamID can contain https://base64.guru/standards/base64url characters, so we can use e.g. colon or hash if we want. Maybe hash char would be great as our use case is be very similar to to URL hash anchor concept (a pointer into a document).

packages/protocol/test/unit/utils/StreamPartitionID.test.ts
```diff
@@ -0,0 +1,188 @@
...
+        expect(key).toBe(spid.toString())
+    })
+
+    it('defaults partition to 0', () => {
+        const spid = new SPID(STREAM_ID)
+        expect(spid.id).toEqual(STREAM_ID)
+        expect(spid.partition).toBe(0)
+    })
+
+    it('lowercases streamId', () => {
```
harbu, line 21: As I expressed in an earlier comment, I believe the intention currently is not to touch the casing at all.

packages/protocol/test/unit/utils/StreamPartitionID.test.ts
```diff
@@ -0,0 +1,188 @@
...
+        const spid = new SPID(STREAM_ID.toUpperCase())
+        expect(spid.id).toEqual(STREAM_ID)
+        expect(spid.partition).toBe(0)
+    })
+
+    it('can use array destructuring with split', () => {
```
harbu, line 27: Thinking out loud here: if we can already do object destructuring, i.e.
```ts
const { id, partition} = spid
```
why do we need array destructing and thus method `split`?

packages/protocol/test/unit/utils/StreamPartitionID.test.ts
```diff
@@ -0,0 +1,188 @@
...
+describe('StreamPartitionID', () => {
+    const STREAM_ID = 'test-stream-id'
+    const PARTITION = 1
+
+    it('can use object destructuring', () => {
+        const spid = new SPID(STREAM_ID, PARTITION)
+        const { id, partition, key } = spid
+        expect(id).toBe(STREAM_ID)
+        expect(partition).toBe(PARTITION)
+        expect(key).toBe(spid.toString())
```
harbu, line 12: I think it would be useful to actually expect for a literal value of `key` here? In case of a situation where `toString` and `key` would be `undefined` etc. for whatever weird reason.

packages/protocol/test/unit/utils/StreamPartitionID.test.ts
```diff
@@ -0,0 +1,188 @@
...
+    })
+
+    describe('from', () => {
+        it('returns self if already a spid', () => {
+            const spid = new SPID(STREAM_ID, PARTITION)
+            expect(SPID.from(spid)).toBe(spid)
+        })
+
+        it('can parse valid strings', () => {
+            // @ts-expect-error SPID.SEPARATOR is protected
```
harbu, line 108: You can probably make `SPID.SEPARATOR` `public` to avoid having to do this? There is no harm if it is `readonly` also.

packages/protocol/test/unit/utils/StreamPartitionID.test.ts
```diff
@@ -0,0 +1,188 @@
...
+                // @ts-expect-error testing bad input
+                new SPID(STREAM_ID, partition)
+            }).toThrow()
+        }
+    })
+
+    describe('from', () => {
+        it('returns self if already a spid', () => {
+            const spid = new SPID(STREAM_ID, PARTITION)
+            expect(SPID.from(spid)).toBe(spid)
```
harbu, line 104: I'm assuming the `toBe` here is checking equality of identity as opposed to value equality, right?

packages/protocol/test/unit/utils/StreamPartitionID.test.ts
```diff
@@ -0,0 +1,188 @@
...
+            const spid = SPID.from({id: `not${STREAM_ID}`, partition: PARTITION + 1, streamId: STREAM_ID, streamPartition: PARTITION})
+            expect(spid.id).toBe(STREAM_ID)
+            expect(spid.partition).toBe(PARTITION)
+        })
+    })
+
+    describe('toString', () => {
+        it('creates a string', () => {
+            const spid = new SPID(STREAM_ID, PARTITION)
+            // @ts-expect-error SPID.SEPARATOR is protected
```
harbu, line 142: Same as above, you could make it public.

harbu (commented):
Yeah `SPID` conjures images of potatoes in my head too (now that you said it), however, it is such a prominent concept in the code that it does warrant a short name.

harbu:
> Very nice implementation! +1
> 
> How about naming the class to `StreamPart`? It would describe that this object represents a part of a stream. The term would be short and precise.
> 
> Another alternative would be `Partition`. It would describe the content with a basic English word. I guess it is quite obvious that in our project the Partition refers to a part of a stream, and not to some other partial object. E.g. Shopify uses same kind of semantic. The two main classes there are `Product` and `Variant`.
> 
> I agree that the class doesn't contain the actual stream or part of it, and is just a reference to it. Therefore appending an "ID" suffix would be valid. But to shorten the term, it would maybe be OK to left that suffix out, too.
> 
> If we shortened the term to SPID abbreviation, that would be one extra term that users need to know when they start to use the Streamr project. Those kind of important but bit unclear key terms make the learning curve very steep. That's why I'd like to use some other, more descriptive term.

Our official documentation refers to this concept as Partition, from https://streamr.network/docs/streams/partitioning:

> To be precise, when messages are published to a stream, they are actually published to a partition within that stream. Partitions can be understood as sub-streams, or in other words, parts of a stream

packages/protocol/src/utils/StreamPartitionID.ts
```diff
@@ -0,0 +1,154 @@
...
+        Object.freeze(this)
+        Object.freeze(this.cachedSplit)
+    }
+
+    /**
+     * Throws if data is invalid.
+     */
+    private validate() {
+        if (typeof this.id !== 'string' || this.id.length === 0) {
+            throw new Error('SPID validation failed: id must be non-empty string')
```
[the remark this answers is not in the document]
timoxley: You're right, it should.

packages/protocol/src/utils/StreamPartitionID.ts
```diff
@@ -0,0 +1,154 @@
...
+     * @param id - stream id
+     * @param partition - stream partition
+     */
+    constructor(id: string, partition = 0) {
+        this.id = typeof id === 'string' ? id.toLowerCase() : id
+        this.partition = partition
+        this.validate()
+
+        // static values
+        const separator = (this.constructor as typeof SPID).SEPARATOR
```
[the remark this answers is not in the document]
timoxley: `SPID` isn't really correct though, it should always read `this.constructor` when referencing static values from the current class, this prevents issues when subclassing.

packages/protocol/src/utils/StreamPartitionID.ts
timoxley: yes perhaps

packages/protocol/test/unit/utils/StreamPartitionID.test.ts
timoxley: array destructuring is a bit cleaner if you don't want to use the defined `id`, `partition` names. Paritition is probably fine but `id` might already exist.
```js
const { id: streamId, partition: streamPartition } = spid
// vs
const [streamId, streamPartition] = spid.split()
```

packages/protocol/test/unit/utils/StreamPartitionID.test.ts
timoxley: I feel like it should be opaque to resist urge to do build or parse or create your own id+partition strings

packages/protocol/src/utils/StreamPartitionID.ts
timoxley: what "cost" though? 

Idea is that since these SPID things are to be used everywhere, ideally the whole class has minimal runtime cost  compared to not using the class, specifically, makes no unnecessary allocations. I'd agree it's premature optimisation if the implementation was remotely complicated, but it seems trivial to just make it do no allocations so why not.

packages/protocol/src/utils/StreamPartitionID.ts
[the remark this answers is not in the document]
timoxley: Not that subclassing is likely or desirable here, just you should always reference `this.constructor` as a habit so you never have to think about whether you're going to break subclassing or not.

packages/protocol/src/utils/StreamPartitionID.ts
```diff
@@ -0,0 +1,154 @@
...
+    /** string key representing SPID */
+    public readonly key: string
+
+    protected readonly cachedSplit: [string, number] // cached split
+
+    /**
+     * @param id - stream id
+     * @param partition - stream partition
+     */
+    constructor(id: string, partition = 0) {
```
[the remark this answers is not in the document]
timoxley: Mainly for IDE auto completion to make the stream id distinct from e.g. an instance `id`

timoxley:
> It would describe that this object represents a part of a stream.

@teogeb But what is a "part"? It's an ID that represents a stream's "partition", not a stream's "part", so it should be `StreamPartition` if anything, and to remove any ambiguity it'd be an ID `StreamPartitionID` i.e. `SPID`.

timoxley:
@harbu @teogeb Regarding stream toLowerCase: usage as a key without lowercasing seems like bugs waiting to happen.

I recall there was a specific issue that blocked @bearni95 from solving this a few weeks back, but I don't remember what the specific issue blocking it was.

packages/protocol/src/utils/StreamPartitionID.ts
timoxley: Right, sure

packages/protocol/src/utils/StreamPartitionID.ts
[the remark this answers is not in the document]
harbu: Coming from Java, static things are never overridden by subclasses. I don't know if this is a general principle in other languages as well.

Be it as it may, it feels quite unidiomatic (and hard to remember) to me to have to call static things with `(this.constructor as typeof BaseClass)`, maybe I'd rather then have them as constants outside the class`

E.g.
const SEPARATOR = '|'
...
class SPID {
    ....
}
```

File: packages/protocol/src/utils/StreamPartitionID.ts (hunk 2 above)
Comment (harbu) [reply packages/protocol/src/utils/StreamPartitionID.ts:98]: The way I see it:

1. We are instantiating a representation of `streamId` _and_ `partition` here. Not too much to ask imo that both be supplied or add a `|0` to the end.

1. Catching accidents
```ts
function performWhatever(userId: string, streamId: string): void {
   const spid = SPID.from(streamId)
   doAsSomething(userId, spid)
}

...

performWhatever('streamId', '0xFFFF...') // whoops
```

packages/protocol/src/utils/StreamPartitionID.ts
harbu: Run-time cost:
Given a single instance of this class
- If `split()` is never invoked, we created an array in constructor that was never used.
- If `split()` is invoked once, well then it does get used but is it any better than
```
split(): [string, number] {
    return [this.id, this.partition]
}
```
Conceptual cost:
I'd argue that the snippet above is conceptually simpler than having an extra field to keep track of. (Albeit that field is simple, yes.)

So the way I see it, is that the case where this provides benefit > cost, is that there is reasonably many usages patterns in the code where this method gets called more than once for a single instance?

packages/protocol/test/unit/utils/StreamPartitionID.test.ts
harbu: Aaah good point. :+1:

packages/protocol/test/unit/utils/StreamPartitionID.test.ts
harbu: Fair point. So basically keeping the separator protected to discourage from bad practices.

Just as a thought: how do you feel about using the `|` separator literally in the tests? The apparent cost being that if we decide to change said separator, tests will break and need to be updated. But could get rid of `@ts-expect-error`? Both sides ma

harbu:
> > It would describe that this object represents a part of a stream.
> 
> @teogeb But what is a "part"? It's an ID that represents a stream's "partition", not a stream's "part", so it should be `StreamPartition` if anything, and to remove any ambiguity it'd be an ID `StreamPartitionID` i.e. `SPID`.

I was thinking about the same thing a bit, or to put it in generic terms `ID !== Entity`. Canvas vs. CanvasId, User vs. UserId, Stream vs. StreamId. Indeed differentiating id(s) and their associated entities is often useful.

harbu:
> @harbu @teogeb Regarding stream id toLowerCase: usage as a key without lowercasing seems like bugs waiting to happen.
> 
> I recall there was a specific issue that blocked @bearni95 from solving this a few weeks back, but I don't remember what the specific issue blocking it was, perhaps something in `core-api`?

Assuming my recollection of past weeks meetings is coherent (:smiley_cat:), I believe the decision was made not to pursue this due to the complexities that were unraveled in the process. It is up to the user of a stream to make sure the get the capitalization correct, if they don't, they are effectively referring to a non-existing stream.

packages/protocol/src/utils/StreamPartitionID.ts
```diff
@@ -0,0 +1,154 @@
+/**
+ * SPID - Stream Partition ID
+ */
+export default class SPID {
+    /** stream id */
```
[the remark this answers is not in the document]
timoxley: https://github.com/streamr-dev/network-monorepo/pull/36/files#r639709738

packages/protocol/src/utils/StreamPartitionID.ts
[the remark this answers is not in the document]
timoxley: >  it feels quite unidiomatic (and hard to remember) 

I guess it feels unidiomatic because people don't subclass much and haven't felt the pain, or they just don't think about how it works? same as most people use `.catch` instead of second argument to `then` in a way that allows double handling or control reentry, or subclass errors without calling `Error.captureStackTrace` or `throw 'some string'`. These all technically "work" but is it good?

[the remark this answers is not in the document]
timoxley: ahhhh well, I say this but I guess the same problem exists in the calls to `SPID.from` etc. 
OK, I'll just change it to `SPID` here.

packages/protocol/src/utils/StreamPartitionID.ts
timoxley: I think the intent here was to give partition a chance to run through parseFloat but yeah perhaps doesn't matter.

packages/protocol/src/utils/StreamPartitionID.ts
timoxley: I think the intention was so the SPID parsing in the implementation of this would works out of the box: `client.subscribe(streamId)` and `client.subscribe({ streamId })`

packages/protocol/src/utils/StreamPartitionID.ts
timoxley: > If split() is never invoked, we created an array in constructor that was never used.

That's fine, it's not the memory I'm trying to optimise for about it's the cost to allocate & GC it. Young generation allocation + GC is fast, generally not something to worry about but seems unnecessary to do any allocation when we can precompute the array upfront. 

> If split() is invoked once, well then it does get used but is it any better …

Intention is that this doesn't require an allocation just to do destructuring:

```js
function onEveryMessage(spid) {
  const [streamId, streamPartition] = spid.split()
  // …
}
```

I was thinking this may be the common usage pattern, but thinking about it now, if everything is using SPIDs then perhaps it won't be necessary and SPIDs could be treated as opaque.

packages/protocol/test/unit/utils/StreamPartitionID.test.ts
timoxley: > But could get rid of @ts-expect-error

I think it's fine to have purposely ignored ts errors when tests poke into internals.

timoxley: It'd be good if there were a way to to say "nothing can access this except subclasses and stuff in the test folder" Maybe marking it as `internal` would be better.

packages/protocol/src/utils/StreamPartitionID.ts
timoxley: having an string marker/prefix that allows us to easily detect a streamId vs some other random string isn't a bad idea though

harbu: The lack of `partition` has a context-dependent interpretation, right? Given stream `s` that has 4 partitions what you'd really want is something as follows.

Subscribing:
```ts
SPID.from('s') // returns [new SPID('s', 0), new SPID('s', 1), new SPID('s', 2), new SPID('s', 3)]
```

Publishing
```ts
SPID.from('s') // returns new SPID('s', randomIntegerBetween(0,3))
```

I think interpreting it as zero automatically is not what is wanted in our use cases.

packages/protocol/src/utils/StreamPartitionID.ts
harbu: Say I create a `SPID` and never use `split` on it? Wouldn't that cause an extra allocation & GC cost that was unnecessary? Are you saying that if you do this allocation in the constructor it's cheaper?

packages/protocol/src/utils/SPID.ts
```diff
@@ -0,0 +1,176 @@
...
+}
+/**
+ * SPID - Stream Partition ID
+ */
+// if you're using JS numbers for
+// ids or keys going over MAX_SAFE_INTEGER has the potential for disasterous
+// effects e.g. serving private user information
+export default class SPID {
+    /** stream id */
+    public readonly id: string
```
teogeb, line 21: If we rename this to `streamId`, we don't need the comment, and maybe not even the `split` method as object destructuring will have clear and unique names?

packages/protocol/test/unit/utils/StreamPartitionID.test.ts
timoxley: done

timoxley:
@teogeb @harbu 

* Removed `SPID#split() and SPID.split()`
* Renamed `SPIDish` to `SPIDLike`, sounds less fun but this seems to be how other tools name such things.
* Changed `SPID.from(…)` & `new SPID(…)` to require a partition to be set. No more partition 0 as default.
* Added `SPID.fromDefaults(input, defaults)` which allows setting a default partition if one is not supplied. See example below.
* Added `SPID#toObject()` which returns `{ streamId, streamPartition }`

```js
SPID.from(streamID) // fail
SPID.from({ id: streamID }) // fail
SPID.fromDefaults(streamID, { partition: 0 }) // ok
SPID.fromDefaults({ id: streamID }, { partition: 0 }) // ok
```

packages/protocol/src/utils/StreamPartitionID.ts
timoxley: done

packages/protocol/src/utils/StreamPartitionID.ts
timoxley: removed split

packages/protocol/test/unit/utils/StreamPartitionID.test.ts
timoxley: done

packages/protocol/src/utils/SPID.ts
```diff
@@ -0,0 +1,176 @@
...
+        this.data = data
+        if (Error.captureStackTrace) {
+            Error.captureStackTrace(this, this.constructor)
+        }
+    }
+}
+/**
+ * SPID - Stream Partition ID
+ */
+// if you're using JS numbers for
```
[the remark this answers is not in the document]
timoxley: heh, I pasted a draft tweet into here to format it, removed.

packages/protocol/src/utils/SPID.ts
```diff
@@ -0,0 +1,202 @@
...
+    /**
+     * Convert to SPID if possible, with defaults.
+     * e.g.
+     * ```ts
+     * fromDefaults(streamId, { partition: 0 })
+     * ```
+     */
+    static fromDefaults(spidLike: SPIDLike, defaultValues?: SPIDLikePartial): SPID
+    static fromDefaults(spidLike: SPIDLikePartial, defaultValues: SPIDLikePartial): SPID // requires id+partition if no defaults
+    static fromDefaults(spidLike: SPIDLikePartial, defaultValues?: SPIDLikePartial): SPID {
```
harbu, line 141: Wouldn't having only the third declaration be enough here since type `SPIDLikePartial` already includes `SPIDLike` as one of its options?

packages/protocol/src/utils/SPID.ts
```diff
@@ -0,0 +1,202 @@
...
+        // return spid if already spid
+        if (spidLike instanceof SPID) {
+            return spidLike
+        }
+
+        // defaults can be partial, e.g. { partition: 0 }
+        // toSPIDObjectPartial can handle undefined input but we want external interface to check for it.
+        const defaults = SPID.toSPIDObjectPartial(defaultValues!)
+        const { streamId = defaults?.streamId, streamPartition = defaults?.streamPartition } = SPID.toSPIDObjectPartial(spidLike)
```
harbu, line 150: Didn't know you can give default values in object destructuring, neat.

harbu (approved):
lgtm

packages/protocol/src/utils/SPID.ts
```diff
@@ -0,0 +1,202 @@
...
+     * Convert SPIDLikePartial to SPIDObjectPartial
+     * i.e. normalizes various input types/shapes to { streamId?, streamPartition? }
+     * Note: does not throw on malformed input
+     */
+    static toSPIDObjectPartial(spidLike: SPIDLikePartial): SPIDObjectPartial {
+        // convert from string
+        if (typeof spidLike === 'string') {
+            const [streamId, partitionStr] = spidLike.split(this.SEPARATOR)
+            // partition is optional, falls back to undefined i.e. default
+            const streamPartition = partitionStr != null ? Number.parseFloat(partitionStr) : undefined
```
teogeb, line 117: Could use `parseInt` instead of `parseFloat`? (also at line 125)

packages/protocol/src/utils/SPID.ts
```diff
@@ -0,0 +1,202 @@
...
+ * SPID or String representing a SPID
+ * Object cases can be typechecked
+ * TODO: SPID string type safety
+ */
+export type SPIDLike = SPID | string | SPIDObject | { id: string, partition: number }
+
+/**
+ * Flexible input type
+ */
+export type SPIDLikePartial = SPIDLike | SPIDObjectPartial | Partial<{ id: string, partition: number }>
```
teogeb, line 27: This definition allows `{ partition: 123 }` to be `SPIDLikePartial`? Intentional?

packages/protocol/src/utils/SPID.ts
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        // defaults can be partial, e.g. { partition: 0 }
+        // toSPIDObjectPartial can handle undefined input but we want external interface to check for it.
+        const defaults = SPID.toSPIDObjectPartial(defaultValues!)
+        const { streamId = defaults?.streamId, streamPartition = defaults?.streamPartition } = SPID.toSPIDObjectPartial(spidLike)
+        try {
+            // constructor can handle partial input but we want external interface to check for it.
+            return new SPID(streamId!, streamPartition!)
+        } catch (err) {
```
teogeb, line 154: Is this catch necessary since constructor already throw a `SPIDValidationError` if the input is not valid?

packages/protocol/src/utils/SPID.ts
```diff
@@ -0,0 +1,202 @@
...
+        // check key matches
+        return this.key === otherSpid.key
+    }
+
+    /**
+     * Convert SPIDLikePartial to SPIDObjectPartial
+     * i.e. normalizes various input types/shapes to { streamId?, streamPartition? }
+     * Note: does not throw on malformed input
+     */
+    static toSPIDObjectPartial(spidLike: SPIDLikePartial): SPIDObjectPartial {
```
teogeb, line 112: The`spidLike` is an optional parameter? We could remove the `!` cast at line 149.

packages/protocol/src/utils/SPID.ts
timoxley: The catch error shows the input to the `from` call.

packages/protocol/src/utils/SPID.ts
timoxley: Intentional

packages/protocol/src/utils/SPID.ts
timoxley: If user tries to use partition 1.5 should we error or pick partition 1?

packages/protocol/src/utils/SPID.ts
timoxley: Right, but that will always be the case in TS, the last overload always has to match all the previous ones.

In TS method/function overloads:
* select the first matching signature
* and the last "implementation" signature always has to be a superset of all the other signatures.

e.g. you can't do something like this
```typescript
class A {
  convert(v: string)
  convert(v: number) { // This overload signature is not compatible with its implementation signature.
      
  } 
}
```
----

In our case, we want TS to error when you give a partial input for the spid, and also nothing for a default, which we can detect is wrong at type compile time:

```js
// partial input, no defaults
fromDefaults({ streamId }) // TypeScript error (wants a `SPIDLike`, not a `SPIDLikePartial`)
```

The first overload matches a non-partial, and keeps the second argument optional. This has to go first since the second overload would also match this signature and thus be selected, and would require default values even when you don't input a partial object.

```typescript
static fromDefaults(spidLike: SPIDLike, defaultValues?: SPIDLikePartial): SPID
```

The second overload tells TS to *require* `defaultValues` to be passed if a partial SPID was passed (or something like it) i.e. `SPIDLikePartitial`. This is the bit we want.
```typescript
static fromDefaults(spidLike: SPIDLikePartial, defaultValues: SPIDLikePartial): SPID // requires id+partition if no defaults
```

The third overload is the implementation signature and should match all prior signatures, and acts as a catch all:
```typescript
static fromDefaults(spidLike: SPIDLikePartial, defaultValues: SPIDLikePartial): SPID // requires id+partition if no defaults
```

packages/protocol/src/utils/SPID.ts
timoxley: if the user calls `toSPIDObjectPartial` we probably want to force them to either verify input is valid or explicitly bang it.

packages/protocol/src/utils/SPID.ts
harbu: I think I'm on to what you're saying, one question though.

Let's say I pass as 1st argument an instance of `SPIDLikePartial` and 2nd argument is left `undefined`. The first overload is skipped  since the 1st argument type doesn't "match" (can't assign `SPIDLikePartial` to `SPIDLike`). We are left thus with
```ts
static fromDefaults(spidLike: SPIDLikePartial, defaultValues: SPIDLikePartial): SPID // requires id+partition if no defaults
static fromDefaults(spidLike: SPIDLikePartial, defaultValues?: SPIDLikePartial): SPID {
```

Is it then so that the given arguments must match _both_ of these overloaded signatures, or is an _or_ style ordeal? I.e. it fails the `fromDefaults(spidLike: SPIDLikePartial, defaultValues: SPIDLikePartial)` "match" (because 2nd argument is undefined), but will "match" the last, catch-all overload? So I'm assuming it is not enough for it to just match the catch-all signature?

harbu: Ahh or is more like that there like a special interpretation of the last signature when giving multiple? Let's say we have some method with abstract arguments:
```ts
static foo(...sig1)
static foo(...sig2)
static foo(...sig3)
static foo(...sig4) {
  ....
}
```

Is there like a special rule that you can't use `static foo(...sig4)` directly to call method, instead you must use one of the above options? But the limiting thing is that the other options must be compatible with `static foo(...sig4)` ?

packages/protocol/src/utils/SPID.ts
harbu: My thinking:
If it were a decimal given by the user:
- It's most likely a typing / copy-paste errors of sort => nice to provide an error to the user so they can have a look at it and make sure they are getting what they intended.
- If they are genuinely giving a decimal they have misunderstood something. Then also an error maybe useful to course correct their understanding.

kare:
Please make sure src contains a mention about what SPID stands for, thanks

PR #37, opened 2021-05-26
Fix InstructionRetryManager test

ptesavol:
Added afterEach() that stops the retrying. Before the fix the test kept on running 100 times per second in the background.


harbu (approved):
Good catch

PR #40, opened 2021-05-27
fix: add show-links.sh and update README

harbu:
Add `show-links.sh` utility to check symlinked packages.

Run `./show-links.sh`

Output
```
broker
	node_modules/streamr-test-utils
	node_modules/streamr-network
cli-tools
	node_modules/streamr-client
client
	node_modules/streamr-test-utils
	node_modules/streamr-client-protocol
network
	node_modules/streamr-test-utils
protocol
	node_modules/streamr-test-utils
test-utils
```

### DIFF
--- a/show-links.sh
+++ b/show-links.sh
@@ -5,13 +5,12 @@
 TITLE_COLOR='\033[0;36m'
 NO_COLOR='\033[0m'
 
-PACKAGES="broker
-cli-tools
-client
-network
-protocol
-test-utils
-"
+LIST_CMD=`npx lerna list --all --parseable | xargs -L1 basename`
+PACKAGES=()
+while IFS= read -r line; do
+    PACKAGES+=("$line")
+done < <(echo $LIST_CMD)
+
 for p in $PACKAGES
 do
     RES=`(cd packages/$p && find node_modules -maxdepth 1 -type l) | sed 's/^/\t/'`


show-links.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+# Go through packages and for each enumrate the node_modules that are symlinked
+
+TITLE_COLOR='\033[0;36m'
+NO_COLOR='\033[0m'
+
+PACKAGES="broker
```
timoxley, line 8: perhaps `npx lerna list --all --parseable`

show-links.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+# Go through packages and for each enumrate the node_modules that are symlinked
```
timoxley, line 3:
```suggestion
# Go through packages and for each enumerate the node_modules that are symlinked
```

show-links.sh
```diff
@@ -0,0 +1,24 @@
...
+
+PACKAGES="broker
+cli-tools
+client
+network
+protocol
+test-utils
+"
+for p in $PACKAGES
+do
```
timoxley, line 16: Oh actually, you probably want `lerna exec`, something like this:

`npx lerna exec --no-bail 'find node_modules -maxdepth 1 -type l -exec basename {} \;'`

timoxley (approved):
Nice output, might consider making it an npm script with `lerna exec`, but either way lgtm. 👍

show-links.sh
harbu: done

show-links.sh
harbu: Running that command gets me the entire list of symlinked packages but without the associated "title", i.e.:

```
streamr-test-utils
streamr-client-protocol
streamr-test-utils
streamr-test-utils
streamr-network
streamr-test-utils
streamr-client-protocol
streamr-client
streamr-client
```

So not sure how I would use this, if you got some idea, feel free to edit.

PR #43, opened 2021-05-27
Disable GapFills if no stream storage assigned.

timoxley:
1. Adds `disable` function to `OrderingUtil` so in-progress gap filling can be cancelled.
2. Disables `OrderingUtil` if it detects a `NO_STORAGE_NODES` error during a gap fill.
3. Disables `OrderingUtil` if `orderMessages` client config is disabled. (fixes #NET-271)
4. Tests resends & resend + subscribe error if no storage node.

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -23,7 +23,7 @@
     "build-production": "npm run clean; npm run bootstrap-dist && npm run build-browser-production && npm run build-node && npm run build:types",
     "prepare": "npm run build",
     "vendor": "SRCDIR=`node -p \"path.dirname(require.resolve('quick-lru'))\"` bash -c 'mkdir -p vendor/quick-lru; cp -n $SRCDIR/index.d.ts vendor/quick-lru; cp -n $SRCDIR/index.js vendor/quick-lru'; true",
-    "copy-package": "mkdir -p dist/; node copy-package.js; cp -f package-lock.json dist; cp -f README.md dist; cp -f LICENSE dist; cp -f readme-header-img.png dist; true",
+    "copy-package": "mkdir -p dist/; node copy-package.js; cp -f package-lock.json dist; cp -f README.md dist; cp -f LICENSE dist; cp -f readme-header-img.png dist; mkdir -p dist/src/stream/encryption/migrations; cp -f ./src/stream/encryption/migrations/* dist/src/stream/encryption/migrations; true",
     "bootstrap-dist": "npm run vendor && npm run copy-package && npm run fix-esm",
     "fix-esm": "mkdir -p dist/src/; cp -Rf src/index-esm.mjs dist/src/index-esm.mjs",
     "build:types": "tsc --incremental --emitDeclarationOnly --stripInternal",


packages/client/src/subscribe/OrderMessages.js
```diff
@@ -16,8 +16,9 @@ let ID = 0
  */
 
 export default function OrderMessages(client, options = {}) {
-    const { gapFillTimeout, retryResendAfter, maxGapRequests } = client.options
+    const { gapFillTimeout, retryResendAfter, maxGapRequests, orderMessages } = client.options
     const { streamId, streamPartition, gapFill = true } = validateOptions(options)
+    let enabled = !!(orderMessages && gapFill && maxGapRequests)
```
harbu, line 21: About this condition. Do we differentiate between ordering and gap filling or is it an all on/off choice?

packages/client/src/subscribe/OrderMessages.js
```diff
@@ -55,11 +55,22 @@ export default function OrderMessages(client, options = {}) {
                 if (done) { return }
                 orderingUtil.add(streamMessage)
             }
+        } catch (err) {
+            if (done) { return }
+
+            if (err.code === 'NO_STORAGE_NODES') {
```
harbu, line 61: Do I remember correctly that checking this beforehand doesn't make sense? I.e. it's better to wait for this error to occur than to check for it before commencing OrderingUtil?

packages/client/src/subscribe/index.ts
```diff
@@ -646,11 +646,15 @@ export class Subscriber {
...
+        async function end(optionalErr?: Error) {
+            const tasks = [
                 resendMessageStream.cancel(optionalErr),
                 realtimeMessageStream.cancel(optionalErr),
-            ])
+                resendSubscribeSub.cancel(optionalErr)
+            ]
+
+            await Promise.allSettled(tasks)
+            await Promise.all(tasks)
```
harbu, line 657: Not sure I understand this pattern of doing first `allSettled` followed by `all`.

packages/client/src/subscribe/resendStream.ts
```diff
@@ -13,10 +14,11 @@ const { ControlMessage } = ControlLayer
  * Sends resend request, handles responses.
  */
 
-export default function resendStream(client, opts = {}, onFinally = async () => {}) {
+export default function resendStream(client: StreamrClient, opts = {}, onFinally = async () => {}) {
     const options = validateOptions(opts)
     const { connection } = client
     const requestId = counterId(`${options.key}-resend`)
+    // @ts-expect-error
```
harbu, line 21: It can be helpful to write after the @ts-expect-error directive why this is being used.

harbu (approved):
I'm not too familiar with this part of the code base, but overall LGTM

packages/protocol/test/unit/utils/OrderedMsgChain.test.ts
```diff
@@ -272,6 +272,55 @@ describe('OrderedMsgChain', () => {
         util.add(msg3)
     })
 
+    it('does not call the gap handler again if disabled', (done) => {
+        let counter = 0
+        const msgs: StreamMessage[] = []
+        util = new OrderedMsgChain('publisherId', 'msgChainId', (msg) => {
+            msgs.push(msg)
```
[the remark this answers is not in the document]
timoxley: yeah you're right that would be better

packages/client/src/subscribe/OrderMessages.js
timoxley: no differentiation, but it may actually be useful to differentiate.

packages/client/src/subscribe/OrderMessages.js
timoxley: I mean, either way it's a round trip to the server so might as well *try* i.e. consider this as the check

packages/client/src/subscribe/index.ts
timoxley: https://twitter.com/timkevinoxley/status/1389940552209256448

packages/client/src/subscribe/resendStream.ts
timoxley: you're right

PR #44, opened 2021-05-28
NET-328: upgrade client version in broker to 5.4.2 series

harbu:
First the simple part: `client.getSessionToken()` is no longer marked as internal, thus associated ts-expect-error comments need be removed.

The more complex part: client now enforces the `storageNode` configuration that it is given. Because broker nodes with metrics enabled use the client to create streams and assign them to storage, these values must be passed from the broker to the client. This means that the passing a `storageNodeRegistry` to the broker is no longer optional but mandatory. (*)

(*) Technically it would be possible to make it so that if metrics are disabled, we allow setting `storageNodeRegistry` to null. However, I don't think we should add more exceptions.


teogeb (approved):
LGTM, could remove null casts in `WebsocketPlugin.ts` (line 46) and `resends-cancelled-on-client-disconnect.test.ts` (line 119).

PR #46, opened 2021-05-28
cli-tools: add command for fetching sessionToken

harbu:
Now that it is public and proper, add command ~`streamr misc fetch-token`~ `streamr misc get-session-token` to ~fetch~ get a session token for the associated wallet.

### DIFF
--- a/packages/cli-tools/bin/streamr-misc.ts
+++ b/packages/cli-tools/bin/streamr-misc.ts
@@ -6,5 +6,5 @@ program
     .version(pkg.version)
     .usage('<command> [<args>]')
     .description('miscellaneous subcommands')
-    .command('fetch-token', 'fetch a session token for the current user')
+    .command('get-session-token', 'get a session token for the current user')
     .parse(process.argv)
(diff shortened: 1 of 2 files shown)


packages/cli-tools/bin/streamr-misc.ts
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env node
+import { program } from 'commander'
+import pkg from '../package.json'
+
+program
+    .version(pkg.version)
+    .usage('<command> [<args>]')
+    .description('miscellaneous subcommands')
+    .command('fetch-token', 'fetch a session token for the current user')
```
teogeb, line 9: How about renaming this to `get-session-token`? Would be in line with the method name. Maybe "token" is not specific enough in a crypto tool.

harbu: Good point, token is a quite overloaded term so better to use `session-token`. And `fetch-session-token` is already pretty long so `get-session-token` better.

harbu:
@teogeb request re-review. Renamed command to `get-session-token`

teogeb (approved):

PR #47, opened 2021-05-31
API key authentication in HTTP server

teogeb:
HTTP server middleware for `apiAuthenticator`. 

If the config lists some API keys in (in `apiAuthentication` block), all requests to httpServer need to be authenticated. The key is provided in the  HTTP header: `Authorization: Bearer <apiKey>`

### DIFF
--- a/packages/broker/test/unit/httpServer.test.ts
+++ b/packages/broker/test/unit/httpServer.test.ts
@@ -68,7 +68,7 @@ describe('HttpServer', () => {
         })
 
         it('unauthorized', async () => {
-            const server = await startTestServer({
+            server = await startTestServer({
                 keys: [MOCK_API_KEY]
             })           
             const response = await createRequest()


packages/broker/test/unit/httpServer.test.ts
```diff
@@ -0,0 +1,75 @@
...
+describe('HttpServer', () => {
+
+    describe('ApiAuthenticator', () => {
+        
+        it('no authentication required', async () => {
+            const server = await startTestServer(null)
+            const response = await createRequest()
+            const body = await response.text()
+            expect(body).toBe('FOO')
+            await stopServer(server)
```
harbu, line 38: `stopServer` might not get invoked if something fails before, leading to subsequent tests in this test suite to all fail. Wrap with `try-finally` or move to `afterEach`?

harbu (changes requested):
LGTM; see comment on test

packages/broker/test/unit/httpServer.test.ts
teogeb: Fixed

packages/broker/test/unit/httpServer.test.ts
```diff
@@ -0,0 +1,84 @@
...
+        it('valid authentication', async () => {
+            const server = await startTestServer({
+                keys: [MOCK_API_KEY]
+            })
+            try {
+                const response = await createRequest({
+                    Authorization: `Bearer ${MOCK_API_KEY}`
+                })
+                const body = await response.text()
+                expect(body).toBe('FOO')
```
harbu, line 50: I belive there is a possibility that these `expect` blocks never get executed yet the test will still show green. I think you need to add a catch into these

```ts
try {
  ...
  expect(...).toEqual(...)
} catch (e) {
    fail(`failed due to ${e}`)
} finally {
   await stopServer(server)
}
```

teogeb: Refactored to use `afterEach`

packages/broker/test/unit/httpServer.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    })
+
+    describe('ApiAuthenticator', () => {
+        
+        it('no authentication required', async () => {
+            server = await startTestServer(null)
+            const response = await createRequest()
+            const body = await response.text()
+            expect(body).toBe('FOO')
+            await stopServer(server)
```
harbu, line 47:
```suggestion
```

packages/broker/test/unit/httpServer.test.ts
```diff
@@ -0,0 +1,81 @@
...
+                keys: [MOCK_API_KEY]
+            })
+            const response = await createRequest({
+                Authorization: 'Bearer invalid-api-key'
+            })
+            expect(response.status).toBe(403)
+        })
+
+        it('unauthorized', async () => {
+            const server = await startTestServer({
```
harbu, line 72:
```suggestion
            server = await startTestServer({
```

harbu (approved):

PR #49, opened 2021-05-31
NET-245: Trusted publishHttp plugin

teogeb:
New `publishHttp` plugin which uses a trusted `StreamrClient`.

Usage:
```
curl \
  --header 'Content-Type: application/json' \
  --request POST \
  --data '{"foo":"bar"}' \
  http://localhost:8891/streams/some-stream-id
```

**Open question**: does not support msgChainId query parameter, could `StreamrClient.publish` support that parameter?

**Breaking change**: renamed the deprecated `publishHttp` plugin to `legacyPublishHttp`.

### DIFF
--- a/packages/broker/src/plugins/publishHttp/publishEndpoint.ts
+++ b/packages/broker/src/plugins/publishHttp/publishEndpoint.ts
@@ -17,7 +17,7 @@ const parseContent = (req: Request) => {
     }
 }
 
-export const createEndpoint = (streamrClient: StreamrClient) => {
+export const createEndpoint = (streamrClient: StreamrClient): express.Router => {
     const router = express.Router()
     router.use(express.raw({
         limit: '1024kb',
--- a/packages/broker/test/integration/plugins/legacyPublishHttp/broker-resistance-to-invalid-data.test.ts
+++ b/packages/broker/test/integration/plugins/legacyPublishHttp/broker-resistance-to-invalid-data.test.ts
@@ -42,7 +42,7 @@ describe('broker resistance to invalid data', () => {
         await tracker.stop()
     })
 
-    test('pushing invalid data returns 400 error & does not crash broker', (done) => {
+    test('pushing invalid data to legacy HTTP plugin returns 400 error & does not crash broker', (done) => {
         const invalidData = '###!!THIS-DATA-IS-NOT-JSON!!###'
 
         const request = http.request({


packages/broker/src/helpers/parser.ts
```diff
@@ -0,0 +1,35 @@
+import { Request } from 'express'
+
+export const parsePositiveInteger = (n: string) => {
```
harbu, line 3:
```suggestion
export const parsePositiveInteger = (n: string): number | never => {
```

packages/broker/src/helpers/parser.ts
```diff
@@ -0,0 +1,35 @@
...
+    const parsed = parseInt(n)
+    if (!Number.isInteger(parsed) || parsed < 0) {
+        throw new Error(`${n} is not a valid positive integer`)
+    }
+    return parsed
+}
+
+export const parseTimestamp = (millisOrString: number|string) => {
```
harbu, line 11:
```suggestion
export const parseTimestamp = (millisOrString: number|string): number|never => {
```

packages/broker/src/helpers/parser.ts
```diff
@@ -0,0 +1,35 @@
...
+            throw new Error(`Invalid timestamp: ${millisOrString}`)
+        } else {
+            return timestamp
+        }
+    } else {
+        throw new Error(`Invalid timestamp: ${millisOrString}`)
+    }
+}
+
+export const getQueryParameter = (name: string, req: Request, parser?: (input: string) => any) => {
```
harbu, line 28: (This change might break compilation)
```suggestion
export const getQueryParameter = <T>(name: string, req: Request, parser?: (input: string) => T): T | string | undefined => {
```

packages/broker/test/integration/plugins/legacyPublishHttp/broker-resistance-to-invalid-data.test.ts
```diff
@@ -43,7 +43,7 @@ describe('broker resistance to invalid data', () => {
         await tracker.stop()
     })
 
-    test('pushing invalid data to HTTP plugin returns 400 error & does not crash broker', (done) => {
+    test('pushing invalid data returns 400 error & does not crash broker', (done) => {
```
harbu, line 46:
```suggestion
    test('pushing invalid data to legacy HTTP plugin returns 400 error & does not crash broker', (done) => {
```

packages/broker/src/plugins/publishHttp/publishEndpoint.ts
```diff
@@ -0,0 +1,62 @@
...
+    }
+    const contentAsString = req.body.toString()
+    try {
+        return JSON.parse(contentAsString)
+    } catch (e) {
+        throw new Error('Request body is not a JSON string')
+    }
+}
+
+export const createEndpoint = (streamrClient: StreamrClient) => {
```
harbu, line 20:
```suggestion
export const createEndpoint = (streamrClient: StreamrClient): express.Router => {
```

packages/broker/src/plugins/publishHttp/publishEndpoint.ts
```diff
@@ -0,0 +1,62 @@
...
+            timestamp = getQueryParameter('timestamp', req, parseTimestamp)
+            partition = getQueryParameter('partition', req, parsePositiveInteger)
+            partitionKey = getQueryParameter('partitionKey', req)
+        } catch (e) {
+            res.status(400).send({
+                error: e.message
+            })
+            return
+        }
+        if ((partition !== undefined) && (partitionKey !== undefined)) {
```
harbu, line 42: :+1:

packages/broker/src/plugins/publishHttp/publishEndpoint.ts
```diff
@@ -0,0 +1,62 @@
...
+            })
+            return
+        }
+        const streamId = req.params.streamId as string
+        const streamPartDefinition = {
+            streamId,
+            streamPartition: partition
+        }
+        try {
+            await streamrClient.publish(streamPartDefinition, content, timestamp, partitionKey)
```
harbu, line 54: Does this work alright when `parameter === undefined` and `partitionKey !== undefined`? It should use the `partitionKey` to figure out a partition in that case.

packages/broker/test/unit/plugins/publishHttp/publishEndpoint.test.ts
```diff
@@ -0,0 +1,127 @@
+import express from 'express'
+import request from 'supertest'
+import { createEndpoint } from '../../../../src/plugins/publishHttp/publishEndpoint'
+
+const MOCK_STREAM_ID = 'mock-stream-id'
+
+describe('PublishEndpoint', () => {
```
harbu, line 7: kudos, very readable tests :+1:

harbu (approved):
LGTM! Mostly I'd add some return types to exported functions / methods. Question about how streamr-client works when given both partition and partitionKey

packages/broker/src/helpers/parser.ts
teogeb: Fixed, removed use case where parser is not defined

packages/broker/src/plugins/publishHttp/publishEndpoint.ts
teogeb: Yes, it does that

PR #52, opened 2021-06-01
fix: make per-node-metrics.test.ts faster

harbu:
43 seconds to 13 seconds on my machine, doesn't in any significant way alter what is getting tested


packages/broker/test/integration/per-node-metrics.test.ts
```diff
@@ -119,9 +120,12 @@ describe('metricsStream', () => {
         })
 
         client2 = createClient(wsPort, tmpAccount.privateKey)
+        await fillMetrics(client2, 60, nodeAddress, 'sec')
```
teogeb, line 123: We could run the three `fillMetrics` calls concurrently?

teogeb (approved):

packages/broker/test/integration/per-node-metrics.test.ts
```diff
@@ -119,9 +120,12 @@ describe('metricsStream', () => {
...
+        await fillMetrics(client2, 60, nodeAddress, 'min')
+        await fillMetrics(client2, 24, nodeAddress, 'hour')
```
harbu, line 125:
```suggestion
        await Promise.all([
            fillMetrics(client2, 60, nodeAddress, 'sec'),
            fillMetrics(client2, 60, nodeAddress, 'min'),
            fillMetrics(client2, 24, nodeAddress, 'hour'),
        ])
            
```

PR #56, opened 2021-06-01
ci: Run all test workflows for every PR.

timoxley:
Discussed with @harbu perhaps turning this on so we can prevent any accidental breakage before a PR hits `main`.

Currently if there are test failures on `main` it's not entirely clear where the failure was introduced, especially if there are multiple PRs merging into `main` before `main` knows if it's CI is green or not.

Also split the slow (30 minute) data union tests into their own workflow so it's more clear when we're just waiting for data unions tests to finish vs rest of client integration tests.

### DIFF
--- a/.github/workflows/client-code.yml
+++ b/.github/workflows/client-code.yml
@@ -98,20 +98,12 @@ jobs:
         test-name: [
             "test-integration-no-resend",
             "test-integration-resend",
-            "test-integration-dataunions",
         ]
         websocket-url:
             - name: "default"
               url:  ""
             - name: "single-node-only"
               url:  "ws://localhost:8690/api/v1/ws"
-
-        exclude:
-          # no need to test different ws urls for dataunion tests
-          - test-name: "test-integration-dataunions"
-            websocket-url:
-                - name: "single-node-only"
-                - url:  "ws://localhost:8690/api/v1/ws"
     env:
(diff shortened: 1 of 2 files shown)


harbu (approved):
Firstly, I'm definitely on board of separating data union tests to their own workflow.

Secondly, for pull requests:

I would keep packages `test-utils` and `protocol` as is, i.e. only touches to their code (or infrastructure code) should cause their respective workflows to run.

The rest perhaps as follows:
- `network` workflow gets run whenever it, `protocol`, or `test-utils` is touched.
- `broker` workflow gets run whenever it, `network`, `protocol` or `test-utils` is touched.
- `client` workflow gets run whenever it, `broker`, `network`, `protocol`, or `test-utils` is touched.

Arguably the above arrangement may become hard to reason about. That being said, I'm slightly worried about wait times here and whether by reserving more jobs it takes even longer for them to finish (since there is a limit to how many concurrent jobs you can have in Github actions?)

With all that being said, I'm willing to try this out and see how it works in practice. Perhaps the aforementioned worries do not come into fruition in practice. If such were the case, I think there would be value in keeping things simple even though it means some extra jobs are run that aren't necessary.

As we discussed it would be great that the `broker` docker image would get built from source and then used in client tests. But that could be a separate PR.

teogeb (approved):

PR #59, opened 2021-06-03
Broker NET-232: Cassandra error handling

teogeb:
Improved error handling of Cassandra errors. Cassandra query errors are propagated to the  returned `Readable` stream as error events. 

Previously if an error occurred, we just closed the stream (`stream.push(null)`) and ignored the error. Now we call `stream.destroy(error)` instead. That call emits the error and closes the stream.

To enable the propagation, stream piping now uses the standard `pipeline` method instead of the `pump` library.

Modified also behavior of empty results: as empty result is a valid output, we must not throw  error for such result. Now we just close the stream in that case.

Storage is used from `DataQueryEndpoint`. If an error occurs, the error is logged there. Therefore duplicate error logging from `Storage` and `BucketManager` was removed.

Added unit tests for storage queries.

To test the error handling manually, do e.g. the following:

1) start broker
2) create stream
3) add stream to a storage node
4) stop cassandra
5) do a resend request from `cli-tools` -> receives `STORAGE_NODE_FETCH_ERROR`, and error is logged in Broker.


packages/broker/test/unit/plugins/storage/Storage.test.ts
```diff
@@ -0,0 +1,255 @@
...
+                messageCounts: [MOCK_MESSAGE_COUNT],
+                messages: [MOCK_MESSAGE_2]
+            }))
+            const resultStream = storage.requestLast(MOCK_STREAM_ID, 0, 1)
+            const onData = jest.fn()
+            resultStream.on('data', onData)
+            resultStream.on('end', () => {
+                expect(onData).toBeCalledTimes(1)
+                expect(onData.mock.calls[0][0].serialize()).toEqual(MOCK_MESSAGE_2.payload)
+                done()
```
harbu, line 120: There's a helper for these sorts of things in `streamr-test-utils`, see this one https://github.com/streamr-dev/streamr-test-utils/blob/master/src/utils.ts#L16-L24. Might or might not be applicable to your use cases here

packages/broker/test/unit/plugins/storage/Storage.test.ts
```diff
@@ -0,0 +1,255 @@
...
+        })
+
+    })
+
+    describe.each([
+        [REQUEST_TYPE_FROM, true],
+        [REQUEST_TYPE_FROM, false],
+        [REQUEST_TYPE_RANGE, true],
+        [REQUEST_TYPE_RANGE, false],
+    ])('%s, publisher: %p', (requestType: string, isPublisher: boolean) => {
```
harbu, line 183: Got a bit confused here while reading variable `isPublisher`. First thought was whether someone is gonna start publishing messages here or something. It got cleared up thought as I read more. Think it is fine as is, but wondering if it was more clear just to define the actual `publisherId` value in the describe table (even though changing it to be different streams is not technically needed)?

packages/broker/test/unit/plugins/storage/Storage.test.ts
```diff
@@ -0,0 +1,255 @@
...
+        } else if (query.includes('payload FROM stream_data')) {
+            return messages!
+        } else {
+            throw new Error(`Assertion failed: query="${query}"`)
+        }
+    }
+}
+
+const createMockStorage = (getResults: (query: string) => any[]|Error) => {
+    const cassandraClient = {
```
harbu, line 53: This may be more in the lines of theoretical rambling, so take it as it as you will. 😄 

Given that you're mocking here, there could be a slight benefit of trying to have this variable be assignable (compatible with?) to `CassandraClient` interface / class as exposed by the library (as opposed to using `as any`). I.e.

```ts
let foo: cassandra.CassandraClientOrWhateverItIsCalled = cassandraClient // compiles
```

Let's say we decide to update the Cassandra driver and it contains a breaking change to `CassandraClient` that we rely on in our code. This test will still be green since we are testing against a mocked version of `CassandraClient`. However, if 
- (a) the implementation here would be assignable to `CassandraClient`
- (b) the breaking change came in the form of a breaking _type signature_ change

then there _could be a chance_ that the code here would not compile because our mocked version is not compatbile with the new `CassandraClient`.

This is of course quite theoretical, and not sure you can even do this in TypeScript. If it were doable without having to write boilerplate implementations of methods, i.e., the code snippet I pasted above "just worked" then could be interesting to try out 🤔

harbu (approved):
LGTM!

packages/broker/test/unit/plugins/storage/Storage.test.ts
```diff
@@ -0,0 +1,255 @@
...
+        it('happy path', (done) => {
+            storage = createMockStorage(createResultFactory({
+                buckets: [MOCK_BUCKET],
+                messageCounts: [MOCK_MESSAGE_COUNT],
+                messages: [MOCK_MESSAGE_2]
+            }))
+            const resultStream = storage.requestLast(MOCK_STREAM_ID, 0, 1)
+            const onData = jest.fn()
+            resultStream.on('data', onData)
+            resultStream.on('end', () => {
```
timoxley, line 117: IMO `done` callback should be avoided at all costs. Typically you end up missing code paths (e.g. error events) meaning if things don't occur as expected the test will just hang.

In saying that, working with async/await and streams is a bit clunky, but you can do it like this using `import { once } from 'events'`: 
```js
const results = []
resultStream.on('data', (results) => results.push(data))
await once(resultStream, 'end') // also attaches error handler
expect(results.map((r) => r.serialize())).toEqual([MOCK_MESSAGE_2.payload])
```

packages/broker/test/unit/plugins/storage/Storage.test.ts
timoxley: The mock also has different async behaviour to the real Cassandra client.

 e.g.
 * `eachRow` here executes synchronously while it certainly doesn't with the real client, 
  * same with the `stream` implementation, the entire stream is created, filled, and ended synchronously
  * and maybe less of an issue but the mock `execute` promise will resolve in the next microtask, between `nextTick` and `setImmediate`, while in reality it's likely to always be some milliseconds later (after both `nextTick` and `setImmediate`).

Might not be an issue but having async behaviour that differs from the real implementation seems like a good way to hide (or even create) subtle async bugs during development.

IMO this should use a real Cassandra client unless there's some critical reason it can't.

timoxley (approved):
wow big improvement, great work on the streams.
I don't think mocking the cassandra client is at all a good idea, but otherwise all this lgtm.

packages/broker/test/unit/plugins/storage/Storage.test.ts
```diff
@@ -0,0 +1,255 @@
...
+    }
+    return new Storage(cassandraClient as any, {})
+}
+
+describe('Storage', () => {
+
+    let storage: Storage
+
+    afterEach(async () => {
+        await storage!.close()
```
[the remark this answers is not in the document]
teogeb: Fixed

packages/broker/test/unit/plugins/storage/Storage.test.ts
teogeb: Converted tests to use waitForEvent and waitForStreamToEnd. Improved readability 👍

packages/broker/test/unit/plugins/storage/Storage.test.ts
teogeb: Fixed

packages/broker/test/unit/plugins/storage/Storage.test.ts
teogeb: Using waitForEvent and waitForStreamToEnd (see @harbu's suggestion)

packages/broker/src/plugins/storage/Storage.ts
```diff
@@ -391,21 +388,21 @@ export class Storage extends EventEmitter {
             const stream2 = this._queryWithStreamingResults(query2, queryParams2)
             const stream3 = this._queryWithStreamingResults(query3, queryParams3)
 
-            return pump(
+            return pipeline(
```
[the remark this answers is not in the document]
timoxley: nvm, I forgot `pipeline` doesn't create a new stream in this case, i.e. `pipeline(src, dest, cb) === dest`

packages/broker/test/unit/plugins/storage/Storage.test.ts
teogeb: Need to have a client where it is easy to trigger all kinds of errors (promise rejections, `Readables` that emit error events). Easier with a mock instance compared to the real client. Agree that now we mock so much of Cassandra client that we don't really test the Cassandra interaction. But the tests validate our own query logic ok.

teogeb: Based on @timoxley's suggestions, refactored the test to use the real Cassandra client for non-error queries. Now it covers more Cassandra logic (and caught a bug https://github.com/streamr-dev/network-monorepo/pull/62). 

It would have been very convenient to implement the real vs. mock Cassandra client using jest `spy`. But as we needed to filter the success/error case by the query types (the parameter of a called method), it was more straightforward to implement the logic by a separate `ProxyClient` class. The class routes the query to the real client, or throws an `Error` if the test setup has called `setError(queryId)` for that query.

teogeb:
The tests will pass once https://github.com/streamr-dev/network-monorepo/pull/62 has been merged to main

harbu (approved):
LGTM

PR #61, opened 2021-06-04
Don't suggest `npm i`

jtakalai:
instead suggest `npm ci`


harbu:
We should probably support both tbh

timoxley (commented):
Correct

PR #64, opened 2021-06-04
NET-249: Read storage nodes from StorageNodeRegistry smart contract

juslesan:
- Fetch storage node information from Storage Node Registry smart contract
- Main functionality in protocol for reuseability


juslesan:
Any opinions on if the production RPC and Smart Contract address should be used as default values somewhere?

packages/broker/src/Plugin.ts
```diff
@@ -17,6 +17,7 @@ export interface PluginOptions {
     apiAuthenticator: ApiAuthenticator
     metricsContext: MetricsContext
     brokerConfig: Config
+    storageNodeRegistry: StorageNodeRegistryItem[]
```
harbu, line 20: Would it make sense for the `broker.ts`  main script to create an instance of `StorageNodeRegistry` and then pass the shared instance to the plugins instead of having each plugin create their own instance. It should work nicely since it is a rather immutable class. Also ping @teogeb .

packages/broker/src/config.ts
```diff
@@ -1,4 +1,4 @@
-export interface TrackerRegistry {
+export interface NetworkSmartContractRegistry {
     registryAddress: string
     jsonRpcProvider: string
 }
```
harbu, line 4: I'd make it even more generic. The nice thing about renaming `registryAddress => contractAddress` is that it matches the name of the protocol util functions that fetch info from smart contracts.
```suggestion
export interface NetworkSmartContract {
    contractAddress: string
    jsonRpcProvider: string
}
```

packages/broker/src/helpers/config.schema.json
```diff
@@ -263,10 +263,46 @@
       "description": "Ethereum address Core (E&E)",
       "pattern": "^0x[a-fA-F0-9]{40}$"
     },
-    "storageNodeRegistry": {
-      "type": "array",
-      "items": {
-        "$ref": "#/definitions/storageNodeRegistryItem"
+    "storageNodeConfig": {
+      "type": "object",
+      "description": "Network node settings",
```
harbu, line 268:
```suggestion
      "description": "Storage node settings",
```

packages/broker/src/helpers/config.schema.json
```diff
@@ -263,10 +263,46 @@
...
+                  "type": "string",
+                  "description": "Ethereum address of tracker registry smart contract",
+                  "pattern": "^0x[a-fA-F0-9]{40}$"
+                },
+                "jsonRpcProvider": {
+                  "type": "string",
+                  "description": "URL for JSON RPC Provider",
+                  "format": "uri"
+                }
+              }
```
harbu, line 302: We should probably make this into a shared JSON schema definition under key "definitions" so that the same validation rule can be used for the tracker registry and this storage node config use case. This would be in line with how the `config.ts` file treats them the same.

harbu (approved):
LGTM! Two code organizational things: 

a. pass shared `StorageNodeRegistry` instance to plugins to make plugins simpler? 
b. possibility of code re-use w.r.t. validation and passing smart contract registry details

packages/broker/configs/docker-1.env.json
```diff
@@ -23,8 +23,12 @@
...
   "streamrAddress": "0xFCAd0B19bB29D4674531d6f115237E16AfCE377c",
-  "storageNodeRegistry": null,
-  "httpServer": {
+  "storageNodeConfig": {
+    "registry": {
+      "contractAddress": "0xEAA002f7Dc60178B6103f8617Be45a9D3df659B6",
+      "jsonRpcProvider": "http://10.200.10.1:8546"
+    }
+  },  "httpServer": {
     "port": 8891
```
harbu, line 32:
```suggestion
  }, 
  "httpServer": {
    "port": 8891
```

harbu (approved):
lgtm

packages/broker/package.json
```diff
@@ -63,7 +63,7 @@
     "qs": "^6.10.1",
     "split2": "^3.2.2",
     "streamr-client": "^5.4.2",
-    "streamr-network": "^24.3.3",
+    "streamr-network": "^24.3.4-alpha.0",
```
timoxley, line 66: Yeah I think we can just leave the alphas versions off, doesn't help a lot and creates a bunch of busywork. Point at whatever version is being used in the local repo.

packages/network/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "streamr-network",
-  "version": "24.3.3",
+  "version": "24.3.4-alpha.0",
```
timoxley, line 3: Should probably rebase or merge `main`

timoxley (approved):
lgtm

PR #65, opened 2021-06-04
Show linked/unlinked package versions

timoxley:
Lists internal packages & their versions on the left.
Linked packages are columns. 
If the package on the left links to the package in the column, it shows a checkmark & the semver range, otherwise it shows the mismatched semver range and prints a warning at the end.

It prints the version ranges so you can double-check that they're formatted as you expect e.g. `^X.Y.Z` vs `X.Y.Z`

image

### Warning Example

Here I changed the network's `package.json` to point at `"streamr-client-protocol": "^8.2.1",`:

image


show-versions.mjs
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env npx zx
+
+import { join } from 'path'
+import semver from 'semver'
+import Table from 'cli-table'
+$.verbose = false
+
+const packages = JSON.parse(await $`lerna list --all --json --loglevel=silent --toposort`)
```
harbu, line 8: Nice way of utilizing `--json` flag 👍

harbu (approved):
Really really nice output, cheers! This is exactly the tool I was looking for 👍

show-versions.mjs
timoxley: Note the silly thing needed `--loglevel=silent` to stop it from spitting out non-json logs.

juslesan (approved):
Good stuff 👍

harbu:
@timoxley remove show-links.sh and add into README section about using this tool?

timoxley:
@harbu Done https://github.com/streamr-dev/network-monorepo/pull/65/commits/7fee7fda290af08b3c37c618461efd4fab6f0b51

PR #66, opened 2021-06-07
Fix broker CI

timoxley:
Something between these commits broke broker's CI

https://github.com/streamr-dev/network-monorepo/compare/aed4a19d251957ce26f6427d466559efa5b3902a...edd163c77a63df487f5c90d409df26df8f14326c

Looks like only package-lock and version number changes though.
The package-lock changes seem the most likely culprit.

### DIFF
--- a/packages/broker/src/plugins/storage/Storage.ts
+++ b/packages/broker/src/plugins/storage/Storage.ts
@@ -436,7 +436,9 @@ export class Storage extends EventEmitter {
     private queryWithStreamingResults(query: string, queryParams: any[]) {
         return this.cassandraClient.stream(query, queryParams, {
             prepare: true,
-            autoPage: false
+            // force small page sizes, otherwise gives RangeError [ERR_OUT_OF_RANGE]: The value of "offset" is out of range.
+            fetchSize: 128,
+            readTimeout: 0,
         }) as Readable
     }
 
--- a/packages/broker/test/integration/plugins/storage/Storage.test.ts
+++ b/packages/broker/test/integration/plugins/storage/Storage.test.ts
@@ -1,3 +1,5 @@
+import { randomFillSync } from 'crypto'
+
 import { Client } from 'cassandra-driver'
 import toArray from 'stream-to-array'
 import { Protocol } from 'streamr-network'
@@ -311,6 +313,7 @@ describe('Storage', () => {
     describe('fast big stream', () => {
         let storedStreamId: string
         const NUM_MESSAGES = 1000
+        const MESSAGE_SIZE = 1e3 // 1k
 
         beforeEach(async () => {
             // slow message setup: run this once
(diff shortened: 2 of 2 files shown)


packages/broker/src/plugins/storage/Storage.ts
```diff
@@ -111,22 +111,22 @@ export class Storage extends EventEmitter {
...
+                    fetchSize: 0 // disable paging
+                })
                 resultSet.rows.reverse().forEach((r: Todo) => {
                     resultStream.write(r)
                 })
-                resultStream.push(null)
-            }).catch((e: Todo) => {
-                resultStream.destroy(e)
-            })
+                resultStream.end()
```
timoxley, line 126: `push` is technically supposed to only be used internal to the stream.

`resultStream.push(null)` pushes null into the readable part of the stream.
`resultStream.write(r)` pushes the value into the writable
In a transform stream, data flows from `Writable` -> `transform(chunk)` -> `Readable`. 

By pushing `null` into the `Readable` side:
* the `null` skips the `transform` function
* it's possible data is still buffered on the `Writable` side. Pushing `null` to the readable side doesn't close the writable. Data from the writable will keep flowing through the `transform()`, but will error with `ERR_STREAM_PUSH_AFTER_EOF` if you don't check `if (this.readableEnded)` before calling `this.push` in the `transform()`.

This might not have broken things before, or maybe it did it's a bit subtle, but it matters now because every `resultStream` is now a `Transform` that implements the repeated "yield to event loop" thing.

packages/broker/src/plugins/storage/Storage.ts
```diff
@@ -277,35 +278,51 @@ export class Storage extends EventEmitter {
...
+                    highWaterMark: 16,
+                    pipeError: true,
                 }),
                 resultStream,
                 (err: Todo) => {
-                    if (err) {
-                        resultStream.destroy(err)
-                    }
+                    resultStream.destroy(err || undefined)
+                    stream1.destroy(err || undefined)
```
timoxley, line 324: I'm not 100% sure these other destroy calls are necessary, have left them in as it's probably better to explicitly clean up than to not?

packages/broker/src/plugins/storage/Storage.ts
```diff
@@ -351,19 +367,40 @@ export class Storage extends EventEmitter {
...
+            'WHERE stream_id = ? AND partition = ? AND bucket_id IN ?',
+            'AND ts > ? AND ts < ? AND publisher_id = ? AND msg_chain_id = ?',
+            'ALLOW FILTERING'
+        ].join(' ')
+        const query3 = [
+            'SELECT payload FROM stream_data',
+            'WHERE stream_id = ? AND partition = ? AND bucket_id IN ? AND ts = ?',
+            'AND sequence_no <= ? AND publisher_id = ?  AND msg_chain_id = ?',
+            'ALLOW FILTERING'
+        ].join(' ')
```
timoxley, line 399: These changes are just for line length.

packages/broker/src/plugins/storage/Storage.ts
```diff
@@ -277,35 +278,51 @@ export class Storage extends EventEmitter {
...
-            const stream2 = this._queryWithStreamingResults(query2, queryParams2)
+            const stream1 = this.queryWithStreamingResults(query1, queryParams1)
+            const stream2 = this.queryWithStreamingResults(query2, queryParams2)
 
             return pipeline(
-                // @ts-expect-error TODO is cassandraStream EventEmitter or ReadableStream?
                 merge2(stream1, stream2, {
-                    pipeError: true
+                    // @ts-expect-error options not in type
+                    highWaterMark: 16,
```
timoxley, line 318: TODO: double check whether this affects anything

packages/broker/test/integration/plugins/storage/Storage.test.ts
```diff
@@ -297,7 +314,7 @@ describe('Storage', () => {
 
             const ts = await storage.getFirstMessageTimestampInStream(streamId, 10)
 
-            expect(ts.getTime()).toEqual(2000)
+            expect(ts).toEqual(2000)
```
timoxley, line 317: See comments for `Storage.ts`

packages/broker/src/plugins/storage/Storage.ts
```diff
@@ -396,45 +435,44 @@ export class Storage extends EventEmitter {
...
+                    })
+                } else {
+                    done()
+                }
             }
         })
     }
 
-    async getFirstMessageTimestampInStream(streamId: string, partition: number) {
+    async getFirstMessageTimestampInStream(streamId: string, partition: number): Promise<number> {
```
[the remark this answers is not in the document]
harbu: lgtm, we can tackle any arising issue separately

packages/broker/test/integration/plugins/storage/Storage.test.ts
```diff
@@ -265,59 +310,100 @@ describe('Storage', () => {
         expect(results3).toEqual([])
     }, 20000)
 
-    test('fast big stream', async () => {
+    describe('fast big stream', () => {
+        let storedStreamId: string
+        const NUM_MESSAGES = 1000
+        const MESSAGE_SIZE = 1e3 // 1k
+
+        beforeEach(async () => {
```
harbu, line 318: Could use `beforeAll` here?

harbu (approved):

packages/broker/test/integration/plugins/storage/Storage.test.ts
timoxley: Tried that but `beforeAll` seems to run before any of the other `beforeEach` calls, so client etc aren't set up.

PR #70, opened 2021-06-08
network, broker [NET-339]: change log level on the fly (DRAFT PR)

harbu:
The ability to change logging level on the fly for instances of the `Logger` class.

I looked into the issue starting from here https://github.com/pinojs/express-pino-logger/issues/11 and it led me to this library https://github.com/pinojs/pino-arborsculpture. The only problem with this library is that (a) it reads the log level from a file and (b) it requires loggers to be passed to it upon creation. I want us to be able to change the log level by e.g. calling a HTTP endpoint. We also create new loggers at runtime so we can't pass them all at once. That being said, the library code provided examples of how to actually change the log level at runtime: https://github.com/pinojs/pino-arborsculpture/blob/master/sculpt.js#L15-L20.

I went with the following approach:
- We maintain a `static WeakSet<Logger>` of loggers who have their log level set to the current log level `static logLevel`.
- You can change the global log level by invoking static method `changeLogLevel(newLevel: string)`, this will  re-set `static logLevel` and clear `static WeakSet<Logger>` .
- Each time before logging, the `Logger` checks if it is part of the `static WeakSet<Logger>`, and if not, it updates its log level and add itself to the set. (*)
 
I don't particularly like that this code needs to perform a check (*) before each logging. However:
- Apparently the library does not keep track of created loggers and thus provide a method / function to get them all.
- It is not possible to iterate over a `WeakSet` thereby being able to do a sweep over `Logger`s upon log level change.
- Something like `Array<WeakRef<Logger>>` could allow for this, but seems like `WeakRef` is not still usable in our node.js version?
- Child loggers to the rescue? However sources seem to indicate the updates to log level do not get applied to child loggers after creation: https://github.com/pinojs/pino/issues/959,  https://github.com/pinojs/pino/issues/766.
- UPDATE: Child loggers with event handler: https://github.com/pinojs/pino/blob/master/docs/api.md#event-level-change. We could have the child loggers listen to this event on the root logger and update their state respectively? I.e. "propagating level changes". Ah but then there is the whole mess of cleaning up listeners etc.


timoxley:
@harbu I wonder if the answer is to use a log aggregating tool/service so we don't *need* to change log levels on the fly at all, and can use the aggregator to do filtering and achieve the same result.

harbu:
Closing this ticket, but leaving it to exist for now.

PR #73, opened 2021-06-09
Net-14: Resend logic when there are multiple storage nodes

juslesan:
- Fetch multiple storage nodes from Stream Registry
- Shuffle the array of storage node URLs when performing queries, if a queried storage node returns an error code, try the next storage node in the shuffled array

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,8 @@
         "bootstrap": "lerna bootstrap",
         "bootstrap-pkg": "npx lerna bootstrap --include-dependencies --scope",
         "versions": "node ./show-versions.mjs",
-        "clean": "npx lerna clean"
+        "clean": "npx lerna clean # remove all node_modules",
+        "clean-package-locks": "npx lerna exec --no-bail 'rm package-lock.json' # remove all package-lock files. npm run clean & npm run bootstrap to fully regenerate."
     },
     "dependencies": {
         "@commitlint/cli": "^12.1.4",
--- a/packages/broker/package.json
+++ b/packages/broker/package.json
@@ -46,7 +46,7 @@
     "ajv-formats": "^2.0.1",
     "arraybuffer-to-string": "^1.0.2",
     "body-parser": "^1.19.0",
-    "cassandra-driver": "4.6.1",
+    "cassandra-driver": "^4.6.1",
     "commander": "^7.2.0",
     "cors": "^2.7.1",
     "ethers": "^5.1.4",
--- a/packages/broker/src/plugins/storage/Storage.ts
+++ b/packages/broker/src/plugins/storage/Storage.ts
@@ -111,22 +111,22 @@ export class Storage extends EventEmitter {
             prepare: true, fetchSize: 1
         }
 
-        const resultStream = this._createResultStream()
+        const resultStream = this.createResultStream()
 
-        const makeLastQuery = (bucketIds: BucketId[]) => {
-            const params = [streamId, partition, bucketIds, limit]
-
-            this.cassandraClient.execute(GET_LAST_N_MESSAGES, params, {
-                prepare: true,
-                fetchSize: 0 // disable paging
-            }).then((resultSet: Todo) => {
+        const makeLastQuery = async (bucketIds: BucketId[]) => {
+            try {
+                const params = [streamId, partition, bucketIds, limit]
+                const resultSet = await this.cassandraClient.execute(GET_LAST_N_MESSAGES, params, {
+                    prepare: true,
+                    fetchSize: 0 // disable paging
+                })
                 resultSet.rows.reverse().forEach((r: Todo) => {
                     resultStream.write(r)
                 })
-                resultStream.push(null)
-            }).catch((e: Todo) => {
-                resultStream.destroy(e)
-            })
+                resultStream.end()
+            } catch (err) {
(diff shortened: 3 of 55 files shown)


packages/broker/src/plugins/legacyWebsocket/historicalData.ts
```diff
@@ -62,25 +65,39 @@ export const createResponse = async (
...
-            data,
-            abort: () => {
-                data.destroy()
-                abortController.abort()
-            },
-            startTime: Date.now()
+    const storageNodeUrls = await storageNodeRegistry.getUrlsByStreamId(request.streamId)
+
+    // Form data query endpoints and shuffle the resulting array
+    const urls = storageNodeUrls.map((storageNodeUrl) => {
```
teogeb, line 71: Could use a helper function to shuffle the array, e.g. https://lodash.com/docs/4.17.15#shuffle

Lodash is currently not in `broker`'s dependencies, but IMO we should add it. It's quite "standard" low level JS-library. Once we have that in the toolbox, we are able to focus more to the higher abstraction level 😄

packages/broker/src/StorageNodeRegistry.ts
```diff
@@ -41,7 +48,7 @@ export class StorageNodeRegistry {
     // TODO when we support multiple storage nodes, this method should actually return all
     // addresses. resends should try to fetch from random storage node, and if that fetch fails,
     // it should try to fetch from other addresses, too
-    private async getStorageNodeAddress(streamId: string): Promise<string|undefined> {
+    private async getStorageNodeAddress(streamId: string): Promise<string[]|undefined> {
```
teogeb, line 51: Rename to `getStorageNodeAddresses` and remove the comment. Could return an empty array if there are no storage nodes (then we don't need the special handling on `undefined` when iterating over results).

packages/broker/src/StorageNodeRegistry.ts
```diff
@@ -24,14 +31,14 @@ export class StorageNodeRegistry {
...
-        if (storageNodeAddress !== undefined) {
-            const url = this.getUrlByAddress(storageNodeAddress)
-            if (url !== undefined) {
-                return url
+    async getUrlsByStreamId(streamId: string): Promise<string[]> {
+        const storageNodeAddresses = await this.getStorageNodeAddress(streamId)
+        if (storageNodeAddresses !== undefined) {
+            const urls = this.getUrlsByAddresses(storageNodeAddresses)
+            if (urls.length > 0) {
+                return urls
```
teogeb, line 39: Should we print the warning to the logs if any of the addresses is not in the registry? 

If we'd iterate over each address (instead of urls) in `historicalData.createResponse`, the fetching of the URL for an address could be part of an iteration loop. -> We'd try another address if we can't get an URL for an address, or that URL responses non-200.

packages/broker/src/StorageNodeRegistry.ts
```diff
@@ -41,7 +48,7 @@ export class StorageNodeRegistry {
     // TODO when we support multiple storage nodes, this method should actually return all
     // addresses. resends should try to fetch from random storage node, and if that fetch fails,
     // it should try to fetch from other addresses, too
```
harbu, line 50: TODO can be removed now
```suggestion
```

packages/broker/src/StorageNodeRegistry.ts
```diff
@@ -50,7 +57,7 @@ export class StorageNodeRegistry {
                 return item.storageNodeAddress
             })
             if (addresses.length > 0) {
-                return addresses[Math.floor(Math.random() * addresses.length)]
+                return addresses
             } else if (addresses.length === 0) {
                 return undefined
```
harbu, line 62: Consider replacing with `return addresses` thereby simplifying method signature return type to be `Promise<string[]>` (and having empty array be a natural signal of "no addresses assigned"). User of method then doesn't need to deal with possible undefined check either, which is nice.

packages/broker/src/StorageNodeRegistry.ts
harbu: Looks like I just said the same things, should've read your comment 1st lol.

packages/broker/src/StorageNodeRegistry.ts
harbu: Would be worth a warning yeah if an address is not in the registry.

packages/broker/src/plugins/legacyWebsocket/historicalData.ts
harbu: I see your point, but I think here it still rather reasonable to determine what is going on. (If you're familiar with comparator stuff)

harbu (commented):
Overall LGTM; see my few comments

packages/broker/src/plugins/legacyWebsocket/historicalData.ts
teogeb: True, I just like to avoid mixing two different abstraction levels within one method. Usually the code is easier to read when there is no mixing. Eg. in this method it is important to know that the array is shuffled, but not important to know how we shuffle it.

In the code there is a comment describing "what is going on". To me it is always a trigger to refactor the logic to helper function 👍

harbu (approved):

PR #74, opened 2021-06-09
Fix webrtc: few additional changes (

harbu:
_Notice based on `fix_webrtc` branch_.

Added a few refactorings that I think are worth it it. They are in separated (more-or-less) neatly per commit, if we want to pick and choose. 

## 1st commit (https://github.com/streamr-dev/network-monorepo/commit/9d58f835a7539e4cf6d04060fa9b95fb8d10a1bd)
Pass `DeferredConnectionAttempt` to `Connection` in constructor and remove separate methods for creating and setting it.

I noticed that we always pass an instance of  `DeferredConnectionAttempt` to a connection when creating it. Might as well do it in constructor to get the benefits of type safety (=> less easy to forget to set it after creating a connection). Also imo makes the state change semantics of a `Connection` instance more clear: we start off a `Connection`  always with `DeferredConnectionAttempt != null` and over a period of time it may  go to `DeferredConnectionAttempt = null` but will never be set to non-null after!

## 2nd commit (https://github.com/streamr-dev/network-monorepo/commit/26048fc72399c71c9ea5b2f97999a844265d5817)
Remove `isOffering` parameters and replace with helper function and method.

Now that we no longer have the storage node / resend logic around, we can consistently rely on the peer ids to determine who is offerer and answerer. To prevent any openings for future bugs, just get rid of the parameter and always use the ids.

## 3rd commit (https://github.com/streamr-dev/network-monorepo/commit/52575bdb9aa125a80f87ba8206bcbe433ba9b1b6)
Extract helper method `replaceConnection`  in class `WebRtcEndpoint`. 

We had two places in `WebRtcEndpoint` in which we replace an existing connection with a new one (possibly stealing the deferred connection attempt in the process). The logic was more less similar, with the only slightly different thing being whether a `connectionId` was set in the process or not. I feel like it is worthwhile having this logic in its own method in case we decide to update it in the future as one could easily forget to update one of the places if there were two.

## 4th commit (https://github.com/streamr-dev/network-monorepo/commit/f5ab0f72135375ed9b045a97ced64b61aead6b8d)
Addressed test I broke when doing 1st commit.

Basically the test will pass `DeferredConnectionAttempt` instances manually to respective `Connection` instance. If no one is around to catch the promise rejections, the rejections will cause the test case to fail. It's worth a consideration whether we want to add a default `catch` into  `DeferredConnectionAttempt` in itself. Really depends on whether we want to make handling of this promise rejection mandatory or not.

## 5th commit (https://github.com/streamr-dev/network-monorepo/commit/c53551664c2bd103dc2577b096d937644c4a7ec3)
Extracted (internal) helper method in test-utils and did some general clean up.

Noticed that methods `runAndWaitForEvents` and `runAndRaceEvents` have identical bodies except for the use of `Promise.all` vs. `Promice.race`. Utilized this observation to extract all the commonalities into a helper method.

Decided to make catching the promise optional.

### DIFF
--- a/packages/network/src/connection/Connection.ts
+++ b/packages/network/src/connection/Connection.ts
@@ -120,7 +120,9 @@ export class DeferredConnectionAttempt {
                 reject(reason)
             })
         })
-    
+
+        // allow promise to reject without outside catch
+        this.connectionAttemptPromise.catch(() => {})
     }
 
     getPromise(): Promise<string> {
--- a/packages/network/test/integration/Connection.test.ts
+++ b/packages/network/test/integration/Connection.test.ts
@@ -43,8 +43,6 @@ describe('Connection', () => {
         const messageQueueTwo = new MessageQueue<string>()
         const deferredConnectionAttemptOne = new DeferredConnectionAttempt('two')
         const deferredConnectionAttemptTwo = new DeferredConnectionAttempt('one')
-        deferredConnectionAttemptOne.getPromise().catch(() => {}) // prevent crashing on reject
-        deferredConnectionAttemptTwo.getPromise().catch(() => {}) // prevent crashing on reject
         connectionOne = new Connection({
             selfId: 'one',
             targetPeerId: 'two',


juslesan (approved):

ptesavol (approved):
The code looks ok, and can be merged. 

The explanation of the first commit is a little unclear  "we start off a Connection always with DeferredConnectionAttempt != null", but in the code, connections are also started with DeferredConnectionAttempt == null.

PR #77, opened 2021-06-10
Broker: create tests streams by id

teogeb:
Use `createTestStream` to create test streams, encode `streamId` in URLs.


packages/broker/test/utils.ts
```diff
@@ -212,3 +212,16 @@ export const waitForStreamPersistedInStorageNode = async (streamId: string, part
     }
     await waitForCondition(() => isPersistent(), undefined, 1000)
 }
+
```
teogeb, line 215: These test utils could be in `test-utils` package. But if we do that test-utils would depend on `streamr-client` and Node.

harbu (approved):

PR #79, opened 2021-06-11
Client NET-340: Publish to specific partition

teogeb:
It is possible to specify a partition in `client.publish()` call, e.g. 

```
client.publish({ 
  streamId: 'foobar', 
  streamPartition: 123
},  message)
```

The client throws an error if both `partition` and `partitionKey` (the fourth parameter of `publish()`) are provided. Otherwise partitionKey logic has not been changed.

Added unit tests for `MessageCreator`, where the actual parameter parsing is done.


timoxley (approved):
yep, looks good

PR #81, opened 2021-06-11
fix: attempt to fix cross client tests with new core API

harbu:
Need to pass id to createStream explicitly

### DIFF
--- a/packages/cross-client-testing/src/main/java/com/streamr/client_testing/StreamTester.java
+++ b/packages/cross-client-testing/src/main/java/com/streamr/client_testing/StreamTester.java
@@ -63,7 +63,7 @@ public class StreamTester {
             // TODO: consider cleaning below when 3.0.0 is landed (and the private static helper method)
             String address = publicKeyFromPrivateKey(privateKey);
             Stream protoStream = new Stream(streamName, "");
-            protoStream.setId(address + "/" + System.nanoTime());
+            protoStream.setId(address + "/" + UUID.randomUUID());
             stream = creator.createStream(protoStream);
             creator.addStreamToStorageNode(stream.getId(), new StorageNode(DEV_STORAGE_NODE_ADDRESS));
         } catch (Exception e) {


teogeb (approved):

packages/cross-client-testing/src/main/java/com/streamr/client_testing/StreamTester.java
```diff
@@ -50,13 +53,18 @@
...
                 SigningOptions.getDefault(), EncryptionOptions.getDefault(), websocketApiUrl, restApiUrl);
         options.setStorageNodeAddress(DEV_STORAGE_NODE_ADDRESS);
         options.setStorageNodeUrl(DEV_STORAGE_NODE_URL);
         this.creator = new StreamrClient(options);
         try {
-            stream = creator.createStream(new Stream(streamName, ""));
+            // TODO: consider cleaning below when 3.0.0 is landed (and the private static helper method)
+            String address = publicKeyFromPrivateKey(privateKey);
+            Stream protoStream = new Stream(streamName, "");
+            protoStream.setId(address + "/" + System.nanoTime());
```
timoxley, line 66: is it possible these execute in same nanosecond

timoxley (approved):
nice one, lgtm

packages/cross-client-testing/src/main/java/com/streamr/client_testing/StreamTester.java
harbu: I read a bit about these and basically https://stackoverflow.com/questions/56809879/can-i-use-nanotime-instead-of-randomuuid:

> In my experience with the particular Java implementations I have used, the origin is the moment when the JVM starts up. This means I will most certainly see the same numbers all over again after the next JVM restart. 

So you have an already plausible chance to hit the same numbers when restarting JVM and running this test multiple times.

**Changed to use UUID.**

harbu: Ah it actually generates a random private key for which the public key is used as the first part of the stream id. So the above argument doesn't really apply. Well, I still feel better having it as UUID than nanoTime so will keep it as is.

PR #87, opened 2021-06-16
broker: upgrade cassandra driver & fix test

harbu:
- Upgrade cassandra driver to 4.6.3
- If you increased constant `MESSAGE_SIZE` in Storage.test.ts you could break the test, made sure to keep 1000 as lower bound for argument `fromTimestamp`.

### DIFF
--- a/packages/broker/test/integration/plugins/storage/Storage.test.ts
+++ b/packages/broker/test/integration/plugins/storage/Storage.test.ts
@@ -358,13 +358,13 @@ describe('Storage', () => {
         })
 
         it('can requestFrom', async () => {
-            const streamingResults = storage.requestFrom(storedStreamId, 0, NUM_MESSAGES, 0, null)
+            const streamingResults = storage.requestFrom(storedStreamId, 0, 1000, 0, null)
             const results = await toArray(streamingResults)
             expect(results.length).toEqual(NUM_MESSAGES)
         })
 
         it('can requestFrom again', async () => {
-            const streamingResults = storage.requestFrom(storedStreamId, 0, NUM_MESSAGES, 0, null)
+            const streamingResults = storage.requestFrom(storedStreamId, 0, 1000, 0, null)
             const results = await toArray(streamingResults)
             expect(results.length).toEqual(NUM_MESSAGES)
         })


packages/broker/test/integration/plugins/storage/Storage.test.ts
```diff
@@ -358,13 +358,13 @@ describe('Storage', () => {
...
+            const streamingResults = storage.requestFrom(storedStreamId, 0, 1000, 0, null)
             const results = await toArray(streamingResults)
             expect(results.length).toEqual(NUM_MESSAGES)
         })
 
         it('can requestFrom again', async () => {
-            const streamingResults = storage.requestFrom(storedStreamId, 0, NUM_MESSAGES, 0, null)
+            const streamingResults = storage.requestFrom(storedStreamId, 0, 1000, 0, null)
             const results = await toArray(streamingResults)
             expect(results.length).toEqual(NUM_MESSAGES)
```
timoxley, line 369: This seems wrong, should this be `1000` or should they both be `NUM_MESSAGES`?

harbu: Signature of requestFrom:
```ts
requestFrom(streamId: string, partition: number, fromTimestamp: number, fromSequenceNo: number, publisherId: string|null): Readable
```

So the parameter in question is `fromTimestamp`.

If we check how the data is set up:
```ts
for (let i = 0; i < NUM_MESSAGES; i++) {
                randomFillSync(randomBuffer)
                const msg = buildMsg({
                    streamId: storedStreamId,
                    streamPartition: 0,
                    timestamp: (i + 1) * 1000,
                    sequenceNumber: i,
                    publisherId: 'publisher1',
                    content: randomBuffer.toString('hex')
                })
                storePromises.push(storage.store(msg))
            }
```
We see that the timestamps start off from 1000, 2000, 3000, 4000, etc.

So in order to get all messages (`NUM_MESSAGES`) the query needs to start from the first timestamp 1000.

timoxley: Ah of course

timoxley (approved):

PR #88, opened 2021-06-16
NET-327: Nodes configured with the same Ethereum private key need unique WebRTC ids

bearni95:
Currently the id of a node is just the Ethereum address of the node.

This creates problems from WebRTC connectivity and tracker topology creation perspective if someone starts multiple nodes with the same Ethereum private key.

Suggested solution:

When the node instance is created, generate a random UUID instanceId which gets appended to the Ethereum address to form the node's ID for the purposes of WebRTC connectivity and communicating with the tracker.


packages/network/test/integration/tracker-instructions.test.ts
```diff
@@ -91,16 +98,18 @@ describe('check tracker, nodes and statuses from nodes', () => {
             waitForEvent(node2, NodeEvent.NODE_SUBSCRIBED)
         ])
 
+        /*
```
bearni95, line 101: Cannot get the portion:
`waitForEvent(tracker.trackerServer, TrackerServerEvent.NODE_STATUS_RECEIVED)`
of code to run properly, always gets timed out and can't figure out why

packages/network/src/connection/PeerInfo.ts
```diff
@@ -128,7 +128,19 @@ export class PeerInfo {
...
+
+        if (peerType === PeerType.Tracker){
+            this.peerId = peerId
+            return
+        }
+
+        if (peerId.indexOf('_sid_') >= 0){
+            this.peerId = peerId
+        } else {
+            this.peerId = `${peerId}_sid_${uuidv4()}`
```
bearni95, line 141: replaced `#` with `_sid_` as a split between the peerId and sessionId to allow the HTTP endpoints to work as expected

harbu:
Hey, I'll need to look more into this, but I think the approach is not really what we want. AFAIK the network shouldn't need to know anything about the contents of the `id` other than it is just some random unique identifier. I'd look into appending the UUID here instead (broker): https://github.com/streamr-dev/network-monorepo/blob/main/packages/broker/src/broker.ts#L64 and then ensuring that we only append it if we are not acting as a storage node. @bearni95

bearni95:
> 
> 
> Hey, I'll need to look more into this, but I think the approach is not really what we want. AFAIK the network shouldn't need to know anything about the contents of the `id` other than it is just some random unique identifier. I'd look into appending the UUID here instead (broker): https://github.com/streamr-dev/network-monorepo/blob/main/packages/broker/src/broker.ts#L64 and then ensuring that we only append it if we are not acting as a storage node. @bearni95

mhm, seems much easier than the path taken. i'll switch it into the broker id modification as suggested

packages/network/src/connection/PeerInfo.ts
```diff
@@ -128,7 +128,19 @@ export class PeerInfo {
             longitude: null,
             country: null,
             city: null
+        } 
+
+        if (peerType === PeerType.Tracker){
```
harbu, line 133: I'm guessing this is not needed now that id gets generated in the broker?

harbu (changes requested):
@bearni95 can you go thru the tests and revert those that worked before and keep those with meaningful / important / refactor changes?

packages/broker/test/integration/multi-account-brokers.test.ts
```diff
@@ -0,0 +1,800 @@
...
+            client1.subscribe({
+                stream: freshStreamId
+            }, () => {}),
+            client2.subscribe({
+                stream: freshStreamId
+            }, () => {}),
+            client3.subscribe({
+                stream: freshStreamId
+            }, () => {}),
+        ])
```
timoxley: Hm, remind me why it needs to already be subscribed? @harbu I recall something like this before in other tests but don't remember the reason, seems crap. Maybe unnecessary?

packages/broker/test/integration/multi-account-brokers.test.ts
```diff
@@ -0,0 +1,800 @@
...
+                key: 2
+            },
+            {
+                key: 3
+            },
+            {
+                key: 4
+            },
+        ])
+    })
```
timoxley: Lots of duplication and whitespace in these tests.

At a minimum I'd probably collapse these something like this (or whatever the smallest you can get it and keep the linter happy):
 ```js
 expect(client3Messages).toEqual([
  { key: 2 }, { key: 3 }, { key: 4 },
])
```
Maybe this just matches how other tests in here are set up, but I feel like these should probably be collected into a variable and tested against that, rather than repeated? 

Something like this perhaps, your call though:

```js
// collect all messages to be published so we can check against them later
// could do something like your `for (let i = 1; i <= 3; ++i) {` above 
// or get fancy like `Array(4).fill().map((_, key) => ({ key }))`
// but it's not much shorter than just writing it out (once)
const messages = [{ key: 0 }, { key: 1 }, { key: 2 }, { key: 3 }] // start from 0 so indexes match keys

await client1.publish(freshStreamId, messages[0]))
await wait(5) // doesn't need to be long, wait(1) would also probably suffice?
const timeAfterFirstMessagePublished = Date.now()
await client1.publish(freshStreamId, messages[1]))
await wait(5)
await client1.publish(freshStreamId, messages[2])

await Promise.all([
            client1.resend({
                stream: freshStreamId,
                resend: {
                    from: {
                        timestamp: timeAfterFirstMessagePublished
                    }
                }
            }, (message) => {
                client1Messages.push(message)
            }),
            // etc …
])

// do these all together
await Promise.all([
    // messages.length - 1, should skip first message
    waitForCondition(() => client1Messages.length === messages.length - 1),
    waitForCondition(() => client2Messages.length === messages.length - 1),
    waitForCondition(() => client3Messages.length === messages.length - 1),
])

// compare directly against messages we published 
expect(client1Messages).toEqual(messages.slice(1)) // all but first
// etc …

File: packages/broker/test/integration/multi-account-brokers.test.ts [hunk 6]
```diff
@@ -0,0 +1,800 @@
...
+        })
+        await wait(25)
+        const timeAfterThirdMessagePublished = Date.now()
+        await wait(25)
+
+        await client1.publish(freshStreamId, {
+            key: 4
+        })
+
+        await wait(1500) // wait for propagation
```
timoxley: We really need to put some kind of utility into the public client interface that waits for messages to hit storage. I have a utility that polls the `last` endpoint in the client tests but I think it should just be in the main client.

bearni95: @timoxley this whole test is a copy-paste from broker.test.ts, only using the same identities to reproduce and ensure that all those tests are passing. Not sure if there would be a more elegant solution, like providing an alternate befreEach to run the same "it" blocks as in the broker default test?

harbu (approved):
Hey, LGTM. I did remove the multi-account-brokers.test.ts and replaced it with something slightly simpler as I felt that for this feature the test can be somewhat light.

PR #89, opened 2021-06-16
Broker NET-300: Testnet miner plugin

teogeb:
New `testnetMiner` plugin.

**TODO**: When we know the STUN server address, we should mention that in plugin's `config.schema.json`.

To get node neighbor data, two new public methods have been added to `streamr-network.NetworkNode`:
- `getNeighborsForStream`
- `getRtt`


packages/broker/src/helpers/scheduler.ts
```diff
@@ -0,0 +1,20 @@
+export const scheduleAtInterval = async (
```
teogeb, line 1: The scheduler does not do any error handling. The tasks methods **should** catch any errors internally.

As we currently don't have any global error handlers, throwing an error from `task()` will crash the application.

harbu: Add this as a comment to the function description (i.e. JSdoc)

packages/broker/src/helpers/scheduler.ts
```diff
@@ -0,0 +1,20 @@
...
+    task: () => Promise<void>, 
+    interval: number, // number of milliseconds to wait after a task is completed
+    executeAtStart: boolean
+): Promise<{ stop: () => void }> => {
+    let timer: NodeJS.Timer|undefined
+    const scheduleNext = () => {
+        timer = setTimeout(async () => {
+            await task()
+            scheduleNext()
```
harbu, line 10: You need to check whether `stop`  was invoked after the above `await` finishes. Otherwise it's possible that the interval continues.

packages/broker/src/plugins/testnetMiner/TestnetMinerPlugin.ts
```diff
@@ -0,0 +1,132 @@
+import { Wallet } from 'ethers'
+import fetch from 'node-fetch'
+// @ts-expect-error no type definitions
+import fetchNatType from 'nat-type-identifier'
```
harbu, line 4: Add the typing yourself (provided this is a reasonably sized library), see "packages/network/src/types/speedometer/index.d.ts" for an example.

packages/broker/src/plugins/testnetMiner/TestnetMinerPlugin.ts
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+
+    private async getLatency() {
+        const startTime = Date.now()
+        try {
+            await fetch(`${this.pluginConfig.claimServerUrl}/ping`)
+            return Date.now() - startTime
+        } catch (e) {
+            logger.warn('Ping error')
```
harbu, line 110: What actually gets returned when we hit `catch` since we are not re-throwing? Good example of where having the explicit return type visible is worth it imo.

packages/broker/src/plugins/testnetMiner/TestnetMinerPlugin.ts
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+
+    private async getNatType(): Promise<string> {
+        logger.info('Analyzing NAT type')
+        const result = await fetchNatType({ 
+            logsEnabled: false, 
+            sampleCount: NAT_ANALYSIS_SAMPLE_COUNT, 
+            stunHost: this.pluginConfig.stunServerHost 
+        })
```
harbu, line 120: Can this throw? If it can, do we want to crash the app or just fill in a placeholder value?

packages/broker/src/plugins/testnetMiner/TestnetMinerPlugin.ts
```diff
@@ -0,0 +1,132 @@
...
+            waitTime: delay,
+            natType: this.natType,
+            peers
+        }
+        try {
+            await fetch(`${this.pluginConfig.claimServerUrl}/claim`, {
+                body: JSON.stringify({
+                    rewardCode,
+                    ...metadata
+                }),
```
harbu, line 92: Small simplification: include `rewardCode` in metadata so you can just do the following?:
```suggestion
                body: JSON.stringify(metadata),
```

packages/broker/src/plugins/testnetMiner/TestnetMinerPlugin.ts
```diff
@@ -0,0 +1,132 @@
...
+                method: 'POST',
+                headers: {
+                    'Content-Type': 'application/json'
+                }
+            })
+            logger.info('Reward claimed successfully')
```
harbu, line 98: I forget how `node-fetch` works (does it throw on non-200 codes?) but should we check HTTP status code afterwards?

packages/broker/src/plugins/testnetMiner/config.schema.json
```diff
@@ -0,0 +1,33 @@
...
+    "required": [
+        "rewardStreamId",
+        "claimServerUrl",
+        "maxClaimDelay"
+    ],
+    "additionalProperties": false,
+    "properties": {
+        "rewardStreamId": {
+            "type": "string",
+            "description": "streamr.eth/testnets/brubeck-rewards"
```
harbu, line 15: I wonder if you can use keyword "default" here, see: https://json-schema.org/understanding-json-schema/reference/generic.html

packages/broker/test/integration/plugins/testnetMiner/TestnetMinerPlugin.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    })
+
+    it('happy path', async () => {
+        expect(claimServer!.pingEndpointCalled).toBeTruthy()
+        await publishRewardCode(rewardStreamId)
+        await waitForCondition(() => claimServer.claimRequestBody !== undefined)
+        expect(claimServer.claimRequestBody.rewardCode).toBe(MOCK_REWARD_CODE)
+        expect(claimServer.claimRequestBody.nodeAddress).toBe(new Wallet(nodePrivateKey).address)
+        expect(claimServer.claimRequestBody.clientServerLatency).toBeDefined()
+        expect(claimServer.claimRequestBody.waitTime).toBeDefined()
```
harbu, line 111:
```suggestion
        expect(claimServer.claimRequestBody.clientServerLatency).toBeGreaterThanOrEqual(0)
        expect(claimServer.claimRequestBody.waitTime).toBeGreaterThanOrEqual(0)
```

harbu (changes requested):
Good stuff :+1: See my comment especially related to helper

packages/broker/src/helpers/scheduler.ts
teogeb: Fixed

packages/broker/src/helpers/scheduler.ts
teogeb: Fixed

packages/broker/src/plugins/testnetMiner/TestnetMinerPlugin.ts
teogeb: Fixed

packages/broker/src/plugins/testnetMiner/TestnetMinerPlugin.ts
teogeb: It seems that the method doesn't throw in practice. But it can take very long time if passed an invalid `stunHost`. Wrapped the call with timeout handler (new `withTimeout()` helper method). If it times out, we'll explicit set `natType='Unknown'`.

packages/broker/src/plugins/testnetMiner/TestnetMinerPlugin.ts
teogeb: Fixed

packages/broker/src/plugins/testnetMiner/TestnetMinerPlugin.ts
teogeb: Fixed

packages/broker/src/plugins/testnetMiner/config.schema.json
teogeb: Fixed

packages/broker/test/integration/plugins/testnetMiner/TestnetMinerPlugin.test.ts
teogeb: Fixed

packages/broker/src/plugins/testnetMiner/TestnetMinerPlugin.ts
teogeb: Added a helper method which always throws if the status is not successful

packages/broker/src/helpers/withTimeout.ts
```diff
@@ -0,0 +1,16 @@
+export const withTimeout = async <T>(promise: Promise<T>, timeout: number): Promise<T> => {
+    return new Promise(async (resolve, reject) => {
+        const timer = setTimeout(() => {
+            reject(new Error('Timeout'))
```
harbu, line 4: I've found that in practice, it helps a bit giving some context in these error messages whilst e.g. debugging.
```suggestion
            reject(new Error('withTimeout: Timeout'))
```

harbu (commented):
LGTM; there is still one fix to the helper imo

packages/broker/src/helpers/withTimeout.ts
teogeb: Fixed. Didn't change the error message as it is written to the end-user log. But now `withTimeout` throws a `GenericError` which contains an error code (e.g. `NAT_ANALYSIS_TIMEOUT` in this case).

harbu (approved):

PR #90, opened 2021-06-16
NET-350, broker: test that shows that storing messages in random order breaks stuff

harbu:


packages/broker/test/integration/plugins/storage/Storage.test.ts
```diff
@@ -370,6 +370,62 @@ describe('Storage', () => {
         })
     })
 
+    describe('messages pushed in randomized order', () => {
```
harbu:
```suggestion
    // This test proves that NET-350 is still an issue
    describe.skip('messages pushed in randomized order', () => {
```

PR #94, opened 2021-06-18
Move failing resend test into flakey test run. Split client tests across workflows.

timoxley:
Hoping to get feedback faster like this.


timoxley:
not sure if better, merged some of the cleanup into #107

PR #95, opened 2021-06-18
Combine client workflows

timoxley:
Alternative approach to split workflows in #94


timoxley:
Not sure if better, rerunning a subset of tests seems useful. Closing for now.

PR #98, opened 2021-06-18
Fix long resend test.

timoxley:
I think this "works" (update: it doesn't) because when it asks for say, the "last X" items, if the very last item isn't in that last X because the last messages logic is broken, (perhaps due to storing messages out of order, then it will never find the last item because it's (mysteriously) at some position > X. Changing it to always fetch "everything" might work around this bug.

Update: ugh, didn't seem to fix it in CI.


timoxley:
Blegh, didn't work

PR #96, opened 2021-06-18
Merge from streamr-client-javascript

jtakalai:

PR #103, opened 2021-06-21
Update client dependencies & match updated deps in other packages.

timoxley:
Trying to figure out what's going on with tests failing in #96
Something to do with use of `--detectOpenHandles` seems to stop the tests from exiting. 
https://github.com/facebook/jest/issues/9473

Might be fixed in newer version of `jest`.


timoxley:
Currently impossible to use Jest 27, can't build correctly due to some weird issue with CJS/ESM `TypeError: (0 , _prettyFormat.default) is not a function`  https://github.com/facebook/jest/issues/11595

Grabbing other non-upgrade fixes from this branch and opening new PR.

PR #105, opened 2021-06-22
Broker NET-333!: Configurable payload format for WebSocket and MQTT plugin

teogeb:
Unified the payload handling of `websocket` plugin and `mqtt` plugin. Added a new setting `payloadMetadata` (a boolean value) to both plugin configs. 

If `payloadMetadata` is `true`, the publish and subscription payloads are JSONs which have a structure of
```
{
    content: ...,
    metadata: ...
}
```

For publish requests, it is possible to set an optional `timestamp` value in the metadata. For subscription requests, the metadata contains `timestamp`, `sequenceNumber`, `publisherId` and `msgChainId` fields of the message.

If `payloadMetadata` setting is `false`, the payloads are plain content JSONs (and therefore it is not possible to define the timestamp  or read the metadata of the message).

**Test refactoring:** As the payload handling of `websocket` and `mqtt` plugins are similar, the plugin integration tests now use a shared helper method `createMessagePluginTest`. MQTT-specific test of `streamIdDomain` has been refactored to a separate unit test.

**Breaking change:** As the `payloadFormat` setting is a required parameter, this is be a breaking change.

### DIFF
--- a/packages/broker/src/helpers/PayloadFormat.ts
+++ b/packages/broker/src/helpers/PayloadFormat.ts
@@ -1,10 +1,10 @@
 export interface PayloadFormat {
     createMessage: (payload: string) => Message|never
-    createPayload: (content: any, metadata?: Metadata) => string|never
+    createPayload: (content: Record<string,unknown>, metadata?: Metadata) => string|never
 }
 
 export interface Message {
-    content: any
+    content: Record<string,unknown>
     metadata: Metadata
 }
 
@@ -23,22 +23,29 @@ const pickProperties = (fields: string[], from: Record<string,unknown>): Record<
     return result
 }
 
+const isJavascriptObject = (content: any): boolean => {
+    return (content instanceof Object) && (!(content instanceof Array))
+}
+
 const assertContent = (content: any): void|never => {
     if ((content === undefined) || (content === null)) {
         throw new Error('Content missing')
     }
-    if (!(content instanceof Array || content instanceof Object)) {
-        throw new Error('Content is not an object or an array')
+    if (!isJavascriptObject(content)) {
+        throw new Error('Content is not an object')
     }
 }
 
 const assertMetadata = (metadata: any): void|never => {
-    if (!(metadata instanceof Object)) {
+    if (!isJavascriptObject(metadata)) {
         throw new Error('Metadata is not an object')
     }
 }
 
 const parsePayloadJson = (payload: string) => {
+    if (payload.length === 0) {
+        throw new Error('Payload missing')
+    }
     try {
         return JSON.parse(payload)
     } catch (e) {
@@ -48,13 +55,15 @@ const parsePayloadJson = (payload: string) => {
 
 export class PlainPayloadFormat implements PayloadFormat {
     createMessage(payload: string): Message|never {
+        const content = parsePayloadJson(payload)
+        assertContent(content)
         return {
-            content: parsePayloadJson(payload),
+            content,
             metadata: {}
         }
     }
 
-    createPayload(content: any): string|never {
+    createPayload(content: Record<string,unknown>): string|never {
         assertContent(content)
         return JSON.stringify(content)
     }
@@ -76,7 +85,7 @@ export class MetadataPayloadFormat implements PayloadFormat {
         return { content, metadata }
     }
 
-    createPayload(content: any, metadata?: Metadata): string|never {
+    createPayload(content: Record<string,unknown>, metadata?: Metadata): string|never {
         assertContent(content)
         const payload: any = {
             content
--- a/packages/broker/src/helpers/PayloadFormat.ts
+++ b/packages/broker/src/helpers/PayloadFormat.ts
@@ -1,6 +1,6 @@
 export interface PayloadFormat {
-    createMessage: (payload: string) => Message
-    createPayload: (content: any, metadata?: Metadata) => string
+    createMessage: (payload: string) => Message|never
+    createPayload: (content: any, metadata?: Metadata) => string|never
 }
 
 export interface Message {
@@ -15,13 +15,15 @@ export interface Metadata {
     msgChainId?: string
 }
 
-const pickProperties = (fields: string[], from: any) => {
+const METADATA_FIELDS = [ 'timestamp', 'sequenceNumber', 'publisherId', 'msgChainId' ]
+
+const pickProperties = (fields: string[], from: Record<string,unknown>): Record<string,unknown> => {
     const result: any = {}
     fields.forEach((field) => result[field] = from[field])
     return result
 }
 
-const assertContent = (content: any) => {
+const assertContent = (content: any): void|never => {
     if ((content === undefined) || (content === null)) {
         throw new Error('Content missing')
     }
@@ -30,7 +32,7 @@ const assertContent = (content: any) => {
     }
 }
 
-const assertMetadata = (metadata: any) => {
+const assertMetadata = (metadata: any): void|never => {
     if (!(metadata instanceof Object)) {
         throw new Error('Metadata is not an object')
     }
@@ -45,14 +47,14 @@ const parsePayloadJson = (payload: string) => {
 }
 
 export class PlainPayloadFormat implements PayloadFormat {
-    createMessage(payload: string): Message {
+    createMessage(payload: string): Message|never {
         return {
             content: parsePayloadJson(payload),
             metadata: {}
         }
     }
 
-    createPayload(content: any): string {
+    createPayload(content: any): string|never {
         assertContent(content)
         return JSON.stringify(content)
     }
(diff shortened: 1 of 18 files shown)


packages/broker/src/helpers/PayloadFormat.ts
```diff
@@ -0,0 +1,94 @@
...
+}
+
+export interface Metadata {
+    timestamp?: number
+    sequenceNumber?: number
+    publisherId?: string
+    msgChainId?: string
+}
+
+const pickProperties = (fields: string[], from: any) => {
```
harbu:
```suggestion
const pickProperties = (fields: string[], from: Record<string, unknown>): Record<string, unknown> => {
```

packages/broker/src/helpers/PayloadFormat.ts
```diff
@@ -0,0 +1,94 @@
+export interface PayloadFormat {
+    createMessage: (payload: string) => Message
+    createPayload: (content: any, metadata?: Metadata) => string
+}
+
+export interface Message {
+    content: any
```
harbu: Not sure if this applies:
```suggestion
    content: Record<string, unknown>
```

packages/broker/src/helpers/PayloadFormat.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    createPayload(content: any): string {
+        assertContent(content)
+        return JSON.stringify(content)
+    }
+}
+
+export class MetadataPayloadFormat implements PayloadFormat {
+
+    private static FIELDS = [ 'timestamp', 'sequenceNumber', 'publisherId', 'msgChainId' ]
```
harbu: I'd move `FIELDS` up in the file next to `interface Metadata` so it becomes less likely to forget to update this array if any changes are made in the future.

packages/broker/src/helpers/PayloadFormat.ts
```diff
@@ -0,0 +1,94 @@
...
+    msgChainId?: string
+}
+
+const pickProperties = (fields: string[], from: any) => {
+    const result: any = {}
+    fields.forEach((field) => result[field] = from[field])
+    return result
+}
+
+const assertContent = (content: any) => {
```
harbu:
```suggestion
const assertContent = (content: any): void | never => {
```

packages/broker/src/helpers/PayloadFormat.ts
```diff
@@ -0,0 +1,94 @@
...
+    if ((content === undefined) || (content === null)) {
+        throw new Error('Content missing')
+    }
+    if (!(content instanceof Array || content instanceof Object)) {
+        throw new Error('Content is not an object or an array')
+    }
+}
+
+const assertMetadata = (metadata: any) => {
```
harbu:
```suggestion
const assertMetadata = (metadata: any): void | never => {
```

packages/broker/src/helpers/PayloadFormat.ts
```diff
@@ -0,0 +1,94 @@
+export interface PayloadFormat {
+    createMessage: (payload: string) => Message
+    createPayload: (content: any, metadata?: Metadata) => string
```
harbu: May be worth adding `never` to document that error could be thrown.

```suggestion
    createMessage: (payload: string) => Message | never
    createPayload: (content: any, metadata?: Metadata) => string | never
```

packages/broker/src/plugins/websocket/PublishConnection.ts
```diff
@@ -34,15 +27,15 @@ export class PublishConnection implements Connection {
...
     
-    init(ws: WebSocket, streamrClient: StreamrClient) {
-        ws.on('message', (contentPayload: string) => {
+    init(ws: WebSocket, streamrClient: StreamrClient, payloadFormat: PayloadFormat) {
+        ws.on('message', (payload: string) => {
             try {
-                const content = parsePayloadJson(contentPayload)
-                const partitionKey = this.partitionKey ?? (this.partitionKeyField ? content[this.partitionKeyField] : undefined)
+                const { content, metadata } = payloadFormat.createMessage(payload)
+                const partitionKey = this.partitionKey ?? (this.partitionKeyField ? (content as any)[this.partitionKeyField] : undefined)
```
harbu: I don't think this cast should be necessary though `(content as any)`? As isn't `message.content` already defined as `any`?

packages/broker/test/unit/plugins/mqtt/Bridge.test.ts
```diff
@@ -0,0 +1,49 @@
+import { StreamrClient } from 'streamr-client'
+import { Bridge } from '../../../../src/plugins/mqtt/Bridge'
+import { PlainPayloadFormat } from '../../../../src/helpers/PayloadFormat'
+
+const MOCK_TOPIC = 'mock-topic'
+const MOCK_STREAM_ID_DOMAIN = 'mock.ens'
+const MOCK_STREAM_ID = `${MOCK_STREAM_ID_DOMAIN}/${MOCK_TOPIC}`
+const MOCK_CONTENT = { foo: 'bar' }
+
+describe('MQTT Bridge', () => {
```
harbu, line 10: I'd like to ask few questions about this on call to clarify my understanding.

packages/broker/test/integration/createMessagingPluginTest.ts
```diff
@@ -0,0 +1,124 @@
...
+    api: MessagingPluginApi<T>, 
+    pluginPort: number, 
+    testModule: NodeJS.Module,
+    pluginConfig: any = {}
+) => {
+
+    // some free ports
+    const LEGACY_WEBSOCKET_PORT = pluginPort + 1
+    const TRACKER_PORT = pluginPort + 2
+    const NETWORK_PORT = pluginPort + 3
```
harbu: I'd rather these be given explicitly. Reason: port conflicts in tests are not unusual, and being able to simply grep exact port values is useful in identifying them.

packages/broker/test/integration/createMessagingPluginTest.ts
```diff
@@ -0,0 +1,124 @@
...
+            streamrClient = createClient(LEGACY_WEBSOCKET_PORT, brokerUser.privateKey, {
+                autoDisconnect: false
+            })
+            stream = await createTestStream(streamrClient, testModule)
+            messageQueue = new Queue<Message>()
+        })
+
+        afterEach(async () => {
+            if (pluginClient !== undefined) {
+                api.closeClient(pluginClient)
```
harbu:
```suggestion
                await api.closeClient(pluginClient)
```

packages/broker/src/helpers/PayloadFormat.ts
teogeb: Fixed

packages/broker/src/helpers/PayloadFormat.ts
teogeb: Fixed

packages/broker/src/helpers/PayloadFormat.ts
teogeb: Fixed

packages/broker/src/helpers/PayloadFormat.ts
teogeb: Fixed

packages/broker/src/helpers/PayloadFormat.ts
teogeb: Fixed

packages/broker/src/plugins/websocket/PublishConnection.ts
teogeb: Fixed

packages/broker/test/integration/createMessagingPluginTest.ts
teogeb: Now the port numbers are  listed explicitly

packages/broker/test/integration/createMessagingPluginTest.ts
teogeb: Fixed

packages/broker/src/helpers/PayloadFormat.ts
teogeb: If I've understood correctly, the content can be an array. `Record<string, unknown>` would be too strict for that case. 

In `streamr-client` we  use object type, but it is not optimal either: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md.

teogeb: Fixed. Updated tests for cases where the content is an array or payload JSON is an array, or empty string. 

The publishHttp plugin now uses `PayloadFormat` to apply the validation logic.

harbu (approved):

PR #106, opened 2021-06-22
Fix incorrect tracker-node-reconnect-instructions.test test case

juslesan:
- Fix ```integration/tracker-node-reconnect-instructions.test.ts``` to expect correct behaviour


timoxley (approved):
lgtm

PR #108, opened 2021-06-23
NET-358: re-enable and fix broker docker image

harbu:
Few points

- `**/node_modules` in .dockerignore very important. Otherwise the context that needs to be sent to the Docker image becomes 1.5GB (compared to 70MB).
- Dockerfile had to be moved to project root as it is not allowed to walk up the directory structure from "base".
- Add symlink to base directory for `tracker.js` to avoid having to update streamr-docker-dev

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,16 @@
 FROM node:14-buster as build
-WORKDIR /usr/src/broker
+WORKDIR /usr/src/monorepo
 COPY . .
-RUN npm ci --only=production
+RUN npm ci
+RUN npm run bootstrap-pkg streamr-broker
 # Build TypeScript files ("npm ci" doesn't trigger the build automatically: https://github.com/npm/npm/issues/17346)
-RUN npm run build 
+RUN (cd packages/broker && npm run build)
 
 FROM node:14-buster-slim
 RUN apt-get update && apt-get install --assume-yes --no-install-recommends curl \
 	&& apt-get clean \
 	&& rm -rf /var/lib/apt/lists/*
-COPY --from=build /usr/src/broker /usr/src/broker
+COPY --from=build /usr/src/monorepo/packages/broker /usr/src/broker
 WORKDIR /usr/src/broker
 
 # Make ports available to the world outside this container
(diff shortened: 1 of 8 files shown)


packages/broker/configs/docker-1.env.json
```diff
@@ -24,10 +24,10 @@
   "streamrUrl": "http://10.200.10.1",
   "streamrAddress": "0xFCAd0B19bB29D4674531d6f115237E16AfCE377c",
   "storageNodeConfig": {
-    "registry": {
```
hpihkala: The old config seemed correct here. Why not read the storage nodes from the smart contract?

hpihkala: Just to clarify, with the changes in this PR, you're switching from

- read the list of storage nodes from storage node registry at address `0xbAA81A0179015bE47Ad439566374F2Bae098686F`

to

- Override whatever is in the storage node registry with a locally configured list, in this case containing the node `0xde1112f631486CfC759A50196853011528bC5FA0`

This change seems unintentional and unwanted - unless there is currently some problem with reading the registry? But I suspect the errors you were seeing were due to mixing up the tracker registry with the storage node registry, which was fixed on line 9.

Dockerfile
```diff
@@ -1,16 +1,18 @@
...
-WORKDIR /usr/src/broker
+WORKDIR /usr/src/monorepo
 COPY . .
-RUN npm ci --only=production
-# Build TypeScript files ("npm ci" doesn't trigger the build automatically: https://github.com/npm/npm/issues/17346)
-RUN npm run build 
+RUN npm set unsafe-perm true
+RUN npm ci
+RUN npm run bootstrap-pkg streamr-broker
+RUN npx lerna exec -- npm prune --production
```
timoxley:
```suggestion
RUN npx lerna exec -- npm prune --production # image contains all packages, remove devDeps to keep image size down
```

Dockerfile
```diff
@@ -1,16 +1,18 @@
...
+RUN npx lerna link
```
timoxley:
```suggestion
RUN npx lerna link # restore inter-package symlinks removed by npm prune
```

timoxley (approved):
lgtm 👍 

Linking to related slack discussion here https://streamr-team.slack.com/archives/CEE3KF542/p1624889342109000

PR #109, opened 2021-06-23
feat: (network) increase abstraction level of the connection classes …

ptesavol:
Increase abstraction level of the connection classes to allow for browser implementation


packages/network/src/connection/NodeWebRtcEndpoint.ts
```diff
@@ -0,0 +1,43 @@
...
+import { ConstructorOptions } from './WebRtcConnection'
+import nodeDataChannel from 'node-datachannel'
+import { MetricsContext } from '../helpers/MetricsContext'
+import { RtcSignaller } from '../logic/RtcSignaller'
+import { NegotiatedProtocolVersions } from "./NegotiatedProtocolVersions"
+import { WebRtcEndpoint } from './WebRtcEndpoint'
+import { NodeWebRtcConnection } from './NodeWebRtcConnection'
+import { WebRtcConnection } from './WebRtcConnection'
+
+export class NodeWebRtcEndpoint extends WebRtcEndpoint implements IWebRtcEndpoint {
```
harbu: Wondering whether we should favor composition over inheritance here? The only two things needed by `WebRtcEndpoint` is a way to create new connections `doCreateConnection` and a clean up method `doStop`. We could create a factory interface that we pass to `WebRtcEndpoint` which it uses to do the two things avoiding inheritance altogether.

packages/network/src/connection/NodeWebRtcConnection.ts
```diff
@@ -0,0 +1,318 @@
...
+    close: (err?: Error) => void
+    error: (err: Error) => void
+    bufferLow: () => void
+    bufferHigh: () => void
+}
+
+// reminder: only use Connection emitter for external handlers
+// to make it safe for consumers to call removeAllListeners
+// i.e. no this.on('event')
+export const ConnectionEmitter = EventEmitter as { new(): StrictEventEmitter<EventEmitter, Events> }
```
harbu: Duplicated code segment from `WebRtcConnection`, can be removed?

packages/network/src/connection/NodeWebRtcConnection.ts
```diff
@@ -0,0 +1,318 @@
...
+    
+    constructor(opts: ConstructorOptions) {
+        super(opts)
+
+        this.connection = null
+        this.dataChannel = null
+        this.onStateChange = this.onStateChange.bind(this)
+        this.onLocalCandidate = this.onLocalCandidate.bind(this)
+        this.onLocalDescription = this.onLocalDescription.bind(this)
+        this.onStateChange = this.onStateChange.bind(this)
```
harbu: onStateChange already binded above?
```suggestion
```

packages/network/src/connection/NodeWebRtcConnection.ts
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    protected doSendMessage(message: string): boolean {
+        return this.dataChannel!.sendMessage(message)
+    }
+
+    connect(): void {
+        if (this.isFinished) {
+            throw new Error('Connection already closed.')
+        }
```
harbu: The problem that I see here is that `this.isFinished` is a field of the super class. This means that every implementer of a subclass must remember to do this check in the beginning of their `connect()` implementation, which ofc, is very easy to forget to do and its duplicated logic.

Some possible solutions:
1. Make each sub class keep track of their own `isFinished`
2. Make `connect` non-abstract, implement it in the super class, and then provide a new protected method `doConnect` that actually does the connection,i .e.

In (super) class `WebRtcConnection`:
```ts
connect(): void {
    if (this.isFinished) {
        throw new Error('Connection already closed.')
    }
    this.doConnect()
}

...

protected abstract doConnect(): void
```

packages/network/src/connection/WebRtcConnection.ts
```diff
@@ -0,0 +1,402 @@
...
+    }
+    
+    abstract connect(): void
+    abstract setRemoteDescription(description: string, type: string): void
+    abstract addRemoteCandidate(candidate: string, mid: string): void 
+    abstract close(err?: Error): void
+    abstract getBufferedAmount(): number 
+    abstract getMaxMessageSize(): number 
+    abstract isOpen(): boolean
+    protected abstract doSendMessage(message: string): boolean
```
harbu, line 446: Might want to write a JSDoc comment on this method. What does the return value indicate? What happens when it is true or false? Also what guarantees do we have about state when we invoke `doSendMessage`? For example, in `NodeWebRtcConnection` the implementation is as follows

```ts
protected doSendMessage(message: string): boolean {
    return this.dataChannel!.sendMessage(message)
}
```
Clearly there is an assumption that `isOpen() === true` holds for an instance  before `doSendMessage` is ever invoked. This may be worth commenting, e.g., "connectivity is checked before base class invokes `doSendMessage`, i.e., checking `isOpen()`  is done by base class".

juslesan (approved):

packages/network/src/connection/NodeWebRtcConnection.ts
```diff
@@ -0,0 +1,318 @@
...
+        this.connectionEmitter.on('localCandidate', this.onLocalCandidate)
+
+        if (this.isOffering()) {
+            const dataChannel = this.connection.createDataChannel('streamrDataChannel')
+            this.setupDataChannel(dataChannel)
+        } else {
+            this.connectionEmitter.on('dataChannel', this.onDataChannel)
+        }
+
+        this.connectionTimeoutRef = setTimeout(() => {
```
harbu: It's not a good idea to be setting directly a super class's field. Need to decide whether timeout logic should be _shared_ or _subclass-specific_.

1. If subclass-specific, then just move the field up from `WebRtcConnection` up to`NodeWebRtcConnection`.
2. If shared, then make `connectionTimeoutRef` private in `WebRtcConnection` and have it handle the whole thing (e.g. in a `connect` method as I described in previous comment). If the timeout occurs, I believe it will call abstract `close` anyways?

packages/network/src/connection/NodeWebRtcConnection.ts
```diff
@@ -0,0 +1,318 @@
...
+            this.close(new Error(`timed out after ${this.newConnectionTimeout}ms`))
+        }, this.newConnectionTimeout)
+    }
+
+    setRemoteDescription(description: string, type: DescriptionType): void {
+        if (this.connection) {
+            try {
+                this.connection.setRemoteDescription(description, type)
+            } catch (err) {
+                this.logger.warn('setRemoteDescription failed, reason: %s', err)
```
harbu, line 131: Wondering if we should have a separate logger for subclasses just for clarity? E.g. we know which sort of subclass we are looking at based on the log message.

packages/network/src/connection/NodeWebRtcConnection.ts
```diff
@@ -0,0 +1,318 @@
...
+            } catch (err) {
+                this.logger.warn('addRemoteCandidate failed, reason: %s', err)
+            }
+        } else {
+            this.logger.warn('skipped addRemoteCandidate, connection is null')
+        }
+    }
+
+    close(err?: Error): void {
+        if (this.isFinished) {
```
harbu: Same thing as before, this should really be checked by the base class not here.

packages/network/src/connection/NodeWebRtcConnection.ts
```diff
@@ -0,0 +1,318 @@
...
+            // already closed, noop
+            return
+        }
+
+        this.isFinished = true
```
harbu: Yeah this should be base class's responsibility, we should not really set values from subclass.

packages/network/src/connection/NodeWebRtcConnection.ts
```diff
@@ -0,0 +1,318 @@
...
+                this.dataChannel.close()
+            } catch (e) {
+                this.logger.warn('dc.close() errored: %s', e)
+            }
+        }
+
+        this.dataChannel = null
+        this.connection = null
+
+        this.doClose()
```
harbu: I really think it should be the other way around here.

In `WebRtcConnection` we should have the public `close` method and an abstract `doClose` method implemented by subclasses, e.g.
```ts
close(err?: Error): void {
    if (this.isFinished) {
       return
    }
    this.isFinished = true
    ....
    this.doClose()
}

...

protected doClose(err?: Error): void
```

And then the method here in this `NodeWebRtcConnection` would be named `doClose` instead.

This way the implementer does not have to remember or know that they need to call `doClose` at the end of their `close` implementation. Instead it is all taken care of by the superclass!

packages/network/src/connection/NodeWebRtcConnection.ts
```diff
@@ -0,0 +1,318 @@
...
+            return this.dataChannel!.isOpen()
+        } catch (err) {
+            return false
+        }
+    }
+
+    private onStateChange(state: string): void {
+        this.logger.trace('conn.onStateChange: %s -> %s', this.lastState, state)
+
+        this.lastState = state
```
harbu, line 216: It's a bit cumbersome to have to remember as an implementer of a sub class to set this value every time a state changes. My recommended change would be the following.

- Remove field `this.lastState` from `WebRtcConnection`
- Add a new abstract method (to replace the above) into `WebRtcConnection`, e.g. `protected abstract getLastState(): string`
- `WebRtcConnection` should solely rely on method `getLastState()` to get the last state
- Add field `this.lastState` to `NodeWebRtcConnection` and implement abstract method `getLastState()` so that it returns this value

packages/network/src/connection/NodeWebRtcConnection.ts
```diff
@@ -0,0 +1,318 @@
...
+        } else if (state === 'failed') {
+            this.close(new Error('connection failed'))
+        } else if (state === 'connecting') {
+            // reset timeout on connecting
+            clearTimeout(this.connectionTimeoutRef!)
+            this.connectionTimeoutRef = setTimeout(() => {
+                if (this.isFinished) { return }
+                this.logger.warn(`connection timed out after ${this.newConnectionTimeout}ms`)
+                this.close(new Error(`timed out after ${this.newConnectionTimeout}ms`))
+            }, this.newConnectionTimeout)
```
harbu: If the plan is to have a shared timeout logic within `WebRtcEndpoint` class, then my proposal would be to remove this with a method call to `WebRtcEndpoint`, e.g.

In WebRtcEndpoint:
```ts
protected resetConnectionTimeout(): void {
            clearTimeout(this.connectionTimeoutRef!)
            this.connectionTimeoutRef = setTimeout(() => {
                if (this.isFinished) { return }
                this.logger.warn(`connection timed out after ${this.newConnectionTimeout}ms`)
                this.close(new Error(`timed out after ${this.newConnectionTimeout}ms`))
            }, this.newConnectionTimeout)
}
```

And then replace the code here as follows:
```ts
        if (state === 'disconnected' || state === 'closed') {
            this.close()
        } else if (state === 'failed') {
            this.close(new Error('connection failed'))
        } else if (state === 'connecting') {
            this.resetConnectionTimeout()
        }
```

packages/network/src/connection/NodeWebRtcConnection.ts
```diff
@@ -0,0 +1,318 @@
...
+        }
+    }
+
+    private onGatheringStateChange(state: string): void {
+        this.logger.trace('conn.onGatheringStateChange: %s -> %s', this.lastGatheringState, state)
+        this.lastGatheringState = state
```
harbu, line 229: Same idea as with `lastState` above.

packages/network/src/connection/NodeWebRtcConnection.ts
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    private onDataChannel(dataChannel: DataChannel): void {
+        this.setupDataChannel(dataChannel)
+        this.logger.trace('connection.onDataChannel')
+        this.openDataChannel(dataChannel)
+    }
+
+    private onLocalDescription(description: string, type: DescriptionType): void {
+        this.emit('localDescription', type, description)
```
harbu: Let's consider moving the event emission to base class `WebRtcConnection`.

I.e. replace this with
```ts
private onLocalDescription(description: string, type: DescriptionType): void {
        super.onLocalDescription(type, description) // super equivalent?
}
```

And then in the base class `WebRtcConnection`:

```ts
protected onLocalDescription(description: string, type: DescriptionType): void {
    this.emit('localDescription', type, description)
}
```

Additionally we need to add a comment to the class JSDoc of `WebRtcEndpoint` that says that "implementers of sub classes need to take care to call protected method onLocalDescription whenever their underlying webrtc library generates a local description".

packages/network/src/connection/NodeWebRtcConnection.ts
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    private onLocalCandidate(candidate: string, mid: string): void {
+        this.emit('localCandidate', candidate, mid)
```
harbu: Same thinking as previous comment.

packages/network/src/connection/NodeWebRtcConnection.ts
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    private setupDataChannel(dataChannel: DataChannel): void {
+        this.paused = false
```
harbu: Again, we should look into moving the backpressure pausing logic to either be shared or subclass-specific. Shouldn't need to set superclass fields directly.

packages/network/src/connection/NodeWebRtcConnection.ts
```diff
@@ -0,0 +1,318 @@
...
+
+        this.dataChannelEmitter.on('error', (err) => {
+            this.logger.warn('dc.onError: %s', err)
+        })
+
+        this.dataChannelEmitter.on('bufferedAmountLow', () => {
+            if (!this.paused) { return }
+            this.paused = false
+            this.setFlushRef()
+            this.emit('bufferLow')
```
harbu: Again consider whether buffer handling logic should be shared or subclass-specific.

packages/network/src/connection/NodeWebRtcConnection.ts
```diff
@@ -0,0 +1,318 @@
...
+            this.logger.trace('dc.onmessage')
+            if (msg === 'ping') {
+                this.pong()
+            } else if (msg === 'pong') {
+                this.pingAttempts = 0
+                this.rtt = Date.now() - this.rttStart!
+            } else {
+                this.emit('message', msg.toString()) // TODO: what if we get binary?
+            }
+        })
```
harbu: The handling of the message here and checking whether ping should probably live under `WebRtcConnection`.

packages/network/src/connection/NodeWebRtcConnection.ts
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    private openDataChannel(dataChannel: DataChannel): void {
+        if (this.connectionTimeoutRef !== null) {
+            clearTimeout(this.connectionTimeoutRef)
+        }
```
harbu: Should be done by `WebRtcEndpoint`, e.g. within `emitOpen()`.

packages/network/src/connection/WebRtcConnection.ts
```diff
@@ -0,0 +1,402 @@
...
+    protected connectionTimeoutRef: NodeJS.Timeout | null
+    protected readonly newConnectionTimeout: number 
+    protected lastState: string | null
+    protected lastGatheringState: string | null
+    protected paused: boolean
+    protected readonly bufferThresholdHigh: number
+    protected readonly bufferThresholdLow: number
+    protected pingAttempts = 0
+    protected rtt: number | null
+    protected rttStart: number | null
```
harbu: Let's try to make as many as possible to `protected readonly`, if not possible keep `private` and provide getter.

harbu (changes requested):
See comments; I can also implement them if needed

packages/network/src/connection/NodeWebRtcConnection.ts
harbu: I can't use the same variable name `logger` in the subclass even though the variable with the same name is private in the base class. So just gonna skip this, because it feels funny to have to name the logger in the subclass as something else 🤷

harbu (approved):

PR #111, opened 2021-06-23
Refactoring suggestions

harbu:
I did the refactoring we discussed on call with @ptesavol (and which are described in my original PR review for https://github.com/streamr-dev/network-monorepo/pull/109). I'd rather not write a long description at this point, however, I can add a more thorough description of my changes, or we can jump on a call for more info, or feel free to ask questions here as needed.


harbu:
Another change that could be nice is to arrange the methods in the two classes to enhance readability, eg.:

```
// public methods, open, message, close
// abstract methods
// protected methods
// private methods
```

But I left this out so the diffs stay clean

juslesan (approved):

packages/network/src/connection/webRtcConnectionFactories.ts
```diff
@@ -0,0 +1,17 @@
+import { ConstructorOptions, WebRtcConnection } from './WebRtcConnection'
+import { NodeWebRtcConnection } from './NodeWebRtcConnection'
+import nodeDataChannel from 'node-datachannel'
+
+export interface WebRtcConnectionFactory {
```
ptesavol: I would suggest removing this file, and moving the WebRtConnectionFactory interface to the WebRtcEndpoint.ts file, and the NodeWebRtcConnectionFactory object to the NodeWebRtcConnection.ts file. This file is uneccessary, and developers of new WebRtcConnection subclasses might feel unconfortable about editing it (as opposod to merely adding new files to the repository)

ptesavol (changes requested):
Looks ok otherwise, but I would remove the separate webRtcConnectionFactories file (see the comment to that file)

PR #112, opened 2021-06-24
CLI-tools NET-360: Support path notation when defining a stream ID

teogeb:
f there is a stream parameter in a command, it can be defined as a full id (e.g. `0x1234567890123456789012345678901234567890/foo/bar`) or a path (e.g. `/foo/bar`). Using path notation requires authentication (`--private-key <key>`).

Updated also the example stream ID in the `README.md`.

### DIFF
--- a/packages/cli-tools/bin/common.ts
+++ b/packages/cli-tools/bin/common.ts
@@ -1,4 +1,5 @@
 import * as commander from 'commander'
+import { Wallet } from 'ethers'
 import { StreamrClientOptions } from 'streamr-client'
 
 export interface EnvironmentOptions {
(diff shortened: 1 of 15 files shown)


packages/cli-tools/bin/common.ts
```diff
@@ -90,4 +91,20 @@ export function createFnParseInt(name: string): (s: string) => number {
...
+    const PATH_PREFIX = '/'
+    if (!streamIdOrPath.startsWith(PATH_PREFIX)) {
+        return streamIdOrPath
+    }
+    const privateKey = options.privateKey
+    if (privateKey === undefined) {
+        console.error(`relative stream id ${streamIdOrPath} requires authentication`)
+        process.exit(1)
+    }
+    return new Wallet(privateKey).address.toLowerCase() + streamIdOrPath
```
harbu, line 109: Isn't this logic something we wanted to include within streamr-client-javascript itself?

packages/cli-tools/bin/common.ts
```diff
@@ -90,4 +91,20 @@ export function createFnParseInt(name: string): (s: string) => number {
         }
         return n
     }
+}
+
+export const createStreamId = (streamIdOrPath: string|undefined, options: any) => {
```
harbu: Maybe it's just me, but `createStreamId` conjures up an image of an object being created or something getting persisted / external API call. Something like `formStreamId` is more string-like, but dunno?

harbu (approved):

packages/cli-tools/bin/common.ts
teogeb: Yes, in the future this path support should be in the client. Currently it is not possible to add the support (would require too much refactoring). But we could re-check the situation after Brubeck refactoring has been done.

packages/cli-tools/bin/common.ts
teogeb: Renamed to `getStreamId`. The original `createStreamId` name was used because the method creates a new piece of information instead of getting it from some object. But agree that it may be misleading in the sense you described.

PR #113, opened 2021-06-24
NET-354 Broker: Testable plugin configs

teogeb:
It is now possible to validate a plugin config without creating an instance of a Plugin. 

A plugin functionality is exported to Broker via this interface:

```
interface PluginDefinition<T> {
    name: string,
    createInstance(options: PluginOptions): Plugin<T>
    getConfigSchema(): Schema|undefined
}
```

Added a test which validates all configs in `configs/` directory.


harbu (approved):
LGTM; Did you consider a model where we keep the validation in the constructor and for validation just create the instances but never invoke `start`?

teogeb:
With NET-190 (https://github.com/streamr-dev/network-monorepo/pull/100) and NET-29 (https://github.com/streamr-dev/network-monorepo/pull/116), this PR is now easily implementable without refactoring the `Plugin` interface. I'll open a new PR for that.

PR #114, opened 2021-06-28
NET-342: Interactive config init

bearni95:
When saying something like:

streamr-broker init

It would start in interactive mode, asking the following:

Do you want to import an Ethereum private key or generate a new one?

[G]enerate or [I]mport

Which of the [http, mqtt, websocket] plugins do you want to enable?

Then it saves a config file into the current directory

Set Ethereum key and address to whatever the user selected

Enable the plugins that the user selected with default config

Additionally, generate 1 random API key into the config file

Print to console what was done:

Saved config file to config.json 

You can start the broker now with streamr-broker config.json


packages/broker/bin/config-wizard.js
```diff
@@ -0,0 +1,18 @@
...
+const program = require('commander')
+
+const CURRENT_VERSION = require('../package.json').version
+
+program
+    .version(CURRENT_VERSION)
+    .name('broker-config-wizard')
+    .description('Run the configuration wizard for the broker')
+    .action(async () => {
+        console.log('action reached')
```
harbu: Unneeded log?

packages/broker/src/ConfigWizard.ts
```diff
@@ -0,0 +1,195 @@
+import inquirer from 'inquirer'
+import { Wallet } from 'ethers'
+import { writeFileSync } from 'fs'
+import path from 'path'
+
+import {Config} from './config'
```
```suggestion
import { Config } from './config'
```

packages/broker/src/ConfigWizard.ts
```diff
@@ -0,0 +1,195 @@
...
+
+import { Logger } from 'streamr-network'
+
+const logger = new Logger(module)
+
+const TODO_VALUES = {
```
harbu, line 12: Something here that should be done still with `TODO_VALUES`?

packages/broker/src/ConfigWizard.ts
```diff
@@ -0,0 +1,195 @@
...
+
+        return this.config.ethereumPrivateKey
+    }
+
+    async promptNumberWithDefault(prompt: inquirer.InputQuestion, defaultValue: number){
+        const valueString = await this.inquirerSinglePrompt(prompt)
+        let value = parseInt(valueString)
+        if (valueString === ''){
+            value = defaultValue
+        }
```
harbu: No need to run parseInt if it is empty
```suggestion
        const value = valueString === '' ? defaultValue : parseInt(valueString)
```

packages/broker/src/ConfigWizard.ts
```diff
@@ -0,0 +1,195 @@
...
+
+        if (!Number.isNaN(value) && Number.isInteger(value) && value > 1023 && value < 49151){
```
harbu: 1) Would be best if on giving an invalid value we would print an error and prompt user to give a number again.
2) Consider renaming to `promptPortNumberWithDefault` so the range requirement `(1023, 49151)` is more clear.
3) If a user gives a number but it is outside the range, we should print a different error indicating this for clarity.

packages/broker/src/ConfigWizard.ts
```diff
@@ -0,0 +1,195 @@
...
+                {name: 'Websocket'},
+                {name:'MQTT'},
+                {name:'HttpPublish'}
+            ]
+        })
+
+        const selectedPlugins = []
+        for (let i = 0; i < plugins.length; i++){
+            selectedPlugins.push(plugins[i])
+            if (plugins[i] === 'Websocket'){
```
harbu: You might want to configure your IDE to add whitespace after if/else if/else-statements for consistent styling with rest of code base.
```suggestion
            if (plugins[i] === 'Websocket') {
```

packages/broker/src/ConfigWizard.ts
```diff
@@ -0,0 +1,195 @@
...
+                }, this.defaultHttpPort)
+                this.config.plugins['httpPublish'] = { port: httpPort }
+            }
+        }
+
+        return selectedPlugins
+
+    }
+
+    async storeConfig(destinationFolder = '../configs/'){
```
harbu: Let's move the default value up to the exported function for clarity.
```suggestion
    async storeConfig(destinationFolder: string){
```

packages/broker/src/ConfigWizard.ts
```diff
@@ -0,0 +1,195 @@
...
+        const filename = `wizard-config.json`
+        const finalPath = path.join(__dirname, destinationFolder, filename)
+        writeFileSync(finalPath, JSON.stringify(this.config))
+        return finalPath
+    }
+
+    async start(destinationFolder?: string){
```
```suggestion
    async start(destinationFolder: string){
```

packages/broker/src/ConfigWizard.ts
```diff
@@ -0,0 +1,195 @@
...
+        logger.info('Broker Config Wizard ran succesfully')
+        logger.info('Generated configuration:', this.config)
+        logger.info(`Stored config under ${finalConfigPath}`)
+        logger.info(`You can start the broker now with \n streamr-broker ${finalConfigPath}`)
+        return finalConfigPath
+    }
+
+}
+
+export async function startBrokerConfigWizard (destinationFolder?: string): Promise<string> {
```
```suggestion
export async function startBrokerConfigWizard (destinationFolder = '../configs/'): Promise<string> {
```

packages/broker/test/unit/ConfigWizard.test.ts
```diff
@@ -0,0 +1,120 @@
+import {unlinkSync, existsSync} from 'fs'
```
harbu:
```suggestion
import { unlinkSync, existsSync } from 'fs'
```

harbu (changes requested):
Good stuff! See my comments

packages/broker/src/ConfigWizard.ts
bearni95: @harbu the TODO_VALUES are used to prefill fields like the ethereum address as well as containing those values that seem to have no standardized definition.

Here's the config object with the current TODO_VALUES being used, may be worth discussing in today's call what to do with some of them:

```
const DefaultConfig: Config = {
    ethereumPrivateKey: TODO_VALUES.string,//'{ETHEREUM_PRIVATE_KEY}',
    network: {
        name: TODO_VALUES.string,
        hostname: '127.0.0.1', // TODO
        port: TODO_VALUES.number,
        advertisedWsUrl: null,
        trackers: [ // TODO
            "ws://127.0.0.1:30301",
            "ws://127.0.0.1:30302",
            "ws://127.0.0.1:30303"
        ],
        location: {
            latitude: 60.19,
            longitude: 24.95,
            country: "Finland",
            city: "Helsinki"
        } // TODO
    },
    reporting: {
        intervalInSeconds: 0,
        streamr: null,
        perNodeMetrics: null
    },
    streamrUrl: 'http://127.0.0.1', // TODO
    streamrAddress: '0xFCAd0B19bB29D4674531d6f115237E16AfCE377c', // TODO
    storageNodeConfig: {
        registry: {
            contractAddress: '0xbAA81A0179015bE47Ad439566374F2Bae098686F', // TODO
            jsonRpcProvider: 'http://10.200.10.1:8546' // TODO
        }
    },
    httpServer: {
        privateKeyFileName: null,
        certFileName: null,
        port: TODO_VALUES.number//'{HTTP_SERVER_PORT}',
    },
    apiAuthentication: null,
    plugins: {}
}
```

harbu (approved):
lgtm

PR #116, opened 2021-07-01
NET-26: Broker configuration validation

teogeb:
New command line option `--test` to test the validity of the configuration file.

If the option is specified in the command line, a Broker instance is created, but not started:
- if an instance can be successfully created, the configuration file must be valid
- otherwise an Error is thrown (e.g `Error: must have required property 'ethereumPrivateKey'`)

To separate the creation and starting, a new `start()` method has been added to `Broker`. Some broker's initialization steps
have been encapsulated to helper methods, but the functionality has not been modified.

**Note:**  Some tasks executed by `createBroker()` call have external dependencies (at least `getStorageNodeRegistryFromContract()`),  and therefore it is possible that the `broker.js --test` fails even if the configuration is valid. 

**Future improvements:**
- Maybe it is possible to remove all interaction to external dependecies before `start()`, e.g. storage node registry could  be lazy loaded (see https://linear.app/streamr/issue/NET-363/refresh-the-storage-node-registry-information-from-contract)
- If we don't initialize any `setTimeout`s in `createBroker()`, we don't need to explicitly call `process.exit(0)`. Currently there  is only one leaking timeout in `PingPongWs` (created by `ClientWsEndpoint` from the `createNetworkNode()` call)


harbu (approved):
Good work! :+1:

PR #121, opened 2021-07-02
[JS Client] Fix for build error in React Native

kaxline:
There is an issue when adding `streamr-client` to a React Native project, where the build throws a warning:

```
warn Package streamr-client has been ignored because it contains invalid configuration. Reason: Package subpath './package.json' is not defined by "exports" in {ProjectRoot}/node_modules/streamr-client/package.json
```

Which is a problem with the `react-native cli` discussed a bit here:

https://github.com/react-native-community/cli/issues/1168

I found a fix for it here, which seems like a light fix that could preempt a lot of confusion and FUD by new adopters adding `streamr-client` to React Native:

https://github.com/infctr/eslint-plugin-typescript-sort-keys/pull/28/files

This PR applies the changes to `package.json` "exports" that should resolve the error/warning.

Let me know what you think!

### DIFF
--- a/packages/client/copy-package.js
+++ b/packages/client/copy-package.js
@@ -9,9 +9,9 @@ delete pkg.private
 pkg.types = `./${path.relative('./dist', pkg.types)}`
 pkg.main = `./${path.relative('./dist', pkg.main)}`
 pkg.browser = `./${path.relative('./dist', pkg.browser)}`
-pkg.exports.browser = `./${path.relative('./dist', pkg.exports.browser)}`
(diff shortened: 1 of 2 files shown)


timoxley:
Thanks for this. Note that `package.json` currently needs the `dist` in the path which will be rewritten by this script when publishing: https://github.com/streamr-dev/network-monorepo/blob/a336cc3db4f7beb71f1fc78a95aecac642111e4e/packages/client/copy-package.js#L8-L14

kaxline:
@timoxley Thanks for taking a look! Can you check my latest commit and see if that's what you had in mind?

PR #123, opened 2021-07-08
[JS Client] React native persistent store proposal

kaxline:
Hi @timoxley, the issue this attempts to solve is that neither BrowserPersistentStore or ServerPersistentStore works in React Native when encrypting and signing data for publishing to a stream. At least according to my testing and analysis (might be wrong!).

This isn't complete yet, but I wanted to check in to see what you think of the approach. I tried to follow the pattern of how the client handles the differences between the BrowserPersistentStore and ServerPersistentStore.

After implementing ReactNativePersistentStore with `@react-native-async-storage/async-storage`, I'm still running into issues, but I wanted to check in on the approach to see if it's even worth fixing those bugs.

Thanks for looking!


kaxline:
With the latest commits, this is working for me. I had to add `@react-native-async-storage/async-storage` to my project manually, but maybe there's a way to do that from the `streamr-client` code.

fonty1:
@timoxley when you get back up to speed could you make sure this PR resolves in some way :)

Added a Linear ticket, NET-668.

PR #124, opened 2021-07-08
Changing Exhange to Exchange in class and type names

kaxline:
This may have been intentional for some reason, since it was in a few different places, but just in case it's not, I'm submitting this PR.


timoxley (approved):
lgtm

PR #125, opened 2021-07-09
network [NET-368]: get rid of uws in network

harbu:
- Replace uWS with ws in `ServerWsEndpoint`
- tracker http endpoints are now implemented in `express` instead of `uWs`
- added boolean `stopped` to `AbstractWsEndpoint` to prevent events occurring after stopping endpoint


juslesan (approved):
LGTM

PR #127, opened 2021-07-15
NET-337: broker: storage node metrics have null values for Cassandra related fields

bearni95:
**Important changes**
- `VolumeLogger.close` is now an async method, fixing disconnection await
- Added `this.client.isConnected` check before publishing to temporarily address the scenario when the client has been closed but there is still a pending-to-fulfill promise that pushed metrics.
___

Earlier the metrics data was assumed to be in format:
```
{
  "writeBytes": 123,  // seconds
  "readyBytes": 456, // seconds
}
```

Now fixed to read the correct format, which is:
```
{
  "writeBytes": {
    "rate": 123000. // milliseconds
  },
  "readyBytes": {
    "rate": 456000 // milliseconds
  }
}
```

Assuming that data in readBytes and writeBytes is always of type `{rate: number, ...}` instead of raw numbers, as seen in:
https://github.com/streamr-dev/network-monorepo/blob/b3c47960669b2b22803f4ddd56467ec1ab6108ad/packages/network/src/helpers/MetricsContext.ts#L67

When Cassandra is disabled it's metrics will contain 0:
```
storage: {
  bytesWrittenPerSec: 0,
  bytesReadPerSec: 0,
}
```
May be worth considering to remove the `storage` entry altogether from the metrics.

The fix implemented in this PR has not automated tests. We updated `per-node-metrics.test.ts` to check the format of the report, but it doesn't test the actual data generation (lines 129-134 and 147-152) logic. It makes more sense to divide `per-node-metrics.test.ts` to separate tests that assert the actual data, and those tests could be inside `Storage.test.ts` etc.?


packages/broker/src/StreamMetrics.ts
```diff
@@ -138,8 +142,10 @@ export class StreamMetrics {
                     this.report.network.connections = throttledAvg(this.report.network.connections, (metricsReport.metrics.WebRtcEndpoint.connections as any).rate || 0)
 
                     if (metricsReport.metrics['broker/cassandra']) {
-                        this.report.storage.bytesWrittenPerSec = throttledAvg(this.report.storage.bytesWrittenPerSec, (metricsReport.metrics['broker/cassandra']) ? (metricsReport.metrics['broker/cassandra'].writeBytes as any).rate: 0)
-                        this.report.storage.bytesReadPerSec = throttledAvg(this.report.storage.bytesReadPerSec, (metricsReport.metrics['broker/cassandra']) ? (metricsReport.metrics['broker/cassandra'].readBytes as any).rate : 0)
+                        // @ts-expect-error
```
timoxley: What is ts complaining about in these places? Can we fix it at the source?

timoxley: I wonder if using `?.` would make ts happy, e.g. `metricsReport.metrics['broker/cassandra']?.writeBytes?.rate || 0`

timoxley (approved):
needs conflict fix but otherwise lgtm.

packages/broker/src/StreamMetrics.ts
```diff
@@ -124,8 +125,12 @@ export class StreamMetrics {
                     this.report.network.bytesToPeersPerSec = (metricsReport.metrics.WebRtcEndpoint.outSpeed as any).rate || 0
                     this.report.network.bytesFromPeersPerSec = (metricsReport.metrics.WebRtcEndpoint.inSpeed as any).rate || 0
                     this.report.network.connections = metricsReport.metrics.WebRtcEndpoint.connections || 0
-                    this.report.storage.bytesWrittenPerSec = (metricsReport.metrics['broker/cassandra']) ? metricsReport.metrics['broker/cassandra'].writeBytes : 0
-                    this.report.storage.bytesReadPerSec = (metricsReport.metrics['broker/cassandra']) ? metricsReport.metrics['broker/cassandra'].readBytes : 0
+
+                    const cassandraWrittenBytesPerMs = metricsReport.metrics['broker/cassandra'].writtenBytes as {rate: number,total: number, last: number}
+                    this.report.storage.bytesWrittenPerSec = cassandraWrittenBytesPerMs.rate / 1000 || 0
```
harbu: Are we sure that the `rate` provided by the `MetricsContext` is in milliseconds? I recall it being in seconds with a window size of 5.

packages/broker/src/StreamMetrics.ts
```diff
@@ -124,8 +125,12 @@ export class StreamMetrics {
...
+                    const cassandraReadBytesPerMs = metricsReport.metrics['broker/cassandra'].readBytes as {rate: number,total: number, last: number}
+                    this.report.storage.bytesReadPerSec = cassandraReadBytesPerMs.rate / 1000 || 0
```
harbu: Same as above.

packages/broker/src/StreamMetrics.ts
```diff
@@ -137,10 +142,11 @@ export class StreamMetrics {
...
 
-                    if (metricsReport.metrics['broker/cassandra']) {
-                        this.report.storage.bytesWrittenPerSec = throttledAvg(this.report.storage.bytesWrittenPerSec, (metricsReport.metrics['broker/cassandra']) ? (metricsReport.metrics['broker/cassandra'].writeBytes as any).rate: 0)
-                        this.report.storage.bytesReadPerSec = throttledAvg(this.report.storage.bytesReadPerSec, (metricsReport.metrics['broker/cassandra']) ? (metricsReport.metrics['broker/cassandra'].readBytes as any).rate : 0)
-                    }
+                    const cassandraWrittenBytesPerMs = metricsReport.metrics['broker/cassandra'].writtenBytes as {rate: number,total: number, last: number}
+                    const cassandraReadBytesPerMs = metricsReport.metrics['broker/cassandra'].readBytes as {rate: number,total: number, last: number}
+
+                    this.report.storage.bytesWrittenPerSec = throttledAvg(this.report.storage.bytesWrittenPerSec, cassandraWrittenBytesPerMs.rate / 1000 || 0)
```
harbu: Same as above.

packages/broker/src/StreamMetrics.ts
```diff
@@ -137,10 +142,11 @@ export class StreamMetrics {
...
+                    this.report.storage.bytesReadPerSec = throttledAvg(this.report.storage.bytesReadPerSec, cassandraReadBytesPerMs.rate / 1000 || 0)
```
harbu: Same as above.

packages/broker/test/integration/per-node-metrics.test.ts
```diff
@@ -3,10 +3,12 @@ import {startTracker, Tracker} from 'streamr-network'
 import { startBroker, createClient, STREAMR_DOCKER_DEV_HOST, createTestStream } from '../utils'
 import { Wallet } from 'ethers'
 import { Broker } from '../broker'
+import { Todo } from '../types'
```
harbu: Let's avoid using `Todo` in new code. It is there as a placeholder type for old code.

packages/broker/test/integration/per-node-metrics.test.ts
```diff
@@ -48,9 +50,40 @@ const fillMetrics = async (client: StreamrClient, count: number, nodeAddress: st
     return Promise.allSettled(promises)
 }
 
+const waitForMessage = (
+    stream: string,
+    client: StreamrClient, 
+): Promise<Todo> => {
```
harbu:
```suggestion
): Promise<any> => {
```

packages/broker/test/integration/per-node-metrics.test.ts
```diff
@@ -48,9 +50,40 @@ const fillMetrics = async (client: StreamrClient, count: number, nodeAddress: st
...
+    return new Promise((resolve, reject) => {
+        try {
+            client.subscribe({stream}, (res: Todo) => {
+                resolve(res)
+            })
```
harbu: Unfortunately `any` is the type that streamr-client gives us for `subscribe` cb function parameter :shrug: 
```suggestion
            client.subscribe({stream}, (res: any) => {
                resolve(res)
            })
```

packages/broker/test/integration/per-node-metrics.test.ts
```diff
@@ -48,9 +50,40 @@ const fillMetrics = async (client: StreamrClient, count: number, nodeAddress: st
...
+        } catch (e){
+            reject(e)
+        }
+    })
+}
+
+const expectMetrics = (res: Todo) => {
```
harbu:
```suggestion
const expectMetrics = (res: Record<string, unknown>) => {
```

packages/broker/test/integration/per-node-metrics.test.ts
```diff
@@ -81,21 +115,13 @@ describe('per-node metrics', () => {
         })
 
         storageNode = await startBroker({
-            name: 'storageNode',
+            name: 'storageNode', 
             privateKey: storageNodeAccount.privateKey,
+            networkPort: networkPort,
```
harbu: AFAIK `networkPort` is no longer used, where did this come from?

harbu (changes requested):

packages/broker/test/integration/per-node-metrics.test.ts
bearni95: without it the storage node doesn't start and the tests crash on the `before` sections

packages/broker/src/StreamMetrics.ts
bearni95: The logic ended up being a bit more involved but the @ts-expect-error is gone anyways

packages/broker/test/integration/per-node-metrics.test.ts
bearni95: Actually Eric was right, without the network port it does work fine, removed it

packages/broker/src/StreamMetrics.ts
bearni95: After checking the measurements are provided in seconds, with the logic around the 5s window being handled before the metrics are populated

packages/broker/src/StreamMetrics.ts
```diff
@@ -137,9 +143,11 @@ export class StreamMetrics {
                     this.report.network.bytesFromPeersPerSec = throttledAvg(this.report.network.bytesFromPeersPerSec, (metricsReport.metrics.WebRtcEndpoint.inSpeed as any).rate || 0)
                     this.report.network.connections = throttledAvg(this.report.network.connections, (metricsReport.metrics.WebRtcEndpoint.connections as any).rate || 0)
 
-                    if (metricsReport.metrics['broker/cassandra']) {
-                        this.report.storage.bytesWrittenPerSec = throttledAvg(this.report.storage.bytesWrittenPerSec, (metricsReport.metrics['broker/cassandra']) ? (metricsReport.metrics['broker/cassandra'].writeBytes as any).rate: 0)
-                        this.report.storage.bytesReadPerSec = throttledAvg(this.report.storage.bytesReadPerSec, (metricsReport.metrics['broker/cassandra']) ? (metricsReport.metrics['broker/cassandra'].readBytes as any).rate : 0)
+                    if (metricsReport.metrics['broker/cassandra']){
+                        const bytesWrittenPerSec = (metricsReport.metrics['broker/cassandra'].writeBytes as {rate: number}).rate ||  0
```
teogeb: As you mentioned "Assuming that data in readBytes and writeBytes is always of type {rate: number, ...}", this fallback `0` is most likely redundant

packages/broker/test/integration/per-node-metrics.test.ts
```diff
@@ -81,21 +121,12 @@ describe('per-node metrics', () => {
         })
 
         storageNode = await startBroker({
-            name: 'storageNode',
+            name: 'storageNode', 
             privateKey: storageNodeAccount.privateKey,
             trackerPort,
+            wsPort,
```
teogeb: The common setup is separate the storage nodes and publishe nodes. Maybe worth considering to use the same architecture?

teogeb (approved):
Good, let's add the tests in a later PR?

packages/broker/src/StreamMetrics.ts
```diff
@@ -253,7 +261,7 @@ export class StreamMetrics {
     }
 
     private async publishReport(): Promise<unknown> {
-        if (!this.stopped && this.targetStreamId) {
+        if (!this.stopped && this.client.isConnected && this.targetStreamId) {
```
harbu: What is this `this.client.isConnected` check for?

packages/broker/src/VolumeLogger.ts
```diff
@@ -168,7 +168,7 @@ export class VolumeLogger {
         const report = await this.metricsContext.report(true)
 
         // Report metrics to Streamr stream
-        if (this.client instanceof StreamrClient && this.legacyStreamId !== undefined) {
+        if (this.client instanceof StreamrClient && this.client.isConnected && this.legacyStreamId !== undefined) {
```
harbu: What is this `this.client.isConnected` check for?

packages/broker/src/VolumeLogger.ts
```diff
@@ -290,7 +290,7 @@ export class VolumeLogger {
         }
     }
 
-    close() {
+    async close() {
```
teogeb: Some test runs show a warning in when I run  `per-node-metrics.test.ts`:
```
WARN [2021-08-12T12:17:36.013] (StreamMetrics       ): Failed to publish to stream 0x4F2693C21088D490774f9FBa7d135707894E2875/streamr/node/metrics/hour due to: Error: Needs connection but – connection: disconnected & wants: DISCONNECTED
autoConnect: false & autoDisconnect: false with 1 handles.
autoConnect disabled automatically after explicit call to connect/disconnect().
```

The warning is not show every time, but in about 50% of the runs.

Were these warnings the reason why `stop` + `close` handling was modified in the latest commit? 

If this error is not related to Cassandra `null` values, we could maybe open another PR to fix the behaviour.

teogeb: Could the root cause be that there is a ongoing publish task when the Broker is closed?

E.g. would the following be a possible scenario:

1) A timer task triggers the report publish task. It calls `this.client.publish(this.targetStreamId, this.report)`. That method returns a promise,  so the task it is completed only once the promise returns. At any moment before that, we may need an open connection. 

2) Some time after publish a user or a test stops the Broker. The `broker.stop()` methods is called. Among other things it closes the `VolumeLogger`. There we  set `StreamMetrics` flag `stopped=true` and disconnect the client . Setting the `stopped=true` disables any future publishing, but it doesn't affect any ongoing publish tasks.

So if `broker.stop()` is called when `client.publish()` promise is pending, the `publish()` fail as `this.client.ensureDisconnected()` already disconnected the client.

bearni95: The check for `client.isConnected()` should deal with this scenario until the spin-off PR can fulfill this fix.

teogeb (approved):
Maybe worth adding a note to the PR description about the changes that were good improvements but not related to this PR:

- VolumeLogger.close is now an async method, fixed disconnection await
- added this.client.isConnected check before publishing (and what this change fixes)

PR #128, opened 2021-07-21
NET-329: Storage crash in production due to null payloads on cassandra

bearni95:
Null payloads on Cassandra messages are causing the following error upon resend:
```
TypeError: Cannot read property 'toString' of null
    at Storage._parseRow (/home/ubuntu/.nvm/versions/node/v14.15.0/lib/node_modules/streamr-broker/dist/src/storage/Storage.js:340:96)
    at Transform.transform [as _transform] (/home/ubuntu/.nvm/versions/node/v14.15.0/lib/node_modules/streamr-broker/dist/src/storage/Storage.js:348:33)
    at Transform._read (_stream_transform.js:205:10)
    at Transform._write (_stream_transform.js:193:12)
    at writeOrBuffer (_stream_writable.js:352:12)
    at Transform.Writable.write (_stream_writable.js:303:10)
    at ResultStream.ondata (_stream_readable.js:719:22)
    at ResultStream.emit (events.js:327:22)
    at ResultStream.Readable.read (_stream_readable.js:519:10)
    at flow (_stream_readable.js:992:34)
 ```

Since querying for the unindexed column `payload` is not performant more debug info should be provided whenever encountering the error.

The source of the null payloads is currently unknown and the proposed solution should ease it's detection.

Upon finding a null `payload` we log the querying parameters and not return the null-payload found message as part of the `resend` call.

### CODE REVIEW

File: packages/broker/test/sequential/plugins/storage/CassandraNullPayloads.test.ts [hunk 1]
```diff
@@ -34,50 +34,36 @@ const insertNullData = async (
...
 }) {
     const storePromises = []
     for (let i = 0; i < count; i++) {
         const timestamp = Math.floor((i / (count - 1)) * (1E10))
-        const msg = buildMsg({ streamId, streamPartition, timestamp, sequenceNumber: 0, publisherId: 'publisher1' })
+
+        const msg = new StreamMessage({
+            messageId: new MessageIDStrict(
+                streamId,
+                0, //streamPartition
```
teogeb: Maybe could remove comments as the data is just mock data and the test doesn't rely on specific values.

teogeb (approved):
LGTM 👍

PR #130, opened 2021-07-22
Net-372: allow dummy messages

juslesan:
Works with https://github.com/streamr-dev/countdown-script

### DIFF
--- a/packages/broker/src/plugins/testnetMiner/TestnetMinerPlugin.ts
+++ b/packages/broker/src/plugins/testnetMiner/TestnetMinerPlugin.ts
@@ -142,8 +142,4 @@ export class TestnetMinerPlugin extends Plugin<TestnetMinerPluginConfig> {
     getConfigSchema() {
         return PLUGIN_CONFIG_SCHEMA
     }
-
-    getDummyMessagesReceived(): number {
-        return this.dummyMessagesReceived
-    }
 }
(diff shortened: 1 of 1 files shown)


hpihkala (commented):
Can we also have a message type that gets printed to console (=logged at INFO level I guess)? For example check for `message.log` field.

packages/broker/src/plugins/testnetMiner/TestnetMinerPlugin.ts
```diff
@@ -57,7 +58,14 @@ export class TestnetMinerPlugin extends Plugin<TestnetMinerPluginConfig> {
             this.natType = await this.getNatType()
         }
         await this.streamrClient!.subscribe(this.pluginConfig.rewardStreamId, (message: any) => {
-            this.onRewardCodeReceived(message.rewardCode)
+            if (message.rewardCode) {
+                this.onRewardCodeReceived(message.rewardCode)
+            } if (message.info) {
+                logger.info(`Info message received: ${message.info}`)
```
hpihkala: Since we control the message content, maybe better to just drop the preceding text:
```suggestion
                logger.info(message.info)
```

hpihkala (changes requested):

hpihkala (approved):

PR #132, opened 2021-07-26
NET-390: broker (config wizard): safety measures to prevent overwriting private keys

bearni95:
If an existing config file exists in the selected path, ask the user if they want to overwrite it

Show the private key to the user and encourage them to store it safely after generate

Display the associated address after both generate and import

### DIFF
--- a/packages/broker/bin/config-wizard.js
+++ b/packages/broker/bin/config-wizard.js
@@ -12,4 +12,3 @@ program
     .description('Run the configuration wizard for the broker')
     
 startBrokerConfigWizard()
-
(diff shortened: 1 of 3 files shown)


timoxley (approved):
lgtm

harbu (approved):

harbu (approved):
LGTM; can you run eslint / tests and double-check they pass and then squash+merge?

bearni95:
> LGTM; can you run eslint / tests and double-check they pass and then squash+merge?

sure thing! getting to it now @harbu

PR #133, opened 2021-07-26
NET-384: Config should accept ethereumPrivateKey with and without 0x prefix

bearni95:
Currently the config requires the ethereumPrivateKey to have the 0x prefix, although this is not customary in the Ethereum space. Versions with and without the prefix should be accepted.

Currently, when attempting to give a private key without the prefix, the following error is printed:

Error: /ethereumPrivateKey must match pattern "^0x[a-f0-9]{64}$"


timoxley:
@bearni95 can we get a test on this

teogeb (approved):

harbu (approved):

PR #134, opened 2021-07-28
NET-289: Update to node16

juslesan:
- Run replace NodeJS 12 run times with NodeJS 16 (NET-353)
- Use 14 for publishing and browser run times
  - Separate run time for networks browser tests
- Fix failing tests in CI due to github-actions default Chrome update
- Fix protocol TrackerRegistry test to match docker


timoxley (changes requested):
LGTM, though the lint/build/browser tests probably should stay without the matrix, not sure it adds any value to run them on multiple node versions.

.github/workflows/protocol.yml
```diff
@@ -33,7 +33,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x, 14.x]
+        node-version: [12.x, 14.x, 16.x]
```
harbu: Do we need to keep 12 here?

harbu (approved):

.github/workflows/protocol.yml
juslesan: Don't think so I'll remove it

PR #136, opened 2021-08-03
Net 398: add STUN/TURN config to broker

juslesan:


packages/broker/src/broker.ts
```diff
@@ -53,6 +53,22 @@ const getStorageNodes = async (config: Config): Promise<StorageNodeRegistryItem[
     }
 }
 
+const getStunTurnUrls = (config: Config) => {
```
harbu:
```suggestion
const getStunTurnUrls = (config: Config): string[] | undefined => {
```

harbu (approved):
lgtm!

PR #142, opened 2021-08-10
Add node utils

juhah:
Adds two new methods to `Utils` object:

* `generateMnemonicFromAddress` to generate three word mnemonic from ETH address
* `parseAddressFromNodeId` to get a valid address from a node id


juslesan (approved):
LGTM!

PR #143, opened 2021-08-10
NET-320: broker metrics into a plugin

harbu:
Moved all metrics related logic under a single metrics plugin.

Further development ideas:
- Turn individual features into their own plugins (e.g. console logging, pm2 error & metrics collection, legacy metrics, and per node metrics could all be their own plugins)
- General refactoring of code


packages/broker/src/plugins/metrics/MetricsPlugin.ts
```diff
@@ -1,16 +1,99 @@
...
+        this.volumeLogger = createVolumeLogger(
+            this.brokerConfig.ethereumPrivateKey,
+            this.pluginConfig,
+            this.metricsContext,
+            this.nodeId,
+            this.storageNodeRegistry.getStorageNodes()
+        )
+    }
+
+    async start(): Promise<unknown> {
```
teogeb, line 40: In `publishHttp` plugin etc. there is no conditionality for httpServer but the Broker will throw an error if there is no config:
```
if (config.httpServer === null) {
    throw new Error('HTTP server config not defined')
}
```

Maybe the same logic could be applied here? Note also that in https://github.com/streamr-dev/network-monorepo/pull/145 we'll add a default port -> a `httpServer` config will always be available.

teogeb (approved):
Great 👍 Could maybe list the changed config property names to PR description? E.g. `reporting.intervalInSeconds` is now plugin's `consoleLogIntervalInSeconds`

packages/broker/src/plugins/metrics/MetricsPlugin.ts
```diff
@@ -1,16 +1,97 @@
...
+    let storageNodeAddress
+    if (config.perNodeMetrics && config.perNodeMetrics.intervals) {
+        reportingIntervals = config.perNodeMetrics.intervals
+        storageNodeAddress = config.perNodeMetrics.storageNode
     }
-}
\ No newline at end of file
+
+    return new VolumeLogger(
+        config.consoleLogIntervalInSeconds,
```
teogeb: This argument is used also in reportingIntervalSeconds for LegacyMetrics

harbu:
Approved

PR #144, opened 2021-08-10
NET-414: Refactor broker's ConfigWizard

bearni95:
Functionality changes:
- default plugin is `publishHttp` instead of `legacyPublishHttp`
- using [updated broker defaults](https://github.com/streamr-dev/network-monorepo/commit/dbc86d2b7df34ebdc30d55f31df45dce702e81c5)
___
**Important note for Windows**
Windows compatibility at some point. All path separators should use https://nodejs.org/api/path.html#path_path_sep e.g. via https://nodejs.org/api/path.html#path_path_join_paths
___ 

Abandon the class-driven implementation for the ConfigWizard:

```typescript
export class ConfigWizard{
    async generateOrImportPrivateKey(): Promise<Wallet>
    async selectPlugins(): Promise<Array<string>>
}
```

In favor of the suggested approach by `inquirer`'s author, much more flexible for testing and abstracting the user answers from the logic and config generation:

```typescript
let prompts: Array<inquirer.Question | inquirer.ListQuestion | inquirer.CheckboxQuestion> = [
    {
        type: 'list',
        name:'generateOrImportEthereumPrivateKey',
        message: 'Do you want to generate a new Ethereum private key or import an existing one?',
        choices: [PRIVATE_KEY_SOURCE_GENERATE, PRIVATE_KEY_SOURCE_IMPORT]
    },
    {
        type: 'input',
        name:'importPrivateKey',
        message: 'Please provide the private key to import',
        when: (answers: inquirer.Answers) => {
            return answers.generateOrImportEthereumPrivateKey === PRIVATE_KEY_SOURCE_IMPORT
        },
        validate: (input: string): string | boolean => {
            try {
                new Wallet(input)
                return true
            } catch (e) {
                if (e.message.includes('invalid hexlify value')){
                    return `Invalid privateKey provided for import: ${input}`
                } else {
                    return e.message
                }
            }
        }
    }
]

const pluginSelectorPrompt = {...}
const pluginPrompts = [{...}, {...}]

prompts = prompts.concat(pluginSelectorPrompt).concat(pluginPrompts)

const answers = await inquirer.prompt(prompts)
const config = getConfigFromAnswers(answers)
```


packages/broker/src/ConfigWizard.ts
```diff
@@ -1,38 +1,46 @@
...
+import * as os from 'os'
+import chalk from "chalk"
 
-import { Config } from './config'
+function logger(...args: any[]) {
+    console.log(chalk.bgWhite.black(':'), ...args)
+}
 
-import * as os from 'os'
+logger.info = (...args: any[]) => {
```
teogeb: Maybe all logger methods could be functions of a logger object? If there is no need that this "plain" logger is an exceptional case.

packages/broker/src/ConfigWizard.ts
```diff
@@ -1,38 +1,46 @@
...
-const DEFAULT_LEGACY_WS_PORT = 7173
+const MIN_PORT_VALUE = 1024
+const MAX_PORT_VALUE = 49151
 
-const DefaultConfig: Config = {
-    ethereumPrivateKey: '',
-    generateSessionId: false,
+import * as WebsocketConfigSchema from './plugins/websocket/config.schema.json'
+import * as MqttConfigSchema from './plugins/mqtt/config.schema.json'
+//import * as HttpConfigSchema from './plugins/publishHttp/config.schema.json'
```
teogeb: Remove comment?

packages/broker/src/ConfigWizard.ts
```diff
@@ -1,38 +1,46 @@
...
+import * as LegacyWebsocketConfigSchema from './plugins/legacyWebsocket/config.schema.json'
+
+
+const DEFAULT_WS_PORT = WebsocketConfigSchema.properties.port.default
+const DEFAULT_MQTT_PORT = MqttConfigSchema.properties.port.default
+const DEFAULT_HTTP_PORT = 8585//HttpConfigSchema.properties.port.default
```
teogeb: Why this is not read from configs? And the value is different from the Broker's default `httpServer` port?

bearni95: this is the one we were commenting on, the path is not the indicated since the plugin shouldn't have a default value for it.

packages/broker/src/ConfigWizard.ts
```diff
@@ -82,200 +84,164 @@ const DefaultConfig: Config = {
...
-            logger.info('This is your generated private key. Please store it in a secure location')
-            logger.alert(this.config.ethereumPrivateKey)
-        } else if (generateOrImport === 'import') {
-            const privateKey = await this.inquirerSinglePrompt({
-                type: 'input',
-                name: 'privateKey',
-                message: "'Please provide the private key to import'",
-            })
-
+export let prompts: Array<inquirer.Question | inquirer.ListQuestion | inquirer.CheckboxQuestion> = [
```
teogeb: Use `const` instead of `let` (there can be a temporary non-exported value to evaluate the prompts value, but export it as a const)

packages/broker/src/ConfigWizard.ts
```diff
@@ -82,200 +84,164 @@ const DefaultConfig: Config = {
...
+        name:'generateOrImportEthereumPrivateKey',
+        message: 'Do you want to generate a new Ethereum private key or import an existing one?',
+        choices: ['Generate', 'Import']
+    },
+    {
+        type: 'input',
+        name:'importPrivateKey',
+        message: 'Please provide the private key to import',
+        when: (answers: inquirer.Answers) => {
+            return answers.generateOrImportEthereumPrivateKey === 'Import'
```
teogeb: 'Generate' and 'Import' could be constants are we are referring to the value in multiple places

packages/broker/src/ConfigWizard.ts
```diff
@@ -82,200 +84,164 @@ const DefaultConfig: Config = {
...
+            return true
+        },
+        default: plugin.port
+    }
+})
+
+prompts = prompts.concat(pluginSelectorPrompt).concat(pluginPrompts)
+
+export const getConfigFromAnswers = (answers: any): Config => {
+    const config = DEFAULT_CONFIG
```
teogeb: Should we clone the DEFAULT_CONFIGS to the config variable? Now we are modifying the exported constant value (which is used e.g. in tests)

packages/broker/src/ConfigWizard.ts
```diff
@@ -82,200 +84,164 @@ const DefaultConfig: Config = {
                 }
             }
         },
-    }
+    },
 }
 
-export class ConfigWizard{
-    config: Config
-    // required to mock results on tests
```
teogeb, line 90: Is this mocking still used?

packages/broker/src/ConfigWizard.ts
bearni95: something like:
```typescript
function logger(...args: any[]) {
    console.log(chalk.bgWhite.black(':'), ...args)
}

logger.prototype = {
    info: (...args: any[]) => {
        console.log(chalk.bgWhite.black(':', ...args))
    },
    alert: (...args: any[]) => {
        console.log(chalk.bgYellow.black('!', ...args))
    },
    warn: (...args: any[]) => {
        console.log(chalk.bgYellow.black('!'), ...args)
    },
    error: (...args: any[]) => {
        console.log(chalk.bgRed.black('!'), ...args)
    }
}
```

packages/broker/src/ConfigWizard.ts
```diff
@@ -82,200 +84,164 @@ const DefaultConfig: Config = {
...
-            })
+export const selectDestinationPathPrompt = {
+    type: 'input',
+    name: 'selectDestinationPath',
+    message: `Select a path to store the generated config in `,
+    default: path.join(os.homedir(), '.streamr/broker-config.json'),
+    validate: (input: string, answers: inquirer.Answers = {}): string | boolean => {
+        try {
+            const filePath = input || answers.selectDestinationPath
+            const parentDirPath = filePath.substring(0, filePath.lastIndexOf('/'))
```
teogeb: Could use https://nodejs.org/api/path.html#path_path_dirname_path

packages/broker/src/ConfigWizard.ts
```diff
@@ -82,200 +84,164 @@ const DefaultConfig: Config = {
...
-                type: 'confirm',
-                name: 'overwrite',
-                message: `Config file ${finalPath} already exists. Overwrite?`,
-                default: false
-            })
+export const selectDestinationPathPrompt = {
+    type: 'input',
+    name: 'selectDestinationPath',
+    message: `Select a path to store the generated config in `,
+    default: path.join(os.homedir(), '.streamr/broker-config.json'),
```
teogeb, line 201: Not related to this PR, but worth checking the Windows compatibility at some point. All path separators should use https://nodejs.org/api/path.html#path_path_sep e.g. via https://nodejs.org/api/path.html#path_path_join_paths

packages/broker/src/ConfigWizard.ts
```diff
@@ -82,200 +84,164 @@ const DefaultConfig: Config = {
...
+            const filePath = input || answers.selectDestinationPath
+            const parentDirPath = filePath.substring(0, filePath.lastIndexOf('/'))
 
-            if (!overwrite) {
-                return this.storeConfig()
+            answers.parentDirPath = parentDirPath
+            answers.parentDirExists = existsSync(parentDirPath)
+            answers.fileExists = existsSync(filePath)
+
+            if (!answers.parentDirExists && !answers.fileExists){
```
teogeb: Minor enhancement: just return `(!answers.parentDirExists && !answers.fileExists)` without the if statement

packages/broker/src/ConfigWizard.ts
```diff
@@ -82,200 +84,164 @@ const DefaultConfig: Config = {
...
-            type: 'checkbox',
-            name:'selectedItems',
-            message: 'Select the plugins to enable',
-            choices: [
-                {name: 'Websocket'},
-                {name:'MQTT'},
-                {name:'HttpPublish'}
-            ]
-        })
+const pluginTemplates = {
```
teogeb: This is a constant -> rename to PLUGIN_TEMPLATES?

packages/broker/test/unit/ConfigWizard.test.ts
```diff
@@ -1,178 +1,202 @@
...
-    wizard.inquirer.prompt = () => Promise.resolve(mockedPromptResult)
-}
 describe('ConfigWizard', () => {
-    let wizard: ConfigWizard
-    let tmpDataDir: string
-    beforeAll(() => {
-        tmpDataDir = mkdtempSync(path.join(os.tmpdir(), 'broker-test-config-wizard'))
-    })
+    const importPrivateKeyPrompt = prompts[1]
+    const validatePortPrompt = prompts[3]
```
teogeb: Rename to `portPrompt`?

packages/broker/src/ConfigWizard.ts
bearni95: Something like `const config = Object.assign({}, DEFAULT_CONFIG)`?

packages/broker/test/unit/ConfigWizard.test.ts
```diff
@@ -1,178 +1,202 @@
...
+
+    describe('importPrivateKey validate', () => {
+        it ('happy path, prefixed', () => {
+            const validate = importPrivateKeyPrompt.validate!
+            const privateKey = Wallet.createRandom().privateKey
+            expect(privateKey.length).toBe(66)
```
teogeb: No need to assert the internal variable (also at line 23)

packages/broker/src/ConfigWizard.ts
```diff
@@ -82,200 +84,164 @@ const DefaultConfig: Config = {
...
+    message: `Select a path to store the generated config in `,
+    default: path.join(os.homedir(), '.streamr/broker-config.json'),
+    validate: (input: string, answers: inquirer.Answers = {}): string | boolean => {
+        try {
+            const filePath = input || answers.selectDestinationPath
+            const parentDirPath = filePath.substring(0, filePath.lastIndexOf('/'))
 
-            if (!overwrite) {
-                return this.storeConfig()
+            answers.parentDirPath = parentDirPath
```
teogeb, line 207: Do we need to store parentDirPath separately? If we already store the selectDestinationPath

packages/broker/src/ConfigWizard.ts
```diff
@@ -82,200 +84,164 @@ const DefaultConfig: Config = {
...
+            answers.parentDirExists = existsSync(parentDirPath)
```
teogeb, line 208: Also are these two variables needed? The facts about file existence are not answers that the user provided

packages/broker/src/ConfigWizard.ts
bearni95: We do in order to create, if necessary, the folder on the later logic step

packages/broker/src/ConfigWizard.ts
bearni95: Once again, needed later to generate the `~/.streamr` path, by default, if it didn't exist

packages/broker/test/unit/ConfigWizard.test.ts
```diff
@@ -1,178 +1,202 @@
...
+            const isValid = validate(invalidPath, answers)
+            expect(isValid).toBe(false)
+            expect(answers.parentDirExists).toBe(false)
+            expect(answers.fileExists).toBe(false)
+
         })
 
-        const importedWallet: Wallet = await wizard.generateOrImportPrivateKey()
-        expect(importedWallet.privateKey).toEqual(privateKey)
+        it ('happy path with overwrite destination', () => {
```
teogeb, line 63: The tested functionality was bit unclear (if the similar test at line 101 does test the overwrite functionality then maybe this could be removed)

packages/broker/src/ConfigWizard.ts
bearni95: Maybe renaming the previous approach is easier?

```typescript
const logger = {
    print: (...args: any[]) => {
        console.log(chalk.bgWhite.black(':'), ...args)
    },
    info: (...args: any[]) => {
        console.log(chalk.bgWhite.black(':', ...args))
    },
    alert: (...args: any[]) => {
        console.log(chalk.bgYellow.black('!', ...args))
    },
    warn: (...args: any[]) => {
        console.log(chalk.bgYellow.black('!'), ...args)
    },
    error: (...args: any[]) => {
        console.log(chalk.bgRed.black('!'), ...args)
    }
}
```

packages/broker/src/ConfigWizard.ts
teogeb: Or https://lodash.com/docs/4.17.15#cloneDeep, if you change e.g. the contents of config.plugins

packages/broker/src/ConfigWizard.ts
teogeb: But currently it is used in `publishHttp: { port: DEFAULT_HTTP_PORT }`. That http port question should be written into the root level httpServer block if user wants to change the default. And if user doesn't want to change, we shouldn't write any port value

packages/broker/src/ConfigWizard.ts
teogeb: Yes, we can calculate it there easily. If I understood correctly that selectDestinationPath is always path.dirname(selectDestinationPath), e.g. `/foo/bar/file.json` -> `/foo/bar/`

packages/broker/src/ConfigWizard.ts
teogeb: But less state to maintain if we calculate it when we need it

packages/broker/src/ConfigWizard.ts
bearni95: is it worth installing a whole new package just for this functionality?

packages/broker/src/ConfigWizard.ts
bearni95: Actually `(answers.parentDirExists || answers.fileExists)` but yup, good improvement

packages/broker/src/ConfigWizard.ts
```diff
@@ -82,200 +84,164 @@ const DefaultConfig: Config = {
...
-                    type: 'input',
-                    name: 'mqttPort',
-                    message: `Select a port for the MQTT Plugin [Enter for default: ${DEFAULT_MQTT_PORT}]`,
-                }, DEFAULT_MQTT_PORT)
-                this.config.plugins['mqtt'] = {
-                    port: mqttPort,
-                    payloadMetadata: false
-                }
+const pluginPrompts = Object.values(pluginTemplates).map((plugin, i) => {
+    const pluginName = Object.keys(pluginTemplates)[i]
```
teogeb: We can't know the iteration order of the object keys

packages/broker/test/unit/ConfigWizard.test.ts
```diff
@@ -1,178 +1,202 @@
...
-            destinationFolder: tmpDataDir,
-            overwrite: true,
+    describe('getConfigFromAnswers', () => {
+        it ('should exercise the `generate` path', () => {
+            const answers = {
+                generateOrImportEthereumPrivateKey: 'Generate',
+            }
+            const config = getConfigFromAnswers(answers)
+            expect(config.ethereumPrivateKey).toBeDefined()
+            expect(config.ethereumPrivateKey.match(/^(0x)?[a-f0-9]{64}$/)).not.toBe(null)
```
teogeb: Could use Jest's `stringMatching` (https://jestjs.io/docs/expect#expectstringmatchingstring--regexp) for all regexp checks

packages/broker/src/ConfigWizard.ts
bearni95: Can't we?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values

the `for...in` loop for `Object.values` should do the same string sorting as the one provided in `Object.keys`, right?

packages/broker/src/ConfigWizard.ts
```diff
@@ -82,200 +87,168 @@ const DefaultConfig: Config = {
...
-            type: 'input',
-            name: 'destinationFolder',
-            message: `Select a path to store the generated config in [Enter for default: ${defaultDestinationFolder}]`,
-        }, defaultDestinationFolder)
+export const createStorageFile = (config: Config, answers: inquirer.Answers): string => {
+    if (!answers.parentDirExists) {
+        mkdirSync(answers.parentDirPath)
     }
+   
+    writeFileSync(answers.selectDestinationPath, JSON.stringify(config))
```
teogeb: Should write the JSON with indentations (`JSON.stringify(this.config, null, 2)`), like it was previously

packages/broker/src/ConfigWizard.ts
```diff
@@ -82,200 +87,168 @@ const DefaultConfig: Config = {
...
-        }
-
-        if (!existsSync(returnedValue)) {
-            logger.warn(`Path [${returnedValue}] does not exist`)
-            return await this.promptExistingPathWithDefault(prompt, defaultValue)
-        }
+const PLUGIN_TEMPLATES = {
+    websocket: { port: DEFAULT_WS_PORT },
+    mqtt: { port: DEFAULT_MQTT_PORT },
+    publishHttp: { port: DEFAULT_HTTP_PORT }
```
teogeb: The `publishHttp` plugin doesn't have a `port` property in the plugin configuration. It uses whatever port is configured to `httpServer`. So we should maybe set the value to httpServer.port (line 62) if user wants to override the port. 

And if user doesn't override some port (a plugin port or `httpServer` port) it would be maybe easiest to leave the configuration out. For HTTP server we can leave the whole `httpServer` block out in that case.

packages/broker/src/ConfigWizard.ts
```diff
@@ -57,7 +63,6 @@ const DefaultConfig: Config = {
         privateKeyFileName: null,
```
teogeb: These `null` values are not needed as null is the default

packages/broker/src/ConfigWizard.ts
```diff
@@ -82,200 +87,168 @@ const DefaultConfig: Config = {
...
-
-    private async promptNumberWithDefault(prompt: inquirer.InputQuestion, defaultValue: number): Promise<number>{
-        const valueString = await this.inquirerSinglePrompt(prompt)
-        const value = valueString === '' ? defaultValue: parseInt(valueString)
-
-        if (Number.isNaN(value) || !Number.isInteger(value)) {
-            logger.warn(`Non-numeric value [${valueString}] provided`)
-            return await this.promptNumberWithDefault(prompt, defaultValue)
-        }
+const promptValues = {
```
teogeb: Would maybe give some more specific name? Or just two constants, e.g. `PRIVATE_KEY_SOURCE_GENERATE` and `PRIVATE_KEY_SOURCE_IMPORT`

packages/broker/test/unit/ConfigWizard.test.ts
```diff
@@ -1,178 +1,189 @@
...
+
+        beforeAll(() => {
+            tmpDataDir = mkdtempSync(path.join(os.tmpdir(), 'broker-test-config-wizard'))
+        })
+
+        beforeEach(() => {
+            config = DEFAULT_CONFIG as Config
+        })
+
+        it ('happy path with overwrite destination', () => {
```
teogeb: This test used to be a test the actual storing (`expect(existsSync(configFileLocation)).toBe(true)`) and now it tests the validation logic. Could move into "storagePrompt validation" section if the main focus is on validation.

packages/broker/src/ConfigWizard.ts
teogeb: True, maybe easiest solution is to clone both config and plugins:
```
const config = { ... DEFAULT_CONFIG, plugins: { ... DEFAULT_CONFIG.plugins } }
```

packages/broker/src/ConfigWizard.ts
```diff
@@ -82,200 +87,168 @@ const DefaultConfig: Config = {
...
+    }
+})
+
+prompts = prompts.concat(pluginSelectorPrompt).concat(pluginPrompts)
+
+export const getConfigFromAnswers = (answers: any): Config => {
+    const config = Object.assign({}, DEFAULT_CONFIG)
+    const pluginNames = Object.keys(PLUGIN_TEMPLATES)
+    const PLUGIN_TEMPLATESArray = Object.values(PLUGIN_TEMPLATES)
+    for (let i = 0; i < PLUGIN_TEMPLATESArray.length; i++){
```
teogeb: This loop could be simplified? Something like this:
```
pluginNames.forEach((name) => {
  const template = PLUGIN_TEMPLATES[name]
  ...
})
```

packages/broker/src/ConfigWizard.ts
teogeb: You are right 👍 Wasn't aware of that the matching ordering is a documented feature. 

(Still it would be maybe more readable and more straightforward if it would be something like)
```
Object.keys(pluginTemplates).map((pluginName) => {
  const plugin = pluginTemplates[name]
})
```

packages/broker/src/ConfigWizard.ts
bearni95: since the `httpServer` property has it's own type the values are required to be present, even if null

packages/broker/src/ConfigWizard.ts
bearni95: Added the note to the PR description, will make a better explanation out of it since Windows support is a whole thing

packages/broker/src/ConfigWizard.ts
```diff
@@ -82,200 +84,170 @@ const DefaultConfig: Config = {
...
+
+export const getConfigFromAnswers = (answers: any): Config => {
+    const config = { ... DEFAULT_CONFIG, plugins: { ... DEFAULT_CONFIG.plugins } }
+
+    const pluginNames = Object.keys(PLUGIN_TEMPLATES)
+    pluginNames.forEach((pluginName) => {
+        const template = PLUGIN_TEMPLATES[pluginName]
+        if (answers.selectPlugins && answers.selectPlugins.includes(pluginName)){
+            // the publishHttp plugin is special, it needs to be added to the config after the other plugins
+            if (pluginName === 'publishHttp') {
```
teogeb: If we refer to this value twice, it should be a constant (maybe all plugin names could be)

bearni95: with the latest commit it is not repeated anymore, seems quite redundant to implement all plugin names as constants for single usage

packages/broker/src/ConfigWizard.ts
```diff
@@ -164,12 +171,17 @@ export const getConfigFromAnswers = (answers: any): Config => {
...
+                config.plugins![pluginName] = {}
+            } else if (pluginName === 'publishHttp') {
+                // the publishHttp plugin is special, it needs to be added to the config after the other plugins
+                config.plugins![pluginName] = {}
                 config.httpServer = {
                     port: answers[`${pluginName}Port`]
                 }
             } else {
+                // use provided a custom value, fill in
                 config.plugins![pluginName] = { ...template, port: answers[`${pluginName}Port`] }
```
teogeb: Is the `...template` part needed anymore?

bearni95: if there ever were fields to be added into the template that should be carried along to the generated config the current structure would allow them to

packages/broker/src/ConfigWizard.ts
```diff
@@ -164,12 +171,17 @@ export const getConfigFromAnswers = (answers: any): Config => {
     pluginNames.forEach((pluginName) => {
         const template = PLUGIN_TEMPLATES[pluginName]
         if (answers.selectPlugins && answers.selectPlugins.includes(pluginName)){
-            // the publishHttp plugin is special, it needs to be added to the config after the other plugins
-            if (pluginName === 'publishHttp') {
+            if (answers[`${pluginName}Port`] === template.port){
+                // user selected default value, enable without specifying the port
+                config.plugins![pluginName] = {}
```
teogeb: If we modify `config.plugins![pluginName]` in each section, maybe that could be outside the `if`s

E.g.
```
let pluginConfig = {}
if ... {
...
}
config.plugins![pluginName] = pluginConfig
```

packages/broker/src/ConfigWizard.ts
```diff
@@ -52,12 +67,6 @@ const DefaultConfig: Config = {
...
     },
-    httpServer: {
-        port: DEFAULT_HTTP_PORT,
-        privateKeyFileName: null,
-        certFileName: null
-    },
-    apiAuthentication: null,
     plugins: {
         legacyWebsocket: {
             port: DEFAULT_LEGACY_WS_PORT
```
teogeb: This could be removed

bearni95: They've been, hence the red coloring

teogeb: Mean the `DEFAULT_LEGACY_WS_PORT`

packages/broker/src/ConfigWizard.ts
teogeb: Ok 👍 Do we support that future feature for `publishHttp` too?

bearni95: not at the moment since it has it's own special path having to do with `httpServer`

packages/broker/src/ConfigWizard.ts
bearni95: 👍

packages/broker/src/ConfigWizard.ts
teogeb: This may be confusing if someone assumes that the feature is working ok for all plugins. Maybe easiest is to drop that redundant feature

bearni95: Applied proposed solution

packages/broker/src/ConfigWizard.ts
```diff
@@ -82,200 +88,175 @@ const DefaultConfig: Config = {
...
+        default: plugin.port
+    })
+})
+
+prompts = prompts.concat(pluginSelectorPrompt).concat(pluginPrompts)
+
+export const getConfigFromAnswers = (answers: any): Config => {
+    const config = { ... DEFAULT_CONFIG, plugins: { ... DEFAULT_CONFIG.plugins } }
+
+    const pluginNames = Object.keys(PLUGIN_TEMPLATES)
```
teogeb: Now this is not a template anymore, rename to PLUGIN_DEFAULT_PORTS or something? E.g.

```
{
  websocket: DEFAULT_WS_PORT
}
```

packages/broker/src/ConfigWizard.ts
teogeb: It is still at line 126 and line 173
```
publishHttp: { port: DEFAULT_HTTP_PORT }
if (pluginName === 'publishHttp') {
```

packages/broker/src/ConfigWizard.ts
```diff
@@ -82,200 +88,181 @@ const DefaultConfig: Config = {
...
-            type: 'checkbox',
-            name:'selectedItems',
-            message: 'Select the plugins to enable',
-            choices: [
-                {name: 'Websocket'},
-                {name:'MQTT'},
-                {name:'HttpPublish'}
-            ]
-        })
+    return config as Config
```
teogeb: The JSON is not a `Config` type (which is the type definition for the configuration file that we get **after** we parse it by `Ajv`)

packages/broker/test/unit/ConfigWizard.test.ts
```diff
@@ -1,178 +1,207 @@
...
+    describe('createStorageFile', () => {
+        let tmpDataDir: string
+        let config: any
+
+        beforeAll(() => {
+            tmpDataDir = mkdtempSync(path.join(os.tmpdir(), 'broker-test-config-wizard'))
+        })
+
+        beforeEach(() => {
+            config = { ... DEFAULT_CONFIG, plugins: { ... DEFAULT_CONFIG.plugins } }
```
teogeb: Why is this needed? The contents of the config is not relevant for these tests, could use just a constant `MOCK_CONFIG = {}`

bearni95: A `plugins` key would be needed to satisfy the logic. Can we revert to  `config = DEFAULT_CONFIG` ?

teogeb: But the value is assigned only in beforeEach, so it is a constant (and if you don't change the contents in any way, why to make a copy)

bearni95: So can we revert to `config = DEFAULT_CONFIG`?

teogeb: `createStorageFile` is the method which we are testing here. It accepts `config: any`. So an empty JSON is a valid input (if no need to be more complicated, the simplest input is the best)

bearni95: Terrific, implemented on the next commit!

teogeb (approved):
LGTM!

PR #145, opened 2021-08-11
NET-404 broker: config defaults

teogeb:
Many config options for Broker and plugins have now some default values. 

Most of the defaults are not environment specific. If a default value is environment specific, the value is for the production environment.

The minimal config is currently like this:
```
{
    "ethereumPrivateKey": "foo",
    "network": {
        "name": "bar"
    },
    "reporting": {
        "intervalInSeconds": 0,
        "streamr": null,
        "perNodeMetrics": null
    },
    "plugins": {}
}
```

Most of the plugins can be initialized with empty setup:
```
    "plugins": {
        "websocket": {},
        "mqtt": {},
        "publishHttp": {},
        "legacyWebsocket": {},
        "testnetMiner": {},
        "metrics": {},
        "storage": {
            "cassandra": {
                "hosts": ["127.0.0.1"],
                "username": "",
                "password": "",
                "keyspace": "streamr_dev_v2",
                "datacenter": "datacenter1"
              }
        }
    }
}
```

**TODO:**
- Reporting config has yet not been modified as it will be moved into a plugin (https://github.com/streamr-dev/network-monorepo/pull/143)
- Config wizard config template has not yet been simplified as there is a pending refactor PR (https://github.com/streamr-dev/network-monorepo/pull/144)

**Future improvements:**
- The `websocket` plugin could use Broker's `httpServer`. That would simplify maintenance as there is be one port less to configure, and possible SSL certificates could be shared.

### DIFF
--- a/packages/broker/plugins.md
+++ b/packages/broker/plugins.md
@@ -218,16 +218,11 @@ plugins: {
 }
 ```
 
-The plugin  uses the Broker's global `httpServer`. It is configured in the root level:
+The default HTTP server port is `8080`. You can change it by specifying a `port` value for the root level `httpServer` option:
+
 ```
 "network": ...
 "plugins": ...
-"httpServer": {}
-```
-
-The default httpServer port is `8080`. You can change it with `port` config option:
-
-```
 "httpServer": {
     "port": 1234
 }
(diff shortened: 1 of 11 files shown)


harbu (approved):
LGTM; can you run eslint / tests and double-check they pass and then squash+merge?

PR #147, opened 2021-08-12
NET-416: pass nat type and broker version metadata to tracker

harbu:
- Package `network` now provides a way to share metadata with the tracker. The latest metadata gets shared with tracker(s) via status messages. The tracker will store the metadata (per node) in memory and provide it to the outside world via HTTP endpoint `/tracker/`.
- `TestnetMinerPlugin` in package `broker` uses the new metadata sharing feature to share `natType` and `brokerVersion` fields with tracker (and the rest of the world).

### DIFF
--- a/packages/broker/src/plugins/testnetMiner/TestnetMinerPlugin.ts
+++ b/packages/broker/src/plugins/testnetMiner/TestnetMinerPlugin.ts
@@ -7,6 +7,7 @@ import { Metrics } from '../../../../network/dist/helpers/MetricsContext'
 import { scheduleAtInterval } from '../../helpers/scheduler'
 import { withTimeout } from '../../helpers/withTimeout'
 import { fetchOrThrow } from '../../helpers/fetch'
+import { version as CURRENT_VERSION } from '../../../package.json'
 
 const REWARD_STREAM_PARTITION = 0
 const LATENCY_POLL_INTERVAL = 30 * 60 * 1000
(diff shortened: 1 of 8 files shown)


packages/broker/src/plugins/testnetMiner/TestnetMinerPlugin.ts
```diff
@@ -57,6 +58,10 @@ export class TestnetMinerPlugin extends Plugin<TestnetMinerPluginConfig> {
         if (this.pluginConfig.stunServerHost !== null) {
             this.natType = await this.getNatType()
         }
+        this.networkNode.setExtraMetadata({
+            natType: this.natType || null,
```
harbu, line 62: `this.natType` may be `undefined`. The problem was that when this was the case, the serialized version of JSON did not contain the key `natType` at all!

packages/broker/test/integration/plugins/testnetMiner/TestnetMinerPlugin.test.ts
```diff
@@ -99,23 +101,31 @@ describe('TestnetMinerPlugin', () => {
         })
     })
 
+    afterAll(async () => {
```
harbu, line 104: Moved this block here to be consistent with rest of code base:
```ts
beforeAll(() => {
...
})

afterAll(() => {
...
})

// tests
```

packages/network/src/logic/Tracker.ts
```diff
@@ -67,6 +68,7 @@ export class Tracker extends EventEmitter {
         this.overlayConnectionRtts = {}
         this.locationManager = new LocationManager()
         this.instructionCounter = new InstructionCounter()
+        this.extraMetadatas = Object.create(null)
```
harbu, line 71: `Object.create()` is preferable to `{}` due to less surprises.

E.g. with literal
```ts
const o1 = {}
o1['toString'] !== undefined // true

const o2 = Object.create(null)
o2['toString'] === undefined // true
```

juslesan (approved):

PR #149, opened 2021-08-16
fix: config-wizard default stun

juslesan:
- Point ConfigWizard default stun url to the correct address (Required change before next internal testnet)


harbu (approved):
LGTM; feel free to squash+merge

PR #154, opened 2021-08-19
fix: broker default ports

hpihkala:
Let's use the following default ports for the plugins. This was changed already once, but the chosen ports (such as 8080) unfortunately clash with commonly used http servers (such as the tomcat we run in our own dev env).

- websocket: 7170
- http: 7171
- mqtt: 1883

### DIFF
--- a/packages/broker/plugins.md
+++ b/packages/broker/plugins.md
@@ -52,7 +52,7 @@ socket.addEventListener('open', () => {
 
 #### Port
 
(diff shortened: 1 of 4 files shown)


teogeb (approved):

hpihkala:
CI failed due to some TS problems in protocol package, seems unrelated

timoxley:
Fixed those TS type issues here: https://github.com/streamr-dev/network-monorepo/pull/157

PR #155, opened 2021-08-19
Simplify running broker in docker: Run broker with no args as default Dockerfile command.

timoxley:
Needs this to land first: https://github.com/streamr-dev/streamr-docker-dev/pull/139

* Run brokers with no args as default command. Removes `CONFIG_FILE` env var. This simplifies running broker in docker for testnet users.
* Removes unneeded EXPOSE statements from Dockerfile.
* Uses npm v6 instead of npm v7. Installs with npm v7 do not seem to succeed reliably. 🤷  Our package-locks are set up for v6 anyway.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,7 @@ FROM node:16-buster as build
 WORKDIR /usr/src/monorepo
 COPY . .
 RUN npm set unsafe-perm true
+RUN npm install -g npm@6 # explicitly use npm v6
 RUN npm ci
 RUN npm run bootstrap-pkg streamr-broker
 RUN npx lerna exec -- npm prune --production # image contains all packages, remove devDeps to keep image size down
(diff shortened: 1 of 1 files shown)


hpihkala (approved):
LGTM assuming you tried building it or CI tests that the Dockerfile works

Dockerfile
```diff
@@ -2,6 +2,7 @@ FROM node:16-buster as build
 WORKDIR /usr/src/monorepo
 COPY . .
 RUN npm set unsafe-perm true
+RUN npm install -g npm@6 # explicitly use npm v6
```
harbu, line 5: :+1:

harbu (approved):

timoxley:
> LGTM assuming you tried building it or CI tests that the Dockerfile works

Built it locally. Seems to work.

PR #156, opened 2021-08-20
network, broker [NET-347]: report webrtc failed connections

harbu:


packages/network/src/connection/WebRtcConnection.ts
```diff
@@ -162,6 +163,7 @@ export abstract class WebRtcConnection extends ConnectionEmitter {
             if (this.isFinished) { return }
             this.baseLogger.warn(`connection timed out after ${this.newConnectionTimeout}ms`)
             this.close(new Error(`timed out after ${this.newConnectionTimeout}ms`))
+            this.emit('timeout')
```
juslesan: The metric for failed connections should be more accurate if it is measured in the close() function when errors are present. This would include ie. 'failed' states for WebRTC. With the current implementation in such situations only the other end will report the error

juslesan (changes requested):

packages/network/src/connection/WebRtcConnection.ts
harbu: Check my newest commit. I made it include more error cases. The ticket asked for "The metric should be incremented if a WebRTC connection fails to open.". To test for this, I added a boolean field `hasOpened`. This way we don't count failures that happened after connection was successfully established.

juslesan (approved):

PR #158, opened 2021-08-23
NET-342: ConfigWizard generation of apiAuthentication.keys

bearni95:
Added `apiAuthentication` field with the following structure to the DEFAULT_CONFIG:
```typescript
"apiAuthentication": {
    "keys": [
      "MTEzNzFkMDQ4Mjg3NDhlNTk4M2FhMGFhOGNhMDI1MjQ"
    ]
  },
```

The value of the keys comes from the `generateApiKey` method, an alphanumeric string encoded in base64:
```typescript
function generateApiKey(){
    const hex = uuid().split('-').join('')
    return Buffer.from(hex).toString('base64').replace(/[^0-9a-z]/gi, '')
}
```

___ 

Modified tests to be system-agnostic, as the `fs` error codes for Ubuntu and Mac are not the same, and the error code irrelevant for our testing purposes:
```typescript
// before
 } catch(e){
    expect(e.code).toBe('EEXIST')
    expect(e.syscall).toBe('mkdir')
// now
} catch(e) {
    expect(e).not.toBeUndefined()
}
```

### DIFF
--- a/packages/broker/src/ConfigWizard.ts
+++ b/packages/broker/src/ConfigWizard.ts
@@ -50,7 +50,7 @@ const logger = {
 
 function generateApiKey(){
     const hex = uuid().split('-').join('')
-    return Buffer.from(hex).toString('base64')
+    return Buffer.from(hex).toString('base64').replace(/[^0-9a-z]/gi, '')
 }
 
 export const DEFAULT_CONFIG: any = {
--- a/packages/broker/src/ConfigWizard.ts
+++ b/packages/broker/src/ConfigWizard.ts
@@ -48,7 +48,7 @@ const logger = {
     }
 }
 
-const generateApiKey = ():string => {
+const generateApiKey = (): string => {
     const hex = uuid().split('-').join('')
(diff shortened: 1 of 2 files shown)


packages/broker/src/ConfigWizard.ts
```diff
@@ -48,6 +48,11 @@ const logger = {
     }
 }
 
+function generateApiKey(){
+    const hex = uuid().split('-').join('')
+    return Buffer.from(hex).toString('base64')
```
teogeb: It would be good to generate alpha-numeric API keys. This can generate a key "ODUwMjg0YmVkMDMwNGNlYmIxOThlZGNhNzNiM2M2M2U="

packages/broker/src/ConfigWizard.ts
```diff
@@ -48,6 +48,11 @@ const logger = {
     }
 }
 
+function generateApiKey(){
```
```suggestion
function generateApiKey(): string {
```

harbu (approved):

teogeb (approved):

PR #161, opened 2021-08-24
NET-437: Broker's config wizard UX/UI polishing

bearni95:
- The wizard now asks the user if they want their generated private key displayed. 
- The logger has been simplified and colors reduced to the minimal.
- Import of private keys is now executed as a `password` input rather than plaintext.
- The ordering of the flow has changed for a better user experience.

### DIFF
--- a/packages/broker/src/ConfigWizard.ts
+++ b/packages/broker/src/ConfigWizard.ts
@@ -32,11 +32,17 @@ const generateApiKey = (): string => {
     return Buffer.from(hex).toString('base64').replace(/[^0-9a-z]/gi, '')
 }
 
-export const DEFAULT_CONFIG_PORTS = {
+export const DEFAULT_CONFIG_PORTS: { [plugin: string]: number } = {
     WS: WebsocketConfigSchema.properties.port.default,
     MQTT: MqttConfigSchema.properties.port.default,
     HTTP: BrokerConfigSchema.properties.httpServer.properties.port.default,
     LEGACY_WS: LegacyWebsocketConfigSchema.properties.port.default,
+}
+
+const PLUGIN_NAMES: {[pluginName: string]: string} = {
+    WS: 'websocket',
+    MQTT: 'mqtt',
+    HTTP: 'publishHttp'
 }
 
 const PRIVATE_KEY_SOURCE_GENERATE = 'Generate'
@@ -104,7 +110,7 @@ const PRIVATE_KEY_PROMPTS: Array<inquirer.Question | inquirer.ListQuestion | inq
             try {
                 new Wallet(input)
                 return true
-            } catch (e) {
+            } catch (e: any) {
                 if (e.message.includes('invalid hexlify value')){
                     return `Invalid privateKey provided for import: ${input}`
                 } else {
@@ -124,18 +130,6 @@ const PRIVATE_KEY_PROMPTS: Array<inquirer.Question | inquirer.ListQuestion | inq
     }
 ]
 
-const PLUGIN_DEFAULT_PORTS: {[pluginName: string]: number} = {
-    websocket: DEFAULT_CONFIG_PORTS.WS,
-    mqtt: DEFAULT_CONFIG_PORTS.MQTT,
-    publishHttp: DEFAULT_CONFIG_PORTS.HTTP,
-}
-
-const PLUGIN_NAMES: {[pluginName: string]: string} = {
-    WEBSOCKET: 'websocket',
-    MQTT: 'mqtt',
-    PUBLISH_HTTP: 'publishHttp'
-}
-
 const createPluginPrompts = (): Array<inquirer.Question | inquirer.ListQuestion | inquirer.CheckboxQuestion> => {
     const selectPrompt: inquirer.CheckboxQuestion = {
         type: 'checkbox',
@@ -144,8 +138,8 @@ const createPluginPrompts = (): Array<inquirer.Question | inquirer.ListQuestion 
         choices: Object.values(PLUGIN_NAMES)
     }
 
-    const portPrompts: Array<inquirer.Question> = Object.keys(PLUGIN_DEFAULT_PORTS).map((name) => {
-        const defaultPort = PLUGIN_DEFAULT_PORTS[name]
+    const portPrompts: Array<inquirer.Question> = Object.keys(DEFAULT_CONFIG_PORTS).map((name) => {
+        const defaultPort = DEFAULT_CONFIG_PORTS[name]
         const MIN_PORT_VALUE = 1024
         const MAX_PORT_VALUE = 49151
         return {
@@ -191,7 +185,7 @@ export const selectStoragePathPrompt = {
             answers.fileExists = existsSync(input)
 
             return true
-        } catch (e) {
+        } catch (e: any) {
             return e.message
         }
     }
@@ -226,15 +220,17 @@ export const getConfig = (privateKey: string, pluginsAnswers: inquirer.Answers):
     const config = { ... CONFIG_TEMPLATE, plugins: { ... CONFIG_TEMPLATE.plugins } }
     config.ethereumPrivateKey = privateKey
 
-    const pluginNames = Object.values(PLUGIN_NAMES)
-    pluginNames.forEach((pluginName) => {
-        const defaultPluginPort = PLUGIN_DEFAULT_PORTS[pluginName]
+    const pluginNames = Object.keys(PLUGIN_NAMES)
+    pluginNames.forEach((pluginKey) => {
+        const pluginName = PLUGIN_NAMES[pluginKey]
+        const defaultPort = DEFAULT_CONFIG_PORTS[pluginKey]
         if (pluginsAnswers.selectPlugins && pluginsAnswers.selectPlugins.includes(pluginName)){
             let pluginConfig = {}
             const portNumber = parseInt(pluginsAnswers[`${pluginName}Port`])
-            if (portNumber !== defaultPluginPort){
+            console.log('portNumber', portNumber, defaultPort)
+            if (portNumber !== defaultPort){
                 const portObject = { port: portNumber }
-                if (pluginName === PLUGIN_NAMES.PUBLISH_HTTP) {
+                if (pluginName === PLUGIN_NAMES.HTTP) {
                     // the publishHttp plugin is special, it needs to be added to the config after the other plugins
                     config.httpServer = portObject
                 } else {
@@ -287,7 +283,7 @@ export const start = async(): Promise<void> => {
         logger.info(networkExplorerUrl)
         logger.info('You can start the broker now with')
         logger.info(`streamr-broker ${storagePath}`)
-    } catch (e) {
+    } catch (e: any) {
         logger.warn('Broker Config Wizard encountered an error:')
         logger.error(e.message)
     }
--- a/packages/broker/src/ConfigWizard.ts
+++ b/packages/broker/src/ConfigWizard.ts
@@ -149,11 +149,10 @@ const createPluginPrompts = (): Array<inquirer.Question | inquirer.ListQuestion 
         choices: Object.values(PLUGIN_NAMES)
     }
 
-    const MIN_PORT_VALUE = 1024
-    const MAX_PORT_VALUE = 49151
-
     const portPrompts: Array<inquirer.Question> = Object.keys(PLUGIN_DEFAULT_PORTS).map((name) => {
         const defaultPort = PLUGIN_DEFAULT_PORTS[name]
+        const MIN_PORT_VALUE = 1024
+        const MAX_PORT_VALUE = 49151
         return {
             type: 'input',
             name: `${name}Port`,
--- a/packages/broker/src/ConfigWizard.ts
+++ b/packages/broker/src/ConfigWizard.ts
@@ -282,8 +282,7 @@ export const startBrokerConfigWizard = async(): Promise<void> => {
         if (privateKeyAnswers.revealGeneratedPrivateKey) {
             logger.info(`This is your node\'s private key: ${privateKey}`)
         }
-        const pluginPrompts = createPluginPrompts()
-        const pluginsAnswers = await inquirer.prompt(pluginPrompts)
+        const pluginsAnswers = await inquirer.prompt(createPluginPrompts())
         const config = getConfig(privateKey, pluginsAnswers)
         const storageAnswers = await selectStoragePath()
         const destinationPath = await createStorageFile(config, storageAnswers)
--- a/packages/broker/src/ConfigWizard.ts
+++ b/packages/broker/src/ConfigWizard.ts
@@ -138,8 +138,9 @@ const createPluginPrompts = (): Array<inquirer.Question | inquirer.ListQuestion 
         choices: Object.values(PLUGIN_NAMES)
     }
 
-    const portPrompts: Array<inquirer.Question> = Object.keys(DEFAULT_CONFIG_PORTS).map((name) => {
-        const defaultPort = DEFAULT_CONFIG_PORTS[name]
+    const portPrompts: Array<inquirer.Question> = Object.keys(DEFAULT_CONFIG_PORTS).map((key) => {
+        const name = PLUGIN_NAMES[key]
+        const defaultPort = DEFAULT_CONFIG_PORTS[key]
         return {
             type: 'input',
             name: `${name}Port`,
--- a/packages/broker/src/ConfigWizard.ts
+++ b/packages/broker/src/ConfigWizard.ts
@@ -12,6 +12,17 @@ import * as MqttConfigSchema from './plugins/mqtt/config.schema.json'
 import * as BrokerConfigSchema from './helpers/config.schema.json'
 import * as LegacyWebsocketConfigSchema from './plugins/legacyWebsocket/config.schema.json'
 
+const createLogger = () => {
+    return {
+        info: (...args: any[]) => {
+            console.log(chalk.bgWhite.black(':'), ...args)
+        },
+        error: (...args: any[]) => {
+            console.error(chalk.bgRed.black('!'), ...args)
+        }
+    }
+}
+
 const generateApiKey = (): string => {
     const hex = uuid().split('-').join('')
     return Buffer.from(hex).toString('base64').replace(/[^0-9a-z]/gi, '')
@@ -259,17 +270,7 @@ export const start = async (
     getPrivateKeyAnswers = () => inquirer.prompt(PRIVATE_KEY_PROMPTS),
     getPluginAnswers = () => inquirer.prompt(createPluginPrompts()),
     getStorageAnswers = selectStoragePath,
-    logger = {
-        info: (...args: any[]) => {
-            console.log(chalk.bgWhite.black(':'), ...args)
-        },
-        warn: (...args: any[]) => {
-            console.warn(chalk.bgYellow.black('!'), ...args)
-        },
-        error: (...args: any[]) => {
-            console.error(chalk.bgRed.black('!'), ...args)
-        }
-    }
+    logger = createLogger()
 ): Promise<void> => {
     try {
         const privateKeyAnswers = await getPrivateKeyAnswers()
@@ -289,7 +290,6 @@ export const start = async (
         logger.info('You can start the broker now with')
         logger.info(`streamr-broker ${storagePath}`)
     } catch (e: any) {
-        logger.warn('Broker Config Wizard encountered an error:')
-        logger.error(e.message)
+        logger.error('Broker Config Wizard encountered an error:' + e.message)
     }
 }
--- a/packages/broker/src/ConfigWizard.ts
+++ b/packages/broker/src/ConfigWizard.ts
@@ -148,41 +148,44 @@ const PLUGIN_NAMES: {[pluginName: string]: string} = {
     PUBLISH_HTTP: 'publishHttp'
 }
 
-const pluginPrompts: Array<inquirer.Question | inquirer.ListQuestion | inquirer.CheckboxQuestion> = [{
-    type: 'checkbox',
-    name:'selectPlugins',
-    message: 'Select the plugins to enable',
-    choices: Object.values(PLUGIN_NAMES)
-}]
+const createPluginPrompts = (): Array<inquirer.Question | inquirer.ListQuestion | inquirer.CheckboxQuestion> => {
+    const selectPluginsPrompt: inquirer.CheckboxQuestion = {
+        type: 'checkbox',
+        name:'selectPlugins',
+        message: 'Select the plugins to enable',
+        choices: Object.values(PLUGIN_NAMES)
+    }
 
-Object.keys(PLUGIN_DEFAULT_PORTS).map((pluginName) => {
-    const defaultPluginPort = PLUGIN_DEFAULT_PORTS[pluginName]
-    pluginPrompts.push({
-        type: 'input',
-        name: `${pluginName}Port`,
-        message: `Select a port for the ${pluginName} Plugin [Enter for default: ${defaultPluginPort}]`,
-        when: (answers: inquirer.Answers) => {
-            return answers.selectPlugins.includes(pluginName)
-        },
-        validate: (input: string | number): string | boolean => {
-            const portNumber = (typeof input === 'string') ? Number(input) : input
+    const pluginPortPrompts: Array<inquirer.Question> = Object.keys(PLUGIN_DEFAULT_PORTS).map((pluginName) => {
+        const defaultPluginPort = PLUGIN_DEFAULT_PORTS[pluginName]
+        return {
+            type: 'input',
+            name: `${pluginName}Port`,
+            message: `Select a port for the ${pluginName} Plugin [Enter for default: ${defaultPluginPort}]`,
+            when: (answers: inquirer.Answers) => {
+                return answers.selectPlugins.includes(pluginName)
+            },
+            validate: (input: string | number): string | boolean => {
+                const portNumber = (typeof input === 'string') ? Number(input) : input
+                if (Number.isNaN(portNumber)) {
+                    return `Non-numeric value provided`
+                }
 
-            if (Number.isNaN(portNumber)) {
-                return `Non-numeric value provided`
-            }
+                if (!Number.isInteger(portNumber)) {
+                    return `Non-integer value provided`
+                }
 
-            if (!Number.isInteger(portNumber)) {
-                return `Non-integer value provided`
-            }
-
-            if (portNumber < MIN_PORT_VALUE || portNumber > MAX_PORT_VALUE) {
-                return `Out of range port ${portNumber} provided (valid range ${MIN_PORT_VALUE}-${MAX_PORT_VALUE})`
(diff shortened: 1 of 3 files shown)


packages/broker/src/ConfigWizard.ts
```diff
@@ -171,7 +165,17 @@ Object.keys(PLUGIN_DEFAULT_PORTS).map((pluginName) => {
     })
 })
 
-prompts = prompts.concat(pluginSelectorPrompt).concat(pluginPrompts)
+const revealGeneratedPrivateKeyPrompt = {
+    type: 'confirm',
+    name: 'revealGeneratedPrivateKey',
+    message: 'We strongly recommend backing up your private key. It will be written into the config file, but would you also like to see it on screen now?',
```
hpihkala:
```suggestion
    message: 'We strongly recommend backing up your private key. It will be written into the config file, but would you also like to see this sensitive information on screen now?',
```

hpihkala: Small text edit to emphasize the sensitivity of the information

packages/broker/src/ConfigWizard.ts
```diff
@@ -171,7 +165,17 @@ Object.keys(PLUGIN_DEFAULT_PORTS).map((pluginName) => {
...
+    default: true,
```
hpihkala: On second thought, let's default to secure -> make the default "no"

```suggestion
    default: false,
```

packages/broker/src/ConfigWizard.ts
```diff
@@ -31,20 +31,14 @@ const PRIVATE_KEY_SOURCE_GENERATE = 'Generate'
...
 const logger = {
-    print: (...args: any[]) => {
-        console.log(chalk.bgWhite.black(':'), ...args)
-    },
     info: (...args: any[]) => {
-        console.log(chalk.bgWhite.black(':', ...args))
-    },
-    alert: (...args: any[]) => {
-        console.log(chalk.bgYellow.black('!', ...args))
+        console.log(chalk.bgWhite.black(':'), ...args)
```
```suggestion
        // eslint-disable-next-line no-console
        console.log(chalk.bgWhite.black(':'), ...args)
```

packages/broker/src/ConfigWizard.ts
```diff
@@ -31,20 +31,14 @@ const PRIVATE_KEY_SOURCE_GENERATE = 'Generate'
...
     },
     warn: (...args: any[]) => {
-        console.log(chalk.bgYellow.black('!'), ...args)
+        console.warn(chalk.bgYellow.black('!'), ...args)
```
```suggestion
        // eslint-disable-next-line no-console
        console.warn(chalk.bgYellow.black('!'), ...args)
```

packages/broker/src/ConfigWizard.ts
```diff
@@ -31,20 +31,14 @@ const PRIVATE_KEY_SOURCE_GENERATE = 'Generate'
...
     },
     error: (...args: any[]) => {
-        console.log(chalk.bgRed.black('!'), ...args)
+        console.error(chalk.bgRed.black('!'), ...args)
```
```suggestion
        // eslint-disable-next-line no-console
        console.error(chalk.bgRed.black('!'), ...args)
```

packages/broker/src/ConfigWizard.ts
```diff
@@ -171,7 +168,17 @@ Object.keys(PLUGIN_DEFAULT_PORTS).map((pluginName) => {
...
+    type: 'confirm',
+    name: 'revealGeneratedPrivateKey',
+    message: 'We strongly recommend backing up your private key. It will be written into the config file, but would you also like to see this sensitive information on screen now?',
+    default: false,
+    when: (answers: inquirer.Answers) => {
+        return answers.generateOrImportEthereumPrivateKey === PRIVATE_KEY_SOURCE_GENERATE
+    }
+}
+
+prompts = prompts.concat(pluginSelectorPrompt).concat(pluginPrompts).concat(revealGeneratedPrivateKeyPrompt)
```
teogeb: Would it make sense to ask all private key related question before the plugin questions? Now it asks:

1. private key generation
2. plugins
2. plugin ports
3. private key revealing
4. config file location

bearni95: Yeah, the sorting got pretty out of hand, thing is:
- We're collecting the answers and then applying logic
- Inquirer doesn't allow any `console` call to make it to the actual console

My initial thought was to add a simple `console.log` on the validating function for the new question but it just won't show up!

Is it maybe worth it to split the inquirer answers->config process and apply it in "blocks"?  This way we could ask all the private key questions, apply any logic we require and display info back to the user outside the questions inquirer allows. We'd have a second block doing it's own `generateConfigFromAnswers`, and so on for the storage bit.

packages/broker/src/ConfigWizard.ts
```diff
@@ -168,21 +177,16 @@ Object.keys(PLUGIN_DEFAULT_PORTS).map((pluginName) => {
...
 })
 
-const revealGeneratedPrivateKeyPrompt = {
-    type: 'confirm',
-    name: 'revealGeneratedPrivateKey',
-    message: 'We strongly recommend backing up your private key. It will be written into the config file, but would you also like to see this sensitive information on screen now?',
-    default: false,
-    when: (answers: inquirer.Answers) => {
-        return answers.generateOrImportEthereumPrivateKey === PRIVATE_KEY_SOURCE_GENERATE
+// prompts = prompts.concat(pluginSelectorPrompt).concat(pluginPrompts).concat(revealGeneratedPrivateKeyPrompt)
```
teogeb: Remove comment

packages/broker/src/ConfigWizard.ts
```diff
@@ -168,21 +177,16 @@ Object.keys(PLUGIN_DEFAULT_PORTS).map((pluginName) => {
...
+export const getEthereumConfigFromAnswers = (answers: inquirer.Answers, config: any) => {
+    config.ethereumPrivateKey = (answers.importPrivateKey) ? answers.importPrivateKey : Wallet.createRandom().privateKey
+    if (answers.revealGeneratedPrivateKey) {
+        logger.info(`This is your node\'s private key: ${config.ethereumPrivateKey}`)
```
teogeb: It would be great if all `getX` methods just return a value a don't do any side effects. Therefore maybe good to move the `logger.info` call outside of this method, and also remove the manipulation of `config` object. 

To me it makes sense to group all config generation into a single methods (as it was in `getConfigFromAnswers`), and pass `privateKeyAnswers` and `pluginAnswers` to that method.

packages/broker/src/ConfigWizard.ts
```diff
@@ -260,11 +262,11 @@ export const createStorageFile = async (config: any, answers: inquirer.Answers):
...
-        const config = getConfigFromAnswers(answers)
-        if (answers.revealGeneratedPrivateKey) {
-            logger.info(`This is your node\'s private key: ${config.ethereumPrivateKey}`)
-        }
+        let config = { ... CONFIG_TEMPLATE, plugins: { ... CONFIG_TEMPLATE.plugins } }
+        const ethereumAnswers = await inquirer.prompt(privateKeyPrompts)
+        config = getEthereumConfigFromAnswers(ethereumAnswers, config)
+        const pluginAnswers = await inquirer.prompt([pluginSelectorPrompt, ...pluginPrompts])
+        config = getPluginsConfigFromAnswers(pluginAnswers, config)
         const nodeAddress = new Wallet(config.ethereumPrivateKey).address
```
teogeb: While we are changing the order of messaging, maybe it would make sense to ask the config location before we show the "`Welcome to the Streamr Network`" message. 

Then we could maybe remove the "Broker Config Wizard ran succesfully" as the success is implied by the "`Welcome to...`" message. And "`Stored config under ${destinationPath}"` is not needed, because we show the location in the "`streamr-broker ${destinationPath}`" line below.

packages/broker/src/ConfigWizard.ts
```diff
@@ -260,11 +262,11 @@ export const createStorageFile = async (config: any, answers: inquirer.Answers):
 
 export const startBrokerConfigWizard = async(): Promise<void> => {
     try {
-        const answers = await inquirer.prompt(prompts)
-        const config = getConfigFromAnswers(answers)
-        if (answers.revealGeneratedPrivateKey) {
-            logger.info(`This is your node\'s private key: ${config.ethereumPrivateKey}`)
-        }
+        let config = { ... CONFIG_TEMPLATE, plugins: { ... CONFIG_TEMPLATE.plugins } }
+        const ethereumAnswers = await inquirer.prompt(privateKeyPrompts)
```
teogeb: Could rename `ethereumAnswers` to `privateKeyAnswers` so that one concept has only one name. (Also e.g. at line `293`)

packages/broker/src/ConfigWizard.ts
```diff
@@ -260,11 +262,11 @@ export const createStorageFile = async (config: any, answers: inquirer.Answers):
...
+        config = getEthereumConfigFromAnswers(ethereumAnswers, config)
+        const pluginAnswers = await inquirer.prompt([pluginSelectorPrompt, ...pluginPrompts])
```
teogeb: Maybe the global `pluginPrompts` variable would include all plugin-related prompts. (If the initialization of that variable is not trivial, wrap the logic into a method.)

packages/broker/src/ConfigWizard.ts
bearni95: There's the call to the method `getEthereumPrivateKeyFromAnswers` that should allow us to display the generated pkey in "runtime" but it still makes the whole logic seem a bit weird.

packages/broker/src/ConfigWizard.ts
```diff
@@ -265,6 +265,16 @@ export const createStorageFile = async (config: any, answers: inquirer.Answers):
     return answers.selectDestinationPath
 }
 
+export const generateMiscFromConfig = (config: any) => {
```
teogeb: Could improve the naming of this function. Also the type of `config` parameter could be more specific.

packages/broker/src/ConfigWizard.ts
```diff
@@ -265,6 +265,16 @@ export const createStorageFile = async (config: any, answers: inquirer.Answers):
...
+    const nodeAddress = new Wallet(config.ethereumPrivateKey).address
+    const mnemonic = Protocol.generateMnemonicFromAddress(nodeAddress)
+    const nodeExplorerUrl = `https://streamr.network/network-explorer/nodes/${nodeAddress}`
+    return {
+        mnemonic,
+        nodeExplorerUrl
```
teogeb: This could be just `explorerUrl` as there a no other `explorers`. Or `networkExplorerUrl` if we want to use the official term.

packages/broker/src/ConfigWizard.ts
```diff
@@ -256,28 +265,42 @@ export const createStorageFile = async (config: any, answers: inquirer.Answers):
...
-        logger.print('View your node in the Network Explorer:')
-        logger.print(`https://streamr.network/network-explorer/nodes/${nodeAddress}`)
-        logger.info('This is your node\'s private key. Please store it in a secure location:')
-        logger.alert(config.ethereumPrivateKey)
+        const privateKeyAnswers = await inquirer.prompt(privateKeyPrompts)
+        const privateKey = getPrivateKeyFromAnswers(privateKeyAnswers)
+        if (privateKeyAnswers.revealGeneratedPrivateKey) {
+            logger.info(`This is your node\'s private key: ${privateKey}`)
+        }
+        const pluginPrompts = createPluginPrompts()
```
teogeb: Could remove internal variables which are just aliases to direct function calls, e.g. `pluginPrompts`.

packages/broker/src/ConfigWizard.ts
```diff
@@ -225,7 +234,7 @@ export const selectDestinationPathPrompt = {
     }
 }
 
-const selectValidDestinationPath = async (): Promise<inquirer.Answers> => {
+const selectStoragePath = async (): Promise<inquirer.Answers> => {
     const answers = await inquirer.prompt([selectDestinationPathPrompt])
```
teogeb: If we unify the naming, could modify all places where those old names are used (e.g. `destinationPath`, `generateOrImportEthereumPrivateKey`, `getEthereumConfig`, `assertValidMisc`). 

The test use the old name for `generateOrImportPrivateKey`, but they don't fail. Maybe they should?

packages/broker/src/ConfigWizard.ts
teogeb: Maybe worth adding a bullet point to PR description about this simplification.

packages/broker/src/ConfigWizard.ts
```diff
@@ -97,19 +94,19 @@ export const CONFIG_TEMPLATE: any = {
     }
 }
 
-let prompts: Array<inquirer.Question | inquirer.ListQuestion | inquirer.CheckboxQuestion> = [
+const privateKeyPrompts: Array<inquirer.Question | inquirer.ListQuestion | inquirer.CheckboxQuestion> = [
```
teogeb: Use the correct naming convention for a constant

packages/broker/src/ConfigWizard.ts
```diff
@@ -138,54 +148,55 @@ const PLUGIN_NAMES: {[pluginName: string]: string} = {
...
-        type: 'input',
-        name: `${pluginName}Port`,
-        message: `Select a port for the ${pluginName} Plugin [Enter for default: ${defaultPluginPort}]`,
-        when: (answers: inquirer.Answers) => {
-            return answers.selectPlugins.includes(pluginName)
-        },
-        validate: (input: string | number): string | boolean => {
-            const portNumber = (typeof input === 'string') ? Number(input) : input
+const createPluginPrompts = (): Array<inquirer.Question | inquirer.ListQuestion | inquirer.CheckboxQuestion> => {
+    const selectPluginsPrompt: inquirer.CheckboxQuestion = {
```
teogeb: This method will be easier to read if we don't prefix most of the variables with word `plugin`. The context should be clear from the method name.

packages/broker/test/unit/ConfigWizard.test.ts
```diff
@@ -174,16 +184,22 @@ describe('ConfigWizard', () => {
...
+    })
 
+    describe('end-to-end', () => {
         it ('should exercise the happy path with default answers', () => {
-            const answers = {
+            const privateKeyAnswers = {
                 generateOrImportEthereumPrivateKey: 'Generate',
+                revealGeneratedPrivateKey: false
+            }
+            const pluginsAnswers = {
```
teogeb, line 189: Remove some redundancy from `DEFAULT_CONFIG_PORTS` field names?

packages/broker/src/ConfigWizard.ts
bearni95: `generateOrImportPrivateKey` is used for the `when` methods in inquirer, which we don't test for. The validation function takes the imported private key from the `input` field rather than from the `answers`, hence the passing tests.

This aside the renamings seem alright

teogeb: We could remove all which don't effect tests (e.g. `generateOrImportPrivateKey` and `revealGeneratedPrivateKey` in `should exercise the happy path with user input`)

packages/broker/src/ConfigWizard.ts
```diff
@@ -133,7 +133,7 @@ const privateKeyPrompts: Array<inquirer.Question | inquirer.ListQuestion | inqui
 ]
 
 export const getPrivateKey = (answers: inquirer.Answers): string => {
-    return (answers.importPrivateKey) ? answers.importPrivateKey : Wallet.createRandom().privateKey
+    return (answers.generateOrImportPrivateKey === PRIVATE_KEY_SOURCE_IMPORT && answers.importPrivateKey) ? answers.importPrivateKey : Wallet.createRandom().privateKey
```
teogeb: Is `&& answers.importPrivateKey` check redundant?

packages/broker/test/unit/ConfigWizard.test.ts
```diff
@@ -6,14 +6,11 @@ import { PROMPTS, DEFAULT_CONFIG_PORTS, selectStoragePathPrompt, createStorageFi
...
 const assertValidPort = (port: number | string, pluginName = 'websocket') => {
     const numericPort = (typeof port === 'string') ? parseInt(port) : port
-    const privateKeyAnswers = {
-        generateOrImportPrivateKey: 'Generate',
-    }
     const pluginsAnswers = {
         selectPlugins:[pluginName],
         websocketPort: port,
     }
-    const privateKey = getPrivateKey(privateKeyAnswers)
```
teogeb: When we validate the port in assertValidPort, we don't need any info about the port. Would 
`const config = getConfig(undefined as any, pluginsAnswers)` be a minimalistic valid way to get the port configs?

packages/broker/test/unit/ConfigWizard.test.ts
```diff
@@ -153,11 +149,7 @@ describe('ConfigWizard', () => {
 
     describe('getEthereumConfig', () => {
```
teogeb: The name of the test?

packages/broker/test/unit/ConfigWizard.test.ts
```diff
@@ -218,7 +207,6 @@ describe('ConfigWizard', () => {
             const importPrivateKey = Wallet.createRandom().privateKey
```
teogeb: The name of the variable? As it has the verb prefix in the name, it sounds like an action (similar to `generateOrImportPrivateKey`) and not the content itself

packages/broker/src/ConfigWizard.ts
```diff
@@ -156,6 +155,9 @@ const createPluginPrompts = (): Array<inquirer.Question | inquirer.ListQuestion
         choices: Object.values(PLUGIN_NAMES)
     }
 
+    const MIN_PORT_VALUE = 1024
```
teogeb: If we move these, let's move to the most local place possible

packages/broker/src/ConfigWizard.ts
```diff
@@ -12,22 +12,6 @@ import * as MqttConfigSchema from './plugins/mqtt/config.schema.json'
...
-const DEFAULT_HTTP_PORT = BrokerConfigSchema.properties.httpServer.properties.port.default
-const DEFAULT_LEGACY_WS_PORT = LegacyWebsocketConfigSchema.properties.port.default
-
-export const DEFAULT_CONFIG_PORTS = {
-    WS: DEFAULT_WS_PORT,
-    MQTT: DEFAULT_MQTT_PORT,
-    HTTP: DEFAULT_HTTP_PORT,
-    LEGACY_WS: DEFAULT_LEGACY_WS_PORT
-}
-
```
teogeb, line 26: Eslint errors on this

packages/broker/src/ConfigWizard.ts
```diff
@@ -225,59 +203,103 @@ export const selectDestinationPathPrompt = {
...
-        logger.print(`Stored config under ${destinationPath}`)
-        logger.print('You can start the broker now with')
-        logger.info(`streamr-broker ${destinationPath}`)
+        const {mnemonic, networkExplorerUrl} = getNodeIdentity(config)
+        logger.info(`Your node's generated name is ${mnemonic}.`)
+        logger.info('View your node in the Network Explorer:')
+        logger.info(networkExplorerUrl)
+        logger.info('You can start the broker now with')
+        logger.info(`streamr-broker ${storagePath}`)
     } catch (e) {
```
teogeb: Fix TypeScript warning:  the type of e is unknown (all 3 places where exception handling is done)

packages/broker/test/unit/ConfigWizard.test.ts
```diff
@@ -174,16 +170,28 @@ describe('ConfigWizard', () => {
...
+    })
 
+    describe('end-to-end', () => {
         it ('should exercise the happy path with default answers', () => {
-            const answers = {
-                generateOrImportEthereumPrivateKey: 'Generate',
+            const privateKeyAnswers = {}
+            const pluginsAnswers = {
                 selectPlugins: [ 'websocket', 'mqtt', 'publishHttp' ],
-                websocketPort: DEFAULT_CONFIG_PORTS.DEFAULT_WS_PORT,
```
teogeb, line 182: The variable name is bit misleading as the ports are not defaults only for the config, but for the Broker or the IT industry (the MQTT port)

packages/broker/src/ConfigWizard.ts
bearni95: tried grepping "catch (e" and it's type is either not given or `any` everywhere else, is the `any` typing here enough as well?

packages/broker/test/unit/ConfigWizard.test.ts
bearni95: The naming here is due to where the imports come from:

```typescript
const DEFAULT_WS_PORT = WebsocketConfigSchema.properties.port.default
const DEFAULT_MQTT_PORT = MqttConfigSchema.properties.port.default
const DEFAULT_HTTP_PORT = BrokerConfigSchema.properties.httpServer.properties.port.default
const DEFAULT_LEGACY_WS_PORT = LegacyWebsocketConfigSchema.properties.port.default
``

File: packages/broker/src/ConfigWizard.ts [hunk 27]
```diff
@@ -53,6 +32,21 @@ const generateApiKey = (): string => {
     return Buffer.from(hex).toString('base64').replace(/[^0-9a-z]/gi, '')
 }
 
+const DEFAULT_WS_PORT = WebsocketConfigSchema.properties.port.default
```
teogeb: These 4 temporary variables are not needed? Each component could access the info via DEFAULT_CONFIG_PORTS

packages/broker/src/ConfigWizard.ts
teogeb: Typing it `any` is fine

packages/broker/src/ConfigWizard.ts
```diff
@@ -123,13 +112,22 @@ let prompts: Array<inquirer.Question | inquirer.ListQuestion | inquirer.Checkbox
...
+        name: 'revealGeneratedPrivateKey',
+        message: 'We strongly recommend backing up your private key. It will be written into the config file, but would you also like to see this sensitive information on screen now?',
+        default: false,
+        when: (answers: inquirer.Answers) => {
+            return answers.generateOrImportPrivateKey === PRIVATE_KEY_SOURCE_GENERATE
+        }
     }
 ]
 
 const PLUGIN_DEFAULT_PORTS: {[pluginName: string]: number} = {
```
teogeb: Do we need both `DEFAULT_CONFIG_PORTS` and `PLUGIN_DEFAULT_PORTS`?

packages/broker/src/ConfigWizard.ts
```diff
@@ -131,9 +130,6 @@ const PRIVATE_KEY_PROMPTS: Array<inquirer.Question | inquirer.ListQuestion | inq
     }
 ]
 
-export const getPrivateKey = (answers: inquirer.Answers): string => {
```
teogeb: When we reorganize the code, it is maybe easiest to choose one primary factor which guides the order. It could be anything, e.g.
- by the user flow (all prompting first, then config generation, then storing the config)
- by content (all private key handing first, then all plugin handling)
- ...

In all cases, it is good to have the very generic code at the top and the main at the end, as you currently have.

bearni95: the idea behind the code sorting is:
- imports
- helper methods
- static values
- inquirer prompts
- post-question logic handling
- start fn

packages/broker/src/ConfigWizard.ts
```diff
@@ -219,65 +185,111 @@ export const selectDestinationPathPrompt = {
...
+    config.ethereumPrivateKey = privateKey
+
+    const pluginNames = Object.keys(PLUGIN_NAMES)
+    pluginNames.forEach((pluginKey) => {
+        const pluginName = PLUGIN_NAMES[pluginKey]
+        const defaultPort = DEFAULT_CONFIG_PORTS[pluginKey]
+        if (pluginsAnswers.selectPlugins && pluginsAnswers.selectPlugins.includes(pluginName)){
+            let pluginConfig = {}
+            const portNumber = parseInt(pluginsAnswers[`${pluginName}Port`])
+            console.log('portNumber', portNumber, defaultPort)
```
teogeb: Remove this debug logging

packages/broker/test/unit/ConfigWizard.test.ts
teogeb: Broker stores the default values in the schema file so that users can easily override the values. If there wouldn't be any configuration options, we'd definitely run the MQTT plugin at port 1883.

packages/broker/src/ConfigWizard.ts
```diff
@@ -219,65 +185,111 @@ export const selectDestinationPathPrompt = {
...
 
+export const getPrivateKey = (answers: inquirer.Answers): string => {
+    return (answers.generateOrImportPrivateKey === PRIVATE_KEY_SOURCE_IMPORT) ? answers.importPrivateKey : Wallet.createRandom().privateKey
+}
+
+export const getConfig = (privateKey: string, pluginsAnswers: inquirer.Answers): any => {
+    const config = { ... CONFIG_TEMPLATE, plugins: { ... CONFIG_TEMPLATE.plugins } }
+    config.ethereumPrivateKey = privateKey
+
+    const pluginNames = Object.keys(PLUGIN_NAMES)
```
teogeb: The keys are not plugin names, but some IDs (compare to line `225` where we get the plugin name by that id)

packages/broker/src/ConfigWizard.ts
teogeb: Sounds good. Then maybe:
- `PROMPTS` should be in the static values section
- `getPrivateKey` in helper methods (or locally near `start`)
- `selectStoragePath` and `createStorageFile` near each other

packages/broker/test/unit/ConfigWizard.test.ts
```diff
@@ -2,22 +2,21 @@ import { Wallet } from 'ethers'
 import { writeFileSync, mkdtempSync, existsSync } from 'fs'
 import os from 'os'
 import path from 'path'
-import { CONFIG_WIZARD_PROMPTS, DEFAULT_CONFIG_PORTS, selectDestinationPathPrompt, createStorageFile, getConfigFromAnswers } from '../../src/ConfigWizard'
+import { PROMPTS, DEFAULT_CONFIG_PORTS, selectStoragePathPrompt, createStorageFile, getConfig, getPrivateKey, getNodeIdentity } from '../../src/ConfigWizard'
 
 const assertValidPort = (port: number | string, pluginName = 'websocket') => {
```
teogeb: This could be a local helper method near `getPrivateKey and getConfig` block

packages/broker/test/unit/ConfigWizard.test.ts
```diff
@@ -147,24 +146,21 @@ describe('ConfigWizard', () => {
 
     })
 
-    describe('getConfigFromAnswers', () => {
+    describe('getPrivateKey and getConfig', () => {
```
teogeb: Could divide this to two separate describe blocks

teogeb (approved):
Great improvements! 👍

PR #162, opened 2021-08-24
NET-441: include nodes with null location data to /locations EP

juslesan:
- Make the locations endpoint more consistent with the rest of the API


packages/network/src/logic/trackerSummaryUtils.ts
```diff
@@ -1,8 +1,9 @@
-import { StreamIdAndPartition } from '../identifiers'
+import { Location, StreamIdAndPartition } from '../identifiers'
 import { OverlayPerStream, OverlayConnectionRtts } from './Tracker'
 
 type OverLayWithRtts = { [key: string]: { [key: string]: { neighborId: string, rtt: number | null }[] } }
 type OverlaySizes = { streamId: string, partition: number, nodeCount: number }[]
+type NodesWithLocations = { [key: string]: Location | null }[]
```
harbu: Would a more tighter type be?
```suggestion
type NodesWithLocations = { [key: string]: Location }[]
```

Since below in the implementation of `getNodesWithLocationData` if a location is not found we use default value
```
{
                latitude: null,
                longitude: null,
                country: null,
                city: null,
            }
```

packages/network/src/logic/trackerSummaryUtils.ts
```diff
@@ -66,6 +67,19 @@ export function addRttsToNodeConnections(
     }
 }
 
+export function getNodesWithLocationData(nodes: ReadonlyArray<string>, locations: Readonly<{[key: string]: Location}>): NodesWithLocations {
```
harbu: Maybe a test for this? E.g. in "trackerSummaryUtils.test.ts" or (indirectly) thru integraiton test  "tracker-endpoints.test.ts". The latter file contains a test for the usual case, but it would be good to get the "location does not exist" case under test as well.

packages/network/src/logic/trackerSummaryUtils.ts
```diff
@@ -1,8 +1,9 @@
-import { StreamIdAndPartition } from '../identifiers'
+import { Location, StreamIdAndPartition } from '../identifiers'
 import { OverlayPerStream, OverlayConnectionRtts } from './Tracker'
 
 type OverLayWithRtts = { [key: string]: { [key: string]: { neighborId: string, rtt: number | null }[] } }
 type OverlaySizes = { streamId: string, partition: number, nodeCount: number }[]
+type NodesWithLocations = { [key: string]: Location }[]
```
harbu: Hmm wait is this intentionally an array of objects? Or just an object, i.e.?

```ts
type NodesWithLocations = { [key: string]: Location }
```

juslesan: Yeah, seems I forgot to remove the brackets they are not required

harbu (approved):

harbu:
Should this be merged @juslesan ?

juslesan:
> Should this be merged @juslesan ?

This fixes some inconsistency issues with the endpoint so it's a good idea to merge it so that API users won't run to similar issues that the NE had.

harbu:
Feel free to resole & merge or close @juslesan

PR #164, opened 2021-08-25
NET-444: Config wizard; user-provided ports written as strings instead of numbers

bearni95:
The wizard's lack of type structure on the temporary configs allowed for the ports on the plugins to be provided as either strings or numbers. 

This PR assumes all given inputs will be strings for the ports and modifies the tests accordingly

**Enhancement:** Located and patched a bug that allowed floats to be provided as port numbers, with the according tests

### DIFF
--- a/packages/broker/src/ConfigWizard.ts
+++ b/packages/broker/src/ConfigWizard.ts
@@ -164,10 +164,9 @@ Object.keys(PLUGIN_DEFAULT_PORTS).map((pluginName) => {
             if (portNumber < MIN_PORT_VALUE || portNumber > MAX_PORT_VALUE) {
                 return `Out of range port ${portNumber} provided (valid range ${MIN_PORT_VALUE}-${MAX_PORT_VALUE})`
             }
-
             return true
         },
-        default: defaultPluginPort
+        default: defaultPluginPort.toString()
     })
 })
 
@@ -181,15 +180,15 @@ export const getConfigFromAnswers = (answers: inquirer.Answers): any => {
         const defaultPluginPort = PLUGIN_DEFAULT_PORTS[pluginName]
         if (answers.selectPlugins && answers.selectPlugins.includes(pluginName)){
             let pluginConfig = {}
-            if (answers[`${pluginName}Port`] !== defaultPluginPort){
+            if (answers[`${pluginName}Port`] !== defaultPluginPort.toString()){
                 if (pluginName === PLUGIN_NAMES.PUBLISH_HTTP) {
                     // the publishHttp plugin is special, it needs to be added to the config after the other plugins
                     config.httpServer = {
-                        port: answers[`${pluginName}Port`]
+                        port: parseInt(answers[`${pluginName}Port`])
                     }
                 } else {
                     // user provided a custom value, fill in
-                    pluginConfig = { port: answers[`${pluginName}Port`] }
+                    pluginConfig = { port: parseInt(answers[`${pluginName}Port`]) }
                 }
             }
             config.plugins![pluginName] = pluginConfig
--- a/packages/broker/src/ConfigWizard.ts
+++ b/packages/broker/src/ConfigWizard.ts
@@ -155,8 +155,8 @@ Object.keys(PLUGIN_DEFAULT_PORTS).map((pluginName) => {
         when: (answers: inquirer.Answers) => {
             return answers.selectPlugins.includes(pluginName)
         },
-        validate: (input: any): string | boolean => {
-            const portNumber = parseInt(input)
+        validate: (input: string | number): string | boolean => {
+            const portNumber = (typeof input === 'string')  ? parseInt(input) : input
             if (!Number.isInteger(portNumber)) {
                 return `Non-numeric value provided`
             }
@@ -207,7 +207,7 @@ export const selectDestinationPathPrompt = {
     default: path.join(os.homedir(), '.streamr/broker-config.json'),
     validate: (input: string, answers: inquirer.Answers = {}): string | boolean => {
         try {
-            const filePath = input || answers.selectDestinationPath
+            const filePath = input
             const parentDirPath = path.dirname(filePath)
 
             answers.parentDirPath = parentDirPath
--- a/packages/broker/src/ConfigWizard.ts
+++ b/packages/broker/src/ConfigWizard.ts
@@ -181,14 +181,13 @@ export const getConfigFromAnswers = (answers: inquirer.Answers): any => {
         if (answers.selectPlugins && answers.selectPlugins.includes(pluginName)){
             let pluginConfig = {}
             if (answers[`${pluginName}Port`] !== defaultPluginPort){
+                const portObject = { port: parseInt(answers[`${pluginName}Port`]) }
                 if (pluginName === PLUGIN_NAMES.PUBLISH_HTTP) {
                     // the publishHttp plugin is special, it needs to be added to the config after the other plugins
-                    config.httpServer = {
-                        port: parseInt(answers[`${pluginName}Port`])
-                    }
+                    config.httpServer = portObject
                 } else {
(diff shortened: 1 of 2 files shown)


packages/broker/src/ConfigWizard.ts
```diff
@@ -164,10 +164,9 @@ Object.keys(PLUGIN_DEFAULT_PORTS).map((pluginName) => {
             if (portNumber < MIN_PORT_VALUE || portNumber > MAX_PORT_VALUE) {
                 return `Out of range port ${portNumber} provided (valid range ${MIN_PORT_VALUE}-${MAX_PORT_VALUE})`
             }
-
             return true
         },
-        default: defaultPluginPort
+        default: defaultPluginPort.toString()
```
teogeb: If we wan't the end result to be a `Number`, is this reverse conversion needed? I mean, if we receive a String, we should convert it to a `Number`. But this seems to do the reverse for the default. (Also at line `183`)

teogeb (changes requested):
If we'd use the `number` input type instead of `input`, we could remove most of this conversion logic?

packages/broker/src/ConfigWizard.ts
bearni95: You're totally right, didn't even consider that `input` could be read as a number! 

Made the corresponding changes, no more casting up and down `numbers` and `strings`

packages/broker/src/ConfigWizard.ts
```diff
@@ -155,7 +155,7 @@ Object.keys(PLUGIN_DEFAULT_PORTS).map((pluginName) => {
         when: (answers: inquirer.Answers) => {
             return answers.selectPlugins.includes(pluginName)
         },
-        validate: (input: string, answers: inquirer.Answers): string | boolean => {
+        validate: (input: number, answers: inquirer.Answers): string | boolean => {
```
teogeb: If the input type is number, the `parseInt` is needed. But actually the type definition is wrong, I guess? Maybe worth using inquire's `filter` hook to do the conversion from string to number?

packages/broker/src/ConfigWizard.ts
```diff
@@ -155,8 +155,11 @@ Object.keys(PLUGIN_DEFAULT_PORTS).map((pluginName) => {
         when: (answers: inquirer.Answers) => {
             return answers.selectPlugins.includes(pluginName)
         },
-        validate: (input: number, answers: inquirer.Answers): string | boolean => {
-            const portNumber = parseInt(input || answers[`${pluginName}Port`])
+        filter: (input: any, answers: inquirer.Answers) => {
+            return parseInt(input || answers[`${pluginName}Port`])
```
teogeb: The `answers[`${pluginName}Port`]` is always `undefined`. 

Also if an invalid number is given, the prompt will look like this:
`Select a port for the websocket Plugin [Enter for default: 7170] (7170) NaN` 
-> it is maybe impossible to input a corrected value as backspace doesn't erase the `NaN`part`

At lines 190 and 194 we don't need `parseInt` as the value is already a `number`. Maybe it would make sense the define a custom interface for `answers` to document the answer types and get warnings if there are any type annotation mismatches?

bearni95: The `filter` method is the one interfering with the correction of values, pushing a fix without `filter` nor the crazy amount of casting we used to have

teogeb: Did this solve the `Select a port for the websocket Plugin [Enter for default: 7170] (7170) NaN`, or that `answers[${pluginName}Port]` is always undefined. Also isn't the `input` type of `filter` known?

packages/broker/src/ConfigWizard.ts
```diff
@@ -185,11 +184,11 @@ export const getConfigFromAnswers = (answers: inquirer.Answers): any => {
                 if (pluginName === PLUGIN_NAMES.PUBLISH_HTTP) {
                     // the publishHttp plugin is special, it needs to be added to the config after the other plugins
                     config.httpServer = {
-                        port: answers[`${pluginName}Port`]
+                        port: parseInt(answers[`${pluginName}Port`])
```
teogeb: If the same modification is needed for two places, it is maybe ideal to extract the value to a separate variable so that the logic is in one place

packages/broker/src/ConfigWizard.ts
```diff
@@ -208,7 +206,7 @@ export const selectDestinationPathPrompt = {
     default: path.join(os.homedir(), '.streamr/broker-config.json'),
     validate: (input: string, answers: inquirer.Answers = {}): string | boolean => {
         try {
-            const filePath = input || answers.selectDestinationPath
+            const filePath = input
```
teogeb: Could remove the redundant variable?

packages/broker/src/ConfigWizard.ts
```diff
@@ -155,16 +155,15 @@ Object.keys(PLUGIN_DEFAULT_PORTS).map((pluginName) => {
         when: (answers: inquirer.Answers) => {
             return answers.selectPlugins.includes(pluginName)
         },
-        validate: (input: string, answers: inquirer.Answers): string | boolean => {
-            const portNumber = parseInt(input || answers[`${pluginName}Port`])
+        validate: (input: string | number): string | boolean => {
+            const portNumber = (typeof input === 'string') ? Number(input) : input
```
teogeb, line 159: This change solved a bug, which happened if input has decimal fractions, e.g. `1234.56`. Worth mentioning in the PR description and maybe also to add a test so that we are protected from that in the future.

packages/broker/test/unit/ConfigWizard.test.ts
```diff
@@ -146,7 +146,7 @@ describe('ConfigWizard', () => {
         })
 
         it ('should exercise the plugin port assignation path', () => {
-            const port = '3737'
+            const port = 3737
```
teogeb: Currently `getConfigFromAnswers` expects that the port can be a `number` or a `string`. It could test the both situations.

teogeb:
If user gave input `7170` to the `websocket` port, it shouldn't be stored to the config file as it is the default port. Currently it stores it:
```
"websocket": {
  "port": 7170
 }
```

If user just accepts the default, it works ok and doesn't store the value.

bearni95:
> If user gave input `7170` to the `websocket` port, it shouldn't be stored to the config file as it is the default port. Currently it stores it:
> 
> ```
> "websocket": {
>   "port": 7170
>  }
> ```
> 
> If user just accepts the default, it works ok and doesn't store the value.

Should be accounted for in the last commit

packages/broker/test/unit/ConfigWizard.test.ts
```diff
@@ -31,15 +31,21 @@ describe('ConfigWizard', () => {
...
             const validate = portPrompt.validate!
-            const port = '10000000000'
+            const port = 10000000000
             expect(validate(port)).toBe(`Out of range port ${port} provided (valid range 1024-49151)`)
         })
 
+        it ('invalid data: float-point number', () => {
+            const validate = portPrompt.validate!
+            const port = 55.55
+            expect(validate(port)).toBe(`Non-numeric value provided`)
```
teogeb: Bit misleading as 55.55 is numeric value, it is just not an integer

packages/broker/test/unit/ConfigWizard.test.ts
```diff
@@ -31,15 +31,21 @@ describe('ConfigWizard', () => {
     describe('plugin port validation', () => {
         it ('happy path', () => {
             const validate = portPrompt.validate!
-            expect(validate('7070')).toBe(true)
+            expect(validate(7070)).toBe(true)
```
teogeb, line 45: The validate method still supports the string input, could keep the original test and add this string test

teogeb (approved):

PR #165, opened 2021-08-25
ci: Build linux/arm64 Docker container with GitHub Actions

samt1803:


samt1803:
I also bumped the github action versions!

.github/workflows/broker.yml
```diff
@@ -98,20 +98,36 @@ jobs:
     if: github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/')
     steps:
       - uses: actions/checkout@v2.3.4
+      - name: Setup qemu
+        uses: docker/setup-qemu-action@v1.2.0
```
kare: Let's install only the required platforms with `platforms` attribute. https://github.com/docker/setup-qemu-action#inputs
And let's move it down below closer to line 186.

.github/workflows/broker.yml
```diff
@@ -98,20 +98,36 @@ jobs:
...
             ${{ runner.os }}-buildx-
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1.1.1
+        id: buildx
+        uses: docker/setup-buildx-action@v1.5.1
         with:
           driver-opts: |
             network=host
+      - name: available platforms
+        run: echo ${{ steps.buildx.outputs.platforms }}
```
kare: Remove debug print.

.github/workflows/broker.yml
```diff
@@ -155,16 +171,19 @@ jobs:
...
             latest=auto
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1.1.1
+        id: buildx2
+        uses: docker/setup-buildx-action@v1.5.1
         with:
           driver-opts: |
             network=host
+      - name: available platforms
+        run: echo ${{ steps.buildx2.outputs.platforms }}
```
kare: Remove debug print.

.github/workflows/broker.yml
```diff
@@ -98,20 +98,36 @@ jobs:
...
+        run: echo ${{ steps.buildx.outputs.platforms }}
       - name: Build
-        uses: docker/build-push-action@v2.3.0
+        uses: docker/build-push-action@v2.7.0
+        with:
+          context: .
+          file: ./Dockerfile
+          platforms: linux/amd64,linux/arm64
+          push: false
+          load: false
```
kare, line 125: Set `load: true` and remove the step below:
`- name: Build (is cached!) and load x86 image`

.github/workflows/broker.yml
```diff
@@ -155,16 +171,19 @@ jobs:
           flavor: |
             latest=auto
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1.1.1
+        id: buildx2
+        uses: docker/setup-buildx-action@v1.5.1
```
kare, line 175: Looks like these two lines can be removed?

.github/workflows/broker.yml
samt1803: I tried this before, it doesn't work because you can't load multiple outputs of the step, it is a known issue and this is the workaround: https://github.com/docker/buildx/issues/59

.github/workflows/broker.yml
```diff
@@ -98,20 +98,36 @@ jobs:
...
+          context: .
+          file: ./Dockerfile
+          platforms: linux/amd64,linux/arm64
+          push: false
+          load: false
+          tags: streamr/broker-node:dev
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache
+      - name: Build (is cached!) and load x86 image
+        uses: docker/build-push-action@v2.7.0
```
kare: At this point we are building a Docker image to run integration tests against. Only `linux/amd64` image is required. Let's remove these lines and build the `linux/arm64` version below on line 186.

.github/workflows/broker.yml
```diff
@@ -155,16 +171,19 @@ jobs:
           flavor: |
             latest=auto
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1.1.1
+        id: buildx2
```
kare: Remove this line.

.github/workflows/broker.yml
```diff
@@ -155,16 +171,19 @@ jobs:
...
+      - name: available platforms
+        run: echo ${{ steps.buildx2.outputs.platforms }}
       - name: Build
-        uses: docker/build-push-action@v2.3.0
+        uses: docker/build-push-action@v2.7.0
         with:
           context: .
           file: ./Dockerfile
-          platforms: linux/amd64
+          platforms: linux/amd64,linux/arm64
```
kare, line 171: If tests above were successful then Docker should only build `linux/arm64` image at this point.

.github/workflows/broker.yml
```diff
@@ -98,20 +98,36 @@ jobs:
...
-        uses: actions/cache@v2.1.4
+        uses: actions/cache@v2.1.6
         with:
           path: /tmp/.buildx-cache
           key: ${{ runner.os }}-buildx-${{ github.sha }}
           restore-keys: |
             ${{ runner.os }}-buildx-
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1.1.1
+        id: buildx
```
kare: Remove line

.github/workflows/broker.yml
samt1803: but will it then not also only push the arm version?

.github/workflows/broker.yml
kare: My point is that this image is for testing only and not to be pushed to Docker registry. The actual multi platform building happens around line 186 (which will use the cached image linux/amd64 image build at this step).

kare (approved):
LGTM

PR #166, opened 2021-08-26
network [NET-436]: custom speedometer

teogeb:
Replaced the `speedometer` package with a custom utility. The utility is used to analyze the throughput  rate of various metrics in `MetricsContext` and `PerStreamMetrics`.

The returned throughput rate results of the custom implementation are more accurate compared to the previous solution. Also, the new implementation doesn't need `NodeJS.Timeout` timers.

Removed also the redundant `PerStreamMetrics` class.


packages/network/src/helpers/Speedometer.ts
```diff
@@ -0,0 +1,91 @@
...
+ * E.g. if you send a total of 1000 messages, you could call record(1) after each message. Or you could 
+ * call record(msg.getSizeInBytes()) after each message to provide the data for a bytes per second 
+ * throughput rate.
+ * 
+ * At any time you can call getRate() to get the current throughput rate per second. 
+ * 
+ * The utility uses n second window to store the recorded data. If less than n seconds have been elapsed
+ * from the construction of the utility, the shorter window is used temporarily.
+ * 
+ * Internally the window is divided into one-second slots, each containing one UTC second of data. There are
```
harbu, line 19: Well said :+1:

packages/network/src/helpers/MetricsContext.ts
```diff
@@ -26,6 +26,8 @@ export class Metrics {
     }
     private readonly recordedMetrics: {
         [key: string]: {
+            // eslint-disable-next-line no-underscore-dangle
+            _speedometer: Speedometer
```
harbu, line 30: Any way around this? E.g. keep speedometer in a separate map?

packages/network/src/helpers/Speedometer.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    private static WINDOW_SIZE_DEFAULT = 5
+
+    // The first item contains a value for UTC second firstSecStartTime and items after that are consecutive 
+    // seconds in descending order (windowSizeInSeconds items). The last item of an array is at least 
+    // partially out of the window
+    private secValues: number[]
+    // UTC time of a second (millisecond part always 0)
+    private firstSecStartTime: number
+    private startupTime: number
```
harbu:
```suggestion
    private readonly startupTime: number
```

packages/network/src/logic/PerStreamMetrics.ts
```diff
@@ -9,9 +9,11 @@ interface AllMetrics<M> {
 }
 
 interface Metric {
+    // eslint-disable-next-line no-underscore-dangle
```
harbu: Hmmm I actually believe this class may not be used anymore

packages/network/test/unit/MetricsContext.test.ts
```diff
@@ -217,22 +217,22 @@ describe('metrics', () => {
...
+        jest.advanceTimersByTime(2000)
 
         const rep3 = await context.report()
         expect(rep3.metrics).toEqual({
             metricOne: {
                 a: 15,
                 b: 666,
                 c: {
-                    rate: 308,
+                    rate: 308 / 4,
```
harbu: Where does the divider come from?

packages/network/test/unit/Speedometer.test.ts
```diff
@@ -0,0 +1,129 @@
+import { Speedometer } from '../../src/helpers/Speedometer'
+
+const WINDOW_SIZE_IN_SECONDS = 5
+
+const createMockCurrentTimeProvider = () => {
```
harbu: Might wanna add a type to this return value for clarity

packages/network/test/unit/Speedometer.test.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    it('empty', () => {
+        currentTimeProvider.set('2000-01-02T03:04:05.678Z')
+        expect(speedometer.getRate()).toBe(0)
+    })
+
+    it('get immedatelly', () => {
+        currentTimeProvider.set('2000-01-02T03:04:05.678Z')
+        speedometer.record(100)
+        expect(speedometer.getRate()).toBe(20)
```
harbu, line 32: Would this also work?
```suggestion
        expect(speedometer.getRate()).toBe(100 / WINDOW_SIZE_IN_SECONDS)
```

packages/network/test/unit/Speedometer.test.ts
```diff
@@ -0,0 +1,129 @@
...
+        speedometer.record(70)
+        currentTimeProvider.set('2000-01-02T03:04:08.999Z')
+        expect(speedometer.getRate()).toBe(380 / WINDOW_SIZE_IN_SECONDS)
+    })
+
+    it('get partially expired', () => {
+        currentTimeProvider.set('2000-01-02T03:04:05.678Z')
+        speedometer.record(100)
+        currentTimeProvider.set('2000-01-02T03:04:10.123Z')
+        expect(speedometer.getRate()).toBeCloseTo((100 - 12.3) / WINDOW_SIZE_IN_SECONDS)
```
harbu, line 60: Might be worth spelling out the calculation of 12.3 for clarity

packages/network/test/unit/Speedometer.test.ts
```diff
@@ -0,0 +1,129 @@
...
+    })
+
+    it('get partially expired: data too old but within same UTC second', () => {
+        currentTimeProvider.set('2000-01-02T03:04:05.678Z')
+        speedometer.record(100)
+        currentTimeProvider.set('2000-01-02T03:04:10.888Z')
+        expect(speedometer.getRate()).toBeCloseTo((100 - 88.8) / WINDOW_SIZE_IN_SECONDS)
```
harbu, line 70: Same as above?

harbu (approved):
lgtm; see few comments

packages/network/src/logic/PerStreamMetrics.ts
teogeb: Removed `PerStreamMetrics`. The Node class recorded some metrics, but `report()` was never called.

packages/network/test/unit/MetricsContext.test.ts
teogeb: Elapsed full seconds: there are `advanceTimersByTime` calls for 0.2s + 2s + 2s. 

Refactored this to calculate the value from `Date.now()`

packages/network/src/helpers/MetricsContext.ts
teogeb: It would make the state unnecessary complicated, maybe not worth it. This is just a private data structure which is not accessible outside of `MetricsContext` class.

PR #167, opened 2021-08-26
style: Document exposed ports in Dockerfile

kare:
This information is useful for the person who is running the container.


Dockerfile
```diff
@@ -19,5 +19,9 @@ ENV LOG_LEVEL=info
 
 RUN ln -s packages/broker/tracker.js tracker.js
 
+EXPOSE 1883/tcp
```
kare, line 25: @harbu These are TCP ports, right? Is UDP used?

harbu (approved):

PR #169, opened 2021-08-26
NET-438: Log improvements on WebRTCConnection timeout

bearni95:
Silences the  connection time-out `logger.warn` on every individual connection and adds a periodic reporting `logger.info` for 'open' and 'connecting' peer states every 5 minutes

### DIFF
--- a/packages/network/src/connection/WebRtcEndpoint.ts
+++ b/packages/network/src/connection/WebRtcEndpoint.ts
@@ -124,7 +124,7 @@ export class WebRtcEndpoint extends EventEmitter implements IWebRtcEndpoint {
     }
 
     private startConnectionStatusReport(): void {
-        const STATUS_REPORT_TIMER_MS = 5 * 60 * 1000
+        const STATUS_REPORT_INTERVAL_MS = 5 * 60 * 1000
         this.statusReportTimer = setInterval(() => {
             let connectedPeerCount = 0
             const pendingPeerIds = []
@@ -137,7 +137,7 @@ export class WebRtcEndpoint extends EventEmitter implements IWebRtcEndpoint {
                 }
             }
             this.logger.info(`Successfully connected to ${connectedPeerCount} peers. Still trying to connect to the following peers: [${pendingPeerIds.join(', ')}]`)
-        }, STATUS_REPORT_TIMER_MS)
+        }, STATUS_REPORT_INTERVAL_MS)
     }
 
     private createConnection(
--- a/packages/network/src/connection/WebRtcEndpoint.ts
+++ b/packages/network/src/connection/WebRtcEndpoint.ts
@@ -49,8 +49,7 @@ export class WebRtcEndpoint extends EventEmitter implements IWebRtcEndpoint {
     private readonly bufferThresholdHigh: number
     private readonly maxMessageSize
 
-    private statusReportTimeout?: NodeJS.Timeout
-    private statusReportIntervalMs: number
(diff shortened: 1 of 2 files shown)


packages/network/src/connection/WebRtcEndpoint.ts
```diff
@@ -117,6 +120,29 @@ export class WebRtcEndpoint extends EventEmitter implements IWebRtcEndpoint {
             .addQueriedMetric('messageQueueSize', () => {
                 return Object.values(this.connections).reduce((total, c) => total + c.getQueueSize(), 0)
             })
+
+        this.statusReportIntervalMs = 5 * 60 * 1000
```
teogeb: The variable could be a constant as it is not configurable.

packages/network/src/connection/WebRtcEndpoint.ts
```diff
@@ -117,6 +120,29 @@ export class WebRtcEndpoint extends EventEmitter implements IWebRtcEndpoint {
...
+        this.startConnectionStatusReport()
+    }
+
+    private startConnectionStatusReport(): void {
+        if (this.statusReportTimeout) {
+            clearTimeout(this.statusReportTimeout)
+        }
+        this.statusReportTimeout = setTimeout(() => {
```
teogeb: Maybe `setInterval` instead of `setTimeout`, to keep this as simple as possible?

packages/network/src/connection/WebRtcEndpoint.ts
```diff
@@ -117,6 +120,29 @@ export class WebRtcEndpoint extends EventEmitter implements IWebRtcEndpoint {
...
+            let connectedPeerCount = 0
+            const pendingConnectionIds = []
```
teogeb: Rename to `pendingPeerIds` as these the ids of `peers`, not ids of `connections` (https://github.com/streamr-dev/network-monorepo/blob/14e94a72c4b96a25f306e8e5526061cf8f7c2925/packages/network/src/connection/WebRtcConnection.ts#L125)

packages/network/src/connection/WebRtcEndpoint.ts
```diff
@@ -49,8 +49,7 @@ export class WebRtcEndpoint extends EventEmitter implements IWebRtcEndpoint {
     private readonly bufferThresholdHigh: number
     private readonly maxMessageSize
 
-    private statusReportTimeout?: NodeJS.Timeout
-    private statusReportIntervalMs: number
+    private statusReportInterval?: NodeJS.Timeout
```
teogeb: The `statusReportInterval` name is maybe misleading, as this variable contains the timer id of `setInterval` call, not the interval length.  Could be e.g.`statusReportTimer`.

packages/network/src/connection/WebRtcEndpoint.ts
```diff
@@ -121,28 +120,24 @@ export class WebRtcEndpoint extends EventEmitter implements IWebRtcEndpoint {
...
-        this.statusReportIntervalMs = 5 * 60 * 1000
         this.startConnectionStatusReport()
     }
 
     private startConnectionStatusReport(): void {
-        if (this.statusReportTimeout) {
-            clearTimeout(this.statusReportTimeout)
-        }
-        this.statusReportTimeout = setTimeout(() => {
+        const statusReportIntervalMs = 5 * 60 * 1000
```
teogeb: Use the correct naming convention for a constant

teogeb (approved):

PR #171, opened 2021-08-26
NET-458: Configuration wizard creates configuration file with open permissions

bearni95:
Currently the configuration wizard creates the configuration file (file contains private key) with permissions of 644. It should be created with permission of 600 (only owner can read/write).


kare (approved):
LGTM

PR #172, opened 2021-08-26
NET-459: Configuration wizard should store files in standard location

bearni95:
## IMPORTANT: Do NOT merge until all testnets are over!

Moving the location of the default config file from `~/.streamr/` to the Linux platform as defined by the XDG Base Directory Specification, `~/.config/streamr/.

On macOS the corresponding directory should be `~/Library/Application Support/streamr` as defined by the Mac App Programming Guide.

Windows is currently unsupported.


harbu (commented):
I agree that this would be better way of storing configs. But a few points need addressing:

1. How do we migrate existing users, i.e. can we move existing configs to this new path somehow? Do we even need to?
2. Updating user-facing documentation on running broker nodes to take into account new path
3. "Windows is currently unsupported." What does this mean for running the broker on Windows, will it not start up?

teogeb (commented):
Note the file name change In https://github.com/streamr-dev/network-monorepo/pull/324. Therefore the file path in this PR could be `path.join(PATHS.config, 'default.json')` as it is a generic config which is usable both in Broker and in cli-tools.

bearni95:
@harbu 
1. Importing the private key (currently supported) should be enough.
2. Good to go with docs update.
3. Without `env-paths` the config wizard shouldn't work on Windows, that should get fixed now

@teogeb applying the filename change

harbu:
Was decided in meeting that will not be implemented

PR #173, opened 2021-08-27
network: cachedJsonGet cache the JSON itself

harbu:
Cache the result of `JSON.stringify()` to avoid having to stringify on each call.


juslesan (approved):

juslesan:
We might need to use this with the rest of the endpoints that send larger payloads eventually

packages/network/src/helpers/trackerHttpEndpoints.ts
```diff
@@ -49,17 +49,18 @@ const cachedJsonGet = (
...
         if ((cache === undefined) || (Date.now() > (cache.timestamp + maxAge))) {
             cache = {
-                json: jsonFactory(),
+                json: JSON.stringify(jsonFactory()),
                 timestamp: Date.now()
             }
         }
-        res.json(cache.json)
+        res.setHeader('Content-Type', 'application/json')
+        res.end(cache.json)
```
teogeb: Could use `res.send(cache.json)` to send the response as it generates the `ETag` header  (https://stackoverflow.com/a/49242271)

PR #174, opened 2021-08-27
No longer send "full" status messages to reduce the size of WS payloads

juslesan:
- Instead of sending a single large JSON status message send multiple small per stream status messages

- A node sends multiple status messages when:

 - the node first connects to the tracker and is subscribed to streams or wants to subscribe
 - a peer disconnects from the node

- This change is backwards compatible with previous versions where nodes still send full status messages as the logic to handle such messages is not been removed from the tracker's side

- Removed unused variable sendStatusToAllTrackersInterval from Node.ts

- Rtts are no longer sent on each status message, instead there is a default 15 second timeout for updating rtts to the tracker.

- Default interval for retrying instruction messages increased 1 -> 3 minutes to reduce status message traffic

### DIFF
--- a/packages/network/src/logic/Node.ts
+++ b/packages/network/src/logic/Node.ts
@@ -37,12 +37,12 @@ export interface NodeOptions {
     trackers: Array<TrackerInfo>
     metricsContext?: MetricsContext
     connectToBootstrapTrackersInterval?: number
-    sendStatusToAllTrackersInterval?: number
     bufferTimeoutInMs?: number
     bufferMaxSize?: number
     disconnectionWaitTime?: number
     nodeConnectTimeout?: number
     instructionRetryInterval?: number
+    rttUpdateTimeout?: number
 }
 
 const MIN_NUM_OF_OUTBOUND_NODES_FOR_PROPAGATION = 1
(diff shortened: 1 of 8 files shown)


packages/network/src/logic/Node.ts
```diff
@@ -410,17 +410,18 @@ export class Node extends EventEmitter {
...
-            }),
-            started: this.started,
-            rtts: this.nodeToNode.getRtts(),
-            location: this.peerInfo.location,
-            singleStream: false,
-            extra: this.extraMetadata
+    // Gets statuses of all streams assigned to a tracker by default
+    private getMultipleStatusMessages(tracker: string, predefinedStreams: StreamIdAndPartition[]): Status[] {
+        let streams
+        if (predefinedStreams.length === 0) {
```
teogeb: The default value handling could be simplified? Now it is a bit surprise that if you provide an empty array, it will handle all streams (instead of handling none of the streams). 

E.g. rename the parameter to `explicitStreams` and annotate it as` StreamIdAndPartition[] | undefined`. Then we would be able to remove the fallback parameter value of `[]` in `prepareAndSendMultipleStatuses`.

packages/network/src/logic/Node.ts
```diff
@@ -447,9 +443,11 @@ export class Node extends EventEmitter {
         }
     }
 
-    private prepareAndSendFullStatus(tracker: string): void {
-        const status = this.getFullStatus(tracker)
-        this.sendStatus(tracker, status)
+    private prepareAndSendMultipleStatuses(tracker: string, streams?: StreamIdAndPartition[]): void {
+        const statusMessages = this.getMultipleStatusMessages(tracker, streams)
+        statusMessages.map((status) => {
```
harbu:
```suggestion
        statusMessages.forEach((status) => {
```

harbu (approved):

harbu (approved):

PR #176, opened 2021-08-27
fix: separate peer and tracker ping intervals + less aggressive defaults

hpihkala:


juslesan (approved):

packages/network/src/composition.ts
```diff
@@ -95,15 +96,15 @@ export const createNetworkNode = ({
...
+    peerPingInterval,
     disconnectionWaitTime,
     newWebrtcConnectionTimeout,
     webrtcDatachannelBufferThresholdLow,
     webrtcDatachannelBufferThresholdHigh,
     stunUrls = ['stun:stun.l.google.com:19302']
 }: NetworkNodeOptions): NetworkNode => {
     const peerInfo = PeerInfo.newNode(id, name, undefined, undefined, location)
-    const endpoint = new NodeClientWsEndpoint(peerInfo, metricsContext, pingInterval)
+    const endpoint = new NodeClientWsEndpoint(peerInfo, metricsContext, peerPingInterval)
```
harbu: Given a node, was the intention to be able to define a separate ping interval for connection to tracker (WebSocket) and connection to other nodes (WebRTC)? If so then should this  be `trackerPingInterval`? (For clarity: the interval parameter here defines the ping-pong interval from node -> tracker).

packages/network/src/composition.ts
```diff
@@ -35,7 +35,8 @@ export interface AbstractNodeOptions {
     name?: string
     location?: Location | null
     metricsContext?: MetricsContext
-    pingInterval?: number
+    peerPingInterval?: number
```
harbu: `peerPingInterval` only used with `startNetworkNode`, move under `NetworkNodeOptions`type?

harbu (approved):

hpihkala:
Noting some seemingly unrelated test failures:
- Protocol unit tests are failing with some TS issues
- Browser integration tests are failing, but the same tests pass in node.js

PR #177, opened 2021-08-29
Overlay topology optimizations

harbu:
> Made a performance test where I setup a topology of 4000 nodes. I then repeatedly randomly pick one of the nodes, make it leave and then rejoin. Idea is to try to approximate what is happening in test net.
> 
> With the original algorithm leaving and rejoining (of a single node) took around 9.65ms on average.With the first optimization (avoiding the O(n) scan in update) the same took 7.46ms.Adding the set of nodes with less than max neighbors on top brought this down to 3.55ms. (edited) 

This PR contains the performance test and the above mentioned optimizations.

### DIFF
--- a/packages/network/src/logic/OverlayTopology.ts
+++ b/packages/network/src/logic/OverlayTopology.ts
@@ -61,27 +61,15 @@ export class OverlayTopology {
             const neighborsToRemove = [...this.nodes[nodeId]].filter((n) => !newNeighbors.includes(n))
             neighborsToRemove.forEach((n) => {
                 this.nodes[n].delete(nodeId)
-                if (this.numOfMissingNeighbors(n) > 0) {
-                    this.nodesWithOpenSlots.add(n)
-                }
+                this.checkOpenSlots(n)
             })
         }
 
         this.nodes[nodeId] = new Set(newNeighbors)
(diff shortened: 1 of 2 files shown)


packages/network/src/logic/OverlayTopology.ts
```diff
@@ -51,26 +53,49 @@ export class OverlayTopology {
...
+        }
+
+        this.nodes[nodeId] = new Set(newNeighbors)
+        newNeighbors.forEach((neighbor) => this.nodes[neighbor].add(nodeId))
+
+        if (this.numOfMissingNeighbors(nodeId) > 0) {
+            this.nodesWithOpenSlots.add(nodeId)
+        } else {
+            this.nodesWithOpenSlots.delete(nodeId)
+        }
```
hpihkala: Could be made DRY by wrapping this into a function `checkOpenSlots` (or similar), can be reused several times in various places:

In `update`:
- For `neighborsToRemove`
- For `nodeId`
- For `newNeighbors`

In `leave`:
- for `neighbor`s

In `formInstructions`:
- for `nodeId`

hpihkala (commented):
Looks solid, only one suggestion to make it DRY

PR #178, opened 2021-08-29
feat(tracker): accept x-forwarded-for header on connections from 127.0.0.1

hpihkala:
Enables the tracker to run behind nginx and still read the original websocket connection's IP address by reading the `X-Forwarded-For` http header. To prevent IP spoofing, the header is only trusted on connections that originate from `127.0.0.1` (ipv4 and ipv6 should both work).


teogeb (approved):

juslesan (approved):
👍

PR #180, opened 2021-09-02
Storage node cluster

hpihkala:
This PR proposes a change which allows a set of storage nodes connected to the same cassandra cluster to share load.

### Conceptual changes

Currently, streams are assigned to a storage node via the `StreamStorageRegistry`. However, the registry is actually just a mapping between streams and an Ethereum address; there is no meaning attached to the assigned address as such. We have previously interpreted that address to be the address corresponding to the private key the storage node is configured with, but this does not need to be the case.

Instead, we can interpret the address to be an identifier for a storage _cluster_, meaning a group of nodes within a single storage facility. The address of the nodes is decoupled from the address of the storage cluster. 

Each node in the cluster fetches the stream assignments for the `clusterAddress`, while each node's own address need not be that address.

The idea is illustrated on [this Miro board](https://miro.com/app/board/o9J_lzVCrnU=/).

### Configuring a storage cluster
The following settings are added to storage plugin config:

```
    "storage": {
      ...
      "cluster": {
        "clusterAddress": "0x12345",
        "clusterSize": 3,
        "myIndexInCluster": 0
      }
    }
```

All nodes in the storage cluster **must** agree on the `clusterAddress` and `clusterSize` as well as the `cassandra` settings. `myIndexInCluster` must be unique and run from `0` to `clusterSize-1`. For example, these could be the configs of three individual nodes in a storage cluster `0x12345`:

```
      // Node 0xaaaaaa
      "cluster": {
        "clusterAddress": "0x12345",
        "clusterSize": 3,
        "myIndexInCluster": 0
      }

      // Node 0xbbbbb
      "cluster": {
        "clusterAddress": "0x12345",
        "clusterSize": 3,
        "myIndexInCluster": 1
      }

      // Node 0xccccc
      "cluster": {
        "clusterAddress": "0x12345",
        "clusterSize": 3,
        "myIndexInCluster": 2
      }
```

In a cluster of size `1`, all stream-partitions obviously map to self. In fact, the default config for `cluster` is the single-node case, making the change 100% backwards compatible:

```
      // Node 0x12345
      "cluster": {
        "clusterAddress": null, // if null, uses node address 0x12345 for assignment discovery like before
        "clusterSize": 1,
        "myIndexInCluster": 0
      }
```

### Load balancing within the cluster

The nodes subscribe and store streams as follows:
- As usual, each node queries what streams are assigned to storage node `0x12345` (which is now the address of none of the nodes themselves, but a virtual address of the cluster).
- Instead of subscribing to _all_ of them, each node subscribes to _part_ of them.
- The selection is _deterministic_ and based on hashing a key containing `streamId` and `partition`, and deriving an index between `[0, clusterSize-1]` from the hash. The mapped index is compared to `myIndexInCluster`, and the node only subscribes to the stream if they match. The determinism guarantees that all nodes in the cluster agree on who should store what. Simplified, the formula is:

```
whoShouldStoreIt = hash(streamId + streamPartition) % clusterSize
```

Note that this update does not add any fault tolerance feature, only a load balancing feature. However, it does reduce the chance of the whole storage facility failing at once: if there are three nodes in the cluster and one fails, the two remaining ones can continue to answer queries (for all streams assigned to the cluster) as well as subscribe to and store their share of the load (two thirds of all stream-partitions assigned to the cluster).

### Other changes in this PR

- Refactor parts of the message partitioning logic from `client` to `protocol` in order to share the same algorithm with the storage code
- Add some more unit test coverage for `StorageConfig`


packages/broker/src/plugins/storage/StorageConfig.ts
```diff
@@ -29,36 +29,45 @@ const getKeyFromStream = (streamId: string, streamPartition: number): StreamKey
...
+const getKeysFromStream = (streamId: string, partitions: number): StreamKey[] => {
+    const keys: StreamKey[] = []
     for (let i = 0; i < partitions; i++) {
-        keys.add(getKeyFromStream(streamId, i))
+        keys.push(getKeyFromStream(streamId, i))
     }
     return keys
 }
 
+const belongsToMeInCluster = (key: StreamKey, clusterSize: number, myIndexInCluster: number) => {
```
teogeb: This could be a method of `StorageConfig` class as it needs access to `clusterSize` and `myIndexInCluster`

packages/broker/src/plugins/storage/StoragePlugin.ts
```diff
@@ -78,7 +83,12 @@ export class StoragePlugin extends Plugin<StoragePluginConfig> {
     private async createStorageConfig() {
         const brokerAddress = new Wallet(this.brokerConfig.ethereumPrivateKey).address
         const apiUrl = this.brokerConfig.streamrUrl + '/api/v1'
-        const storageConfig = await StorageConfig.createInstance(brokerAddress, apiUrl, this.pluginConfig.storageConfig.refreshInterval)
+        const storageConfig = await StorageConfig.createInstance(
+            this.pluginConfig.cluster.clusterAddress || brokerAddress, 
+            this.pluginConfig.cluster.clusterSize || 1,
```
teogeb: The schema specifies that `clusterSize` and `myIndexInCluster` are always non-null `numbers` -> no need to have fallback values

teogeb (approved):
LGTM

packages/broker/src/plugins/storage/StorageConfig.ts
harbu:
```suggestion
const belongsToMeInCluster = (key: StreamKey, clusterSize: number, myIndexInCluster: number): boolean => {
```

packages/broker/src/plugins/storage/StoragePlugin.ts
```diff
@@ -21,6 +21,11 @@ export interface StoragePluginConfig {
     storageConfig: {
         refreshInterval: number
     }
+    cluster: {
+        clusterAddress: string,
```
harbu: When would the `clusterAddress` be null here?

packages/broker/src/plugins/storage/StoragePlugin.ts
harbu: Yeah it might be better to have one source for default values to avoid any potential confusion.

packages/broker/src/plugins/storage/config.schema.json
```diff
@@ -56,6 +56,32 @@
...
+                },
+                "clusterSize": {
+                    "type": "number"
+                },
+                "myIndexInCluster": {
+                    "type": "number"
+                }
+            },
+            "default": {
+                "clusterAddress": null,
```
harbu, line 81: What does it mean for `clusterAddress` to be null?

packages/broker/src/plugins/storage/StoragePlugin.ts
hpihkala: @harbu the logic is that if `clusterAddress` is null (which it is by default), the the address of the node is used. I'm not sure if it's possible to configure such logic in the json schema and/or TS file?

The idea was that the default values should result in a singleton storage node (= a cluster of size 1, subscribing to streams assigned to the node's own address).

packages/broker/src/plugins/storage/StoragePlugin.ts
hpihkala: Removed the fallbacks for `clusterSize` and `myIndexInCluster`. I guess there's no way to have the logic

> if `clusterAddress` is null, use the `brokerAddress`

...in the schema - so that should still stay here?

packages/broker/src/plugins/storage/StoragePlugin.ts
hpihkala: Added comment about it

hpihkala:
I think the test failures in this branch are now the same as in `main`... 😁

packages/broker/src/plugins/storage/StorageConfig.ts
```diff
@@ -118,40 +124,62 @@ export class StorageConfig {
...
+        this.streamKeys = new Set([...this.streamKeys].filter((x) => !keysToRemove.has(x)))
         this.listeners.forEach((listener) => {
-            streamKeys.forEach((key: StreamKey) => listener.onStreamRemoved(getStreamFromKey(key)))
+            keysToRemove.forEach((key: StreamKey) => listener.onStreamRemoved(getStreamFromKey(key)))
         })
     }
 
+    private belongsToMeInCluster(key: StreamKey): boolean {
+        const hashedIndex = Protocol.Utils.keyToArrayIndex(this.clusterSize, key.toString())
+        return hashedIndex === this.myIndexInCluster
```
timoxley, line 152: So if the cluster size changes, data may go to a different node, will this break data retrieval?

packages/broker/src/plugins/storage/StorageConfig.ts
```diff
@@ -43,22 +43,26 @@ export class StorageConfig {
...
+    clusterId: string
+    clusterSize: number
+    myIndexInCluster: number
     apiUrl: string
     private _poller!: ReturnType<typeof setTimeout>
     private _stopPoller: boolean
 
     // use createInstance method instead: it fetches the up-to-date config from API
-    constructor(nodeId: string, apiUrl: string) {
+    constructor(clusterId: string, clusterSize: number, myIndexInCluster: number, apiUrl: string) {
```
timoxley, line 54: What happens if the `clusterSize` config differs between nodes? e.g. config updated but nodes not restarted at the same time

What happens if there are multiple nodes with the same `myIndexInCluster`?

packages/broker/src/plugins/storage/StorageConfig.ts
hpihkala: It won't break data retrieval, because all the data still goes into the same Cassandra cluster. All of the nodes in the cluster can equally well retrieve data from that Cassandra cluster. 

All nodes must be reconfigured and restarted if the cluster size is changed, so that they are aligned on who subscribes to what - see the answers to your other question.

packages/broker/src/plugins/storage/StorageConfig.ts
hpihkala: > What happens if the clusterSize config differs between nodes? e.g. config updated but nodes not restarted at the same time

In that case the nodes aren't aligned on who handles what. Some streams may be subscribed to by multiple nodes (not a fatal problem, see below), while some streams may not be subscribed to by any nodes (the bigger problem). It's the responsibility of the storage cluster operator to ensure correct configuration (which is acceptable for now IMO).

> What happens if there are multiple nodes with the same myIndexInCluster?

In that case the nodes will subscribe to the exact same subset of streams, and all that data gets written to Cassandra multiple times (first written and then overwritten). It will work, but produces unnecessary load and cleanup overhead for Cassandra.

packages/broker/src/plugins/storage/StorageConfig.ts
timoxley: Ahh I didn't catch the "same Cassandra cluster" part, makes sense now.

packages/broker/src/plugins/storage/StoragePlugin.ts
harbu: Ah got it!

PR #182, opened 2021-09-03
NET-493: Don't assign 'sec' metrics streams to storage

bearni95:
Avoid storing the 'sec' metric on any storage node.

### DIFF
--- a/packages/broker/src/plugins/metrics/node/MetricsPublisher.ts
+++ b/packages/broker/src/plugins/metrics/node/MetricsPublisher.ts
@@ -72,21 +72,24 @@ export class MetricsPublisher {
         const stream = await this.client.getOrCreateStream({
             id: this.getStreamId(periodLegth)
         })
-        // TODO: pretify this error handler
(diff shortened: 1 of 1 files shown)


packages/broker/src/plugins/metrics/node/MetricsPublisher.ts
```diff
@@ -72,21 +72,24 @@ export class MetricsPublisher {
...
         })
-        // TODO: pretify this error handler
-        // https://linear.app/streamr/issue/BACK-155/assign-a-stream-to-a-storage-node-when-it-has-already-been-assigned
-        try {
-            await stream.addToStorageNode(this.storageNodeAddress)
-        } catch (e) {
-            if (!e.body) { throw e }
-            let parsedBody
+
+        if (periodLegth > PERIOD_LENGTHS.FIVE_SECONDS) {
```
teogeb: Could use exact comparison `(periodLength !== PERIOD_LENGTHS.FIVE_SECONDS)` as that is the only exception and we are just comparing two IDs.

harbu (approved):

PR #184, opened 2021-09-06
network [NET-461]: only connect to trackers that are needed

harbu:
Change behavior of network node so that it only maintains connections to trackers that correspond to streams it has joined. It used to work so that the network node would connect to all trackers and maintain connections regardless of streams.


packages/network/src/composition.ts
```diff
@@ -102,7 +103,8 @@ export const createNetworkNode = ({
     newWebrtcConnectionTimeout,
     webrtcDatachannelBufferThresholdLow,
     webrtcDatachannelBufferThresholdHigh,
-    stunUrls = ['stun:stun.l.google.com:19302']
+    stunUrls = ['stun:stun.l.google.com:19302'],
+    trackerConnectionMaintenanceInterval
```
harbu, line 110: Added this to constructor (as opposed to constant) as it is useful to be able to make the interval very small in tests for fast(er) tests.

packages/network/src/logic/Node.ts
```diff
@@ -192,6 +192,7 @@ export class Node extends EventEmitter {
         if (!this.streams.isSetUp(streamId)) {
             this.logger.trace('add %s to streams', streamId)
             this.streams.setUpStream(streamId)
+            this.maintainTrackerConnections()
```
harbu, line 199: This is here so the node can immediately react when a new stream has been joined to (as opposed to waiting for next interval execution)

packages/network/test/integration/ws-endpoint.test.ts
```diff
@@ -55,7 +55,7 @@ describe('ws-endpoint', () => {
 
     it('server and client form correct peerInfo on connection', async () => {
         const client = new NodeClientWsEndpoint(PeerInfo.newNode('client'))
-        const server = await startServerWsEndpoint('127.0.0.1', 30696, PeerInfo.newNode('server'))
+        const server = await startServerWsEndpoint('127.0.0.1', 30697, PeerInfo.newNode('server'))
```
harbu, line 58: Noticed port conflict, fixed

juslesan (approved):
LGTM, still need to fix merge conflicts against per stream status messages

packages/network/src/logic/Node.ts
```diff
@@ -101,13 +101,13 @@ export class Node extends EventEmitter {
...
         this.peerInfo = opts.peerInfo
 
-        this.connectToBootstrapTrackersInterval = opts.connectToBootstrapTrackersInterval || 5000
         this.sendStatusToAllTrackersInterval = opts.sendStatusToAllTrackersInterval || 1000
         this.bufferTimeoutInMs = opts.bufferTimeoutInMs || 60 * 1000
         this.bufferMaxSize = opts.bufferMaxSize || 10000
         this.disconnectionWaitTime = opts.disconnectionWaitTime || 30 * 1000
         this.nodeConnectTimeout = opts.nodeConnectTimeout || 15000
         this.instructionRetryInterval = opts.instructionRetryInterval || 60000
+        this.trackerConnectionMaintenanceInterval = opts.trackerConnectionMaintenanceInterval ?? 5000
```
timoxley, line 115: 👍

timoxley (approved):
lgtm

timoxley (commented):
Ah, async work in a `setInterval` could be a problem, check comment: https://github.com/streamr-dev/network-monorepo/pull/184#discussion_r702987068

packages/network/src/logic/Node.ts
```diff
@@ -525,12 +526,21 @@ export class Node extends EventEmitter {
...
     }
 
-    private connectToBootstrapTrackers(): void {
-        this.trackerRegistry.getAllTrackers().forEach((trackerInfo) => {
-            this.trackerNode.connectToTracker(trackerInfo.ws, PeerInfo.newTracker(trackerInfo.id))
-                .catch((err) => {
-                    this.logger.warn('could not connect to tracker %s, reason: %j', trackerInfo.ws, err)
-                })
+    private maintainTrackerConnections(): void {
+        const activeTrackers = new Set<string>()
```
timoxley, line 541: what happens if `maintainTrackerConnections` is called while `connectToTracker` or `disconnectFromTracker` is still pending from previous `setInterval`?

is it safe to call `connectToTracker` twice with the same info? or `disconnectFromTracker` while it's currently trying to `connectToTracker`? 

Maybe this should be a `setTimeout(maintainTrackerConnections, interval)` after a `Promise.allSettled`

packages/network/src/logic/Node.ts
timoxley: Hm, does this mean there could be up to a 5s delay before a node will connect to the stream-appropriate tracker, and thus up to a 5s delay before it will start connecting to peers on a different tracker?

packages/network/src/logic/Node.ts
harbu: > what happens if `maintainTrackerConnections` is called while `connectToTracker` or `disconnectFromTracker` is still pending from previous `setInterval`?

It should work fine. The Promise returned by `connectToTracker` is optional to `await` on. It's there for unit / integration test purposes where you are interested for a connection to be formed before continuing to do something (e.g. publish messages). Generally outside testing, it's more of a fire-and-forget command style ordeal. 

> is it safe to call `connectToTracker` twice with the same info? or `disconnectFromTracker` while it's currently trying to `connectToTracker`?

Yeah if you call `connectToTracker` twice, the 2nd call is effectively a NOP. There may be a bug now that I read the code wherein a _pending_ connection doesn't properly get killed when calling `disconnectFromTracker`. That should be killed, however.

> Maybe this should be a `setTimeout(maintainTrackerConnections, interval)` after a `Promise.allSettled`

I think we should rather not return a Promise at all (to avoid confusion) and just have it as fire-and-forget style. This way the logic layer is better separated from whatever the connection layer is doing (closer to the connection-manager style of thinking).

PR #186, opened 2021-09-06
NET-496: select a random stream from a configured list (single tracker)

juslesan:
- Subscribe to a random stream from a given list
- Add selected the stream id to the claim request
- [NET-499](https://linear.app/streamr/issue/NET-499/use-multiple-trackers-on-miner-plugin) for using multiple trackers

### DIFF
--- a/packages/broker/src/plugins/testnetMiner/config.schema.json
+++ b/packages/broker/src/plugins/testnetMiner/config.schema.json
@@ -5,9 +5,11 @@
     "description": "Testnet miner plugin configuration",
     "additionalProperties": false,
     "properties": {
-        "rewardStreamId": {
-            "type": "string",
-            "default": "streamr.eth/testnets/brubeck-rewards"
+        "rewardStreamIds": {
+            "type": "array",
+            "items": { "type": "string" },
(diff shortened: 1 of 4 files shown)


harbu (approved):

packages/broker/src/plugins/testnetMiner/config.schema.json
```diff
@@ -5,9 +5,11 @@
...
     "additionalProperties": false,
     "properties": {
-        "rewardStreamId": {
-            "type": "string",
-            "default": "streamr.eth/testnets/brubeck-rewards"
+        "rewardStreamIds": {
+            "type": "array",
+            "items": { "type": "string" },
+            "uniqueItems": true,
+            "default": ["streamr.eth/testnets/brubeck-rewards"]
```
teogeb, line 12: The default value should contain list of all testnet reward streams? @juslesan Do we know the stream IDs already?

juslesan: No, the real names for the reward streams are not known yet. There's a need to make a configuration PR once all urgent  changes are completed

PR #188, opened 2021-09-07
network: type annotations for PeerId, NodeId and TrackerId

teogeb:
Added type aliases for `PeerId`, `NodeId` and `TrackerId`. 

All those types are `strings`, but as they have aliases, it is now easier to understand some of the data structures. E.g.
```
private readonly disconnectionTimers: { [key: string]: NodeJS.Timeout }
private readonly trackerBook: { [key: string]: string } // address => id
private readonly consecutiveDeliveryFailures: { [key: string]: number } // id => counter
```
are now 
```
private readonly disconnectionTimers: Record<NodeId,NodeJS.Timeout>
private readonly trackerBook: { [key: string]: TrackerId } // address => id
private readonly consecutiveDeliveryFailures: Record<NodeId,number> // id => counter
```

Renamed also some variables e.g. in `InstructionRetryManager` to reflect the types more accurately.


packages/network/src/connection/WebRtcEndpoint.ts
```diff
@@ -120,7 +120,7 @@ export class WebRtcEndpoint extends EventEmitter implements IWebRtcEndpoint {
     }
 
     private createConnection(
-        targetPeerId: string,
+        targetPeerId: PeerId,
         routerId: string,
```
juslesan, line 124: RouterId could be annotated as signalerId, but a I prefer the more generic string type for now

packages/network/src/identifiers.ts
```diff
@@ -50,13 +51,11 @@ export interface Location {
...
-        inboundNodes: string[]
-        outboundNodes: string[]
-        counter: number
-    }
-}
+export type StatusStreams = Record<StreamKey, {
+    inboundNodes: NodeId[]
+    outboundNodes: NodeId[]
+    counter: number
+}>
```
juslesan, line 58: 👍

packages/network/src/protocol/NodeToNode.ts
```diff
@@ -44,26 +45,26 @@ export class NodeToNode extends EventEmitter {
     }
 
     connectToNode(
-        receiverNodeId: string,
+        receiverNodeId: NodeId,
         trackerAddress: string,
         trackerInstructed = true
-    ): Promise<string> {
+    ): Promise<PeerId> {
```
juslesan: NodeId is returned here

packages/network/src/protocol/TrackerNode.ts
```diff
@@ -134,11 +136,11 @@ export class TrackerNode extends EventEmitter {
         return requestId
     }
 
-    async send<T>(receiverNodeId: string, message: T & TrackerLayer.TrackerMessage): Promise<void> {
-        await this.endpoint.send(receiverNodeId, message.serialize())
+    async send<T>(receiverPeerId: PeerId, message: T & TrackerLayer.TrackerMessage): Promise<void> {
```
juslesan: Just a thought, the receiver here should always be a tracker, see second parameter type. However, once we move towards a decentralized tracker algorithm the receiver would be a super node (or a node with a public ip)

Given that in most cases in this class the "PeerId" is "TrackerId" maybe it's the better option for now.

packages/network/src/protocol/TrackerNode.ts
```diff
@@ -157,7 +159,7 @@ export class TrackerNode extends EventEmitter {
         }
     }
 
-    connectToTracker(trackerAddress: string, trackerPeerInfo: PeerInfo): Promise<string> {
+    connectToTracker(trackerAddress: string, trackerPeerInfo: PeerInfo): Promise<PeerId> {
```
juslesan: Always returns a TrackerId

packages/network/src/protocol/NodeToNode.ts
```diff
@@ -117,7 +118,7 @@ export class NodeToNode extends EventEmitter {
         return this.endpoint.getRtts()
     }
 
-    getNegotiatedProtocolVersionsOnNode(peerId: string): [number, number] {
+    getNegotiatedProtocolVersionsOnNode(peerId: PeerId): [number, number] {
```
juslesan: Change peerId: PeerId to nodeId: NodeId

juslesan (changes requested):
In general, I think that it's best to avoid using the "PeerId" type in the Logic + Protocol layers. Instead use "NodeId" and "TrackerId" as much as possible there. In the connection layer, always using "PeerId" is fine as it is agnostic to the higher layer functionality.

packages/network/src/protocol/NodeToNode.ts
teogeb: Fixed

packages/network/src/protocol/NodeToNode.ts
teogeb: Fixed

packages/network/src/protocol/TrackerNode.ts
teogeb: Fixed

packages/network/src/protocol/TrackerNode.ts
teogeb: Changed to `TrackerId`

juslesan (approved):

PR #191, opened 2021-09-07
Fix tracker hashing

hpihkala:
Use our "standard" string to array index hashing method for selecting tracker, instead of the simple home-made one.

The first commit was cherry-picked from #180 to get it into use faster.

This change is only breaking if there is a network where old and new code are mixed. But that won't be the case in testnet2 and beyond.


juslesan (approved):

timoxley (approved):
👍 nice one

hpihkala:
The single cross-client test failure seems unrelated to changes. Merging

PR #192, opened 2021-09-08
network [NET-472]: debounced instruction send for Tracker

teogeb:
Tracker instructions are collected to buffers and sent after a short delay. For each stream there is a separate buffer.

We use debouncing to delay the sending. We send all the instructions in a stream buffer, when either when of these conditions is satisfied: 
 - the topology stabilizes: no new instructions has been formed for the stream in X milliseconds
- the buffer times out: we have buffered an instruction for Y milliseconds

Added new command line options for the two configuration options. The default values are currently 100 ms and 2000 ms (we can optimize the values later, if needed).

### DIFF
--- a/packages/network/src/logic/InstructionSender.ts
+++ b/packages/network/src/logic/InstructionSender.ts
@@ -1,0 +1,108 @@
+import _ from 'lodash'
+import { Logger } from '../helpers/Logger'
+import { Metrics } from '../helpers/MetricsContext'
+import { StreamIdAndPartition, StreamKey } from '../identifiers'
+import { TrackerServer } from '../protocol/TrackerServer'
+import { NodeId } from './Node'
+import { TopologyStabilizationOptions } from './Tracker'
+
+/**
+ * Instructions are collected to buffers and sent after a short delay. For each stream
+ * there is a separate buffer.
+ * 
+ * We use debouncing to delay the sending. It means that we send the buffered instructions 
+ * when either of these conditions is satisfied: 
+ * - the topology stabilizes: no new instructions has been formed for the stream in 
+ *   X milliseconds
+ * - the buffer times out: we have buffered an instruction for Y milliseconds
+ * 
+ * When an instruction is added to a stream buffer, it may overwrite an existing
+ * instruction in the buffer if the both instructions share the same nodeId. In that 
+ * situation we expect that the previous instruction is no longer valid (it has a lower
+ * counterValue) and can be ignored.
+ */
+
+const DEFAULT_TOPOLOGY_STABILIZATION: TopologyStabilizationOptions = {
+    debounceWait: 100,
+    maxWait: 2000
+}
+
+const logger = new Logger(module)
+
+export interface Instruction {
+    nodeId: NodeId,
+    streamKey: StreamKey,
+    newNeighbors: NodeId[],
+    counterValue: number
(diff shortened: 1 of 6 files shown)


packages/network/src/logic/InstructionSender.ts
```diff
@@ -0,0 +1,108 @@
...
+        this.instructions.set(instruction.nodeId, instruction)
+        this.debouncedOnReady()
+    }
+
+    getInstructions(): IterableIterator<Instruction> {
+        return this.instructions.values()
+    }
+}
+
+export class InstructionSender {
```
juslesan, line 60: Maybe InstructionDebouncer or Buffer would be a more accurate name

juslesan (approved):
After testing this change significantly reduced the churn cycles required to form d-regular random graphs.

Locally, dumping around 100 nodes to the stream results in a completed graph in the first instruction cycle, reducing the amount of strain on the tracker significantly.

The branch was also tested locally in a high churn setting, where it maintained a fully connected graph for subscribing nodes, and fully resolved after the churn was over. 

During churn testing an unrelated bug to this PR [was found](https://linear.app/streamr/issue/NET-502/unused-webrtc-connections-are-not-cleaned-properly). Unused connections are not properly cleaned up by the nodes in some situations, requiring further investigation. The bug does not affect message propagation but could explain high memory consumption of some nodes.

packages/network/src/logic/InstructionSender.ts
```diff
@@ -0,0 +1,108 @@
...
+                this.streamBuffers.delete(streamKey)
+                this.sendInstructions(newBuffer)
+            })
+            this.streamBuffers.set(streamKey, newBuffer)
+            return newBuffer
+        }
+    }
+
+    private async sendInstructions(buffer: StreamInstructionBuffer) {
+        for (const instruction of buffer!.getInstructions()) {
```
timoxley, line 92: what happens, or should happen if it's still executing another `sendInstructions`? I guess the counterValue disambiguates and gives an ordering, but perhaps it would be good to cancel any pending instructions that are superseded?

packages/network/src/logic/InstructionSender.ts
teogeb: Yes, if this class would do only the debouncing logic and not the actual sending. Now it encapsulates the whole send logic (not much) and specifically the fact that debouncing is used. The `Tracker` class is quite large, so maybe it is good we some details can be refactored out from that class.

One alternative would be to divide this class to `InstructionSender` + `InstructionDebouncer`. Maybe we should do that if/when the class grows.

packages/network/src/logic/InstructionSender.ts
teogeb: If the node receives an old instruction (a `counterValue` which is smaller than the most recent one), it will just ignore it. Also, if we form a new instruction for the same stream+node while we buffer the previous instruction, it is automatically overwritten in the buffer. 

Maybe it is quite unlikely that we'd be sending a buffer at the same time when a new instruction is formed?
- the deboucing reduces the amount of instructions significantly
- the buffer has just stabilized or timeouted -> if it stabilized, but we still form a new instruction right after the stabilization, we should maybe increase the `debounceWait` configuration option
- the sending of the buffer is a relatively fast operation

PR #195, opened 2021-09-10
NET-488: auto migrate config testnet1 -> testnet2

juslesan:
- Add auto-migration from testnet1 config to testnet2 config to broker.js executable
- Update config wizard


packages/broker/src/helpers/ConfigAutoMigrate.ts
```diff
@@ -0,0 +1,116 @@
...
+            config['network']['trackers'] = MigrateTrackerRegistry(config['network']['trackers'] as TrackerRegistryItem[])
+            delete config['plugins']['testnetMiner']['rewardStreamId']
+            config['plugins']['testnetMiner']['rewardStreamIds'] = TESTNET2_STREAM_IDS
+            validateConfig(config.plugins.testnetMiner, schemaTestnet2)
+
+            console.info('Updating config file at ' + configFilePath)
+            fs.writeFileSync(configFilePath, JSON.stringify(config, null, 2))
+            return config
+        } catch (err) {
+            console.warn('Testnet1 -> Testnet2 configuration failed', err)
```
harbu:
```suggestion
            console.warn('Testnet1 -> Testnet2 configuration migration failed', err)
```

packages/broker/src/helpers/ConfigAutoMigrate.ts
```diff
@@ -0,0 +1,116 @@
...
+export const testnet2AutoMigrate = (config: any, configFilePath: string): Config => {
+    try {
+        validateConfig(config.plugins.testnetMiner, schemaTestnet2)
+        return config
+    } catch (err) {
+        try {
+            console.info('Migrating testnet1 config to testnet2 config')
+            validateConfig(config, schema)
+
+            if (config.plugins.testnetMiner.rewardStreamIds) {
```
harbu, line 95: A bit unclear why this check is needed. Is it in case of a previously abruptly ended migration process?

packages/broker/src/helpers/ConfigAutoMigrate.ts
```diff
@@ -0,0 +1,116 @@
...
+    'streamr.eth/brubeck-testnet/rewards/z15g13',
+    'streamr.eth/brubeck-testnet/rewards/111249',
+    'streamr.eth/brubeck-testnet/rewards/0g2jha',
+    'streamr.eth/brubeck-testnet/rewards/fijka2',
+    'streamr.eth/brubeck-testnet/rewards/91ab49',
+    'streamr.eth/brubeck-testnet/rewards/giab22',
+    'streamr.eth/brubeck-testnet/rewards/25kpf4'
+]
+
+const MigrateTrackerRegistry = (trackerRegistry: TrackerRegistryItem[]): TrackerRegistryItem[] => {
```
harbu:
```suggestion
const migrateTrackerRegistry = (trackerRegistry: TrackerRegistryItem[]): TrackerRegistryItem[] => {
```

packages/broker/src/helpers/ConfigAutoMigrate.ts
juslesan: The validateConfig function seems to have a side effect of adding the default values to the config object. This happens in the first try { validateConfig }. The added rewardsStreamIds has to be removed here to validate against the previous version's schema.

harbu: Ahh interesting...

harbu (approved):

PR #198, opened 2021-09-13
network, broker [NET-440]: improve logging

teogeb:
Small improvements for logging:
- If a node is unable to connect a tracker, the warning is logged only once. Previously the warning was logged once in every 5 seconds. To implement that, the connection logic was extracted to `TrackerConnector` class.
- Changed the log levels of some connection/ping errors from from `warn` to `info`. Closed connections are normal behavior of a node -> no need to warn about that.

Removed also deprecated `connectToBootstrapTrackersInterval` configuration option (https://github.com/streamr-dev/network-monorepo/pull/184)

### DIFF
--- a/packages/network/src/logic/TrackerConnector.ts
+++ b/packages/network/src/logic/TrackerConnector.ts
@@ -1,0 +1,65 @@
+import { Utils } from 'streamr-client-protocol'
+import { StreamIdAndPartition, TrackerInfo } from '../identifiers'
+import { NodeToTracker } from '../protocol/NodeToTracker'
+import { Logger } from '../helpers/Logger'
+import { PeerInfo } from '../connection/PeerInfo'
+import { TrackerId } from './Tracker'
+
+export class TrackerConnector {
+
+    private readonly getStreams: () => ReadonlyArray<StreamIdAndPartition>
+    private readonly nodeToTracker: NodeToTracker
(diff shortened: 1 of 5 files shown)


packages/network/src/logic/TrackerConnector.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export class TrackerConnector {
+
+    private readonly getStreams: () => ReadonlyArray<StreamIdAndPartition>
+    private readonly nodeToTracker: NodeToTracker
+    private readonly trackerRegistry: Utils.TrackerRegistry<TrackerInfo>
+    private readonly logger: Logger
+    private maintenanceTimer?: NodeJS.Timeout | null
+    private readonly maintenanceInterval: number
+    private connectionErrors: Map<TrackerId,boolean>
```
harbu: Would a `Set` be more idiomatic here?  Also name could be more indicative of what the role is?
```suggestion
    private trackersWithErrorsSilenced: Set<TrackerId>
```

harbu (approved):

juslesan (approved):

packages/network/src/logic/TrackerConnector.ts
teogeb: Fixed

PR #200, opened 2021-09-15
NET-507: Node tracker logic refactor & performance increases

harbu:
- Performance improvement: `maintainConnections()` no longer has to loop thru all streams
    - Previous implementation would loop through all streams and collect associated trackers into a set
    - Now we loop all trackers, and for each tracker, check whether a stream exists that is assigned to it. This is done in a short-circuit way, which should make it quite efficient compared to the old way.
    - We know that `#streams >> #trackers` so this way of doing things is more efficient.
- Performance improvement: now when we join a stream, instead of going and maintaining _all_ tracker connections (via `maintainConnections()`) we only connect to the associated tracker.
- Refactor: Found way to simplify `StatusMessage` forming & sending logic.
- Refactor: Moved all Node's tracker related logic to `TrackerManager`.

### DIFF
--- a/packages/network/src/logic/TrackerConnector.ts
+++ b/packages/network/src/logic/TrackerConnector.ts
@@ -8,6 +8,11 @@ import { StreamManager } from './StreamManager'
 
 const logger = new Logger(module)
 
+enum ConnectionState {
+    SUCCESS,
+    ERROR
+}
+
 export class TrackerConnector {
 
     private readonly streamManager: StreamManager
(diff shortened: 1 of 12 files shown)


packages/network/src/logic/TrackerConnector.ts
```diff
@@ -62,4 +59,32 @@ export class TrackerConnector {
...
+                    // -> now it doesn't log anything if the connection error reason changes
+                    this.connectionStates.set(id, ConnectionState.ERROR)
+                    logger.warn('Could not connect to tracker %s, reason: %s', id, err.message)
+                }
+            })
+    }
+
+    private isActiveTracker(trackerId: TrackerId): boolean {
+        for (const streamKey of this.streamManager.getStreamKeys()) {
+            const { id: streamId, partition } = StreamIdAndPartition.fromKey(streamKey)
```
timoxley, line 89: `SPID`

packages/network/src/logic/TrackerManager.ts
```diff
@@ -1,17 +1,25 @@
...
+import { NodeId, NodeOptions } from './Node'
+import { InstructionThrottler } from './InstructionThrottler'
+import { InstructionRetryManager } from './InstructionRetryManager'
+import { Metrics } from '../helpers/MetricsContext'
 
 const logger = new Logger(module)
 
 type FormStatusFn = (streamId: StreamIdAndPartition, includeRtt: boolean) => Status
 
+interface Subscriber {
```
harbu, line 17: name?

packages/network/src/logic/TrackerConnector.ts
harbu: I'd rather we do the conversion from `StreamIdAndPartition` to `SPID` in a separate PR (to avoid having yet another standard -type of a situation). I made an issue on this https://linear.app/streamr/issue/NET-518/network-and-broker-replace-streamidandpartition-with-spid.

teogeb (approved):

juslesan (approved):
LFTM, maybe its time to split the tracker and node logic to separate directories inside `src/logic/`

PR #202, opened 2021-09-16
Net 511: unix sockets in websocket server

juslesan:
- Allow using Unix sockets in tracker's WebSocket server

### DIFF
--- a/packages/broker/test/integration/SubscriptionManager.test.ts
+++ b/packages/broker/test/integration/SubscriptionManager.test.ts
@@ -28,7 +28,7 @@ describe('SubscriptionManager', () => {
 
     beforeEach(async () => {
         tracker = await startTracker({
-            listenConfig: {
+            listen: {
                 hostname: '127.0.0.1',
                 port: trackerPort
             },
--- a/packages/broker/test/integration/SubscriptionManager.test.ts
+++ b/packages/broker/test/integration/SubscriptionManager.test.ts
@@ -28,8 +28,10 @@ describe('SubscriptionManager', () => {
 
     beforeEach(async () => {
         tracker = await startTracker({
-            host: '127.0.0.1',
-            port: trackerPort,
+            listenConfig: {
+                hostname: '127.0.0.1',
+                port: trackerPort
+            },
             id: 'tracker'
         })
 
--- a/packages/network/bin/tracker.js
+++ b/packages/network/bin/tracker.js
@@ -12,6 +12,7 @@ program
     .option('--trackerName <trackerName>', 'Human readable name', undefined)
     .option('--port <port>', 'port', '27777')
     .option('--ip <ip>', 'ip', '0.0.0.0')
+    .option('--unixSocket <unixSocket>', 'unixSocket', undefined)
     .option('--maxNeighborsPerNode <maxNeighborsPerNode>', 'maxNeighborsPerNode', '4')
     .option('--metrics <metrics>', 'output metrics to console', false)
     .option('--metricsInterval <metricsInterval>', 'metrics output interval (ms)', '5000')
@@ -23,6 +24,10 @@ program
 const id = program.opts().id || `TR${program.opts().port}`
 const name = program.opts().trackerName || id
 const logger = new Logger(module)
+const listenConfig = program.opts().unixSocket ? program.opts().unixSocket : {
+    hostname: program.opts().ip,
+    port: program.opts().port
+}
 
 const getTopologyStabilization = () => {
     const debounceWait = program.opts().topologyStabilizationDebounceWait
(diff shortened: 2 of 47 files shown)


packages/broker/test/integration/SubscriptionManager.test.ts
```diff
@@ -28,8 +28,10 @@ describe('SubscriptionManager', () => {
 
     beforeEach(async () => {
         tracker = await startTracker({
-            host: '127.0.0.1',
-            port: trackerPort,
+            listenConfig: {
```
timoxley: maybe this should be `listen` instead of `listenConfig`, I feel like the `config` is implied

packages/network/bin/tracker.js
```diff
@@ -23,6 +24,10 @@ program
 const id = program.opts().id || `TR${program.opts().port}`
 const name = program.opts().trackerName || id
 const logger = new Logger(module)
+const listenConfig = program.opts().unixSocket ? program.opts().unixSocket : {
+    hostname: program.opts().ip,
+    port: program.opts().port
```
timoxley: does this need the `Number.parseInt(program.opts().port, 10)` that was there before?

timoxley (dismissed):
if the program is terminated abormally, e.g. with `SIGKILL`,  the socket file will likely not get cleaned up and the program will refuse to boot until the file is manually removed.

https://stackoverflow.com/questions/16178239/gracefully-shutdown-unix-socket-server-on-nodejs-running-under-forever/16502680

> I suggest you clean up on start. When you get EADDRINUSE, try to connect to the socket. If the socket connection succeeds, another server is running and so this instance should exit. If the connection fails then it is safe to unlink the socket file and create a new one.

packages/network/src/connection/ws/ServerWsEndpoint.ts
```diff
@@ -188,8 +189,42 @@ export async function startHttpServer(
...
+                        fs.unlinkSync(config)
+                        httpServer.listen(config, function() {
+                            staticLogger.info('server recovered')
+                            return httpServer
+                        })
+                    }
+                })
+                clientSocket.connect({path: config},function() {
+                    staticLogger.trace('Server is already running in ' + config)
+                    process.exit()
```
timoxley: should probably exit(1)

packages/network/src/connection/ws/ServerWsEndpoint.ts
```diff
@@ -177,9 +189,42 @@ export async function startHttpServer(
...
+                })
+            }
+        })
+    }
+    try {
+        httpServer.listen(config)
+        await once(httpServer, 'listening')
+        staticLogger.info(`listening on %s`, JSON.stringify(config))
+    } catch (err) {
```
timoxley, line 239: What throws? the `once`?

packages/network/test/integration/UnixSocketWsServer.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    let serverWsEndpoint: ServerWsEndpoint | undefined = undefined
+
+    afterEach(async () => {
+        try {
+            await serverWsEndpoint?.stop()
+        } catch (err) {
+        }
+    })
+
+    test('WS connection can established with unix sockets', async () => {
```
timoxley, line 16: Should probably test these two socket cleanup cases:

socket file exists but nothing listening
socket file exists but another server listening

timoxley (commented):
I refactored that code a little bit (untested/unlinted) and added some unlink error handling, process.exit(1), feel free to use some/all/none of this:

```typescript
function cleanSocket() {
    httpServer.on('error', (err: any) => {
        // rethrow if unexpected error
        if (!err.message.includes('EADDRINUSE')) { throw err }

        staticLogger.info('socket in use, trying to recover: %s', config)
        staticLogger.trace('checking if socket in use by another server')
        const clientSocket = new net.Socket()
        // socket will automatically close on error
        clientSocket.on('error', (err: any) => {
            // rethrow if unexpected error
            if (!err.message.includes('ECONNREFUSED')) { throw err }
            
            // No other server listening
            try {
                staticLogger.trace('cleaning unused socket: %s', config)
                fs.unlinkSync(config)
            } catch (unlinkErr) {
                // ignore error if somehow file was already removed
                if (unlinkErr.code !== 'ENOENT') {
                    throw unlinkErr
                }
            }

            // retry listening
            httpServer.listen(config)
        })

        clientSocket.once('connect', () => {
            // bad news if we are able to connect
            staticLogger.trace('Another server already running on socket: %s', config)
            process.exit(1)
        })

        clientSocket.connect({ path: config })
    })
}

if (typeof config === 'string') {
    cleanSocket()
}
```

packages/network/src/connection/ws/ServerWsEndpoint.ts
juslesan: Seems that the await once() throws here when the httpServer errors with EADDRINUSE

timoxley (approved):

PR #203, opened 2021-09-17
NET-502: clean up unused connections

juslesan:
- Clean up leftover unused connections that remain after high churn

### DIFF
--- a/packages/network/src/logic/Node.ts
+++ b/packages/network/src/logic/Node.ts
@@ -86,6 +86,7 @@ export class Node extends EventEmitter {
     private readonly consecutiveDeliveryFailures: Record<NodeId,number> // id => counter
     private readonly rttUpdateTimeoutsOnTrackers: { [key: string]: NodeJS.Timeout } // trackerId => timeout
     private readonly metrics: Metrics
+    private connectionCleanUpInterval: NodeJS.Timeout | null
     private handleBufferedMessagesTimeoutRef?: NodeJS.Timeout | null
     protected extraMetadata: Record<string, unknown> = {}
 
@@ -167,6 +168,8 @@ export class Node extends EventEmitter {
             .addRecordedMetric('onUnsubscribeRequest')
             .addRecordedMetric('onNodeDisconnect')
             .addFixedMetric('latency')
+
+        this.connectionCleanUpInterval = this.startConnectionCleanUpInterval()
     }
 
     start(): void {
@@ -392,6 +395,11 @@ export class Node extends EventEmitter {
             this.handleBufferedMessagesTimeoutRef = null
         }
 
+        if (this.connectionCleanUpInterval) {
+            clearInterval(this.connectionCleanUpInterval)
+            this.connectionCleanUpInterval = null
+        }
+
         Object.values(this.disconnectionTimers).forEach((timeout) => clearTimeout(timeout))
         Object.values(this.rttUpdateTimeoutsOnTrackers).forEach((timeout) => clearTimeout(timeout))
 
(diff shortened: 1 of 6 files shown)


packages/network/src/logic/Node.ts
```diff
@@ -131,6 +132,7 @@ export class Node extends EventEmitter {
         )
         this.consecutiveDeliveryFailures = {}
         this.trackerConnector = new TrackerConnector(() => this.streams.getStreams(), this.nodeToTracker, this.trackerRegistry, this.logger, opts.trackerConnectionMaintenanceInterval ?? 5000)
+        this.connectionCleanUpInterval = this.startConnectionCleanUpInterval(2 * 60 * 1000)
```
teogeb: Move this to `Node.start()` method?

packages/network/src/protocol/NodeToNode.ts
```diff
@@ -125,4 +125,8 @@ export class NodeToNode extends EventEmitter {
             || this.endpoint.getDefaultControlLayerProtocolVersion()
         return [controlLayerVersion, messageLayerVersion]
     }
+
+    getAllConnectionNodeIds(): PeerId[] {
```
teogeb:
```suggestion
    getAllConnectionNodeIds(): NodeId[] {
```

teogeb (approved):

packages/network/src/logic/Node.ts
```diff
@@ -534,6 +542,19 @@ export class Node extends EventEmitter {
...
     }
 
+    private startConnectionCleanUpInterval(interval: number): NodeJS.Timeout {
+        return setInterval(() => {
+            const peerIds = this.nodeToNode.getAllConnectionNodeIds()
+            const unusedConnections = peerIds.filter((peer) => !this.isNodePresent(peer))
+            if (unusedConnections.length > 0) {
+                this.logger.debug(`Disconnecting from ${unusedConnections.length} unused connections`)
+                unusedConnections.forEach((peerId) => {
+                    this.nodeToNode.disconnectFromNode(peerId, 'Unused connection')
```
timoxley, line 553: any harm in calling this while already disconnecting?

packages/network/src/logic/Node.ts
```diff
@@ -534,6 +542,19 @@ export class Node extends EventEmitter {
         }
     }
 
+    private startConnectionCleanUpInterval(interval: number): NodeJS.Timeout {
+        return setInterval(() => {
+            const peerIds = this.nodeToNode.getAllConnectionNodeIds()
+            const unusedConnections = peerIds.filter((peer) => !this.isNodePresent(peer))
```
timoxley, line 549: is it possible this cleans up nodes that are involved in a currently executing instruction?
I wonder if this should be invoked in a debounced function that's called after each tracker instruction is completed, rather than on an interval?

packages/network/src/logic/Node.ts
```diff
@@ -172,6 +174,7 @@ export class Node extends EventEmitter {
     start(): void {
         this.logger.trace('started')
         this.trackerConnector.start()
+        this.connectionCleanUpInterval = this.startConnectionCleanUpInterval(2 * 60 * 1000)
```
timoxley, line 178: Recommend adding `clearInterval(this.connectionCleanUpInterval!)` before setting it for good measure. Accidentally having multiple intervals running can silently cause all sorts of weird problems and it's easy to prevent entirely if you *always* clear the timeout before you set a timeout/interval

packages/network/src/logic/Node.ts
juslesan: Shouldn't be a problem:
https://github.com/streamr-dev/network-monorepo/blob/4ae0467d8472109de70a5e93523978df31a0b305/packages/network/src/connection/WebRtcConnection.ts#L179

packages/network/src/logic/Node.ts
juslesan: That could be possible. However, when churn testing locally with 40 nodes, 5 minutes of churn, and a 1 minute interval on the clean up function the the network operated as expected and did not gather high numbers of unused connections as described in the ticket.

PR #204, opened 2021-09-17
SSR friendly web build

mondoreale:
In this PR I'm making the web/browser build compatible with gatsby.

- I place `window`-touching things inside functions (see uuids), initialize them on first use.
- I go with a custom `globalObject`, use `this` if the default `self` isn't there.
- Add missing "is `window` defined" checks.

Hope that works well and can be pushed out very soon. **I need this for the pub site launch.**

cc @hpihkala


jtakalai (commented):
lots of lint complaints for missing types

mondoreale:
@jtakalai anything there about what I did here or generally?

packages/client/src/utils/index.ts
```diff
@@ -16,14 +16,18 @@ import Scaffold from './Scaffold'
 
 export { AggregatedError, Scaffold }
 
-const UUID = uuidv4()
+let UUID: any
```
timoxley: `:string`

packages/client/src/StreamrClient.ts
```diff
@@ -149,8 +149,7 @@ class StreamrCached {
     }
 }
 
-// use process id in node uid
-const uid = process.pid != null ? process.pid : `${uuid().slice(-4)}${uuid().slice(0, 4)}`
+let uid: any
```
timoxley: `:string`

timoxley (approved):
I wonder if `globalThis` would "just work"

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis

e.g. 

node:
`globalThis.process`

browser:
`globalThis.fetch`

timoxley:
> I place window-touching things inside functions (see uuids), initialize them on first use.

Why

timoxley:
also wondering if maybe we need to shim `uuidv4` for browser?

mondoreale:
> > I place window-touching things inside functions (see uuids), initialize them on first use.
> 
> Why

Because we don't shim `uuidv4` for browser. ;)

mondoreale:
`globalThis` works.

packages/client/src/StreamrClient.ts
```diff
@@ -192,6 +191,12 @@ export class StreamrClient extends EventEmitter { // eslint-disable-line no-rede
     // TODO annotate connection parameter as internal parameter if possible?
     constructor(options: StreamrClientOptions = {}, connection?: StreamrConnection) {
         super()
+
+        if (typeof uid === 'undefined') {
+            // use process id in node uid
+            uid = process.pid != null ? `${process.pid}` : `${uuid().slice(-4)}${uuid().slice(0, 4)}`
+        }
```
timoxley: Please add some comments as to why this is done here otherwise it's likely it'll end up accidentally refactored back to how it was

packages/client/src/utils/index.ts
```diff
@@ -16,14 +16,18 @@ import Scaffold from './Scaffold'
...
 
 export const SEPARATOR = '-'
 /*
  * Incrementing + human readable uuid
  */
 
 export function uuid(label = '') {
+    if (!UUID) {
+        UUID = uuidv4()
+    }
```
timoxley: same

timoxley (approved):

packages/client/src/StreamrClient.ts
mondoreale: Good point.

mondoreale:
@timoxley, can you have another look? It's sort of compromise between a proper shim and touching things when they're needed.

mondoreale:
It was a bit weird to describe reasons behind the previous approach without telling the whole story so I did something more self-explanatory.

mondoreale:
Any chance we could get that merged and released today?

mondoreale:
@timoxley can you assist me here? I'd like to merge that one in and move on with the site. Can I have a 5.4.5 maybe for it? I see there's a v6 in the cooking. LMK

timoxley (approved):
add comments then lgtm.

mondoreale:
@timoxley, done. If it's all good please merge and publish. Thanks!

packages/client/src/utils/uuid.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const SEPARATOR = '-'
+
+let UUID: string
+
+export default function uuid(label = ''): string {
+    if (typeof UUID === 'undefined') {
+        // Create UUID on the first use of the function in order to avoid premature `uuidv4` calls.
+        // Doing it otherwise will break browser projects that utilize server-side rendering (no
+        // `window` while build's target is `web`).
```
mondoreale: Added a similar one here.

timoxley (approved):

PR #205, opened 2021-09-17
Report storage refresh failure correctly.

timoxley:
Changes errors like this:
```
Unable to refresh storage config: TypeError: json.flatMap is not a function
```

into something more clear:
```
Unable to refresh storage config: Error: Refresh failed: 500 {"code":"MySQLSyntaxErrorException","message":"Unknown column 'this_.id' in 'field list'"}
```

image

### DIFF
--- a/packages/broker/src/plugins/storage/StorageConfig.ts
+++ b/packages/broker/src/plugins/storage/StorageConfig.ts
@@ -97,7 +97,14 @@ export class StorageConfig {
 
     async refresh(): Promise<void> {
         const res = await fetch(`${this.apiUrl}/storageNodes/${this.nodeId}/streams`)
+        if (!res.ok) {
+            throw new Error(`Refresh failed: ${res.status} ${await res.text()}`)
+        }
         const json = await res.json()
+        if (!Array.isArray(json)) {
+            throw new Error(`Invalid response. Refresh failed: ${json}`)
+        }
+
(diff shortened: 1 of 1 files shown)


juslesan (approved):

packages/broker/src/plugins/storage/StorageConfig.ts
```diff
@@ -97,7 +97,10 @@ export class StorageConfig {
 
     async refresh(): Promise<void> {
         const res = await fetch(`${this.apiUrl}/storageNodes/${this.nodeId}/streams`)
-        const json = await res.json()
+        if (!res.ok) {
+            throw new Error(`Refresh failed: ${res.status} ${await res.text()}`)
+        }
+        const json = (await res.json()) || []
```
teogeb, line 103: Is this `[]` fallback needed? In which situation it would be applied in practice?

teogeb: Is this `[]` fallback needed? In which situation it would be applied in practice?

If such error happens, would it be better we e.g. throw an `Error` and therefore ignore the refresh(). Now we remove all storage assignments.

teogeb (dismissed):

packages/broker/src/plugins/storage/StorageConfig.ts
timoxley: Good point.

PR #208, opened 2021-09-22
Unify types for StorageNode, Tracker and SmartContract configs.

timoxley:
This is some low priority typing/refactoring.

### Problem

Currently there's multiple definitions + names for tracker & storage node records.

* `SmartContractRecord` (protocol) is identical to `TrackerRegistryItem` (broker).
* `StorageNodeInfo` (protocol) is identical to `StorageNodeRegistryItem` (broker).
* `SmartContractRecord` is confusingly named as we have multiple smart contracts.

Likely this duplication is because of the way `Protocol` is exported from `network` seems to lose type exports, so they needed to be redefined. Going forward it's possible we should just make protocol a direct dependency of broker.

### Fixes

This is a subset of some changes I've been working with on the `NET-201-brubeck-client` branch.

* `TrackerRegistryItem` and `SmartContractRecord` are now `TrackerRecord` from protocol.
* `StorageNodeRegistryItem` and `StorageNodeInfo` are now `StorageNodeRecord` from protocol.
* Removed the `T extends TrackerInfo` generic parameter from `TrackerRegistry` as it was unused (was always `TrackerInfo` or equivalent). 
* Added a new file `network/src/streamr-client-protocol.ts` that allows you to import the protocol via the network and use it as if it were just the protocol module. e.g.
```typescript
import { StorageNodeRecord } from 'streamr-network/dist/streamr-client-protocol'
```
* Added `getTrackerRecord()` method to the `Tracker` class. This gives the correct shape for consumption of a `tracker` instance in `NetworkNode` `trackers` config. e.g.
```typescript
publisherNode = createNetworkNode({
    id: 'publisherNode',
    trackers: [tracker.getTrackerRecord()]
})
```
* `{ contractAddress: string, jsonRpcProvider: string | ConnectionInfo }` is now `SmartContractConfig` from protocol. This also addresses some inconsistency with the typing of `jsonRpcProvider` as `string | ConnectionInfo` vs just `string`. 
* Also updated the broker config JSON schema to be consistent with this change using the typescript definition of `ConnectionInfo` (via `typescript-json-schema`).

### DIFF
--- a/packages/broker/src/config.ts
+++ b/packages/broker/src/config.ts
@@ -1,13 +1,4 @@
-export interface NetworkSmartContract {
-    contractAddress: string
-    jsonRpcProvider: string
-}
-
-export interface TrackerRegistryItem {
-    id: string
-    ws: string
-    http: string
-}
+import { SmartContractConfig, TrackerRecord, StorageNodeRecord } from 'streamr-network/dist/streamr-client-protocol'
 
 export interface TurnConfig {
     url: string,
@@ -17,7 +8,7 @@ export interface TurnConfig {
 
 export interface NetworkConfig {
     name: string,
-    trackers: TrackerRegistryItem[] | NetworkSmartContract,
+    trackers: TrackerRecord[] | SmartContractConfig,
     stun: string | null,
     turn: TurnConfig | null
     location: {
@@ -29,18 +20,13 @@ export interface NetworkConfig {
 }
 
 export interface HttpServerConfig {
-    port: number, 
+    port: number,
     privateKeyFileName: string | null,
     certFileName: string | null
 }
 
-export interface StorageNodeRegistryItem {
-    address: string
-    url: string
-}
-
 export interface StorageNodeConfig {
-    registry: StorageNodeRegistryItem[] | NetworkSmartContract
+    registry: StorageNodeRecord[] | SmartContractConfig
 }
 
 export interface Config {
--- a/packages/broker/src/helpers/config.schema.json
+++ b/packages/broker/src/helpers/config.schema.json
@@ -226,7 +226,7 @@
               "type": "array",
               "description": "List of storageNode HTTP URLs to connect to",
               "items": {
-                "$ref": "#/definitions/storageNodeRegistryItem"
+                "$ref": "#/definitions/storageNodeRecord"
               }
             },
             {
@@ -264,7 +264,7 @@
       "minimum": 0,
       "maximum": 65353
     },
-    "storageNodeRegistryItem": {
+    "storageNodeRecord": {
       "type": "object",
       "required": [
         "address",
@@ -277,7 +277,7 @@
         },
         "url": {
           "type": "string"
-        }  
+        }
       }
     },
     "smartContractConfig": {
--- a/packages/broker/src/helpers/config.schema.json
+++ b/packages/broker/src/helpers/config.schema.json
@@ -1,304 +1,350 @@
 {
-  "$id": "config.schema.json",
-  "$schema": "http://json-schema.org/draft-07/schema#",
-  "description": "Broker configuration format",
-  "type": "object",
-  "required": [
-    "plugins",
-    "ethereumPrivateKey",
-    "network"
-  ],
-  "additionalProperties": false,
-  "properties": {
-    "plugins": {
-      "type": "object",
-      "description": "Pluging configurations",
-      "additionalProperties": true
+    "$id": "config.schema.json",
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "description": "Broker configuration format",
+    "type": "object",
+    "required": [
+        "plugins",
+        "ethereumPrivateKey",
+        "network"
+    ],
+    "additionalProperties": false,
+    "properties": {
+        "plugins": {
+            "type": "object",
+            "description": "Pluging configurations",
+            "additionalProperties": true
+        },
+        "httpServer": {
+            "type": [
+                "object"
+            ],
+            "description": "HTTP server configuration",
+            "default": {},
+            "additionalProperties": false,
+            "properties": {
+                "port": {
+                    "$ref": "#/definitions/port",
+                    "description": "Port to start HTTP server on",
+                    "default": 7171
+                },
+                "certFileName": {
+                    "type": [
+                        "string",
+                        "null"
+                    ],
+                    "description": "Path of certificate file to use for SSL",
+                    "default": null
+                },
+                "privateKeyFileName": {
+                    "type": [
+                        "string",
+                        "null"
+                    ],
+                    "description": "Path of private key file to use for SSL",
+                    "default": null
+                }
+            }
+        },
+        "ethereumPrivateKey": {
+            "type": "string",
+            "description": "Ethereum private key to establish broker identity",
+            "pattern": "^(0x)?[a-f0-9]{64}$"
+        },
+        "generateSessionId": {
+            "type": "boolean",
+            "description": "Whether to generate a session id that gets attached to node id (or not). Disable for consistent node id.",
+            "default": true
+        },
+        "network": {
+            "type": "object",
+            "description": "Network node settings",
+            "required": [
+                "name",
+                "location"
+            ],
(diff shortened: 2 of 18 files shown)


packages/network/src/identifiers.ts
```diff
@@ -1,7 +1,9 @@
-import { TrackerLayer, SmartContractRecord } from 'streamr-client-protocol'
+import { TrackerLayer } from 'streamr-client-protocol'
 import { NodeId } from './logic/Node'
 import { RtcSubTypes } from './logic/RtcMessage'
 
+export { TrackerRecord } from 'streamr-client-protocol'
```
juslesan, line 5: Would it make more sense to import TrackerRecord directly from streamr-client-protocol to modules with these changes?

packages/network/src/streamr-client-protocol.ts
```diff
@@ -0,0 +1,3 @@
+import Protocol from 'streamr-client-protocol'
+export * from 'streamr-client-protocol'
+export default Protocol
```
juslesan, line 3: I feel like there is a need to get rid of the instanceof checking for messages in the protocol to avoid having to strictly use the network's protocol instance in the broker and elsewhere.

juslesan (approved):

packages/broker/src/helpers/config.schema.json
```diff
@@ -1,304 +1,350 @@
 {
```
harbu, line 1: Why these changes? Are these just indentation?

packages/network/src/identifiers.ts
harbu: Yeah re-exporting seems like needless indirection? Just import from protocol directly where needed?

packages/network/src/streamr-client-protocol.ts
```diff
@@ -0,0 +1,3 @@
+import Protocol from 'streamr-client-protocol'
```
harbu, line 1: Maybe the contents of this file should live in `composition.ts` so instead of
```ts
import { ... } from 'streamr-network/dist/streamr-client-protocol'
```

We can just do 
```ts
import { ... } from 'streamr-network'
```

(Which I find better because the latter is relying on purposefully exported things instead of referring to an arbitrary file within the project.)

We could also consider just adding `streamr-client-protocol` as a dependency of broker and using stuff from there directly, which would be arguably less confusing. Why we haven't done so before? Becuase of concerns of potential version mismatches between network and broker which could lead to weird behaviour.

harbu (commented):
Good stuff; see my comments on import

packages/broker/src/config.ts
```diff
@@ -1,13 +1,4 @@
...
-    contractAddress: string
-    jsonRpcProvider: string
-}
-
-export interface TrackerRegistryItem {
-    id: string
-    ws: string
-    http: string
-}
+import { SmartContractConfig, TrackerRecord, StorageNodeRecord } from 'streamr-network/dist/streamr-client-protocol'
```
teogeb, line 1: Maybe we could avoid importing the internals of `streams-network`? E.g. 
```
import { Protocol } from 'streamr-network'
trackers: Protocol.TrackerRecord[] | Protocol.SmartContractConfig,
```

packages/protocol/src/utils/TrackerRegistry.ts
```diff
@@ -37,40 +31,27 @@ export class TrackerRegistry<T extends TrackerInfo> {
...
-
-    const contract = new Contract(contractAddress, trackerRegistryConfig.abi, provider)
-    // check that contract is connected
-    await contract.addressPromise
-
+async function fetchTrackers(config: SmartContractConfig) {
+    const contract = await initContract(config, trackerRegistryConfig.abi)
     if (typeof contract.getNodes !== 'function') {
-        throw Error(`getNodes function is not defined in smart contract (${contractAddress})`)
+        throw Error(`getNodes function is not defined in smart contract (${config.contractAddress})`)
```
teogeb, line 42: It seems that the checking of available functions is a typical use case. Maybe `initContract` could implement this checking if we'd add a third parameter which lists all required functions?

packages/protocol/src/utils/StorageNodeRegistry.ts
```diff
@@ -1,20 +1,16 @@
...
+import { initContract, SmartContractConfig } from './SmartContractUtil'
 
 // storageNodeAddress => HTTP address
-export type StorageNodeInfo = {
+export type StorageNodeRecord = {
     address: string
     url: string
 }
 
 export class StorageNodeRegistry {
```
teogeb, line 10: Maybe we could move this class to `Broker` package as storage functionality is nowadays considered to be an implementation detail of a broker node, and not protocol level logic.

(At some point we should update the https://github.com/streamr-dev/streamr-specs/blob/master/PROTOCOL.md#storagenodesrequest spec and officially deprecate `StorageNodesRequest` and `StorageNodesRequest`)

There is already a class `network-monorepo/packages/broker/src/StorageNodeRegistry.ts`. This functionality could be merged to that file.

packages/network/src/logic/Tracker.ts
```diff
@@ -254,4 +254,12 @@ export class Tracker extends EventEmitter {
     getOverlayPerStream(): Readonly<OverlayPerStream> {
         return this.overlayPerStream
     }
+
+    getTrackerRecord(): TrackerRecord {
+        return {
+            id: this.peerInfo.peerId,
+            http: this.getUrl().replace(/^ws/, 'http'),
```
teogeb, line 261: If the tracker doesn't have HTTP endpoints, this field should be `undefined`? See https://github.com/streamr-dev/network-monorepo/blob/06137024a0e56bd24e601fe27ac6811d0ffe3dc6/packages/network/src/composition.ts#L90

Currently the `TrackerRecord` type definition doesn't allow `undefined` field value. Does the smart contract allow it?

packages/protocol/src/utils/StorageNodeRegistry.ts
samt1803: In the decentralised version I am working on, the StorageNodeRegistry is residing (like the StreamRegistry) in the client and the one in the broker is being replaced by just calling the client. So this StorageNodeRecord is needed in the client for CRUD operations on storagenodes or adding streams to storagenodes. So it could be moved to the client, and then the broker uses the definition from the client instead of the protocol..

samt1803 (approved):

packages/broker/src/helpers/config.schema.json
timoxley: Ah must have autoformatted the whitespace. It's now 4 space indentation like rest of code.

packages/protocol/src/utils/TrackerRegistry.ts
timoxley: Good thought, but let's add it if/when there's more than just these two.

packages/network/src/logic/Tracker.ts
timoxley: > if the tracker doesn't have HTTP endpoints

Good point

packages/network/src/streamr-client-protocol.ts
timoxley: > concerns of potential version mismatches between network and broker which could lead to weird behaviour

Yeah I figured there might be some issues with module-level/class static state or `instanceof` and didn't want to audit every file so just left that part as-is.

timoxley: Though I guess there may be similar issues with client's internal protocol vs broker + network protocol, either versioning or simply a different instance.

timoxley: related https://github.com/streamr-dev/network-monorepo/pull/208#discussion_r714559277

packages/network/src/logic/Tracker.ts
timoxley: > Does the smart contract allow it?

Supports anything, the data is a json blob as a string.

harbu:
Status of this, will be resolving conflicts or close PR? @timoxley

PR #209, opened 2021-09-24
fix: attempts to fix it with AutononceWallet and with NonceManager

samt1803:


samt1803:
this is an attempt to fix the "nonce is too low" or "there already is a transaction with that nonce in the mempool"  errors when trying to run tests, that use the same privatekey in parallell

both approaches did not work yet..

PR #210, opened 2021-09-27
NET-419: Update node-datachannel to v0.1.10

paullouisageneau:
This PR fixes "Aborted" errors when running the network CI by updating node-datachannel to v0.1.10. The errors were caused by a possible race condition when closing a Peer Connection early in its lifecycle, which should be fixed in libdatachannel v0.15.2.

The fix should apply to different kind of failures linked to transport initialization, like termination without exception and segfaults, which often followed `[rtc::impl::PeerConnection::initDtlsTransport@251] Connection is closed` errors in the log.

### DIFF
--- a/packages/network/package-lock.json
+++ b/packages/network/package-lock.json
@@ -9344,9 +9344,9 @@
 			"dev": true
 		},
 		"node-datachannel": {
-			"version": "0.1.5",
-			"resolved": "https://registry.npmjs.org/node-datachannel/-/node-datachannel-0.1.5.tgz",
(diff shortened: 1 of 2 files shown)


juslesan (approved):
LGTM, I gave the CI a few extra spins and it no longer crashed!

harbu (approved):

paullouisageneau:
Are some cross-client tests expected to fail?

juslesan:
> Are some cross-client tests expected to fail?

Yes, those are flaky and are not strictly related to the Network package

timoxley:
> Are some cross-client tests expected to fail?

Worth restarting them, if they fail consistently then there's a problem.

paullouisageneau:
It seems the `stream-encrypted-exchanged-rotating-revoking-signed` and `default stream-encrypted-exchanged-rotating-signed ` ones fail most of the time.

paullouisageneau:
There is always one or two randomly failing but I'm merging anyway as it looks unrelated.

PR #211, opened 2021-09-27
network [NET-264]: rework node propagation logic

harbu:
## Motivation
While revisiting NET-264 #16 it came to our attention that lack of buffering on the Node (logic layer) does cause some real issues. This came up while studying an integration test related to the storage node assignment publish / subscribe flow. Basically there is no way of a user of `streamr-client` to ensure that neighbors are present when calling only `publish` (and not explicitly subscribing). In response, we came up with an idea of a more purposeful "cache" or message propagation logic (and the ability to turn off "caching" effectively if desired later).

It can also be considered good service that node attempts to propagate a node to at least a certain number of nodes before "giving up" albeit it does involve more code (and complexity).

## Idea
We've had message buffering before but here I extracted and encapsulated the logic into its own thing, attempting to make it more clear how the propagation logic works.

The core idea is that a node will attempt to propagate a message within a certain `ttl` to a total of  `minPropagationTargets` neighbors. Only upon the `ttl` expiration or `minPropagationTargets` being send to (or buffer becoming full), is a node allowed to "give up" on propagating a message.

## Implementation
There are three new classes from top to bottom by usage:
- Class `Propagation` is responsible for message propagation.
- Class `PropagationTaskStore` keeps track of _active_ propagation tasks and allows querying a set of tasks by stream for example. Class `Propagation` uses this.
- Class `FifoMapWithTtl` is a bounded map with FIFO replacement policy and max size. Entries also have a TTL after which they are considered stale and not returned by queries.


teogeb (approved):
Good improvement 👍

PR #212, opened 2021-09-27
NET-519: move tracker and node classes to respective folders

harbu:
Node stuff under `/src/logic/node` and tracker stuff under `/src/logic/tracker`


juslesan (approved):
👍

PR #223, opened 2021-09-29
network: remove support for full status messages

teogeb:
Currently all status messages that nodes send include only one stream. In https://github.com/streamr-dev/network-monorepo/pull/174 we modified the sending logic in node, but didn't change the tracker behavior (to keep the backwards compatibility).

In this PR we drop the backwards compatibility. The trackers is no longer able to process status messages that contain multiple streams.


juslesan (approved):
LGTM, ready to merge after eslint fix

PR #228, opened 2021-09-30
broker [NET-530]: Increase client cache to 6hrs

harbu:


timoxley (approved):
lgtm

harbu:
I'm assuming this is no longer needed with built-in data

PR #233, opened 2021-10-06
fix(broker): handle generateSessionId=false

harbu:
Noticed that sessionIds are still getting appended to testnet miner nodes. Added check for `!generateSessionId` in broker logic. (It is true by default but testnet miners override to false)

### DIFF
--- a/packages/broker/src/broker.ts
+++ b/packages/broker/src/broker.ts
@@ -84,12 +84,7 @@ export const createBroker = async (config: Config): Promise<Broker> => {
     const storageNodes = await getStorageNodes(config)
     const storageNodeRegistry = StorageNodeRegistry.createInstance(config, storageNodes)
 
-    // leave undefined to let client generate session id
-    let sessionId: string | undefined
-    if (config.plugins['storage']) { // storage node needs consistent id
-        sessionId = brokerAddress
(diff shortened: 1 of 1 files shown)


teogeb:
There is a risk that the provided `network.id `is not a public key of the given private key. How about if we just add a new config option of `sessionId: true/false`? The client would generate nodeId from the private key: if sessionId=false it wouldn't append anything, and if true it would append `'#' + uuid()`

packages/broker/src/broker.ts
```diff
@@ -84,20 +84,15 @@ export const createBroker = async (config: Config): Promise<Broker> => {
     const storageNodes = await getStorageNodes(config)
     const storageNodeRegistry = StorageNodeRegistry.createInstance(config, storageNodes)
 
-    // leave undefined to let client generate session id
-    let sessionId: string | undefined
-    if (config.plugins['storage']) { // storage node needs consistent id
-        sessionId = brokerAddress
-    }
-
+    const usePredeterminedNetworkId = !config.generateSessionId || config.plugins['storage']
```
timoxley, line 87: Better

timoxley (commented):
Looks good 👍

PR #235, opened 2021-10-06
Fairer isOffering logic

juslesan:
- Concatenate nodeIds and and compare their hashes when selecting the offering peer for WebRTC
- This reduces the likelyhood of nodes always being an offerer or non-offerer if their nodeIds are too "high" or "low"

### DIFF
--- a/packages/network/src/connection/WebRtcConnection.ts
+++ b/packages/network/src/connection/WebRtcConnection.ts
@@ -51,7 +51,7 @@ export function isOffering(myId: PeerId, theirId: PeerId): boolean {
 
 function offeringHash(idPair: string): number {
     const buffer = crypto.createHash('md5').update(idPair).digest()
-    return buffer.readInt32LE(0)
+    return buffer.readInt32LE()
 }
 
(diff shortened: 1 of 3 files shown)


packages/network/src/connection/WebRtcConnection.ts
```diff
@@ -45,7 +46,12 @@ interface Events {
 export const ConnectionEmitter = EventEmitter as { new(): StrictEventEmitter<EventEmitter, Events> }
 
 export function isOffering(myId: PeerId, theirId: PeerId): boolean {
-    return myId < theirId
+    return offeringHash(myId + theirId) < offeringHash(theirId + myId)
+}
+
+function offeringHash(idPair: string): number {
+    const buffer = crypto.createHash('md5').update(idPair).digest()
+    return buffer.readInt32LE(0)
```
teogeb: The `0` parameter is not needed, as it is the default. The logic is simpler if we remove the parameter.

teogeb (approved):

PR #241, opened 2021-10-08
broker: function return types

teogeb:
Now all functions have return types.  Removed eslint exception `"@typescript-eslint/explicit-module-boundary-types": "warn"` so that eslint will show an error if we miss a return type in the future.

Changed Todo types to actual types in `Stream`, `StreamStateManager` and `DeleteExpiredCmd`. `Stream` and `StreamStateManager` are now generic classes (the connection type is a generic type).

### DIFF
--- a/packages/broker/src/RequestAuthenticatorMiddleware.ts
+++ b/packages/broker/src/RequestAuthenticatorMiddleware.ts
@@ -8,6 +8,7 @@ const logger = new Logger(module)
 /**
  * Middleware used to authenticate REST API requests
  */
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
 export const authenticator = (streamFetcher: StreamFetcher, permission = 'stream_subscribe') => (req: Todo, res: Todo, next: Todo) => {
     let sessionToken
 
--- a/packages/broker/src/RequestAuthenticatorMiddleware.ts
+++ b/packages/broker/src/RequestAuthenticatorMiddleware.ts
@@ -1,4 +1,5 @@
 import { Todo } from './types'
+import { Request, Response, NextFunction } from 'express'
 import { HttpError } from './errors/HttpError'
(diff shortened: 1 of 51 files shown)


packages/broker/src/RequestAuthenticatorMiddleware.ts
```diff
@@ -8,6 +8,7 @@ const logger = new Logger(module)
 /**
  * Middleware used to authenticate REST API requests
  */
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
```
harbu, line 12: Could we add express request / response types here?

packages/broker/src/helpers/fetch.ts
```diff
@@ -1,6 +1,6 @@
-import fetch, { RequestInit } from 'node-fetch'
+import fetch, { RequestInit, Response } from 'node-fetch'
 
-export const fetchOrThrow = async (url: string, init?: RequestInit) => {
+export const fetchOrThrow = async (url: string, init?: RequestInit): Promise<Response|never> => {
```
harbu:
```suggestion
export const fetchOrThrow = async (url: string, init?: RequestInit): Promise<Response> => {
```

packages/broker/src/plugins/legacyWebsocket/WebsocketServer.ts
```diff
@@ -248,7 +248,7 @@ export class WebsocketServer extends EventEmitter {
         this.connections.delete(connection)
 
         // Unsubscribe from all streams
-        connection.forEachStream((stream: Stream) => {
+        connection.forEachStream((stream: Stream<any>) => {
```
harbu: Is it possible to have a tighter generic type here, `Stream<Connection>`?

harbu (approved):

PR #242, opened 2021-10-08
protocol: function return types

teogeb:
Added missing return types. Also disabled/fixed `max-len` warnings in some test files.

### DIFF
--- a/packages/protocol/src/protocol/control_layer/ControlMessage.ts
+++ b/packages/protocol/src/protocol/control_layer/ControlMessage.ts
@@ -55,7 +55,7 @@ export default class ControlMessage {
         this.requestId = requestId
     }
 
-    static registerSerializer(version: number, type: ControlMessageType, serializer: Serializer<ControlMessage>) {
+    static registerSerializer(version: number, type: ControlMessageType, serializer: Serializer<ControlMessage>): void {
         // Check the serializer interface
         if (!serializer.fromArray) {
             throw new Error(`Serializer ${JSON.stringify(serializer)} doesn't implement a method fromArray!`)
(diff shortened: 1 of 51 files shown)


packages/protocol/src/protocol/control_layer/ControlMessage.ts
```diff
@@ -91,18 +91,19 @@ export default class ControlMessage {
...
-    serialize(version = this.version, ...typeSpecificSerializeArgs: any[]) {
+    serialize(version = this.version, ...typeSpecificSerializeArgs: any[]): string {
         return JSON.stringify(ControlMessage.getSerializer(version, this.type).toArray(this, ...typeSpecificSerializeArgs))
     }
 
     /**
      * Takes a serialized representation (array or string) of a message, and returns a ControlMessage instance.
      */
-    static deserialize(msg: any, ...typeSpecificDeserializeArgs: any[]) {
+    // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
```
timoxley, line 105: curious why

timoxley (approved):
lgtm

packages/protocol/src/protocol/control_layer/ControlMessage.ts
teogeb: The `allowArgumentsExplicitlyTypedAsAny=false` option of `@typescript-eslint/explicit-module-boundary-types` doesn't allow `any` as a parameter type for an exported method:
https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md#allowargumentsexplicitlytypedasany

We don't specify the value for `allowArgumentsExplicitlyTypedAsAny` in the base config. The default is `false`. Could consider to set the option to `true` there:
https://github.com/streamr-dev/eslint-config-streamr/blob/master/ts/.eslintrc.js
https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md#options

PR #244, opened 2021-10-11
Brubeck Client for Browser

timoxley:
Adds support for browser builds of client.

* Added some additional steps + verification to browser tests. Now uses two clients with different keys and adds grant permissions + assign storage steps.
* Fixed CORS requests for requests coming from client to storage node.
* Removed rootDirs from typescript config, was confusing and used inconsistently anyway.
Now you have to `import … from ../../src/file` instead of `import … from ../../file`
* *Made protocol a direct dependency of broker.*

#### Improved bundling process and reduced size

* Split Logger, Tracker & Network Node so browser build can import node without any tracker dependencies e.g. express
* Replaced ethers & bip39 (mnemonic) packages in protocol with equivalent individual `ethersproject/*` packages. This reduces bundle size.
* Removed double bundling of core-js, just use usage plugin from `@babel/preset-env (See https://github.com/zloirock/core-js/issues/848#issuecomment-717208058)
* Updated dev dependencies & unified dependency versions across packages.

#### Improved bootstrap time: 1m30s to 12s 
* Removed unnecessary build steps e.g. clean before build
* Reusable tsc builds using `tsc --build` mode. https://www.typescriptlang.org/docs/handbook/project-references.html
* Added caching to eslint

#### Also Note
* ~Some diff noise due to reverting some merges from main~ Reverted those reverts.

### New Updates
* Integrated npm workspaces https://github.com/streamr-dev/network-monorepo/pull/259
* Fixed all eslint warnings.
* Fixed network browser/karma tests.
* Removed unused client dependencies.
* Updated client dependencies.
* Added a fix for resend flakiness.
* Prevented dist, build, node_modules, etc from being copied into docker image, it installs rebuilds things anyway.
* Set up client integration tests & cross client tests to run against monorepo brokers/trackers.
* Configured husky to run eslint before pushing
* Enabled client export tests in CI.
* Removed lerna usage everywhere but the show-versions script.


teogeb:
Good improvements 👍 

> #### Also Note
> * Some diff noise due to reverting some merges from main.

If this PR shouldn't revert the changes when we merge it to  main, should we exclude those reverts in this PR (that is, revert the reverts)?

timoxley:
> If this PR shouldn't revert the changes when we merge it to main, should we exclude those reverts in this PR (that is, revert the reverts)?

Yep, we should revert the reverts, though I'll try just rebasing the revert commits out. Hopefully that won't generate any hairy conflicts.

.github/workflows/broker.yml
```diff
@@ -26,21 +26,19 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [ 14.x ]
+        node-version: [16.x]
```
teogeb, line 29: Many tests are now running on Node 16 instead of Node 14. Would it make drop the Node 14 support for the remaining test matrixes, too (benchmark tests in `client-build`, flakey and memory tests in `client-code`, lint and unit tests in `protocol`, and `test-utils`)

packages/broker/src/plugins/legacyWebsocket/RequestHandler.ts
```diff
@@ -65,27 +71,28 @@ export class RequestHandler {
...
                 return this.publish(connection, request as PublishRequest)
-            case ControlLayer.ControlMessage.TYPES.ResendLastRequest:
-            case ControlLayer.ControlMessage.TYPES.ResendFromRequest:
-            case ControlLayer.ControlMessage.TYPES.ResendRangeRequest:
+            case ControlMessage.TYPES.ResendLastRequest:
+            case ControlMessage.TYPES.ResendFromRequest:
+            case ControlMessage.TYPES.ResendRangeRequest:
                 return this.resend(connection, request as (ResendLastRequest|ResendFromRequest|ResendRangeRequest))
-            default:
+            default: {
```
teogeb, line 87: Is this `{ }` wrapper needed?

packages/client/src/utils/Context.ts
```diff
@@ -12,7 +12,7 @@ export class ContextError extends Error {
     code?: string
     constructor(context: Context, message: string = '', ...args: any[]) {
         // @ts-expect-error inspectOpts not in debug types
-        super(`${context.id}: ${formatWithOptions({ ...context.debug.inspectOpts, colors: false }, message, ...args)}`)
+        super(`${context?.id}: ${formatWithOptions({ ...(context?.debug?.inspectOpts || {}), colors: false }, message, ...args)}`)
```
teogeb, line 15: Should the `context` parameter be optional if `?` operation is needed?

packages/client/src/utils/ConvertBrowserStream.ts
```diff
@@ -0,0 +1,36 @@
+import { PassThrough, Readable } from 'stream'
+
+export function ConvertBrowserStream(browserStream: ReadableStream | Readable) {
```
teogeb: The name could describe what it is conversion target? Now bit unclear what kind of transformation is done. (Also as it is just a function, the name could be lowercased)

packages/client/src/utils/log.ts
```diff
@@ -41,6 +41,14 @@ if (typeof window === 'undefined') {
     })
 }
 
+// e.g. in browser
```
teogeb: Could this fallback handling be part of the `export function formatWithOptions` implementation at line 73 so that we don't need to modify an external object? It a bit surprising side-effect of the `log.ts` import.

packages/client/src/BrubeckNode.ts
```diff
@@ -32,7 +30,7 @@ export default class BrubeckNode implements Context {
         @inject(Config.Network) options: NetworkNodeOptions
     ) {
         this.options = options
-        this.id = instanceId(this, uid)
+        this.id = instanceId(this)
```
teogeb, line 33: Currently there is a `id` parameter in options, and it if the value is not set, we generate a fallback value. Maybe we could use that same id in here so that there's only one ID to refer to a BrubeckNode? 

(Maybe a separate PR, if we want to implement something like that)

packages/client/src/Ethereum.ts
```diff
@@ -66,7 +67,7 @@ export abstract class EthereumConfig {
 }
 
 @scoped(Lifecycle.ContainerScoped)
-export default class StreamrEthereum {
+class StreamrEthereum {
```
teogeb, line 70: Is this change of export style needed for browser compatibility? If so, why only this class was changed?

packages/client/src/ResendSubscribe.ts
```diff
@@ -57,7 +57,10 @@ export class ResendSubscription<T> extends Subscription<T> {
         try {
             yield* await this.getResent()
         } catch (err) {
-            await this.handleError(err)
+            if (err.code !== 'NO_STORAGE_NODES') {
```
teogeb, line 60: Maybe worth mentioning this change of functionality in the PR description?

packages/client/test/integration/GroupKeyPersistence.test.ts
```diff
@@ -145,10 +145,10 @@ describeRepeats('Group Key Persistence', () => {
             })
 
             published.push(...await publishTestMessages(3))
-            const received2 = await sub2.collect(3)
+            const received2 = await sub2.collect(4)
```
teogeb, line 148: Could be worth commenting why we need to collect 4 messages instead of 3. Is the reason is that the internal group key response will be visible for the subscribers? Should it be?

packages/network/src/helpers/logger/LoggerBrowser.ts
```diff
@@ -0,0 +1,39 @@
...
+            // which can fail when under jest+typescript, due to some CJS/ESM
+            // incompatibility leading to throwing an error like:
+            // "prettyFactory is not a function"
+            prettifier: undefined,
+            prettyPrint: false,
+        }
+        super(options, destinationStream)
+    }
+
+    private static createName(module: NodeJS.Module, context?: string) {
```
teogeb: The `createName` functions in `LoggerBrowser` and `LoggerNode` are almost identical. Would it be possible to move the method e.g. to `LoggerCommon`?

packages/broker/src/plugins/legacyWebsocket/RequestHandler.ts
timoxley: Not required in this case, but putting a block around multi-line switch cases is a good habit as it's low cost but prevents unexpected weirdness triggered by putting declarations in the case body: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#block-scope_variables_within_switch_statements

Likely TS would pick up any issues, so this habit is probably legacy JS, but I prefer indenting with the brackets there anyway.

packages/client/src/ResendSubscribe.ts
timoxley: This fixes a v6 regression.

packages/client/src/BrubeckNode.ts
timoxley: That's the node id, which is a bit different. 

This `id` is really internal and mostly there for logs & debugging. It follows the pattern used in the rest of the client codebase. It allows differentiating between instances in logs e.g.
```javascript
const thing1 = new Thing(opts)
const thing2 = new Thing(opts)
thing1.debug('test') // Thing:1 test
thing2.debug('test') // Thing:2 test
```
and also logging "namespaces" so you can see who generated an instance:
```javascript
const child1 = thing1.createChild()
const child2 = thing2.createChild()
child1.debug('test') // Thing:1 Child: 1 test
child2.debug('test') // Thing:2 Child: 2 test
```

packages/client/src/utils/Context.ts
timoxley: it's not actually optional, but at runtime it could possibly be undefined.

This is often invoked like: 

```tyoescript
class Task implements Context {
  id = instanceId(this)
  debug = debug(this.id)
  run() {
    // …
    throw new ContextError(this, 'an error happened during run')  
  }
}
```

which ideally would spit out errors something like: `"Error: Task:1 an error happened during run"`

but if this is invoked in a manner like so:

```typescript
const task = new Task()
setTimeout(task.run)
```

then at runtime, the useful part of the error will replaced with `Uncaught TypeError: Cannot read properties of undefined (reading 'id')`. 

With the above change errors will instead look like: `"Error: undefined: an error happened during run"`. Ideally this doesn't happen at all, but if it does then the least we can do is to ensure the original error message/stack isn't thrown away or replaced.

**Key points**: Never throw away error information. Ensure errors can't throw errors.

Related to this, it's possible we should enable this typescript-eslint rule in order to help catch cases like `setTimeout(task.run)` at compile time: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/unbound-method.md

packages/client/src/utils/ConvertBrowserStream.ts
timoxley: Will do.

packages/client/src/utils/log.ts
timoxley: Good point. Changed.

packages/client/src/Ethereum.ts
timoxley: IIRC I think there was some circular dependency hell and this resolved it.

packages/client/test/integration/GroupKeyPersistence.test.ts
timoxley: I'd add a comment if I knew why 🤷  it changed at some point. Just matching network/broker behaviour.

packages/network/src/helpers/logger/LoggerBrowser.ts
timoxley: Done in [dd64b882](https://github.com/streamr-dev/network-monorepo/pull/244/commits/0a7622c6dde8527acada1980264611f9febfae41)

.github/workflows/broker.yml
timoxley: Yeah not a bad idea but perhaps in another PR. Should probably switch to 16 + 17 instead of 14 + 16

packages/client/src/utils/ConvertBrowserStream.ts
timoxley: Done

packages/client/test/integration/GroupKeyPersistence.test.ts
teogeb: But is the current behavior a bug? 

If I understood correctly, the use case we are testing is:
- subscribe with `subscriber1` 
- publish 1 test message
- expect `subscriber1` to receive the test message
- subscribe with `subscriber2`
- publish 3 more test messages
- expect `subscriber2` to receive 4 messages (`expect(received2).toEqual(published)`)

As the first message was published before `subscriber2` subscribed, should it receive the first test message (published at line 130). I'd assume it should it just receive 3 messages which were published after it subscriber.

Maybe the test would be more readable if we didn't alter the `const published` variable, but instead have variables like `published1` and `published2`.

packages/client/src/utils/ConvertBrowserStream.ts
teogeb: Is there a reason why this doesn't follow the usual naming convention that the first letter of a function name is lower-cased?

packages/client/src/utils/WebStreamToNodeStream.ts
```diff
@@ -0,0 +1,71 @@
...
+    const reader = fromBrowserStream.getReader()
+    /* eslint-disable no-constant-condition, no-await-in-loop */
+    const ac = new AbortController()
+    const cleanup = () => {
+        ac.abort()
+    }
+    reader.closed.finally(() => { // eslint-disable-line promise/catch-or-return
+        toNodeStream.off('close', cleanup)
+    })
+    // toNodeStream.once('error', cleanup)
```
teogeb, line 34: Remove if not needed?

packages/client/src/Publisher.ts
```diff
@@ -147,6 +147,9 @@ export default class BrubeckPublisher implements Context, Stoppable {
         count?: number
         messageMatchFn?: (msgTarget: StreamMessage, msgGot: StreamMessage) => boolean
     } = {}) {
+        if (!streamMessage || !streamMessage.spid) {
```
teogeb, line 150: As `streamMessage: StreamMessage` is not an optional parameter, it should be caller's responsibility to ensure that the method is called with a valid value. The method should be able to rely on the contract it defined in the parameter list definition.

PR #246, opened 2021-10-12
ci(broker): rm extranous test-sequential run

harbu:
Turns out `npm run test-integration` already runs sequential tests, no need to run again.


timoxley (approved):
👍

PR #247, opened 2021-10-13
network: refactor inbound/outboundNodes to neighbors

teogeb:
Simplify status message content and topology handling:
- Previously each status message contained `inboundNodes` and `outboundNodes` for a stream. 
- Those two arrays had exactly same content
- Now we use one array, called `neighbors`

There is also a backwards compatibility handling `Tracker.ts`, which enables the tracker to receive the statuses in the old format.

### DIFF
--- a/packages/network/src/logic/tracker/OverlayTopology.ts
+++ b/packages/network/src/logic/tracker/OverlayTopology.ts
@@ -50,7 +50,7 @@ export class OverlayTopology {
     }
 
     update(nodeId: NodeId, neighbors: NodeId[]): void {
-        const newNeighbors = _.uniq([...neighbors])
+        const newNeighbors = _.uniq(neighbors)
             .filter((n) => n in this.nodes)
             .filter((n) => n !== nodeId) // in case nodeId is reporting itself as neighbor
 
--- a/packages/network/src/logic/tracker/OverlayTopology.ts
+++ b/packages/network/src/logic/tracker/OverlayTopology.ts
@@ -1,4 +1,5 @@
 import { NodeId } from '../node/Node'
+import _ from 'lodash'
 
 // From: https://gist.github.com/guilhermepontes/17ae0cc71fa2b13ea8c20c94c5c35dc4
 const shuffleArray = <T>(arr: Array<T>): Array<T> => arr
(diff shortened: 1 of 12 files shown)


packages/network/src/logic/tracker/OverlayTopology.ts
```diff
@@ -50,7 +50,7 @@ export class OverlayTopology {
     }
 
     update(nodeId: NodeId, neighbors: NodeId[]): void {
-        const newNeighbors = _.uniq([...neighbors])
+        const newNeighbors = _.uniq(neighbors)
```
timoxley, line 53: could also drop `lodash` entirely for `const newNeighbors = [...new Set(neighbors)]`

also to help keep browser bundle size at a minimum we'd only import the specific lodash utility needed i.e.  `import uniq from 'lodash/uniq'` or `import { throttle, debounce } from 'lodash'`, where the latter can be converted into the former by `babel-lodash-plugin`.

harbu (approved):
LGTM;

Looks like browser tests are not compatible with `jest-extended` atm?

packages/network/src/logic/tracker/Tracker.ts
```diff
@@ -180,8 +186,7 @@ export class Tracker extends EventEmitter {
         if (status.counter === COUNTER_UNSUBSCRIBE) {
             this.leaveAndCheckEmptyOverlay(status.streamKey, this.overlayPerStream[status.streamKey], node)
         } else {
-            const neighbors = new Set([...status.inboundNodes, ...status.outboundNodes])
-            this.overlayPerStream[status.streamKey].update(node, [...neighbors])
+            this.overlayPerStream[status.streamKey].update(node, status.neighbors)
```
timoxley, line 189: Yeah good call, better to have the uniqueness enforced by `update`

timoxley (approved):
Only gripe is neighbours vs neighbors:

image

packages/network/src/logic/tracker/OverlayTopology.ts
teogeb: Seems that the `babel-lodash-plugin` supports also the current `import _ from 'lodash'` style: https://github.com/lodash/babel-plugin-lodash#example

Let's add the plugin in another PR.

PR #248, opened 2021-10-13
broker: remove Todo type

harbu:
Address some technical debt by replacing `Todo` type with actual types. The `Todo` type was introduced originally as a placeholder type for tricky / unclear typing situations when we converted the broker code base from JavaScript to TypeScript.


packages/broker/src/plugins/legacyPublishHttp/DataProduceEndpoints.ts
```diff
@@ -1,21 +1,22 @@
...
 import { LEGACY_API_ROUTE_PREFIX } from '../../httpServer'
 import { parsePositiveInteger, parseTimestamp } from '../../helpers/parser'
-import { Todo } from '../../types'
 
 const logger = new Logger(module)
 
 const { StreamMessage, MessageID, MessageRef } = Protocol.MessageLayer
 const { InvalidJsonError, ValidationError } = Protocol.Errors
 
+type OptionalQueryParam = string | undefined
```
teogeb: We could avoid all query parameter casts if we introduce an interface for the parameters:
```
interface QueryParams {
    ts: string
    seq: string
    prev_ts?: string
    prev_seq?: string
    signatureType?: string
    signature?: string
    pkey?: string
    address?: string
    msgChainId?: string
}
```

It would be a generic type to `AuthenticatedRequest`, if we modify that type to be:
```
export interface AuthenticatedRequest<Q> extends Request<Record<string,any>,any,any,Q,Record<string,any>> {
    stream?: Record<string,unknown>
}
```

Then we can easily use that at `DataProduceEndpoint.ts@43` as `AuthenticatedRequest<QueryParams>`
and in `RequestAuthenticatorMiddleware.ts@16` as `AuthenticatedRequest<any>`.

harbu:
@teogeb can you try out that change you suggested?

juslesan (approved):

packages/broker/src/plugins/legacyPublishHttp/DataProduceEndpoints.ts
teogeb: @harbu pushed a commit https://github.com/streamr-dev/network-monorepo/pull/248/commits/35d39bcf4939b68210ae4d0a47d44cb8ab6995af to implement this, all tests pass

teogeb (approved):
LGTM, added a commit about `Metrics` import. Now all classes import it directly (similarly to `Publisher.ts`)

packages/broker/src/RequestAuthenticatorMiddleware.ts
```diff
@@ -1,18 +1,21 @@
-import { Todo } from './types'
 import { Request, Response, NextFunction } from 'express'
 import { HttpError } from './errors/HttpError'
 import { Logger } from 'streamr-network'
 import { StreamFetcher } from './StreamFetcher'
 
 const logger = new Logger(module)
 
+export interface AuthenticatedRequest<Q> extends Request<Record<string,any>,any,any,Q,Record<string,any>> {
```
timoxley, line 8: you could set a default for `Q` i.e. `Q = any` so you can use `AuthenticatedRequest` instead of `AuthenticatedRequest<any>`

timoxley: This type looks like a crime scene

packages/broker/src/errors/FailedToPublishError.ts
```diff
@@ -1,7 +1,5 @@
-import { Todo } from '../types'
-
 export class FailedToPublishError extends Error {
```
timoxley, line 1: This should probably go away and all the validation go into the client. I see there's some special handling for `FailedToPublishError` instances in broker, it should probably actually check `FailedToPublishError` from the client.

packages/broker/src/plugins/legacyMqtt/Connection.ts
```diff
@@ -120,7 +135,7 @@ export class Connection extends events.EventEmitter {
     }
 
     streamsAsString(): any {
```
timoxley: this ` :any` should probably be `string[]`

packages/broker/src/plugins/legacyMqtt/MqttServer.ts
```diff
@@ -174,13 +162,16 @@ export class MqttServer extends EventEmitter {
...
+    async handlePublishRequest(connection: Connection, packet: mqtt.IPublishPacket): Promise<void> {
         logger.debug('publish request %o', packet)
 
         const { topic, payload, qos } = packet
 
         try {
-            const streamObj = await this.streamFetcher.authenticate(topic, connection.token, 'stream_publish')
+            const streamObj = await this.streamFetcher.authenticate(topic, connection.token, 'stream_publish') as {
+                id: string,
+                partitions: number
```
timoxley, line 173: Maybe it should define this in `streamFetcher.authenticate`?

packages/broker/src/plugins/legacyMqtt/MqttServer.ts
```diff
@@ -262,7 +257,7 @@ export class MqttServer extends EventEmitter {
         logger.debug('closing client "%s" on streams "%o"', connection.id, connection.streamsAsString())
 
         // Unsubscribe from all streams
-        connection.forEachStream((stream: Todo) => {
+        connection.forEachStream((stream) => {
             const object = {
```
timoxley: would `const object: Partial<IPubackPacket> = {` work?

packages/broker/src/plugins/legacyPublishHttp/DataProduceEndpoints.ts
```diff
@@ -1,21 +1,32 @@
...
+    ts?: string
+    seq?: string
+    prev_ts?: string
+    prev_seq?: string
+    signatureType?: string
+    signature?: string
+    pkey?: string
+    address?: string
+    msgChainId?: string
+}
```
timoxley, line 28: Maybe
```typescript
type QueryParams = Partial<{
    ts: string
    seq: string
    …
}>
```

packages/broker/src/plugins/legacyPublishHttp/DataProduceEndpoints.ts
```diff
@@ -77,8 +88,8 @@ export const router = (streamFetcher: StreamFetcher, publisher: Publisher, parti
             try {
                 const streamMessage = new StreamMessage({
                     messageId: new MessageID(
-                        req.stream.id,
-                        partitionFn(req.stream.partitions, req.query.pkey),
+                        req.stream!.id as string,
```
timoxley, line 91: can we put this in `AuthenticatedRequest` somehow

packages/broker/src/plugins/storage/DataQueryEndpoints.ts
```diff
@@ -52,7 +51,7 @@ class ResponseTransform extends Transform {
     }
 }
 
-function parseIntIfExists(x: Todo) {
+function parseIntIfExists(x: string | undefined): number | undefined {
```
timoxley, line 54: This might give better typing:
```typescript
function parseIntIfExists(x: undefined): undefined;
function parseIntIfExists(x: string): number;
function parseIntIfExists(x: string | undefined): number | undefined {

}
```

packages/broker/src/plugins/storage/DataQueryEndpoints.ts
```diff
@@ -148,8 +147,8 @@ export const router = (storage: Storage, streamFetcher: Todo, metricsContext: Me
 
     // eslint-disable-next-line max-len
     createEndpointRoute('from', router, metrics, (req: Request, streamId: string, partition: number, onSuccess: (data: Readable) => void, onError: (msg: string) => void) => {
-        const fromTimestamp = parseIntIfExists(req.query.fromTimestamp)
-        const fromSequenceNumber = parseIntIfExists(req.query.fromSequenceNumber) || MIN_SEQUENCE_NUMBER_VALUE
+        const fromTimestamp = parseIntIfExists(req.query.fromTimestamp as string)
+        const fromSequenceNumber = parseIntIfExists(req.query.fromSequenceNumber as string) || MIN_SEQUENCE_NUMBER_VALUE
```
timoxley: if `req` was `AuthenticatedRequest<QueryParams>` would `req.query` have accurate types?

packages/broker/src/plugins/storage/Storage.ts
```diff
@@ -150,10 +154,10 @@ export class Storage extends EventEmitter {
          * - if enough => get all messages and return
          * - if not => move to the next bucket and repeat cycle
          */
-        this.cassandraClient.eachRow(GET_BUCKETS, [streamId, partition], options, (_n: Todo, row: Todo) => {
+        this.cassandraClient.eachRow(GET_BUCKETS, [streamId, partition], options, (_n, row: types.Row) => {
             bucketId = row.id
             bucketIds.push(bucketId)
-        }, async (err: Todo, result: Todo) => {
+        }, async (err: Error, result: types.ResultSet) => {
```
timoxley: should this be `err: Error | undefined`

i.e. `if (err) {` on line 163 would always be true if `err` was always an `Error`

packages/broker/src/types/mqtt-connection/index.d.ts
```diff
@@ -0,0 +1,80 @@
+// Adapted from here https://github.com/mqttjs/mqtt-connection/issues/20. There doesn't seem to be any official typings
+// available, unfortunately.
+declare module 'mqtt-connection' {
```
timoxley, line 3: wow nice one

packages/broker/test/integration/StreamFetcher.test.ts
```diff
@@ -4,17 +4,24 @@ import bodyParser from 'body-parser'
 import { v4 as uuidv4 } from 'uuid'
 import { StreamFetcher } from '../../src/StreamFetcher'
 import { HttpError } from '../../src/errors/HttpError'
-import { Todo } from '../types'
+import http from "http"
```
timoxley: '

timoxley (approved):
great work 👍

packages/broker/src/errors/FailedToPublishError.ts
harbu: True. Probably in another PR then.

packages/broker/src/plugins/legacyMqtt/MqttServer.ts
harbu: Perhaps, it's a bit of a hard call when it comes to values that are run-time generated. If we think about it, we don't really have a guarantee that the type of `streamFetcher.authenticate` is anything else except `any`. I feel like if method `authenticate` actually validated the response, then we could set the type to be

```
{
                id: string,
                partitions: number
}
```

harbu: So essentially without validation, we have to force the type using `as` somewhere. Decided to do it here for now.

packages/broker/src/plugins/legacyPublishHttp/DataProduceEndpoints.ts
harbu: I guess a matter of taste.

packages/broker/src/plugins/legacyPublishHttp/DataProduceEndpoints.ts
harbu: Probably yeah, however seeing as this is legacy won't bother at this point.

packages/broker/src/plugins/storage/DataQueryEndpoints.ts
harbu: This doesn't work if invoking
```ts
let x: undefined | string
// fill x somehow
parseIntIfExists(x)
```
Which ended up being a problem when doing custom Request types

packages/broker/src/plugins/storage/Storage.ts
harbu: Aaah looks like the cassandra-driver is mistyped (packages/broker/node_modules/cassandra-driver/index.d.ts):
```ts
export type ValueCallback<T> = (err: Error, val: T) => void;
export type EmptyCallback = (err: Error) => void;
export type ArrayOrObject = any[]|{[key: string]: any};
```
Copied it from there. Indeed it should be `Error | undefined`

PR #251, opened 2021-10-15
network: extract DisconnectionManager from Node

harbu:
Again, in furtherance of encapsulation, modularization, and testing, extract node-to-node disconnection logic from `Node` into new class `DisconnectionManager`. 

Wrote a unit test for `DisconnectionManager`.

### DIFF
--- a/packages/network/src/logic/node/DisconnectionManager.ts
+++ b/packages/network/src/logic/node/DisconnectionManager.ts
@@ -27,6 +27,8 @@ export interface DisconnectionManagerOptions {
  *      from.
  */
 export class DisconnectionManager {
+    public static DISCONNECTION_REASON = 'no shared streams'
+
     private readonly disconnectionTimers = new Map<NodeId, NodeJS.Timeout>()
     private readonly getAllNodes: GetAllNodesFn
     private readonly hasSharedStreams: HasSharedStreamsFn
@@ -93,6 +95,6 @@ export class DisconnectionManager {
 
     private loggedDisconnect(nodeId: NodeId): void {
         logger.trace('executing disconnect from %s', NameDirectory.getName(nodeId))
-        this.disconnect(nodeId, 'no shared streams')
+        this.disconnect(nodeId, DisconnectionManager.DISCONNECTION_REASON)
     }
 }
--- a/packages/network/src/logic/node/DisconnectionManager.ts
+++ b/packages/network/src/logic/node/DisconnectionManager.ts
@@ -16,6 +16,16 @@ export interface DisconnectionManagerOptions {
     cleanUpIntervalInMs: number
 }
 
+/**
+ * DisconnectionManager assists a network node in disconnecting from other nodes when streams are
+ * no longer shared between them.
+ *
+ * There are two ways this is achieved:
+ *  1. Manual: a node can schedule (and cancel) disconnections that get executed after `disconnectionDelayInMs` if
+ *      they still don't share streams.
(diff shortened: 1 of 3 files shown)


juslesan (approved):

packages/network/src/logic/node/DisconnectionManager.ts
```diff
@@ -0,0 +1,97 @@
...
+ * no longer shared between them.
+ *
+ * There are two ways this is achieved:
+ *  1. Manual: a node can schedule (and cancel) disconnections that get executed after `disconnectionDelayInMs` if
+ *      they still don't share streams.
+ *  2. Automatic: a clean up interval is ran periodically in which any node without shared streams gets disconnected
+ *      from.
+ */
+export class DisconnectionManager {
+    private readonly disconnectionTimers: Record<NodeId, NodeJS.Timeout> = Object.create(null)
```
teogeb: Could use `Map` as a data type as this is a mapping of one value to another? Operations would be a slightly more readable, e.g. `Object.values(this.disconnectionTimers).forEach` -> `this.disconnectionTimers.forEach`

packages/network/src/logic/node/DisconnectionManager.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    stop(): void {
+        clearInterval(this.connectionCleanUpInterval!)
+        Object.values(this.disconnectionTimers).forEach((timeout) => {
+            clearTimeout(timeout)
+        })
+    }
+
+    scheduleDisconnectionIfNoSharedStreams(nodeId: NodeId): void {
+        if (!this.hasSharedStreams(nodeId)) {
```
teogeb, line 75: Would it make sense if we check the `this.hasSharedStreams(nodeId)` in a caller, and rename this method to `scheduleDisconnection`?

packages/network/test/unit/DisconnectionManager.test.ts
```diff
@@ -0,0 +1,179 @@
...
+
+        it('longer scenario', async () => {
+            getAllNodes.mockReturnValue(['n1', 'n2', 'n3', 'n4'])
+            hasSharedStreams.mockImplementation((nodeId) => ['n1', 'n4'].includes(nodeId))
+            setUpManager(1000, TTL)
+            manager.start()
+
+            await wait(TTL + 1)
+            expect(disconnect.mock.calls).toEqual([
+                ['n2', 'no shared streams'],
```
teogeb: Extract the `'no shared streams'` literal to a constant? Either in this class or in `DisconnectionManager`.

teogeb (approved):

packages/network/test/unit/DisconnectionManager.test.ts
harbu: Fixed

packages/network/src/logic/node/DisconnectionManager.ts
harbu: Fixed

packages/network/src/logic/node/DisconnectionManager.ts
harbu: I thought about such a version of the implementation. However, I ultimately decided not to go that route. My arguments would be:

- The timeout itself eventually checks `hasSharedStreams` so it felt appropriate that it gets checked by method itself before creating the timeout as it can easily do so as it has access to the appropriate method.
- Less burden to the user of this method as they don't have to check it themselves.

Downside is the long name, but imo it's worth the trade-off.

PR #254, opened 2021-10-18
InstructionSender refactor

harbu:
- Some refactoring of `InstructionSender`
- Slight *modification to behaviour*: it will send a stream's instructions to all nodes concurrently as opposed to one-by-one.

### DIFF
--- a/packages/network/src/logic/tracker/InstructionSender.ts
+++ b/packages/network/src/logic/tracker/InstructionSender.ts
@@ -3,7 +3,6 @@ import io from '@pm2/io'
 import { Logger } from '../../helpers/Logger'
 import { Metrics } from '../../helpers/MetricsContext'
 import { StreamIdAndPartition, StreamKey } from '../../identifiers'
-import { TrackerServer } from '../../protocol/TrackerServer'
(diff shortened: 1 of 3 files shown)


packages/network/src/logic/tracker/InstructionSender.ts
```diff
@@ -93,22 +103,29 @@ export class InstructionSender {
         }
     }
 
-    private async sendInstructions(buffer: StreamInstructionBuffer) {
-        for (const instruction of buffer!.getInstructions()) {
-            const { nodeId, streamKey, newNeighbors, counterValue } = instruction
+    private async sendInstructions(buffer: StreamInstructionBuffer): Promise<void> {
+        const promises: Promise<void>[] = []
+        for (const { nodeId, streamKey, newNeighbors, counterValue } of buffer.getInstructions()) {
```
teogeb: Could simplify this if we create `promises` by mapping `Array.from(buffer.getInstructions())`

harbu: fixed

teogeb (approved):

PR #255, opened 2021-10-18
NET-557: validate status messages before processing

juslesan:
- Validate status messages before processing to avoid tracker crashes on conflicting versions.
- New custom WS disconnection reason/code: INVALID_PROTOCOL_MESSAGE
- In the future, there could be a range of backwards compatible protocol versions that the trackers could process.


packages/network/src/connection/ws/NodeClientWsEndpoint.ts
```diff
@@ -56,6 +56,8 @@ export default class NodeClientWsEndpoint extends AbstractClientWsEndpoint<NodeC
             this.onClose(connection, code, reason as DisconnectionReason)
             if (code === DisconnectionCode.DUPLICATE_SOCKET) {
                 this.logger.warn('Connection refused: Duplicate nodeId detected, are you running multiple nodes with the same private key?')
+            } if (code === DisconnectionCode.VERSION_CONFICT) {
```
harbu:
```suggestion
            } else if (code === DisconnectionCode.VERSION_CONFICT) {
```
(or newline)

packages/network/src/connection/ws/AbstractWsEndpoint.ts
```diff
@@ -18,14 +18,16 @@ export enum DisconnectionCode {
     GRACEFUL_SHUTDOWN = 1000,
     FAILED_HANDSHAKE = 4000,
     DEAD_CONNECTION = 4001,
-    DUPLICATE_SOCKET = 4002
+    DUPLICATE_SOCKET = 4002,
+    VERSION_CONFICT = 4003
```
harbu:
```suggestion
    VERSION_CONFLICT = 4003
```

packages/network/src/logic/tracker/Tracker.ts
```diff
@@ -122,50 +122,56 @@ export class Tracker extends EventEmitter {
...
-        this.updateNodeOnStream(source, stream)
-        this.formAndSendInstructions(source, stream.streamKey)
+            // update topology
+            this.createTopology(stream.streamKey)
+            this.updateNodeOnStream(source, stream)
+            this.formAndSendInstructions(source, stream.streamKey)
+        }
+        catch (err: any) {
+            this.logger.warn(`Error while processing status message from node ${source}:\n${err}`)
+            this.trackerServer.disconnectFromPeer(source, DisconnectionCode.VERSION_CONFICT, DisconnectionReason.VERSION_CONFICT)
```
harbu: I don't think we can always attribute this error to be a version conflict. It could be totally something unrelated causing the error as well, e.g., a code change that causes something to break here consistently.

I think we should detect a version conflict _before the error_ occurs. Did we still exchange version information when connecting over WebSocket? Maybe this disconnection should be done at that point?

Secondly, I think we should avoid the `try-catch`-everything here. I would rather validate that variable `status` contains expected data in correct format. If not, disconnect the WebSocket maybe with reason `INVALID_STATUS_MESSAGE` or similar. 

After making sure that `status` is correct, anything that could fail afterwards would be due to a bug in our code. I don't think we should have to be defensive about our own code so I would remove the `catch`. That would make the software crash in case of an error, but I think that would be justified due to it really being a problem with our internal logic at that point.

packages/network/test/integration/tracker-node-status.test.ts
```diff
@@ -194,4 +195,27 @@ describe('check status message flow between tracker and two nodes', () => {
...
+            }
+        })
+        await runAndWaitForEvents([() => {
+            // @ts-expect-error private field
+            nodeOne.trackerManager.nodeToTracker.endpoint.send('tracker', faultyStatus.serialize())
+        }], [
+            // @ts-expect-error private field
+            [nodeOne.trackerManager.nodeToTracker, 'streamr:tracker-node:tracker-disconnected']
+        ])
+    })
```
harbu: I would write this using `NodeToTracker` directly to avoid having to access private methods and since we don't need features of `Node`.

You can check https://github.com/streamr-dev/network-monorepo/blob/main/packages/network/test/integration/counter-filtering-in-tracker.test.ts#L35-L41 for example

harbu (changes requested):

packages/network/src/logic/tracker/Tracker.ts
harbu: If I wanted to implement this quickly, I would actually just skip implementing the version check thing I described above and just validate the contents of `status` variable...

packages/network/test/integration/tracker-node-disconnects-on-invalid-status.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    let nodeToTracker: NodeToTracker
+    const TRACKER_ID = 'tracker'
+
+    beforeEach(async () => {
+        tracker = await startTracker({
+            listen: {
+                hostname: '127.0.0.1',
+                port: 30436
+            },
+            id: 'tracker'
```
harbu, line 20:
```suggestion
            id: TRACKER_ID
```

packages/network/test/integration/tracker-node-disconnects-on-invalid-status.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+        await runAndWaitForEvents([
+            () => nodeToTracker.connectToTracker(tracker.getUrl(), trackerPeerInfo)
+        ], [
+            [nodeToTracker, NodeToTrackerEvent.CONNECTED_TO_TRACKER]
+        ])
+    })
+    beforeEach(async () => {
+        await tracker.stop()
+        await nodeToTracker.stop()
```
harbu: - `allSettled` to ensure that both `stop` methods get invoked (even if `tracker.stop()` rejected)
-  `?.` in case `beforeEach` threw exception in the middle (i.e. `tracker` or `nodeToTracker` was never initialized)
```suggestion
        Promise.allSettled([
            await tracker?.stop()
            await nodeToTracker?.stop()
        ])
```

harbu (approved):
Nice, `ajv` works well here.

packages/network/src/logic/tracker/Tracker.ts
```diff
@@ -62,6 +64,7 @@ export class Tracker extends EventEmitter {
     private readonly logger: Logger
     private readonly metrics: Metrics
     private readonly statusMeter: any
+    private readonly statusSchemaValidator: any
```
teogeb: Would it make sense to validate the format in protocol package? There are a lot of validations already there for other message types. (Maybe we could also specify the format here: https://github.com/streamr-dev/streamr-specs/blob/master/PROTOCOL.md#statusmessage)

packages/network/src/connection/ws/AbstractWsEndpoint.ts
```diff
@@ -18,14 +18,16 @@ export enum DisconnectionCode {
     GRACEFUL_SHUTDOWN = 1000,
     FAILED_HANDSHAKE = 4000,
     DEAD_CONNECTION = 4001,
-    DUPLICATE_SOCKET = 4002
+    DUPLICATE_SOCKET = 4002,
+    INVALID_STATUS = 4003
```
teogeb: Maybe this could be more generic, e.g. `MALFORMED_PROTOCOL_MESSAGE`?

teogeb (changes requested):

packages/network/src/logic/tracker/Tracker.ts
teogeb: Created ticket: https://linear.app/streamr/issue/NET-568/network-move-protocol-level-logic-to-protocol-package

teogeb (approved):

teogeb:
Note that the format of the status message has changed in https://github.com/streamr-dev/network-monorepo/commit/6a453cceb1f063a3ae3d144599818f729f354bba

harbu (approved):
LGTM! :+1:  Feel free to merge to main when good for you. I'll look at the typescript in a separate PR if I have a chance.

PR #256, opened 2021-10-19
network, broker [NET-518]: replace StreamIdAndPartition with SPID

teogeb:
Use `SPID` and `SPIDKey` to identify and a stream partition. Removed deprecated classes/types: `StreamdIdAndPartition`, `StreamKey` and `StreamPart`.

Most changes are modifications to internal data structures.  

External modifications: 

1. Status message format changed from `{ streamKey: '0x123::0, ...' }` to `{ id: '0x123', partition: 0 ... }`. There is a temporary backwards compatibility handler so that Tracker can read the deprecated format.
2. The data format of Tracker's HTTP endpoints changed. The keys are now `SPIDKeys`, not `StreamKeys`.
3. The public accessor method in in `Node.ts` and `Tracker.ts` changed from getStreams() to getSPIDs(). The return value changed from `string` array to Iterable<SPID>.

How SPID is used in this PR:

1. `SPID` is an independent entity, and a preferred way to reference to a stream partition. When we add a stream partition e.g. to a storage config, we use method `addSPID()` (not `getStreams()` or `getStreamPartitions()`). 
2. `SPIDKey` is an internal data format. The public methods of the application (in `Node.ts` and `Tracker.ts`) return `SPIDs`, not `SPIDKeys`. If we need to use textual representation of a SPID (e.g. as a JSON key, or as a parameter in command line), we read and write data in `SPIDKey` format.
3. We don't yet have public API methods which take `SPIDs` as input parameters. Maybe we could allow both: `SPID|SPIDKey`? 
4. In protocol level (the status message), we don't use `SPID` or `SPIDKey`. The format is plain `{ id, partition }`. 
5. In log files we don't use terms `SPID` or `SPIDKey`. There we use the term "partition": e.g.`"Add 123 partitions to storage config"`.

The internal hashing in `TrackerRegistry` are calculated from `SPIDKey` values. Therefore `streamId` -> `trackerIndex` results will be different after this change.

### DIFF
--- a/packages/broker/src/Stream.ts
+++ b/packages/broker/src/Stream.ts
@@ -54,7 +54,7 @@ export class Stream<C> {
     }
 
     getSPIDKey(): Protocol.SPIDKey {
-        return new Protocol.SPID(this.id, this.partition).toKey()
+        return Protocol.SPID.toKey(this.id, this.partition)
     }
 
     getName(): string {
--- a/packages/broker/src/Stream.ts
+++ b/packages/broker/src/Stream.ts
@@ -1,5 +1,4 @@
 import { Protocol } from 'streamr-network'
-import { SPID } from 'streamr-network/dist/streamr-protocol'
 
 type State = 'init'|'subscribing'|'subscribed'
 
@@ -55,7 +54,7 @@ export class Stream<C> {
     }
 
     getSPIDKey(): Protocol.SPIDKey {
-        return new SPID(this.id, this.partition).toKey()
+        return new Protocol.SPID(this.id, this.partition).toKey()
     }
 
     getName(): string {
--- a/packages/broker/src/Stream.ts
+++ b/packages/broker/src/Stream.ts
@@ -1,3 +1,6 @@
+import { Protocol } from 'streamr-network'
+import { SPID } from 'streamr-network/dist/streamr-protocol'
+
 type State = 'init'|'subscribing'|'subscribed'
 
 export class Stream<C> {
@@ -51,8 +54,8 @@ export class Stream<C> {
         return this.state === 'subscribed'
     }
 
-    toString(): string {
-        return `${this.id}::${this.partition}`
+    getSPIDKey(): Protocol.SPIDKey {
+        return new SPID(this.id, this.partition).toKey()
     }
 
     getName(): string {
--- a/packages/broker/src/StreamStateManager.ts
+++ b/packages/broker/src/StreamStateManager.ts
@@ -1,15 +1,11 @@
 import { Stream } from './Stream'
-import { Logger } from 'streamr-network'
+import { Logger, Protocol } from 'streamr-network'
 
 const logger = new Logger(module)
 
-function getStreamLookupKey(streamId: string, streamPartition: number) {
-    return `${streamId}::${streamPartition}`
-}
-
 export class StreamStateManager<C> {
-    private streams: Record<string,Stream<C>> = {}
-    private timeouts: Record<string,NodeJS.Timeout> = {}
+    private streams: Record<Protocol.SPIDKey,Stream<C>> = {}
+    private timeouts: Record<Protocol.SPIDKey,NodeJS.Timeout> = {}
 
     getOrCreate(streamId: string, streamPartition: number, name = ''): Stream<C> {
         const stream = this.get(streamId, streamPartition)
@@ -20,7 +16,8 @@ export class StreamStateManager<C> {
     }
 
     get(streamId: string, streamPartition: number): Stream<C> {
-        return this.streams[getStreamLookupKey(streamId, streamPartition)]
+        const key = new Protocol.SPID(streamId, streamPartition).toKey()
+        return this.streams[key]
     }
 
     getByName(name: string): Stream<C>|null {
@@ -37,7 +34,7 @@ export class StreamStateManager<C> {
             throw new Error('streamId or streamPartition not given!')
         }
 
-        const key = getStreamLookupKey(streamId, streamPartition)
+        const key = new Protocol.SPID(streamId, streamPartition).toKey()
         if (this.streams[key]) {
             throw new Error(`stream already exists for ${key}`)
         }
@@ -62,7 +59,7 @@ export class StreamStateManager<C> {
             }
         }, 60 * 1000)
 
-        logger.debug('Stream object "%s" created', stream.toString())
+        logger.debug('Stream object "%s" created', stream.getSPIDKey())
         return stream
     }
 
@@ -73,13 +70,13 @@ export class StreamStateManager<C> {
 
         const stream = this.get(streamId, streamPartition)
         if (stream) {
-            const key = getStreamLookupKey(streamId, streamPartition)
+            const key = new Protocol.SPID(streamId, streamPartition).toKey()
             clearTimeout(this.timeouts[key])
             delete this.timeouts[key]
             delete this.streams[key]
         }
 
-        logger.debug('Stream object "%s" deleted', stream.toString())
+        logger.debug('Stream object "%s" deleted', stream.getSPIDKey())
     }
 
     close(): void {
--- a/packages/broker/src/StreamStateManager.ts
+++ b/packages/broker/src/StreamStateManager.ts
@@ -16,7 +16,7 @@ export class StreamStateManager<C> {
     }
 
     get(streamId: string, streamPartition: number): Stream<C> {
-        const key = new Protocol.SPID(streamId, streamPartition).toKey()
+        const key = Protocol.SPID.toKey(streamId, streamPartition)
         return this.streams[key]
     }
 
@@ -34,7 +34,7 @@ export class StreamStateManager<C> {
             throw new Error('streamId or streamPartition not given!')
         }
 
-        const key = new Protocol.SPID(streamId, streamPartition).toKey()
+        const key = Protocol.SPID.toKey(streamId, streamPartition)
         if (this.streams[key]) {
             throw new Error(`stream already exists for ${key}`)
         }
@@ -70,7 +70,7 @@ export class StreamStateManager<C> {
 
         const stream = this.get(streamId, streamPartition)
         if (stream) {
-            const key = new Protocol.SPID(streamId, streamPartition).toKey()
+            const key = Protocol.SPID.toKey(streamId, streamPartition)
             clearTimeout(this.timeouts[key])
             delete this.timeouts[key]
             delete this.streams[key]
--- a/packages/broker/src/broker.ts
+++ b/packages/broker/src/broker.ts
@@ -1,7 +1,6 @@
-import { Protocol, MetricsContext } from 'streamr-network'
+import { Logger, Protocol, MetricsContext } from 'streamr-network'
 import StreamrClient from 'streamr-client'
 import { Wallet } from 'ethers'
-import { Logger } from 'streamr-network'
 import { Server as HttpServer } from 'http'
 import { Server as HttpsServer } from 'https'
 import { Publisher } from './Publisher'
@@ -20,7 +19,7 @@ const logger = new Logger(module)
 
 export interface Broker {
     getNeighbors: () => readonly string[]
-    getStreams: () => readonly string[]
+    getSPIDs: () => Iterable<Protocol.SPID>
     getNodeId: () => string
     start: () => Promise<unknown>
     stop: () => Promise<unknown>
@@ -127,7 +126,7 @@ export const createBroker = async (config: Config): Promise<Broker> => {
 
     return {
         getNeighbors: () => networkNode.getNeighbors(),
-        getStreams: () => networkNode.getStreams(),
+        getSPIDs: () => networkNode.getSPIDs(),
         getNodeId: () => networkNode.getNodeId(),
         start: async () => {
             logger.info(`Starting broker version ${CURRENT_VERSION}`)
--- a/packages/broker/src/plugins/storage/StorageConfig.ts
+++ b/packages/broker/src/plugins/storage/StorageConfig.ts
@@ -12,7 +12,7 @@ export interface StorageConfigListener {
 const getSPIDKeys = (streamId: string, partitions: number): Protocol.SPIDKey[] => {
     const keys: Protocol.SPIDKey[] = []
     for (let i = 0; i < partitions; i++) {
-        keys.push(new Protocol.SPID(streamId, i).toKey())
+        keys.push(Protocol.SPID.toKey(streamId, i))
     }
     return keys
 }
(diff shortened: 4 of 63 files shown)


packages/broker/src/Stream.ts
```diff
@@ -1,3 +1,6 @@
+import { Protocol } from 'streamr-network'
+import { SPID } from 'streamr-network/dist/streamr-protocol'
```
harbu: Should import from `Protocol`?

packages/broker/src/StreamStateManager.ts
```diff
@@ -37,7 +34,7 @@ export class StreamStateManager<C> {
             throw new Error('streamId or streamPartition not given!')
         }
 
-        const key = getStreamLookupKey(streamId, streamPartition)
+        const key = new Protocol.SPID(streamId, streamPartition).toKey()
```
harbu: I wonder if we should make static method `SPID#toKey`   also support passing two arguments. So we can replace instances of 
```ts
const key = new Protocol.SPID(streamId, streamPartition).toKey()
```
with
```ts
const key = SPID.toKey(streamId, streamPartition)
```

It seems to be a common enough pattern to perhaps warrant doing?

packages/broker/src/plugins/storage/StorageConfig.ts
```diff
@@ -1,38 +1,19 @@
 import fetch from 'node-fetch'
-import { Logger } from 'streamr-network'
-import { StreamPart } from '../../types'
-import { StreamMessage, keyToArrayIndex } from 'streamr-network/dist/streamr-protocol'
+import { Logger, Protocol } from 'streamr-network'
+import { SPID } from 'streamr-network/dist/streamr-protocol'
```
harbu: Import

packages/broker/src/plugins/storage/StorageConfig.ts
```diff
@@ -1,38 +1,19 @@
...
-    return {
-        id,
-        partition: Number(partitionStr)
-    }
-}
-
-const getKeyFromStream = (streamId: string, streamPartition: number): StreamKey => {
-    return `${streamId}::${streamPartition}`
+    onSPIDAdded: (spid: Protocol.SPID) => void
+    onSPIDRemoved: (spid: Protocol.SPID) => void
```
harbu, line 9: I'm wonder if it would be more clear to refer to the stream partitions by full name?
```suggestion
    onStreamPartitionAdded: (spid: Protocol.SPID) => void
    onStreamPartitionRemoved: (spid: Protocol.SPID) => void
```

packages/broker/src/plugins/storage/StorageConfig.ts
```diff
@@ -49,7 +30,7 @@ export class StorageConfig {
 
     static ASSIGNMENT_EVENT_STREAM_ID_SUFFIX = '/storage-node-assignments'
 
-    streamKeys: Set<StreamKey>
+    private spidKeys: Set<Protocol.SPIDKey>
```
harbu, line 32: Again, wonder whether it would be more clear to use full name?
```suggestion
    private streamPartitions: Set<Protocol.SPIDKey>
```

packages/broker/src/plugins/storage/StorageConfig.ts
```diff
@@ -100,13 +85,13 @@ export class StorageConfig {
         this.poller = setTimeout(() => this.poll(pollInterval), pollInterval)
     }
 
-    hasStream(stream: StreamPart): boolean {
-        const key = getKeyFromStream(stream.id, stream.partition)
-        return this.streamKeys.has(key)
+    hasSPID(spid: Protocol.SPID): boolean {
```
harbu, line 87: Pondering whether full name would be clearer?
```suggestion
    hasStreamPartition(spid: Protocol.SPID): boolean {
```

packages/broker/src/broker.ts
```diff
@@ -20,7 +19,7 @@ const logger = new Logger(module)
 
 export interface Broker {
     getNeighbors: () => readonly string[]
-    getStreams: () => readonly string[]
+    getSPIDs: () => Iterable<Protocol.SPID>
```
harbu, line 22: I wonder if we should refer to them as `StreamPartitions` in names?
```suggestion
    getStreamPartitions: () => Iterable<Protocol.SPID>
```

packages/broker/src/plugins/storage/StorageConfig.ts
```diff
@@ -100,13 +85,13 @@ export class StorageConfig {
...
+        const key = new Protocol.SPID(spid.streamId, spid.streamPartition).toKey()
+        return this.spidKeys.has(key)
     }
 
-    getStreams(): StreamPart[] {
-        return Array.from(this.streamKeys).map((key) => getStreamFromKey(key))
+    getSPIDs(): Protocol.SPID[] {
```
harbu, line 92:
```suggestion
    getStreamPartitions(): Protocol.SPID[] {
```

packages/broker/src/plugins/storage/StoragePlugin.ts
```diff
@@ -46,20 +45,16 @@ export class StoragePlugin extends Plugin<StoragePluginConfig> {
         this.cassandra = await this.getCassandraStorage()
         this.storageConfig = await this.createStorageConfig()
         this.messageListener = (msg) => {
-            const streamPart = {
-                id: msg.messageId.streamId,
-                partition: msg.messageId.streamPartition
-            }
-            if (this.storageConfig!.hasStream(streamPart)) {
+            if (this.storageConfig!.hasSPID(Protocol.SPID.from(msg.messageId))) {
```
harbu: Would it be useful to add a method to `StreamMessage` for retrieving the SPID? Seems like a use case that will come up a lot.

packages/broker/test/unit/plugins/storage/StorageConfig.test.ts
```diff
@@ -1,6 +1,7 @@
 /* eslint-disable no-underscore-dangle, object-curly-newline */
 import { StorageConfig, StorageConfigListener } from '../../../../src/plugins/storage/StorageConfig'
 import nock from 'nock'
+import { SPID } from 'streamr-network/dist/streamr-protocol'
```
harbu: import

packages/network/src/helpers/transformIterable.ts
```diff
@@ -0,0 +1,5 @@
+export const transformIterable = function *<F,T>(from: Iterable<F>, tranform: (f: F) => T): Iterable<T> {
```
harbu, line 1: Is this the equivalent of `Array#map` but of `Iterable`?

packages/network/src/logic/node/StreamManager.ts
```diff
@@ -16,16 +17,16 @@ function keyForDetector({ publisherId, msgChainId }: MessageLayer.MessageID) {
 }
 
 export class StreamManager {
-    private readonly streams: Map<string, StreamState> = new Map<string, StreamState>() // streamKey => {}
+    private readonly streams: Map<SPIDKey,StreamState> = new Map<SPIDKey,StreamState>() // spidKey => {}
```
harbu: Comment no longer necessary (remnant from pre-TS days, I belive)
```suggestion
    private readonly streams: Map<SPIDKey,StreamState> = new Map<SPIDKey,StreamState>()
```

packages/network/src/logic/node/TrackerConnector.ts
```diff
@@ -12,12 +12,12 @@ enum ConnectionState {
     ERROR
 }
 
-type GetStreamsFn = () => Iterable<StreamIdAndPartition>
+type GetSPIDsFn = () => Iterable<SPID>
```
harbu, line 15: Maybe full name?
```suggestion
type GetStreamPartitionsFn = () => Iterable<SPID>
```

packages/network/src/logic/node/TrackerConnector.ts
```diff
@@ -26,22 +26,22 @@ export class TrackerConnector {
...
         this.connectToTracker = connectToTracker
         this.disconnectFromTracker = disconnectFromTracker
         this.trackerRegistry = trackerRegistry
         this.maintenanceInterval = maintenanceInterval
         this.connectionStates = new Map()
     }
 
-    onNewStream(streamId: StreamIdAndPartition): void {
-        const trackerInfo = this.trackerRegistry.getTracker(streamId.id, streamId.partition)
+    onNewStream(streamId: SPID): void {
```
harbu: Re-name to spid or streamPartition?

packages/network/src/logic/node/TrackerConnector.ts
```diff
@@ -26,22 +26,22 @@ export class TrackerConnector {
...
+        const trackerInfo = this.trackerRegistry.getTracker(streamId.streamId, streamId.streamPartition)
```
harbu: Wondering if we should support `TrackerRegistry#getTracker(spid: SPID)` ?

packages/network/src/logic/node/TrackerManager.ts
```diff
@@ -21,14 +21,14 @@ interface NodeDescriptor {
 }
 
 interface Subscriber {
-    subscribeToStreamIfHaveNotYet: (streamId: StreamIdAndPartition, sendStatus?: boolean) => void
+    subscribeToStreamIfHaveNotYet: (spid: SPID, sendStatus?: boolean) => void
```
harbu, line 24: I'm a bit undecided if we should use `streamPartition` as names for these or `spid`?
```suggestion
    subscribeToStreamIfHaveNotYet: (streamPartition: SPID, sendStatus?: boolean) => void
```

packages/network/src/logic/tracker/InstructionCounter.ts
```diff
@@ -1,43 +1,45 @@
-import { Status, StreamKey } from '../../identifiers'
+import { SPID, SPIDKey } from 'streamr-client-protocol'
+import { Status } from '../../identifiers'
 import { NodeId } from '../node/Node'
 
 export const COUNTER_UNSUBSCRIBE = -1
 
-type Counters = Record<NodeId,Record<StreamKey,number>>
+type Counters = Record<NodeId,Record<SPIDKey,number>>
```
harbu: I find it hard to parse these without spaces tbh:
```suggestion
type Counters = Record<NodeId, Record<SPIDKey, number>>
```

packages/network/test/unit/deprecated-tracker-status.test.ts
```diff
@@ -0,0 +1,89 @@
...
+        const actualStatus = updateNodeOnStream.mock.calls[0][1]
+        expect(actualStatus).toContainEntries([
+            ['id', STREAM_ID],
+            ['partition', STREAM_PARTITION],
+            ['neighbors', [TARGET_NODE_1, TARGET_NODE_2]],
+            ['counter', COUNTER],
+        ])
+        expect(formAndSendInstructions).toBeCalledTimes(1)
+        expect(formAndSendInstructions).toBeCalledWith(SOURCE_NODE, spidKey)
+    }
```
harbu: Instead of having to rely on all this mocking (and internals that are bound to change), couldn't you just use something from`trackerSummaryUtils.ts` to check the state of the tracker after a node with deprecated status message joins?

packages/protocol/src/utils/SPID.ts
```diff
@@ -13,6 +13,8 @@ import ValidationError from '../errors/ValidationError'
 
 type RequiredKeys<T, Keys extends keyof T> = Omit<T, Keys> & Required<Pick<T, Keys>>
 
+export type SPIDKey = string
```
harbu, line 16: With template literal types (https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) we can actually do the following
```suggestion
export type SPIDKey = `${string}#${number}`
```

harbu (commented):
Nice work!

### Discussion Points
- <del>How to refer to SPIDs in variable / method names? E.g. `getSpids()` or `getStreamPartitions()`. I'm not convinced which one would be more readable in the end, both have their trade-offs.</del> We will use SPID.

### Suggestions
- Make SPIDKey into template literal type for a bit tighter type checking
- Add static method `SPID#toKey(streamId: string, streamPartition: number)` for ergonomics
- We should probably have method `StreamMessage#getSPID()` for convenience?
- SPID-argument support for `TrackerRegistry`, i.e., `TrackerRegistry#getTracker(spid: SPID)`?

### Fixes
- Some places are importing from `streamr-network/dist/streamr-protocol`
- I think I noticed some `streamId` named variables containing `SPID`. Not critical, but something I think I saw a couple times.

packages/network/test/unit/deprecated-tracker-status.test.ts
teogeb: This backwards compatibility is a temporary feature, which we'll remove very soon. Then can easily remove this test, too, as all the backwards compatibility handling is tested here.

packages/network/src/helpers/transformIterable.ts
teogeb: Yes, and doing the mapping lazily. One possible name for the utility function could be `mapIterable`.

packages/network/src/logic/node/TrackerConnector.ts
teogeb: Changed  `getTracker` to get the parameter as `SPID`. 

Also the internal hashing in TrackerRegistry are calculated from `SPIDKey` values. Therefore `streamId` -> `trackerIndex` results will be different after this change.

packages/broker/src/plugins/storage/StoragePlugin.ts
teogeb: There was already that method, now using it in network and broker code. Added `InstructionMessage.getSPID()`

packages/protocol/src/utils/SPID.ts
teogeb: Seems quite complicated to add that restriction as we have *many* input formats to support. If we can simplify `SPID.fromDefaults()`, it should be possible to add the support for literal type.

If I remember correctly `SPID.fromDefaults()` was planned to be a temporary helper method, which helps us migrate all  input format variants to use `SPID`. Especially there was need for `SPID.fromDefaults(streamId)` to help us to collect all "partition 0 is the default, if not specified otherwise" logic to one place (and later migrate those usages one by one).

@timoxley Do I remember correctly? Can we therefore drop `SPID.fromDefaults()` after we the migration?

harbu (approved):
LGTM!

packages/protocol/src/utils/SPID.ts
teogeb: Created tickets:
- https://linear.app/streamr/issue/NET-566/client-use-spid
- https://linear.app/streamr/issue/NET-567/spidkey-as-template-literal-type

PR #257, opened 2021-10-20
network: rename SPID usages

teogeb:
Renamed many fields/methods/classes which operate on `SPID` object. No functional changes.

Changed also some log messages and the content of `AbstractWsEndpoint.DisconnectionReason.NO_SHARED_STREAMS` (an undocumented error code).

`NetworkNode`
- `getNeighborsForStream`->`getNeighborsForSPID`
- `subscribe` and `unsubscribe` parameters are SPIDs

`StreamManager` -> `SPIDManager`
- fields and methods
- `removeNodeFromStream`->`removeNeighbor`

`StreamStatus`->`SPIDStatus`

`Node.streams`->`spidManager`, `unsubscribeFromStream` -> `unsubscribeFromSPID`

`DisconnectionManager.hasSharedStreams`->`hasSharedSPIDs`

`TrackerManager.Subscriber` methods

`trackerSummaryUtils.getStreamSizes`->`getSPIDSizes`

`DisconnectionManager.scheduleDisconnectionIfNoSharedStreams`->`scheduleDisconnectionIfNoSharedSPIDs`

`InstructionRetryManager/InstructionThrottler/InstructionCounter.removeStream`->`removeSPID`

`TrackerManager` and `TrackerConnector` methods

`trackerSummaryUtils.findStreamsForNode`->`findSPIDsForNode`

`overlayPerStream`->`overlayPerSPID`


packages/broker/src/plugins/storage/BucketManager.ts
```diff
@@ -1,15 +1,14 @@
 import { Client } from 'cassandra-driver'
 import Heap from 'heap'
 import { types as cassandraTypes } from 'cassandra-driver'
-import { Logger } from 'streamr-network'
+import { Logger, Protocol } from 'streamr-network'
 import { Bucket, BucketId } from './Bucket'
+import { SPIDKey } from '../../../../network/node_modules/streamr-client-protocol/dist/src'
```
harbu: import

timoxley (commented):
nice, this definitely makes it clear when things are "per stream" vs "per stream + partition".

packages/broker/src/plugins/storage/BucketManager.ts
```diff
@@ -1,15 +1,14 @@
...
 const { TimeUuid } = cassandraTypes
 
 const logger = new Logger(module)
 
-type StreamPartKey = string
-
-interface StreamPartState {
+interface SPIDState {
```
harbu, line 10: Ping @timoxley 

About naming, we had a long discussion about this PR with @teogeb, and I feel like in many (but not all) instances using `spid` or `SPID` or equivalent shorthand in method / field / parameter names makes the code  less readable.

Let's take this line, for example. For me, `SPIDState` is not as concise of a name as `StreamPartitionState`. In this case, it is not really that we are maintaining state for a "stream partition id" as much as we maintaining a state for a "stream partition". The `spid` is just a way to reference a particular stream partition.

Another example below in this file, we have a field called `spids: Record<SPIDKey,SPIDState>`. It's a bit unclear what the variable is actually about. Perhaps something like `activeStreamPartitions: Record<SPIDKey, StreamPartitionState>` would be more revealing?

A bit of an analogy I thought of. `string` is a type but rarely do we name a string variable itself as `string` instead we opt for something more revealing. Obviously `spid` is far less general of a type as `string`, but I feel like some of the same thought applies here as well.

Another example. Say I created a new data structure that holds strings in a collection, such as a `Queue<string>`, `Heap<string>`, `Cache<string>`, whatever. Let's say I decide to drop the generic type template for whatever reason and just decide it stores strings, period. In that case, the variable and parameter names within the class could reasonably be named "string" since from the data structure's viewpoint they are merely strings without any further meaning. Whereas then something using this data structure would probably want to name its string variables something else for clarity.

But coming back to the question in hand. I wonder if we should favor "streamPartition" at least in method names? E.g.
```ts
getStreamPartitions(): Array<SPID>
```
 vs.
 ```ts
 getSPIDs(): Array<SPID>
 ```
 
How about in parameter names?
 ```ts
subscribe(streamPartition: SPID)
```
 vs.
 ```ts
subscribe(spid: SPID)
 ```
 
Something about naming everything `spid` feels very off / unintuitive to me. Granted this is quite subjective, I wonder what you guys think about this?

timoxley: I think you may be right, especially once you have SID in there too. 

Maybe we should have a rethink and see if there's another option, perhaps something that isn't a derivative of Stream + Partition.

Good thing is we could land this to get centralisation and make the name change after, since SPID will behave nicely with find/replace.

teogeb: Yeah, now when we see all the use cases of `SPID`s, it would be good to time to check the big picture. I totally agree with @harbu that especially the class names (e.g. `SPIDState` and `SPIDState`) are misleading.

If we find a new self-explanatory term, that would be great. On the other hand, the term stream partition is already used in end user UIs and documentation, and it is easy to understand as it is  plain English. 

To me, `SPID` feels like an independent term, not just an abbreviation of a stream partition. Therefore in some sense, we have now two words for the same concept: `SPID` and `stream partition`. Maybe it is just my interpretation. But if that's a more common interpretation, then there's some unnecessary complexity which we could maybe optimize?

When we introduced the `SPID` class few months ago, we didn't yet include the `SPIDKey` type. Now when we have both, it sounds to me that `SPIDKey` represents the ID of a stream partition, and `SPID` is the  actual stream partition. How about if we rename `SPID` to `StreamPartition` and `SPIDKey` to `SPID`/`StreamPartitionID`? 

That solution would make the method parameter types very guessable. An API method `addStreamPartition(streamPartition: StreamPartition)` would be super easy to use.

packages/broker/src/plugins/storage/BucketManager.ts
teogeb: Fixed

PR #258, opened 2021-10-21
NET-554: add option to filter remote ICE candidates with private IPv4

paullouisageneau:
This PR adds the option `network.webrtcDisallowPrivateAddresses` to filter out ICE candidates with private IPv4 to prevent the WebRTC agent from probing those addresses. 

The options is disabled by default (meaning private IPv4 addresses are allowed).

```
"network": {
    "webRtcDisallowPrivateAddresses": true // default false
}
```

### DIFF
--- a/packages/broker/src/broker.ts
+++ b/packages/broker/src/broker.ts
@@ -85,7 +85,7 @@ export const createBroker = async (config: Config): Promise<Broker> => {
 
     const usePredeterminedNetworkId = !config.generateSessionId || config.plugins['storage']
 
-    const webrtcDisallowPrivateAddresses = config.network?.webrtcDisallowPrivateAddresses || false
+    const webrtcDisallowPrivateAddresses = config.network.webrtcDisallowPrivateAddresses
 
     const streamrClient = new StreamrClient({
         auth: {
--- a/packages/broker/src/broker.ts
+++ b/packages/broker/src/broker.ts
@@ -96,7 +96,8 @@ export const createBroker = async (config: Config): Promise<Broker> => {
             trackers,
             location: config.network.location,
             metricsContext,
-            stunUrls: getStunTurnUrls(config)
+            stunUrls: getStunTurnUrls(config),
+            webrtcDisallowPrivateAddresses: config.network?.webrtcDisallowPrivateAddresses || false
         }
     })
     const publisher = new Publisher(streamrClient, metricsContext)
--- a/packages/broker/src/broker.ts
+++ b/packages/broker/src/broker.ts
@@ -84,6 +84,9 @@ export const createBroker = async (config: Config): Promise<Broker> => {
     const storageNodeRegistry = StorageNodeRegistry.createInstance(config, storageNodes)
 
     const usePredeterminedNetworkId = !config.generateSessionId || config.plugins['storage']
+
+    const webrtcDisallowPrivateAddresses = config.network?.webrtcDisallowPrivateAddresses || false
+
     const streamrClient = new StreamrClient({
         auth: {
             privateKey: config.ethereumPrivateKey,
@@ -97,7 +100,7 @@ export const createBroker = async (config: Config): Promise<Broker> => {
             location: config.network.location,
             metricsContext,
             stunUrls: getStunTurnUrls(config),
-            webrtcDisallowPrivateAddresses: config.network?.webrtcDisallowPrivateAddresses || false
+            webrtcDisallowPrivateAddresses
         }
     })
     const publisher = new Publisher(streamrClient, metricsContext)
@@ -146,6 +149,10 @@ export const createBroker = async (config: Config): Promise<Broker> => {
             logger.info(`Configured with trackers: [${trackers.map((tracker) => tracker.http).join(', ')}]`)
             logger.info(`Configured with Streamr: ${config.streamrUrl}`)
             logger.info(`Plugins: ${JSON.stringify(plugins.map((p) => p.name))}`)
+
+            if (!webrtcDisallowPrivateAddresses) {
+                logger.warn('WebRTC private address probing is allowed, this might cause issues for hosted nodes.')
+            }
         },
         stop: async () => {
             if (httpServer !== undefined) {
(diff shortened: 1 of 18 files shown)


packages/broker/src/broker.ts
```diff
@@ -145,6 +149,10 @@ export const createBroker = async (config: Config): Promise<Broker> => {
             logger.info(`Configured with trackers: [${trackers.map((tracker) => tracker.http).join(', ')}]`)
             logger.info(`Configured with Streamr: ${config.streamrUrl}`)
             logger.info(`Plugins: ${JSON.stringify(plugins.map((p) => p.name))}`)
+
+            if (!webrtcDisallowPrivateAddresses) {
+                logger.warn('WebRTC private address probing is allowed, this might cause issues for hosted nodes.')
```
timoxley: Based on how people respond to log messages during the testnet, I feel like this warning generates more questions than it answers:
what is the actionable task for a user?
what kind of issues, for whom?
what is a "hosted node"?
what are the consequences of disabling it?
etc

Maybe something like this?:

"WebRTC private address probing is allowed. This can trigger a false-positive for anti-malware detection on some web hosts e.g. Hetzner. You can disable this by setting `webrtcDisallowPrivateAddresses: true` in your broker config. Note that disabling this will prevent direct connections between peers using IPv4 addresses on your local network."

It's a bit wordy. I feel like we could/should start replacing these types of messages with links to more information so we have the space to elaborate and update the text without redeploying, while keeping the logs clean.

e.g.
```
WebRTC private address probing is allowed. This can trigger a false-positive for anti-malware detection on some web hosts. More Info: https://streamr.network/msg/private-address-probing
```
Or perhaps it could link to a page on the monorepo github wiki.

packages/network/src/helpers/AddressTools.ts
```diff
@@ -0,0 +1,27 @@
...
+                return true
+            }
+        }
+    }
+
+    return false
+}
+
+export function getAddressFromIceCandidate(candidate: string): string | undefined {
+    const fields = candidate.split(' ').filter((field: string) => field.length > 0)
```
timoxley: note that new versions of typescript can now (probably) infer the `:string` in `.filter((field: string) => field.length > 0`
```suggestion
    const fields = candidate.split(' ').filter((field) => field.length > 0)
```
Doesn't matter much either way, just FYI.

packages/network/src/helpers/AddressTools.ts
```diff
@@ -0,0 +1,27 @@
+import ipaddr from 'ipaddr.js'
+
+// IPv4 private address ranges as specified by RFC 1918
+const IPv4PrivateRanges = [
+    '10.0.0.0/8',
+    '172.16.0.0/12',
+    '192.168.0.0/16',
+].map((a: string) => ipaddr.parseCIDR(a))
```
timoxley: same here: 
```suggestion
].map((a) => ipaddr.parseCIDR(a))
```
Up to you though.

packages/network/src/helpers/AddressTools.ts
paullouisageneau: Sure :ok_hand:

packages/broker/src/broker.ts
paullouisageneau: Good point, I changed the message to something more explicit with a link to a wiki page:

`WebRTC private address probing is allowed. This can trigger false-positives for port scanning detection on some web hosts. More info: https://github.com/streamr-dev/network-monorepo/wiki/WebRTC-private-addresses`

packages/broker/src/broker.ts
```diff
@@ -145,6 +149,12 @@ export const createBroker = async (config: Config): Promise<Broker> => {
             logger.info(`Configured with trackers: [${trackers.map((tracker) => tracker.http).join(', ')}]`)
             logger.info(`Configured with Streamr: ${config.streamrUrl}`)
             logger.info(`Plugins: ${JSON.stringify(plugins.map((p) => p.name))}`)
+
+            if (!webrtcDisallowPrivateAddresses) {
+                logger.warn('WebRTC private address probing is allowed. ' +
+                    'This can trigger false-positives for port scanning detection on some web hosts. ' +
+                    'More info: https://github.com/streamr-dev/network-monorepo/wiki/WebRTC-private-addresses')
```
timoxley, line 156: Nice

packages/broker/src/helpers/config.schema.json
```diff
@@ -191,6 +191,11 @@
               }
             }
           ]
+        },
+        "webrtcDisallowPrivateAddresses": {
+            "type": "boolean",
+            "description": "If true, disallow WebRTC from probing private addresses",
+            "default": "false"
```
timoxley: Should this be `false` i.e. without quotes?

timoxley (commented):
👍

packages/broker/src/broker.ts
```diff
@@ -84,6 +84,9 @@ export const createBroker = async (config: Config): Promise<Broker> => {
     const storageNodeRegistry = StorageNodeRegistry.createInstance(config, storageNodes)
 
     const usePredeterminedNetworkId = !config.generateSessionId || config.plugins['storage']
+
+    const webrtcDisallowPrivateAddresses = config.network?.webrtcDisallowPrivateAddresses || false
```
harbu: From what I understand (@teogeb can correct me if I'm wrong) you can rely on the default value of the schema here being present

```suggestion
    const webrtcDisallowPrivateAddresses = config.network.webrtcDisallowPrivateAddresses
```

packages/broker/src/helpers/config.schema.json
harbu: Good question, as this will be used as the default value when instantiating the config object. I'd imagine it should be `false` without quotes so that it is not interpreted as true?

paullouisageneau: Indeed, it should actually be an actual boolean.

harbu (approved):

PR #259, opened 2021-10-21
npm workspaces

timoxley:
Part of some work I'm doing to generate docs. I want to create a new sub-package `packages/docs/`, which depends on the other sub-packages, then uses `tsdoc` to generate the docs for all of them.

However, there's some bug in `lerna` when you have multiple (linked) copies of the same package at different levels, and trying to create the `node_modules/.bin/` symlinks it will error with `EEXIST`. Could not find a workaround short of removing the `bin` scripts from `broker`. 

Lerna is semi-abandoned and was always a temporary solution so I thought it might be time to retry `npm` workspaces. The latest `npm` 8 release appears to be usable. Installs work. Tests work. Isn't significantly slower than `lerna`.

### Key Changes

* Use `npm@8`
* `npm ci` will install everything. `npm run bootstrap` is `npm ci`.
* You can now run `npm run bootstrap-pkg streamr-network` without running anything else beforehand. No `npm ci`, `npm install` or `npm run bootstrap` needed. `bootstrap-pkg` installs both top-level and sub-dependencies. This is equivalent to what `npm ci && npm run bootstrap-pkg streamr-network` used to do.
* `npm run bootstrap-root` Removes built-files & any sub-package dependencies from `node_modules`, leaving `package-lock.json` and `package.json` alone.  Use when you (only) want to be able to run top-level scripts e.g. `npm run versions` without installing anything else.
* To install dependencies into a sub-package you should use `npm install --workspace` from the top-level, with the name of the sub-package you want to update. e.g. `npm install --workspace=streamr-network`, or just regular `npm install` at the top-level to install everything. It's possible these two commands will generate different package-lock results 🤷 .

### Comparison

|  Comand      |  After (using `npm` workspaces) | Before (using `lerna`) |
| ------------ | ----------- | ----------- |
| `npm ci`      |  Installs all top-level dependencies AND sub-packages according to top-level `package-lock.json` | Installs only top-level dependencies e.g. `lerna`  according to top-level `package-lock.json` |
| `npm install`      | Installs all top-level dependencies AND sub-packages and updates `package-lock.json`. Will install new sub-package dependencies. |  Installs only top-level dependencies and updates `package-lock.json` 
| `npm run bootstrap`   | Runs `npm ci`. | Installs dependencies for sub-packages using `lerna`. Unclear whether package-lock is used, but it will update `package-lock.json`.  |
| `npm run bootstrap-pkg streamr-network`      | Installs top-level dependencies AND the dependencies for a sub-package using package-lock. Does not update `package-lock.json`. Will not install new sub-package dependencies. | Installs dependencies for a sub-package using `lerna`. Unclear if package-lock is used. Does not update `package-lock.json`. |
| `npm install --include-workspace-root --workspace streamr-network`       | Installs top-level dependencies (`--include-workspace-root`) AND the dependencies for a sub-package. Will update `package-lock.json`. Will install new sub-package dependencies.  | N/A but roughly equivalent to running this in the old setup: `npm install && npm run bootstrap-pkg streamr-network` |
| `npm run clean-dist`       | Removes cache and `dist` from sub-packages. Removes cache from top-level. | Removes cache and `dist` from sub-packages. |
| `npm run clean`      |  Removes all cache, dist & packages from root & sub-package `node_modules/`, except those packages needed by the top-level `package.json` e.g. `zx`. |  Runs `npm run clean-dist` and `npm run bootstrap-root`. |
| `npm run prune-pkg streamr-broker`      | Removes all packages except the production deps needed by the specified package's `package.json`. Note they will be installed in the top-level `node_modules/` not `packages/*/node_modules/` | N/A, roughly equivalent to `cd packages/streamr-broker/ && npm prune --production` 
| `npm run bootstrap-root`      | Removes all packages except those needed by top-level `package.json`. | N/A, roughly equivalent to `npm ci && npm run clean` 

Left `lerna` in the tree because `lerna` works best for implementing the `npm run versions` script.

Changes are against browser client PR #244.


.github/workflows/broker.yml
```diff
@@ -42,15 +42,14 @@ jobs:
         uses: actions/cache@v2.1.6
         with:
           path: ~/.npm
-          key: ${{ runner.os }}-node-${{ matrix.node-version }}-${{ hashFiles('**/package-lock.json') }}
+          key: ${{ runner.os }}-node-${{ matrix.node-version }}-${{ hashFiles('package-lock.json') }}
```
harbu: What is the purpose of these `**/package-lock.json` => `package-lock.json` changes?

harbu (approved):
LGTM! :+1:

.github/workflows/broker.yml
kare: Current `main` uses `actions/setup-node` cache and `actions/cache` is removed.

kare: I think the Tim's intention here is to cache per lerna|npm workspace-package. Perhaps cache key should be per package also?

timoxley: @harbu @kare `npm` now hoists all non-conflicting dependencies into the top-level `node_modules` directory, e.g. when installed via monorepo, there should now only be a single copy of the `typescript` package, shared across all sub-packages. This means `npm` only creates a single top-level `package-lock.json`, the per-package package-lock files are gone.

Given @kare's changes this can probably be removed.

PR #260, opened 2021-10-22
ci: clean up workflows

kare:


harbu:
Haven't read thoroughly yet but seems to have some conflicting change with https://github.com/streamr-dev/network-monorepo/pull/259/files regarding package-lock file references in github actions .yml file.

PR #268, opened 2021-11-01
client [NET-571]: Using the Brubeck client without providing a private key

teogeb:
Generate random Ethereum address for node if user has not authenticated with a private key.

### DIFF
--- a/packages/client/src/BrubeckNode.ts
+++ b/packages/client/src/BrubeckNode.ts
@@ -50,8 +50,7 @@ export default class BrubeckNode implements Context {
 
         // generate id if none supplied
         if (id == null || id === '') {
-            const address = await this.ethereum.getAddress()
-            id = `${address}#${uuid()}`
+            id = await this.generateId()
         }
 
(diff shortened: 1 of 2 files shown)


harbu (approved):

hpihkala:
A typical use case for 'anonymous' users is this:
- The client is used on a web page to subscribe to a public stream

Can we check: In this case, does the client currently go through the authentication cycle to the Core API?

Would be good if it didn't, as generating a new key for each page load will pollute the centralized database with throw-away users.

teogeb:
Currently that use doesn't generate any throw-away users. In this PR we generate an address for a network node, but the client itself remains a an unauthenticated client. With this setup, any operations that require authentication (e.g. some DU operations) throw an error if no authentication is provided.

The behavior may change if we implement https://linear.app/streamr/issue/NET-579/permissions-checks-when-subscribing-to-a-stream

hpihkala:
> Currently that use doesn't generate any throw-away users.

Ok great!

PR #269, opened 2021-11-02
Fix broken browser tests

juslesan:
- Replace Jest functions that do not work in the browser tests

### DIFF
--- a/packages/network/karma.config.js
+++ b/packages/network/karma.config.js
@@ -15,11 +15,12 @@ module.exports = function (config) {
           
         files: [
             './karma-setup.js',
+
             './test/browser/BrowserWebRtcConnection.test.ts',
             './test/browser/IntegrationBrowserWebRtcConnection.test.ts',
             './test/integration/**/!(NodeWebRtcConnection*|tracker-endpoints*|UnixSocketWsServer*).ts/',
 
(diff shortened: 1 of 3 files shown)


teogeb:
Would it be possible to modify the browser test setup so that it supports `jest-extended`? 

Currently we set the standard jest object to global namespace in `karma-setup.js`:
```
window.jest = jest
```

We also set some additional methods there, e.g.:
```
jest.advanceTimersByTime = timers.advanceTimersByTime
```

Maybe the `jest-extended` methods could be injected to global namespace similarly?

juslesan:
> Would it be possible to modify the browser test setup so that it supports `jest-extended`?
> 
> Currently we set the standard jest object to global namespace in `karma-setup.js`:
> 
> ```
> window.jest = jest
> ```
> 
> We also set some additional methods there, e.g.:
> 
> ```
> jest.advanceTimersByTime = timers.advanceTimersByTime
> ```
> 
> Maybe the `jest-extended` methods could be injected to global namespace similarly?

I've tried to add the `jest-extended` functions to the browser tests but importing the matchers seems to not be compatible with browser. Simply importing the package to `karma-setup.js` results in weird undefined errors that seem to be caused by webpack. @teogeb

timoxley (approved):
👍

teogeb (approved):
Yes, good to get the tests green. Let's integrate `jest-extended` in another PR (or add the similar helper methods manually).

PR #272, opened 2021-11-08
broker [NET-573]: MQTT plugin improvements

teogeb:
Fixes:

1) A message was delivered to a subscriber twice if it was published by a MQTT client. Now we store message chain ids of the message which are published via MQTT server. We don't re-deliver those messages as Aedes mirrors the messages to all subscribers. (Aedes doesn't have a feature for disabling mirroring. If it had, that would be an alternative way to implement the functionality.)

2) A message was delivered multiple times if there were multiple subscribers. Now we subscribe to `streamrClient` only once. Messages are delivered to the MQTT server and it propagates the messages  to all subscribers.

3) A subscription was unsubscribed if any of the clients unsubscribed the topic. Now we store list of `clientId`s and unsubscribes only if all clients have unsubscribed. 

4) Unsubscribes can't interfere other plugins. The MQTT plugin explictly unsubscribes from the  subscriptions it has created.

### DIFF
--- a/packages/broker/src/plugins/mqtt/Bridge.ts
+++ b/packages/broker/src/plugins/mqtt/Bridge.ts
@@ -1,9 +1,17 @@
 import { StreamrClient, Subscription } from 'streamr-client'
-import { Logger } from 'streamr-network'
+import { Logger, Protocol } from 'streamr-network'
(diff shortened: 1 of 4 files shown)


samt1803 (approved):
looks good!

PR #273, opened 2021-11-08
network: add support for jest-extended in browser tests

teogeb:
Added support for `jest-extended` methods (e.g  the`toIncludeSameMembers()` matcher) in browser tests.

The injection in `karma-setup.js` is a bit hack as we need to import an internal file from `jest-extended` library:
```
import jestExtendedMatchers from 'jest-extended/dist/matchers'
```

This PR also brings back the use of `jest-extended` in two network test cases (reverts most of PR https://github.com/streamr-dev/network-monorepo/pull/269).


juslesan (approved):
Good stuff!
It seems that some browser tests are a bit flaky in CI, might be worth taking a look at before merging.

PR #276, opened 2021-11-10
fix(client): unsubscribe from node

teogeb:
Fixed `SubscriptionSession` to call `node.unsubscribe` instead of `node.subscribe`.

The original change is from a commit two months ago: https://github.com/streamr-dev/network-monorepo/commit/cb1436ddddfdc70b51deed858fccc2183a00e488


timoxley (approved):
👍 Good one

PR #277, opened 2021-11-10
broker [NET-581]: Remove legacyPublishHttp and legacyMqtt plugins

teogeb:
Removed `legacyPublishHttp` and `legacyMqtt` plugins. 

Converted some plugin tests to use non-legacy plugins. Most legacy plugin tests were removed because the existing tests for non-legacy plugins covered the same functionality.


juslesan (approved):

samt1803 (approved):
👍

PR #278, opened 2021-11-10
Fix broker docker build so new images compatible with main branch can be pushed.

timoxley:
* CI is failing for all of the cross-client tests that involve the JS client because
* the network nodes are incompatible with the trackers run in docker because
* the Docker images for the broker haven't been updated in a month because
* the broker's CI docker image build fails because
* the client's `sqlite3` dependency is trying to run an npm `install` script during the `npm prune --production`, and that `install` script uses a development dependency `node-pre-gyp` which was removed during that same `npm prune --production`.

😵‍💫 

Not clear how/why this changed, possibly a change in the npm 6 line, or a dependency of npm 6?

### Changes

* Set the Dockerfile to use `--ignore-scripts` with `npm prune`. This seems to fix the issue.
* Set the broker CI config to build & tests images even if they're not tagged or on main branch. i.e. it should build & test the docker image for PRs now. This allows me to test the above change and should help catch a broken docker push step sooner in the future.

It's possible this issue goes away with the move to npm 8 and npm workspaces in #259. The change to run the docker build step for PRs will help.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
 FROM node:16-buster as build
 WORKDIR /usr/src/monorepo
 RUN npm set unsafe-perm true && \
-	# explicitly use npm v6
-	npm install -g npm@6 --prefer-offline
+	# explicitly use npm v8
+	npm install -g npm@8 --no-audit --progress=false
 COPY ["./*.json", "./*.js", "./*.mjs", ".npmrc",  ".gitignore", "./"]
 RUN npm ci
 COPY ["./packages", "./packages"]
@@ -12,7 +12,7 @@ RUN npm run bootstrap-pkg -- streamr-broker
 # --ignore-scripts as sqlite package in the client tries running its
 # 'install' script, which uses node-pre-gyp, which is a devDependency that
 # gets removed by prune.
-RUN npx lerna exec -- npm prune --production --ignore-scripts && \
+RUN npx lerna exec --parallel --include-dependencies --scope "streamr-broker" -- npm prune --production --ignore-scripts --prefer-offline --no-audit && \
 	# restore inter-package symlinks removed by npm prune
 	npx lerna link
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,12 +5,13 @@ RUN npm set unsafe-perm true && \
 	# explicitly use npm v6
 	npm install -g npm@6 && \
 	npm ci && \
-	npm run bootstrap-pkg streamr-broker && \
-	# image contains all packages, remove devDeps to keep image size down
-	# --ignore-scripts as sqlite package in the client tries running its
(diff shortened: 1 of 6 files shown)


kare (approved):
LGTM

Dockerfile
```diff
@@ -1,24 +1,23 @@
...
-RUN npx lerna exec -- npm prune --production --ignore-scripts && \
+RUN npx lerna exec --parallel --include-dependencies --scope "streamr-broker" -- npm prune --production --ignore-scripts --prefer-offline --no-audit && \
 	# restore inter-package symlinks removed by npm prune
 	npx lerna link
 
-FROM node:16-buster-slim
-RUN apt-get update && apt-get install --assume-yes --no-install-recommends curl \
-	&& apt-get clean \
-	&& rm -rf /var/lib/apt/lists/*
+FROM node:16-alpine
```
kare: Please use Debian based images for Docker containers. Currently `buster` or `buster-slim`.

Reasoning is to minimise download times when pulling containers from Hub. This is achieved by sharing Docker container layers. Currently we aim to share the base layer and maybe later we can share more.

Following command creates a single layer that can be shared by other containers:
```
RUN apt-get install -y curl
```

Dockerfile
```diff
@@ -31,4 +30,6 @@ EXPOSE 7170/tcp
 EXPOSE 7171/tcp
 
 WORKDIR /usr/src/monorepo/packages/broker
-CMD ./bin/broker.js # start broker from default config
+
+# start broker from default config
+CMD ["./bin/broker.js"]
```
kare, line 37:
```suggestion
ENTRYPOINT ["./bin/broker.js"]
```

I didn't try, but I think this should fix the issue where `init` is required.

kare (dismissed):

Dockerfile
timoxley: @kare good thought, though I believe this will break the ability to use this image to start other executables

e.g. trackers in the docker-compose https://github.com/streamr-dev/streamr-docker-dev/blob/3c55bab982ed65d30722e65b36a0b0649298f0ce/docker-compose.yml#L128

something like this in the testnet instructions:
```
docker run -it -v $(cd ~/.streamrDocker; pwd):/root/.streamr \
  streamr/broker-node:testnet bin/config-wizard

```

Dockerfile
timoxley: Ah good point, though the reason I switched to alpine is the image is only ~5MB vs ~25MB for buster-slim.

kare: Given that all Docker containers use Debian Buster base image I don't think the size of the base image is significant.

timoxley: Changed.

kare (approved):
LGTM

PR #280, opened 2021-11-16
ETH-187 fix sidechain address query

jtakalai:
The real fix is to use the safeGetDataUnion until v6 where it should be fixed (when it's ok to break interface)

Removed lots of "internal functions" that were only used in tests. There still is lots to remove, e.g. the whole Contracts.ts class.


jtakalai:
all data union tests pass, so I'm calling it good

timoxley (commented):
lgtm

PR #281, opened 2021-11-22
Fix getStats

jtakalai:


timoxley (approved):
lgtm, but still not sure about why branches: '**' needs to be in there

jtakalai:
Data union tests green, calling it good.

PR #284, opened 2021-11-23
Small cleanups to make lint and TS compiler happier

jtakalai:
Changes mostly in data union code, in how ethers is imported

### DIFF
--- a/packages/client/tsconfig.json
+++ b/packages/client/tsconfig.json
@@ -2,8 +2,7 @@
     "extends": "@streamr/dev-config/ts/tsconfig.browser.json",
     "compilerOptions": {
         "outDir": "dist",
-        "rootDirs": ["src", "test"],
-        "resolveJsonModule": true
+        "rootDirs": ["src", "test"]
     },
     "include": [
         "src/**/*",
--- a/packages/client/tsconfig.json
+++ b/packages/client/tsconfig.json
@@ -2,7 +2,8 @@
     "extends": "@streamr/dev-config/ts/tsconfig.browser.json",
     "compilerOptions": {
         "outDir": "dist",
-        "rootDirs": ["src", "test"]
+        "rootDirs": ["src", "test"],
+        "resolveJsonModule": true
     },
     "include": [
         "src/**/*",
--- a/.github/workflows/broker.yml
+++ b/.github/workflows/broker.yml
@@ -6,12 +6,13 @@ on:
       - "*"
     branches:
       - main
+      - v5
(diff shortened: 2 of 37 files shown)


packages/client/tsconfig.json
```diff
@@ -2,7 +2,8 @@
     "extends": "@streamr/dev-config/ts/tsconfig.browser.json",
     "compilerOptions": {
         "outDir": "dist",
-        "rootDirs": ["src", "test"]
+        "rootDirs": ["src", "test"],
+        "resolveJsonModule": true
```
timoxley: `resolveJsonModule` shouldn't be necessary, did you hit an issue? https://github.com/streamr-dev/network-monorepo/blob/193e6eac1e150cd7b3f1a20a7c86e814bcb12dc0/packages/dev-config/ts/tsconfig.browser.json#L22

timoxley (approved):
Great stuff

packages/client/tsconfig.json
kare: It's set in @streamr/dev-config: https://github.com/streamr-dev/network-monorepo/blob/main/packages/dev-config/ts/tsconfig.browser.json#L22

packages/client/tsconfig.json
```diff
@@ -2,7 +2,8 @@
     "extends": "@streamr/dev-config/ts/tsconfig.browser.json",
     "compilerOptions": {
         "outDir": "dist",
-        "rootDirs": ["src", "test"]
+        "rootDirs": ["src", "test"],
```
kare: @jtakalai @timoxley @harbu @teogeb
Any reason for not using these configuration settings?
```
    "incremental": true,                              /* Enable incremental compilation */
    "composite": true,                                /* Enable constraints that allow a TypeScript project to be used with project references. */
    "tsBuildInfoFile": ".tsbuildinfo",                /* Specify the folder for .tsbuildinfo incremental compilation files. */
```
Is the `.tsbuildinfo` cacheable in CI?

kare: I see `incremental` is set in the config, but it requires `composite = true`

packages/client/tsconfig.json
jtakalai: I think tests import contract jsons directly

jtakalai: I can delete it, too

packages/client/tsconfig.json
jtakalai: no idea

PR #285, opened 2021-11-23
fix: sidechain address should NOT come from factory

jtakalai:
factory contracts have changed, so DUs deployed with previous factories wouldn't work. Better just ask the DU contract directly.


jtakalai:
Data union test pass, I'm calling it good.

PR #286, opened 2021-11-24
Fix a bug in node-fetch replacement in browser

aseyfpour:
Following bug occurred in our product, this PR is a solution to solve this issue:

```
Error: Failed to publish to stream STREAM due to: TypeError: Failed to execute 'fetch' on 'Window': Illegal invocation
    at Object.<anonymous> (chrome-extension://jgcccciklffokknddmkbmkfdclnjajhp/background.js:133844:64)
    at Generator.next (<anonymous>)
    at asyncGeneratorStep (chrome-extension://jgcccciklffokknddmkbmkfdclnjajhp/background.js:106463:44)
    at _next (chrome-extension://jgcccciklffokknddmkbmkfdclnjajhp/background.js:106478:33)
    at chrome-extension://jgcccciklffokknddmkbmkfdclnjajhp/background.js:106483:29
    at new Promise (<anonymous>)
    at new Wrapper (chrome-extension://jgcccciklffokknddmkbmkfdclnjajhp/background.js:157801:44)
    at Object.<anonymous> (chrome-extension://jgcccciklffokknddmkbmkfdclnjajhp/background.js:106475:32)
    at Object.authFetch [as default] (chrome-extension://jgcccciklffokknddmkbmkfdclnjajhp/background.js:133827:39)
    at LoginEndpoints.<anonymous> (chrome-extension://jgcccciklffokknddmkbmkfdclnjajhp/background.js:133149:62).
Message was:
```

### DIFF
--- a/packages/client/src/shim/node-fetch.ts
+++ b/packages/client/src/shim/node-fetch.ts
@@ -1,7 +1,7 @@
 type FetchResponse = Response
 // In browsers, the node-fetch package is replaced with this to use native fetch
 export default (
-    ((typeof window !== 'undefined' && window.fetch.bind(window)) || (typeof fetch !== 'undefined' && fetch) || undefined)!
+    ((typeof window !== 'undefined' && typeof window.fetch !== 'undefined' && window.fetch.bind(window)) || (typeof fetch !== 'undefined' && fetch) || undefined)!
 )
 
 export { FetchResponse as Response }
--- a/packages/client/src/shim/node-fetch.ts
+++ b/packages/client/src/shim/node-fetch.ts
@@ -1,7 +1,7 @@
 type FetchResponse = Response
 // In browsers, the node-fetch package is replaced with this to use native fetch
 export default (
-    ((typeof fetch !== 'undefined' && fetch) || (typeof window !== 'undefined' && window.fetch) || undefined)!
+    ((typeof window !== 'undefined' && window.fetch.bind(window)) || (typeof fetch !== 'undefined' && fetch) || undefined)!
 )
 
 export { FetchResponse as Response }


packages/client/src/shim/node-fetch.ts
```diff
@@ -1,7 +1,7 @@
 type FetchResponse = Response
 // In browsers, the node-fetch package is replaced with this to use native fetch
 export default (
-    ((typeof fetch !== 'undefined' && fetch) || (typeof window !== 'undefined' && window.fetch) || undefined)!
+    ((typeof window !== 'undefined' && window.fetch.bind(window)) || (typeof fetch !== 'undefined' && fetch) || undefined)!
```
mondoreale: Is the order change necessary? I'd imagine this should work just fine and make for a smaller diff?
```js
((typeof fetch !== 'undefined' && fetch) || (typeof window !== 'undefined' && window.fetch.bind(window)) || undefined)!
```
We may also want to proactively cover a situation in which `window.fetch` is `undefined`, meaning `window.fetch.bind` will explode if that happens.

Other than that it's a very good change!

aseyfpour: Unfortunately, it is necessary, in our case both fetch and window.fetch had values and if I didn't change the orders same error will raise.

I'm not sure that scenario will happen but I'll add it in a second.

mondoreale: Ah, isn't `fetch` equal to `window.fetch` in the browser? That'd make sense and changing the order is the right thing to do, in the end.

About that extra scenario, right, most `window` objects (96%+) will have `fetch` on them. Hm, I'm fine with both (with and without the extra check).

mondoreale (approved):
👍🏻

packages/client/src/shim/node-fetch.ts
aseyfpour: > Ah, isn't `fetch` equal to `window.fetch` in the browser?

They are the same in the browser. If I want to keep the order (Like below) I had to add a new check in the first condition to be sure window is not undefined. So for keeping the conditions simple I changed the orders. 

```
((typeof fetch !== 'undefined' && typeof window !== 'undefined' && fetch.bind(window)) || (typeof window !== 'undefined' && window.fetch.bind(window)) || undefined)!
```

timoxley (approved):
Looks good

aseyfpour:
What should I do now?

hpihkala:
@aseyfpour Thanks a lot for the contribution! 👍  Next, someone on our side needs to merge this, perhaps after checking what's up with the failing tests. Looks like they might be unrelated.

You are probably using the 5.x client version at Swash, which is the latest official one (current main branch contains upcoming Brubeck milestone stuff which will be 6.x). 

@timoxley or @mondoreale could you please cherry pick this fix to the `v5` branch too, and publish a new patch release of that?

aseyfpour:
Yeah, we are using 5.x version in Swash.

Thanks in advance.

PR #287, opened 2021-11-25
Net-582: Allow one-way connections on the network layer

juslesan:
- Nodes can form publish only stream connections to other nodes that allow one way connections.
- The "subscribing" nodes must be pre-subscribed to the stream to ensure that a tracker connection exists.
- One-way stream connections are created outside of that tracker algorithm and the trackers are only used for WebRTC signalling messages.
- New Protocol layer messages for establishing the one-way stream connections using P2P connections.
- SenderId included to protocol messages to make relaying possible in the future. A future relaying solution using DHTs etc, would make it possible for inOnly nodes to be pre-subscriber to streams. 
- Add a broker plugin to force nodes to subscribe to a given list of streams.

Client: #289


timoxley (approved):
lgtm

PR #289, opened 2021-11-29
Net-582: Client-side

juslesan:
- Fixes unsubscribe bug in one way connections
- New functions in the Client API for openening and closing proxied publish only stream connections

### DIFF
--- a/packages/client/src/BrubeckNode.ts
+++ b/packages/client/src/BrubeckNode.ts
@@ -6,7 +6,7 @@ import { NetworkNodeOptions, createNetworkNode, NetworkNode } from 'streamr-netw
 import { pOnce, uuid, instanceId } from './utils'
 import { Context } from './utils/Context'
 import { Config } from './Config'
-import { StreamMessage } from 'streamr-client-protocol'
+import { StreamMessage, SPID } from 'streamr-client-protocol'
 import { DestroySignal } from './DestroySignal'
 import Ethereum from './Ethereum'
 
@@ -166,29 +166,29 @@ export default class BrubeckNode implements Context {
         }
     }
 
-    openPublishProxyConnectionOnStream(streamId: string, streamPartition: number | undefined, nodeId: string): void | Promise<void> {
-        const partition = streamPartition || 0
+    openPublishProxyConnectionOnStreamPartition(spid: SPID, nodeId: string): void | Promise<void> {
+        const { streamId, streamPartition } = spid
         try {
             if (!this.cachedNode || !this.startNodeComplete) {
                 return this.startNode().then((node) => {
-                    return node.joinStreamAsPurePublisher(streamId, partition, nodeId)
+                    return node.joinStreamAsPurePublisher(streamId, streamPartition, nodeId)
                 })
             }
-            return this.cachedNode.joinStreamAsPurePublisher(streamId, partition, nodeId)
+            return this.cachedNode.joinStreamAsPurePublisher(streamId, streamPartition, nodeId)
         } finally {
-            this.debug('openProxyConnectionOnStream << %o', streamId, partition, nodeId)
+            this.debug('openProxyConnectionOnStream << %o', streamId, streamPartition, nodeId)
         }
     }
 
-    closePublishProxyConnectionOnStream(streamId: string, streamPartition: number | undefined, nodeId: string): void {
-        const partition = streamPartition || 0
+    closePublishProxyConnectionOnStreamPartition(spid: SPID, nodeId: string): void {
+        const { streamId, streamPartition } = spid
         try {
             if (!this.cachedNode || !this.startNodeComplete) {
                 return
             }
-            this.cachedNode.leavePurePublishingStream(streamId, partition, nodeId)
+            this.cachedNode.leavePurePublishingStream(streamId, streamPartition, nodeId)
         } finally {
-            this.debug('closeProxyConnectionOnStream << %o', streamId, partition, nodeId)
+            this.debug('closeProxyConnectionOnStream << %o', streamId, streamPartition, nodeId)
         }
     }
 }
--- a/packages/client/src/BrubeckNode.ts
+++ b/packages/client/src/BrubeckNode.ts
@@ -165,4 +165,30 @@ export default class BrubeckNode implements Context {
             this.debug('publishToNode << %o', streamMessage.getMessageID())
         }
     }
+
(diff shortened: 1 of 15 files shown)


packages/client/test/integration/ProxyPublishing.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    })
+
+    it('Publish only connections work', async () => {
+        const receivedMessagesProxy: any[] = []
+        await proxyClient1.subscribe(stream, (msg) => {
+            receivedMessagesProxy.push(msg)
+        })
+        await wait(1000)
+        await publishingClient.setPublishProxy(stream, proxyNodeId1)
+        publishingClient.publish(stream, {
```
timoxley:
```suggestion
        await publishingClient.publish(stream, {
```
All these publish calls should probably be awaited

packages/client/test/integration/ProxyPublishing.test.ts
```diff
@@ -0,0 +1,160 @@
...
+        await stream.grantPermission(StreamOperation.STREAM_PUBLISH, proxyUser2.username)
+        await stream.grantPermission(StreamOperation.STREAM_SUBSCRIBE, proxyUser2.username)
+    })
+
+    it('Publish only connections work', async () => {
+        const receivedMessagesProxy: any[] = []
+        await proxyClient1.subscribe(stream, (msg) => {
+            receivedMessagesProxy.push(msg)
+        })
+        await wait(1000)
```
timoxley, line 74: Is there a way we can avoid waits in the tests?

packages/client/test/unit/WebStreamToNodeStream.test.ts
```diff
@@ -39,7 +42,9 @@ describe('WebStreamToNodeStream', () => {
     })
 
     it('can work with small buffer', async () => {
+        // eslint-disable-next-line import/no-unresolved
```
timoxley, line 45: 👍

packages/client/src/Publisher.ts
```diff
@@ -201,6 +201,30 @@ export default class BrubeckPublisher implements Context, Stoppable {
...
+    }
+
+    async removePublishProxy(streamObjectOrId: SIDLike, nodeId: string) {
+        const { streamId, streamPartition } = SPID.parse(streamObjectOrId)
+        await this.node.closePublishProxyConnectionOnStream(streamId, streamPartition, nodeId)
+    }
+
+    async setPublishProxies(streamObjectOrId: SIDLike, nodeIds: string[]) {
+        const { streamId, streamPartition } = SPID.parse(streamObjectOrId)
+        nodeIds.forEach(async (nodeId) => {
```
timoxley: Generally not great to put `await` in `forEach`, it executes tasks in parallel and in the background i.e. doesn't wait for these to finish before  `setPublishProxies` resolves, and any errors are uncatchable.

I'd recommend Promise.allSettled + Promise.all. 
`allSettled` runs tasks in parallel but waits for all to complete before resolving. This ensures you don't have tasks executing in the background if one task fails.

```js
const tasks = nodeIds.map(async (nodeId) => await this.node.openPublishProxyConnectionOnStream(streamId, streamPartition, nodeId))
await Promise.allSettled(tasks) // wait for all tasks to succeed or fail
await Promise.all(tasks) // fail if any one task failed
```

Alternatively you could do these in series using a control loop e.g. `for…of`:
```js
for (const nodeId of nodeIds) {
  await this.node.openPublishProxyConnectionOnStream(streamId, streamPartition, nodeId))
}
```

packages/client/test/integration/ProxyPublishing.test.ts
```diff
@@ -0,0 +1,160 @@
...
+        })
+        proxyNodeId1 = await proxyClient1.node.getNodeId()
+        proxyNodeId2 = await proxyClient2.node.getNodeId()
+        stream = await createTestStream(publishingClient, module)
+        const pubUser = await publishingClient.getUserInfo()
+        const proxyUser = await proxyClient1.getUserInfo()
+        const proxyUser2 = await proxyClient2.getUserInfo()
+
+        await stream.grantPermission(StreamOperation.STREAM_GET, pubUser.username)
+        await stream.grantPermission(StreamOperation.STREAM_PUBLISH, pubUser.username)
```
timoxley: There's a new client api that allows you to set multiple permissions at once:

```suggestion
        await stream.grantPermissions([StreamOperation.STREAM_PUBLISH, StreamOperation.STREAM_PUBLISH] pubUser.username)
```

timoxley (changes requested):
Pretty clean, looks good. Just needs the `forEach`/`await` thing fixed, if it's intentional perhaps a comment.

packages/client/test/integration/ProxyPublishing.test.ts
juslesan: Probably by passing events from the network to the client

packages/client/src/BrubeckNode.ts
```diff
@@ -165,4 +165,30 @@ export default class BrubeckNode implements Context {
             this.debug('publishToNode << %o', streamMessage.getMessageID())
         }
     }
+
+    openPublishProxyConnectionOnStream(streamId: string, streamPartition: number | undefined, nodeId: string): void | Promise<void> {
```
teogeb: The public methods could use a `SPID` instance to wrap `streamId`+`streamPartittion`. And rename to `openPublishProxyConnectionOnStreamPartition` as it always targets only one stream partition, not the whole stream.

Similar fix also for line 183, and for new methods in `Publisher`

packages/client/src/BrubeckNode.ts
```diff
@@ -165,4 +165,30 @@ export default class BrubeckNode implements Context {
...
+        const partition = streamPartition || 0
+        try {
+            if (!this.cachedNode || !this.startNodeComplete) {
+                return this.startNode().then((node) => {
+                    return node.joinStreamAsPurePublisher(streamId, partition, nodeId)
+                })
+            }
+            return this.cachedNode.joinStreamAsPurePublisher(streamId, partition, nodeId)
```
teogeb:
```suggestion
            await this.cachedNode.joinStreamAsPurePublisher(streamId, partition, nodeId)
```

juslesan: The functions in the BrubeckNode class have a pattern of being able to be ran synchronously or asynchronously. This is why the function in question is returned instead of awaited.

juslesan:
Merging now to start work on NET-588

PR #290, opened 2021-11-29
Node 16

timoxley:
Removes `node-webcrypto-ossl` from the client.
Tests all packages on node 16 and 17 instead of just node 16 in CI.


juslesan (approved):

packages/client/webpack.config.js
```diff
@@ -113,7 +113,6 @@ module.exports = (env, argv) => {
                 crypto: require.resolve('crypto-browserify'),
                 buffer: require.resolve('buffer/'),
                 'node-fetch': path.resolve('./src/shim/node-fetch.ts'),
```
juslesan, line 115: Are the 'apostrophies' required here?

timoxley:
```
> { node-fetch: path.resolve('./src/shim/node-fetch.ts') }
({ node-fetch: path.resolve('./src/shim/node-fetch.ts') })
       ^

Uncaught SyntaxError: Unexpected token '-'
```

packages/dev-config/ts/tsconfig.node.json
```diff
@@ -1,7 +1,7 @@
 {
     "$schema": "https://json.schemastore.org/tsconfig",
     "display": "Node 14 – Streamr",
```
teogeb:
```suggestion
    "display": "Node 16 – Streamr",
```

teogeb (approved):
👍

PR #295, opened 2021-11-30
broker [NET-268]: Sequence numbers in storage queries

teogeb:
Fix for storage node queries. The `sequenceNumber` parameter was ignored for these queries:
- resend range, when publisher(+chain) was not specified
- resend from, when publisher was not specified

Previously both queries had separate implementations for:
- query with a publisher parameter
- query without a publisher parameter

The implementation with a publisher parameter didn't have the bug. Therefore we now use the unified implementation, which is based on the non-buggy implementation. It is only modified to support the optionality of the publisher parameter.

Also updated some method parameters to use `undefined` instead of `null`.


juslesan (approved):
LGTM!

PR #303, opened 2021-12-08
Net-588: reliability

juslesan:
- If a proxy connection is lost the node behind the proxy will attempt to reconnect as long as the proxy connection is wanted


packages/network/src/logic/node/Node.ts
```diff
@@ -202,8 +214,8 @@ export class Node extends EventEmitter {
             if (sendStatus) {
                 this.trackerManager.sendStreamStatus(spid)
             }
-        } else if (this.streams.isSetUp(spid) && this.streams.isOneDirectional(spid)) {
-            logger.trace(`Could not join stream ${spid.key} as stream is set as one-directional`)
+        } else if (this.streams.isSetUp(spid) && this.streams.isBehindProxy(spid)) {
```
harbu, line 218:
```suggestion
        } else if (this.streams.isBehindProxy(spid)) {
```
The first conditional expression is unnecessary as we are in the `else` clause therefore we know that `this.streams.isSetUp(spid)` must be true.

packages/network/src/logic/node/Node.ts
```diff
@@ -362,7 +259,7 @@ export class Node extends EventEmitter {
             streamMessage.getStreamPartition()
         )
         // Check if the stream is set as one-directional and has inbound connection
-        if (source && this.streams.isSetUp(spid) && this.streams.isOneDirectional(spid) && !this.streams.hasInboundConnection(spid, source)) {
+        if (source && this.streams.isSetUp(spid) && this.streams.isBehindProxy(spid) && !this.streams.hasInboundConnection(spid, source)) {
             // Perhaps the node should be disconnected here if bad behaviour is repeated
```
harbu, line 265: If this is unintended behavior it would be good to `log.warn` it so it can be spotted if it were to happen in practice.

packages/network/src/logic/node/StreamManager.ts
```diff
@@ -22,7 +22,7 @@ function keyForDetector({ publisherId, msgChainId }: MessageLayer.MessageID) {
 export class StreamManager {
     private readonly streams: Map<SPIDKey,StreamState> = new Map<SPIDKey,StreamState>()
 
-    setUpStream(spid: SPID, oneDirectional = false): void {
+    setUpStream(spid: SPID, isBehindProxy = false): void {
```
harbu, line 25: Given that the value of `isBehindProxy` is rather consequential and that this is internal logic, I would advocate not using a default value here. But it's a matter of taste, so I'll leave it up to you.
```suggestion
    setUpStream(spid: SPID, isBehindProxy: boolean): void {
```

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
```diff
@@ -0,0 +1,211 @@
...
+        this.trackerManager = opts.trackerManager
+        this.streamManager = opts.streamManager
+        this.nodeToNode = opts.nodeToNode
+        this.node = opts.node
+        this.nodeConnectTimeout = opts.nodeConnectTimeout
+        this.acceptProxyConnections = opts.acceptProxyConnections
+        this.connections = new Map()
+    }
+
+    addConnection(spid: SPID, nodeId: NodeId): void {
```
harbu: Using stricter access modifiers when possible enhances the reader's ability to reason about the code. (E.g. when used here, I know that this method is only invoked within this class.)
```suggestion
    private addConnection(spid: SPID, nodeId: NodeId): void {
```

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
```diff
@@ -0,0 +1,211 @@
...
+        }
+        this.connections.get(spid.key)!.set(nodeId, {
+            state: State.CONNECTING
+        })
+    }
+
+    removeConnection(spid: SPID, nodeId: NodeId): void {
+        if (this.connections.has(spid.key)) {
+            this.connections.get(spid.key)!.delete(nodeId)
+            if ([...this.connections.get(spid.key)!].length === 0) {
```
harbu: Wouldn't this work?
```suggestion
            if (this.connections.get(spid.key)!.size === 0) {
```

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
```diff
@@ -0,0 +1,211 @@
...
+    addConnection(spid: SPID, nodeId: NodeId): void {
+        if (!this.connections.has(spid.key)) {
+            this.connections.set(spid.key, new Map())
+        }
+        this.connections.get(spid.key)!.set(nodeId, {
+            state: State.CONNECTING
+        })
+    }
+
+    removeConnection(spid: SPID, nodeId: NodeId): void {
```
harbu:
```suggestion
    private removeConnection(spid: SPID, nodeId: NodeId): void {
```

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
```diff
@@ -0,0 +1,211 @@
...
+        if (this.connections.has(spid.key)) {
+            this.connections.get(spid.key)!.delete(nodeId)
+            if ([...this.connections.get(spid.key)!].length === 0) {
+                this.connections.delete(spid.key)
+            }
+        }
+
+        this.streamManager.removeNodeFromStream(spid, nodeId)
+        if (this.streamManager.isSetUp(spid)
```
harbu: Condition `this.streamManager.isSetUp(spid)` is not strictly necessary here (if it were `false` the method invocation `removeNodeFromStream` would've thrown above).

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
```diff
@@ -0,0 +1,211 @@
...
+        return this.connections.get(spid.key)!.get(nodeId)!
+    }
+
+    async openOutgoingStreamConnection(spid: SPID, targetNodeId: string): Promise<void> {
+        const trackerId = this.trackerManager.getTrackerId(spid)
+        const trackerAddress = this.trackerManager.getTrackerAddress(spid)
+        try {
+            if (!this.streamManager.isSetUp(spid)) {
+                this.streamManager.setUpStream(spid, true)
+            } else if (this.streamManager.isSetUp(spid) && !this.streamManager.isBehindProxy(spid)) {
```
harbu:
```suggestion
            } else if (!this.streamManager.isBehindProxy(spid)) {
```

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
```diff
@@ -0,0 +1,211 @@
...
+                const reason = `Could not open a proxy outgoing stream connection ${spid.key}, bidirectional stream already exists`
+                logger.warn(reason)
+                this.node.emit(Event.PUBLISH_STREAM_REJECTED, targetNodeId, spid, reason)
+                return
+            } else if (this.streamManager.isSetUp(spid) && this.streamManager.hasOutOnlyConnection(spid, targetNodeId)) {
```
harbu:
```suggestion
            } else if (this.streamManager.hasOutOnlyConnection(spid, targetNodeId)) {
```

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
```diff
@@ -0,0 +1,211 @@
...
+                const reason = `Could not open a proxy outgoing stream connection ${spid.key}, proxy stream connection already exists`
+                logger.warn(reason)
+                this.node.emit(Event.PUBLISH_STREAM_REJECTED, targetNodeId, spid, reason)
+                return
+            } else if (this.streamManager.isSetUp(spid) && this.hasConnection(targetNodeId, spid)) {
```
harbu:
```suggestion
            } else if (this.hasConnection(targetNodeId, spid)) {
```

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
harbu: What is this condition about actually? Is `this.streamManager.hasOutOnlyConnection(spid, targetNodeId)` true if we have already formed a outoing stream connection (i.e. proxy stream) and are re-trying?

Also how is this different from the below case, i.e., `this.hasConnection(targetNodeId, spid)`?

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
```diff
@@ -0,0 +1,211 @@
...
+        const spid = new SPID(streamId, streamPartition)
+        if (this.streamManager.isSetUp(spid) && this.streamManager.hasInOnlyConnection(spid, nodeId)) {
+            this.removeConnection(spid, nodeId)
+            this.node.emit(Event.ONE_WAY_CONNECTION_CLOSED, nodeId, spid)
+        }
+        if (this.streamManager.isSetUp(spid) && this.streamManager.hasOutOnlyConnection(spid, nodeId)) {
+            this.removeConnection(spid, nodeId)
+            this.node.emit(Event.ONE_WAY_CONNECTION_CLOSED, nodeId, spid)
+            logger.info(`Proxy node ${nodeId} closed one-way stream connection for ${spid}`)
+        }
```
harbu, line 159: Does this code get executed in practice? Looks like method `processLeaveRequest` gets invoked only when we receive a message from the proxy node that they are leaving. But in that situation, won't they always be an "inOnly" node from our perspective (us = being the contact node)?

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
```diff
@@ -0,0 +1,211 @@
...
+            this.node.emit(Event.PUBLISH_STREAM_REJECTED, targetNodeId, spid, err)
+        } finally {
+            this.trackerManager.disconnectFromSignallingOnlyTracker(trackerId)
+        }
+    }
+
+    async closeOutgoingStreamConnection(spid: SPID, targetNodeId: NodeId): Promise<void> {
+        if (this.streamManager.isSetUp(spid) && this.streamManager.hasOutOnlyConnection(spid, targetNodeId)) {
+            clearTimeout(this.getConnection(targetNodeId, spid)!.reconnectionTimer!)
+            await this.nodeToNode.leaveStreamOnNode(targetNodeId, spid)
```
harbu: This can `throw` if the connection is dead. I wonder if we should still proceed with cleanup in that situation?

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
timoxley:
```suggestion
            if (this.connections.get(spid.key)!.size === 0) {
```

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
timoxley: A comment for the `if` condition here might be worthwhile

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
```diff
@@ -0,0 +1,211 @@
...
+    async processPublishStreamRequest(message: PublishStreamConnectionRequest, nodeId: string): Promise<void> {
+        const { streamId, streamPartition } = message
+        const spid = new SPID(streamId, streamPartition)
+
+        // More conditions could be added here, ie. a list of acceptable ids or max limit for number of one-way this
+        const isAccepted = this.streamManager.isSetUp(spid) && this.acceptProxyConnections
+        if (isAccepted) {
+            this.streamManager.addInOnlyNeighbor(spid, nodeId)
+        }
+        return await this.nodeToNode.respondToPublishOnlyStreamConnectionRequest(nodeId, spid, isAccepted)
```
timoxley: generally `return` or `await`, but not both
```suggestion
        await this.nodeToNode.respondToPublishOnlyStreamConnectionRequest(nodeId, spid, isAccepted)
```

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
timoxley: I wonder if this should/could `this.removeConnection(spid, targetNodeId)` before the `leaveStreamOnNode`?

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    stop(): void {
+        this.connections.forEach((stream: Map<NodeId, ProxyConnection>) => {
+            stream.forEach((connection: ProxyConnection) => {
+                if (connection.reconnectionTimer !== undefined) {
+                    clearTimeout(connection.reconnectionTimer)
+                }
+            })
+        })
```
timoxley, line 217: should `stop` perhaps delete the contents of these arrays too?

timoxley (approved):
A few queries but otherwise lgtm

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
teogeb: As we check `!this.streamManager.isSetUp(spid)`at line 96, is there need to recheck the opposite case at line 98, 103 and 108? 

Also the method could be easier to read if we'd move lines 113-116 after line 97, and then remove the redundant `return` statements.

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
```diff
@@ -0,0 +1,211 @@
...
+        if (connection.state !== State.RECONNECTING) {
+            connection.state = State.RECONNECTING
+        }
+        const trackerId = this.trackerManager.getTrackerId(spid)
+        const trackerAddress = this.trackerManager.getTrackerAddress(spid)
+        try {
+            await this.trackerManager.connectToSignallingOnlyTracker(trackerId, trackerAddress)
+            await promiseTimeout(this.nodeConnectTimeout, this.nodeToNode.connectToNode(targetNodeId, trackerId, false))
+            await this.nodeToNode.requestPublishOnlyStreamConnection(targetNodeId, spid)
+            logger.trace(`Successful proxy stream reconnection to ${targetNodeId}`)
```
harbu: Some duplicated logic with `openOutgoingStreamConnection`. Generally when you have connection and re-connection logic, you can collapse the two into one thing.

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
```diff
@@ -0,0 +1,211 @@
...
+    node: Node,
+    nodeConnectTimeout: number,
+    acceptProxyConnections: boolean
+}
+
+enum State {
+    CONNECTING,
+    CONNECTED,
+    RECONNECTING
+}
```
harbu: I would be a bit careful naming the states like this. As we've previously noticed, we don't really know if a connection is connected or not until we attempt to use it. Maybe domain-level terms would be better e.g. `NEGOTIATING`, `ACCEPTED` . Also generally `CONNECTING` and `RECONNECTING` can be just collapsed to one state.

Also I don't see that `state` would be used much here, just for bookkeeping? We mostly write the value but don't read it anywhere?

harbu (changes requested):
good stuff; see my comments

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
juslesan: Line 96 there to create a stream if necessary, it is not use specifically to enter lines 113-116 expect when a stream does not exist. Rest of the time when the function is called (if the stream already exists) it will have to pass the cases on lines 98, 103 and 108 before starting to create the proxy connection

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
juslesan: At the moment, the only state that is really used is the reconnecting state

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
juslesan: This will not be executed in practice currently, so maybe its not required. It is there so that the proxy contact nodes would be able to remove proxied nodes if necessary

harbu (approved):

PR #305, opened 2021-12-08
client: [NET-594]: validate client options

teogeb:
Validate the options JSON given by the user, and throw an error if there are invalid settings. 

Check that there are no missing required properties, and that the JSON doesn't contain any unknown properties (most likely typos). Also validate the content of each field if the content can be validated easily (e.g. an Ethereum address or an URL). The validation is implemented using `ajv` JSON schema validator.

Open questions:

- The validation requires that the `auth.privateKey` is a string and it validates the content with a regexp. In theory the client supported also `BytesLike` format, but that is an undocumented feature. In practice the privateKey was always a string. If `BytesLike` format needs to be supported, the validation would be a lot of weaker and it wouldn't catch many potential misconfigurations. 

- The `tracker.id` is currently annotated as a generic string. If we modify some Broker tests, It could be annotated as an `ethereum-address`. The tests use tracker ids like `tracker-1`.


timoxley (approved):
I think private key as a string is fine for now.
nice work, looks good 👍

timoxley:
> The tracker.id is currently annotated as a generic string. If we modify some Broker tests, It could be annotated as an ethereum-address. The tests use tracker ids like tracker-1.

This is leftover from earlier when tracker ids were just strings. The tests should probably change to use addresses and the ids should probably be validated. See if it's not too much work.

PR #306, opened 2021-12-09
feat(broker, client): Tracker id is an Ethereum address

teogeb:
In Broker and Client the Tracker id is always an Ethereum addresses. 

The network layer implementation has not been modified as we want to keep the network layer chain agnostic. The `startTracker` still generates an UUID for trackerId if no id is given in `startTracker` call. In Broker layer that default is always overriden (`broker/bin/tracker.js:35`).

The only user-visible functionality change in this PR is that the Client validates the Tracker id when a list of Trackers is provided in StreamrClient options (`config.schema.json`)

Other modifications:

- Simplified test tracker initialization in Broker tests: new helper method `startTestTracker(port)` generates a Tracker id from the port value.

- Network tests to use generated tracker ids (`tracker.getTrackerId()`) instead of hardcoded ids (e.g. 'tracker'). 

- Fixed import paths for `network/bin/*.js` utilities

- Removed redundant fallbacks in `broker/bin/tracker.js`


juslesan (commented):
In my opinion it would be best to keep the network layer "chain agnostic" for now. So there is no need to import Ethers or use Ethereum addresses in the trackers by default. 

The chain specifics should be configured to the network from the broker layer. This includes the Ethereum addresses of the trackers, which are brought in the the network by starting the tracker from brokers `bin/tracker`.

teogeb:
> In my opinion it would be best to keep the network layer "chain agnostic" for now. 

@juslesan Modified the PR so that network layer is kept chain agnostic

juslesan (approved):

PR #309, opened 2021-12-10
broker: [NET-593]: all StreamrClient options are configurable

teogeb:
The broker configuration file is restructured. All client options are moved to `client` block. That block can contain most of the existing configuration options. It is now possible to have any `StreamrClient` configuration option in that block: all options are passed to the client constructor when we create a `StreamrClient` instance.

ethereumPrivateKey -> client.auth.privateKey
- In Broker config this is a required value as we need to generate node's Ethereum address from this value.

network.name -> client.network.name
- Broker's network.name was a required parameter, and in client it is an optional parameter.

network.location -> client.network.location
- The Location object is now an optional parameter of `AbstractNodeOptions`, similarly to other fields of that object. No need to set an undefined location to `null`. The same applies to the fields of a Location object.

streamrUrl -> client.restUrl
- The content of the restUrl parameter is `${config.streamrUrl}/api/v1`. The `broker.js` executable had an optional command line argument `streamrUrl`. That is now renamed to `restUrl`, and the content also changed similarly.

streamrAddress -> plugins.storage.storageConfig.streamrAddress
- Removed the `streamrAddress` command line argument from `broker.js` executable (no reason to configure one plugin-specific option with a command line argument).

network.trackers -> client.network.trackers
- Broker supports two formats of tracker definition: a `NetworkSmartContract` and an array. If a contract is provided, Broker will query it and pass an array to `StreamrClient` (which doesn't currently support the contract format)

network.stun, network.turn -> client.network.stunUrls
- The client format is different from Broker format. The `stunUrls` is an array of strings. Use this algorithm to convert a `turn` definition to `stunUrls`: https://github.com/streamr-dev/network-monorepo/blob/004eee1698db96675fb76397abda82eda9245624/packages/broker/src/broker.ts#L61

storageNodeConfig.registry -> client.storageNodeRegistry

network.webrtcDisallowPrivateAddresses -> client.network.webrtcDisallowPrivateAddresses

network.acceptProxyConnections -> client.network.acceptProxyConnections

removed generateSessionId
- If the network. id should not be suffixed without sessionId, provide an explicit value for client.network.id property. The value should be the address of client.auth.privateKey. Note: if `storage` plugin was enabled, the sessionId generation was disabled and this setting was ignored.

removed support for client.network.metricsContext parameter in client, added `getMetricsContext()` method to `Node` class
- The MetricsContext is created in BrubeckNode and it uses network.name or network.id as the peerId. Note: as network.name is an optional parameter, the peerId may contain sessionId as part of the peerId. Therefore if deterministic id is needed, name or id should be provided in the client options. 

Also removed the migration script which converts `testnet1` config to `testnet2` config (`ConfigAutoMigrate.ts`)

### Future improvements
- When BrubeckNode calls `createNetworkNode`, it doesn't pass network.name as a parameter. Instead it passes the id. Should it use the network.name, if user has specified that in the client options?
- If StreamrClient starts to support the `NetworkSmartContract` format, we can remove the configuration transformation from Broker.
- ConfigWizard, development + docker configs and Broker integration tests currently use explicit network.id value as these uses cases were previously configured with `generateSessionId: false`. Maybe we should start to use sessionIds if there is no need to keep the non-standard behavior (the default setting of the removed `generateSessionId` was `true`). Before that we need to add sessionIds support to TestnetMinerBackend. Storage node configurations (`development-1.env.json`, `docker-1.env.json`) need explicit ids as storage nodes don't support sessionIds.
- We could use different fallback for `MetricsContext`, e.g. the nodeAddress without the sessionId. Then the `peerId` would always be deterministic. (The peerId is used in `VolumeEndpoint` and `ConsoleAndPM2Metrics`.)
- Maybe easier way to access `MetricContext`? E.g. `plugin.getMetricsContext() `


packages/broker/src/helpers/config.schema.json
```diff
@@ -1,314 +1,112 @@
...
+            "additionalProperties": true,
+            "properties": {
+                "auth": {
+                    "type": "object",
+                    "properties": {
+                        "privateKey": {
+                            "type": "string",
+                            "pattern": "^(0x)?[a-fA-F0-9]{64}$"
+                        }
+                    }
```
harbu, line 24: I assume the rest are validated within client?

packages/broker/src/plugins/legacyWebsocket/WebsocketPlugin.ts
```diff
@@ -43,12 +45,11 @@ export class WebsocketPlugin extends Plugin<WebsocketPluginConfig> {
         this.websocketServer = new WebsocketServer(
             httpServer,
             this.networkNode,
-            new StreamFetcher(this.brokerConfig.streamrUrl),
+            new StreamFetcher(this.getRestUrl()),
             this.publisher,
-            this.metricsContext,
+            (await (this.streamrClient!.getNode())).getMetricsContext(),
```
harbu, line 50: I noticed this same line before
```ts
(await (this.client.getNode())).getMetricsContext()
```
and it seems a bit cumbersome to get the metrics context. I'm wondering whether it would be possible to have the `metricsContext` already "injected" into the plugin base class instance as a field before `start`?

harbu (approved):

packages/broker/src/helpers/config.schema.json
teogeb: Yes. This is to indicate that the `privateKey` is a required property when we configure the client from Broker. It is optional in Client.

packages/broker/src/plugins/legacyWebsocket/WebsocketPlugin.ts
teogeb: True, added that improvement to "Future improvements" list

PR #310, opened 2021-12-14
fix: adding theGraph to the test environment in ci

samt1803:
- adding theGraph to the test environment in ci
- fixed protocol integrationtests


samt1803:
network integration tests passing now: https://github.com/streamr-dev/network-monorepo/runs/4523022446?check_suite_focus=true

and in client and broker integrationtests the graph can be called, there are now other things failing the test

timoxley (approved):
lgtm

PR #311, opened 2021-12-14
fix(client): null query parameters

teogeb:
Fix `Rest.getUrl` to support `null` / `undefined` query parameter values. The new helper method is also used in `Resends`, no functionality change there.

### DIFF
--- a/packages/client/src/Rest.ts
+++ b/packages/client/src/Rest.ts
@@ -2,6 +2,8 @@
  * More ergonomic wrapper around fetch/authFetch
  */
 import { Lifecycle, scoped, inject, DependencyContainer } from 'tsyringe'
+import omitBy from 'lodash/omitBy'
+import isNil from 'lodash/isNil'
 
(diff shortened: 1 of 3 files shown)


packages/client/src/Rest.ts
```diff
@@ -30,6 +32,10 @@ function serialize(body: any): string | undefined {
     return typeof body === 'string' ? body : JSON.stringify(body)
 }
 
+export const createQueryString = (query: Record<string, any>) => {
+    return new URLSearchParams(omitBy(query, isNil)).toString()
```
timoxley: Using `lodash/omitBy` here likely adds bloat to the browser bundle, and the vanilla version is straightforward enough:

```js
Object.fromEntries(Object.entries(query).filter(([_k, v]) => v != null))
```

timoxley (approved):
I'd avoid adding more lodash into the bundle unless absolutely necessary, but otherwise lgtm 👍

PR #319, opened 2021-12-16
fix: fixed err with storagenode being not synced to theGraph

samt1803:
fix: fixed err with storagenode being not synced to theGraph when accessing its streams;

also **removed detectOpenHandles for integration tests**, since it forces jest to run all tests in band instead of parallel

### DIFF
--- a/packages/client/src/NodeRegistry.ts
+++ b/packages/client/src/NodeRegistry.ts
@@ -116,9 +116,11 @@ export class NodeRegistry {
     async setNode(nodeMetadata: string): Promise<void> {
         log('setNode %s -> %s', nodeMetadata)
         await this.connectToNodeRegistryContract()
-
+        const nodeAddress = await this.ethereum.getAddress()
         const tx = await this.nodeRegistryContract!.createOrUpdateNodeSelf(nodeMetadata)
         await tx.wait()
+        await until(async () => { return this.getStorageNodeUrl(nodeAddress) !== undefined }, 10000, 500,
+            () => `Failed to create/update node ${nodeAddress}, timed out querying fact from theGraph`)
     }
 
     async removeNode(): Promise<void> {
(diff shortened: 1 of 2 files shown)


samt1803:
~~this did't fix it yet~~
fixed it

packages/client/src/NodeRegistry.ts
```diff
@@ -116,9 +116,19 @@ export class NodeRegistry {
...
-
+        const nodeAddress = await this.ethereum.getAddress()
         const tx = await this.nodeRegistryContract!.createOrUpdateNodeSelf(nodeMetadata)
         await tx.wait()
+        await until(async () => {
+            try {
+                const url = await this.getStorageNodeUrl(nodeAddress)
+                return nodeMetadata.indexOf(url) !== -1
+            } catch (err) {
+                log('node not found yet %o', err)
```
timoxley: Logging this every 500ms seems like a lot

packages/client/src/NodeRegistry.ts
```diff
@@ -116,9 +116,19 @@ export class NodeRegistry {
...
         log('setNode %s -> %s', nodeMetadata)
         await this.connectToNodeRegistryContract()
-
+        const nodeAddress = await this.ethereum.getAddress()
         const tx = await this.nodeRegistryContract!.createOrUpdateNodeSelf(nodeMetadata)
         await tx.wait()
+        await until(async () => {
+            try {
+                const url = await this.getStorageNodeUrl(nodeAddress)
+                return nodeMetadata.indexOf(url) !== -1
```
timoxley: Perhaps:
```suggestion
                return nodeMetadata.includes(url)
```

timoxley (approved):
I'd probably get it to log less or not at all for retries, but otherwise lgtm.

packages/client/src/NodeRegistry.ts
samt1803: ah, thats why there is no contains :)

PR #320, opened 2021-12-16
Tricking jest run parallel

samt1803:


samt1803:
closing, did not work, actionrunner was overwhelmed and the maxWorkers in jest also works now without the empty testfiles

PR #322, opened 2021-12-16
cleaned up tests that were commented out / disabled on ocr branch

samt1803:
- re-enabled tests that got disabled on OCR branch (Resends)
- removed commented out tests that are now obsolete (testing mostly authentication in streamfetcher, that now works without authentication and also mostly delegates calls to client)


timoxley (approved):
lgtm

PR #324, opened 2021-12-16
cli-tools [NET-572]: Brubeck client support

teogeb:
Updated CLI-tools to use latest `StreamrClient`. Refactored most of the commands to use new `createClientCommand` helper method, which handles the initialization of `StreamrClient`, and destroying the instance when the command completes.

Updated `commander` dependency to 8.3.0. The functionality of `exitWithHelpIfArgsNotBetween` is now implemented with `allowExcessArguments(false)`.

Divided the three resend commands to separate files, similarly to other commands. Inlined some command-specific helper functions from `src` folder to the command files.

### Configuration management

It is now possible to specify a configuration file with `--config` argument. All `StreamrClient` options can be configured in that file (the format of the file is same as `StreamrClient` constructor options). If no configuration file is specified, default options are read from `~/.streamr/config/default.json`. 

The config file format is also compatible with Broker config format. The configuration file must contain `client` block as a root level element, and additional fields are allowed so that a Broker config file passes the validation.

Removed `--http-url` and `--ws-url` argument which are no longer supported by the client. Dropped also `--stg` argument, but we can add that later when new configuration values for the staging environment are available. The `--dev` and `--privateKey` arguments are still supported.

### Other changes related to StreamrClient v6 update

Updated permission handling:
- enum values: `share` is now `grant`, `get` is removed
- `stream grant-permission`: dropped support for legacy enum names like 'stream_get`
- `stream revoke-permission`: references permissions by user, not by `permissionId`
- `stream list --permission`: modified to use the same enums as other commands

Removed two columns from `stream list` output:
- `lastUpdated` (not supported)
- `name` (deprecated)

Relative streamIds are supported natively by `StreamrClient`. There is no longer need to do that transformation in the CLI-tools.

### Future improvements

- When Broker's Config Wizard generates a new config file, it could use  `~/.streamr/config/default.json` as default location.


packages/cli-tools/CHANGELOG.md
```diff
@@ -7,9 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 - Add permission commands: `stream grant-permission` and `stream revoke-permission`
 - Remove `typescript` and `ts-node` as run-time dependencies
-- Remove `--msg-chain-id` parameter from `stream resend from`
+- (Breaking) Remove `--msg-chain-id` parameter from `stream resend from`
 - (Breaking) `streamr stream create` argument is a stream ID, not a name
 - Support path notation when defining a stream ID
+- (Breaking) Permission names and arguments:
+  - renamed `--operation` to `--permission` in `stream list`
+  - the data format of `permissions` returned by `--include-permissions` of `stream show`
```
harbu, line 16: missing verb or smth?

packages/cli-tools/README.md
```diff
@@ -116,8 +111,28 @@ To fetch data between two date-times
 streamr stream resend range 2019-05-10T17:00:00 2019-05-11T21:00:00 <streamId> --private-key <key>
 ```
 
-Flag `--dev` or `--stg` can be enabled for the command to operate on pre-defined development or staging environment.
+#### Configuration
 
+Use can specify the environment and the authentication with the following command line arguments
```
harbu: Did you mean to write "user"?

```suggestion
User can specify environment and authentication details with the following command line arguments.
```

packages/cli-tools/CHANGELOG.md
```diff
@@ -7,9 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 - Add permission commands: `stream grant-permission` and `stream revoke-permission`
 - Remove `typescript` and `ts-node` as run-time dependencies
-- Remove `--msg-chain-id` parameter from `stream resend from`
+- (Breaking) Remove `--msg-chain-id` parameter from `stream resend from`
```
harbu, line 10: Should we add entry for "(Breaking) Command `get-session-token` removed."?

packages/cli-tools/bin/streamr-storage-node-add-stream.ts
```diff
@@ -1,32 +1,12 @@
...
-import { StreamrClient, Stream } from 'streamr-client'
-import {
-    envOptions,
-    authOptions,
-    exitWithHelpIfArgsNotBetween,
-    formStreamrOptionsWithEnv,
-    getStreamId,
-} from './common'
-import pkg from '../package.json'
+require('../src/logLevel')
```
harbu: why `require` when elsewhere `import`?

packages/cli-tools/bin/streamr-stream-publish.ts
```diff
@@ -1,30 +1,62 @@
...
-                console.error(err)
-                process.exit(1)
-                // process.stdin.pipe(ps) recover pipe to continue execution
+            .once('finish', async () => {
+                // We need to wait some time because the client.publish() may resolve the promise
+                // before the node has propagated the message. That may happend if the node
+                // has not yet connected to Tracker when client.publish() is called. In that case
+                // the message is put to the propagation queue (activeTaskStore.add call in
+                // network Propagation.ts:59) and the client.publish() promise resolves immeditatelly.
+                // TODO Remove this wait when NET-604 has been resolved
```
harbu, line 52: :+1:

packages/cli-tools/bin/streamr-stream-show.ts
```diff
@@ -1,21 +1,19 @@
...
+import { createClientCommand } from '../src/command'
 
-const program = new Command()
-program
+createClientCommand(async (client: StreamrClient, streamId: string, options: any) => {
+    const stream = await client.getStream(streamId)
+    const obj = stream.toObject()
+    if (options.includePermissions) {
+        // TODO the data of stream.getPermissions() should be transformed somehow
+        // @ts-expect-error permissions not on {}
```
harbu: Should this transformation be included in this PR or a separate one?

packages/cli-tools/src/resend.ts
```diff
@@ -1,13 +1,19 @@
-import { StreamrClient, StreamrClientOptions, ResendOptions } from 'streamr-client'
+import { StreamrClient, ResendOptions } from 'streamr-client'
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export const assertBothOrNoneDefined = (option1: string, option2: string, errorMessage: string, commandOptions: any): void | never => {
```
harbu:
```suggestion
export const assertBothOrNoneDefined = (option1: string, option2: string, errorMessage: string, commandOptions: Record<string, unknown>): void | never => {
```

packages/cli-tools/bin/streamr-storage-node-add-stream.ts
teogeb: Fixed. At first there were some problems when importing "nothing" form a module. But now works ok.

packages/cli-tools/bin/streamr-stream-show.ts
teogeb: No need to do in this PR, we'll do in NET-610

packages/cli-tools/CHANGELOG.md
teogeb: Added

harbu (approved):

PR #325, opened 2021-12-17
fix(broker): add io.destroy to appropriate place

harbu:


teogeb (approved):

packages/broker/src/plugins/metrics/ConsoleAndPM2Metrics.ts
```diff
@@ -94,6 +94,7 @@ export class ConsoleAndPM2Metrics {
     }
 
     stop(): void {
+        io.destroy()
```
timoxley, line 97: does this need the typeof function check that petri mentioned?

timoxley (approved):

packages/broker/src/plugins/metrics/ConsoleAndPM2Metrics.ts
harbu: Here the `io.metric` calls in the constructor initialize pm2/io, so I don't think its necessary.

PR #326, opened 2021-12-17
feat: re-enabled grant- and revoke- endpoints and tests

samt1803:
feat: re-enabled grant- and revoke- endpoints taking arrays of permissions, re-enabled their tests

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -46,7 +46,7 @@
         "test-types": "npm run test-browser-types",
         "test-browser-types": "tsc -b tsconfig.browser.json",
         "coverage": "jest --coverage",
-        "test-integration": "jest --verbose --useStderr --maxWorkers=2 --forceExit test/integration",
+        "test-integration": "jest --verbose --useStderr --maxWorkers=8 --forceExit test/integration",
         "test-exports": "cd test/exports && npm run link && tsc --noEmit --project ./tsconfig.json && npm test",
         "test-integration-no-resend": "jest --verbose --useStderr --forceExit --testTimeout=30000 --testPathIgnorePatterns='resend|Resend' --testNamePattern='^((?!(resend|Resend|resent|Resent|gap|Gap)).)*$' test/integration/*.test.*",
         "test-integration-resend": "jest --verbose --useStderr --forceExit --testTimeout=40000 --testNamePattern='(resend|Resend|resent|Resent)' test/integration/*.test.*",
--- a/packages/client/src/Stream.ts
+++ b/packages/client/src/Stream.ts
@@ -203,25 +203,27 @@ class StreamrStream implements StreamMetadata {
         return this._streamRegistry.hasPublicPermission(this.id, permission)
     }
 
-    // async hasUserPermissions(operations: StreamPermission[], userId: string) {
-    //     this.assertUserId(userId)
-    //     const permissions = this._streamRegistry.getPermissionsForUser(this.id, userId, )
-    //     return this.hasPermissions(operations, userId)
-    // }
-    // async hasPermissions(operations: StreamPermission[], userId: string) {
-    //     const permissions = await this._streamRegistry.getPermissionsForUser(this.id, userId)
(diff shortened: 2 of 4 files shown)


teogeb (approved):
Looks good! 👍  If we want to simplify the public API we could now even leave out the singular form of `grant` and `revoke`: it is easy to call `grantUserPermissions([SUBSCRIBE], userId)`, if I want to grant just one permission.

packages/client/src/Stream.ts
```diff
@@ -203,25 +203,27 @@ class StreamrStream implements StreamMetadata {
...
-    // }
-
-    // async hasPermission(operation: StreamPermission, userId: string|undefined) {
-    //     const permissions = await this.hasPermissions([operation], userId)
-    //     if (!Array.isArray(permissions) || !permissions.length) { return undefined }
-    //     return permissions[0]
-    // }
+    async hasUserPermissions(requestedPermissions: StreamPermission[], userId: string): Promise<boolean> {
+        this.assertUserId(userId)
+        const permissions = await this._streamRegistry.getDirectPermissionsForUser(this.id, userId)
```
timoxley, line 208: Maybe could condense the `for..of` to:
```typescript
return requestedPermission.every((requestedPermission) => permissions[requestedPermission])
```
Though perhaps there's an argument to keeping it like this for some kind of consistency with the other methods below?

packages/client/src/Stream.ts
```diff
@@ -203,25 +203,27 @@ class StreamrStream implements StreamMetadata {
...
+                return false
+            }
+        }
+        return true
+    }
+
+    async hasPermissions(requestedPermissions: StreamPermission[], userId: string): Promise<boolean> {
+        this.assertUserId(userId)
+        const permissions = await this._streamRegistry.getPermissionsForUser(this.id, userId)
+        for (const requestedPermission of requestedPermissions) {
```
timoxley, line 220: same here

timoxley (approved):
lgtm 👍

timoxley:
> If we want to simplify the public API we could now even leave out the singular form of grant and revoke: it is easy to call grantUserPermissions([SUBSCRIBE], userId), if I want to grant just one permission.

yep, I agree with @teogeb here. And now I'm thinking about it, it's probably a good principle in general to design APIs to work on multiple items first.

samt1803:
agree with all the points, will implement them probably on Tuesday

packages/client/package.json
```diff
@@ -46,7 +46,7 @@
         "test-types": "npm run test-browser-types",
         "test-browser-types": "tsc -b tsconfig.browser.json",
         "coverage": "jest --coverage",
-        "test-integration": "jest --verbose --useStderr --maxWorkers=2 --forceExit test/integration",
+        "test-integration": "jest --verbose --useStderr --maxWorkers=8 --forceExit test/integration",
```
harbu, line 49: If there is a specific `maxWorkers` setting that works well in CI, I would look into configuring it into Github Actions YAML instead of here. Reason being is that it is very environment specific what value works well. (For example, if you don't have many cores on your machine, setting the value high will actually slow down the test suite.)

You can pass `maxWorkers` to the command as follows, for example:
```
npm run test-integration -- --maxWorkers=8
``´

File: packages/client/src/Stream.ts (hunk 2 above)
Comment (harbu) [reply packages/client/src/Stream.ts:220]: For code re-use, you could extract the common functionality of methods  `hasUserPermissions` and `hasPermissions` as a private helper method, something like the below:

```ts
private containsEveryRequestedPermission(requestedPermissions: StreamPermission[], fetchedPermissions: StreamPermission[]): boolean {
    // or alternatively what Tim suggested
    for (const requestedPermission of requestedPermissions) {
        if (!fetchedPermissions[requestedPermission]) {
            return false
        }
    }
    return true
}
```

harbu (approved):

packages/client/package.json
samt1803: ah yes, will remove this, it was just a test to see if ci would run them in parallel

teogeb:
https://github.com/streamr-dev/network-monorepo/pull/450 implements similar functionality, closing this

PR #327, opened 2021-12-20
ci(broker): remove core-api from test-integration job

harbu:


samt1803 (approved):
yes, broker tests should not need core-api any more

PR #329, opened 2021-12-21
Net-589: Firehose metrics stream pattern 

juslesan:
- Nodes use a firehose stream for publishing metrics
- Nodes use hashing to publishing their metrics to one of n partitions configured for the stream
- This means that multiple nodes are assigned to the same metrics partition, which is good for scalability of any aggregators or storage nodes.


packages/broker/src/plugins/metrics/MetricsPlugin.ts
```diff
@@ -11,8 +11,10 @@ const logger = new Logger(module)
 
 export interface MetricsPluginConfig {
     consoleAndPM2IntervalInSeconds: number
+    firehoseStreamIdHead: string
```
harbu: Is this one used (looks like the one inside `nodeMetrics` section is only used?), can be removed?

harbu (commented):
LGTM; see comments

packages/broker/src/plugins/metrics/MetricsPlugin.ts
juslesan: good catch

packages/broker/src/plugins/metrics/MetricsPlugin.ts
```diff
@@ -13,6 +13,7 @@ export interface MetricsPluginConfig {
     consoleAndPM2IntervalInSeconds: number
     nodeMetrics: {
         storageNode: string
+        firehoseStreamIdHead: string
```
teogeb: Rename to `streamIdPrefix`, as the firehose pattern is the only pattern we'll support? We concatenate this to a `suffix`, so `prefix` could be a good word for the prefix/head of the string.

teogeb (changes requested):
We should also remove `MetricsPublisher.ensureStreamsCreated()` as we expect that the firehose streams have already been created.

juslesan:
> We should also remove MetricsPublisher.ensureStreamsCreated() as we expect that the firehose streams have already been created.

I think this should be left in as the streamId is configurable, it makes local development and adding user configured metrics streams easier. Unless this is something that should no longer be done with OCR. @teogeb

teogeb:
> > We should also remove MetricsPublisher.ensureStreamsCreated() as we expect that the firehose streams have already been created.
> 
> I think this should be left in as the streamId is configurable, it makes local development and adding user configured metrics streams easier. Unless this is something that should no longer be done with OCR. @teogeb

If user configures a non-standard `streamId`, we could assume that he creates that stream manually (and gives permissions / adds to storage node if needed). In `NodeMetrics.tests.ts` we can create some test streams in `beforeAll()`. 

The nodeMetrics plugin is disabled in all pre-defined broker configs (`development-1/2/3.json` and `docker-1/2/3.json`). If we run nodes created by config-wizard, we should maybe disable the plugin anyway so that it doesn't generate local data to the production environment.

packages/broker/src/plugins/metrics/node/MetricsPublisher.ts
```diff
@@ -16,45 +16,26 @@ export class MetricsPublisher {
     private readonly nodeAddress: string
     private readonly client: StreamrClient
     private readonly storageNodeAddress: string
+    private readonly streamIdHead: string
 
-    constructor(nodeAddress: string, client: StreamrClient, storageNodeAddress: string) {
+    constructor(nodeAddress: string, client: StreamrClient, storageNodeAddress: string, streamIdHead: string) {
```
harbu:
```suggestion
    constructor(nodeAddress: string, client: StreamrClient, storageNodeAddress: string, streamIdPrefix: string) {
```

packages/broker/src/plugins/metrics/node/MetricsPublisher.ts
```diff
@@ -16,45 +16,26 @@ export class MetricsPublisher {
     private readonly nodeAddress: string
     private readonly client: StreamrClient
     private readonly storageNodeAddress: string
+    private readonly streamIdHead: string
```
harbu:
```suggestion
    private readonly streamIdPrefix: string
```

packages/broker/src/plugins/metrics/node/MetricsPublisher.ts
```diff
@@ -16,45 +16,26 @@ export class MetricsPublisher {
...
 
-    constructor(nodeAddress: string, client: StreamrClient, storageNodeAddress: string) {
+    constructor(nodeAddress: string, client: StreamrClient, storageNodeAddress: string, streamIdHead: string) {
         this.nodeAddress = nodeAddress
         this.client = client
         this.storageNodeAddress = storageNodeAddress
+        this.streamIdHead = streamIdHead
```
harbu:
```suggestion
        this.streamIdPrefix = streamIdPrefix
```

packages/broker/src/plugins/metrics/node/MetricsPublisher.ts
```diff
@@ -164,7 +145,7 @@ export class MetricsPublisher {
     getStreamId(periodLength: number): string {
         const suffix = STREAM_ID_SUFFIXES[periodLength]
         if (suffix !== undefined) {
-            return `${this.nodeAddress.toLowerCase()}/streamr/node/metrics/${suffix}`
+            return `${this.streamIdHead}${suffix}`
```
harbu:
```suggestion
            return `${this.streamIdPrefix}${suffix}`
```

harbu (approved):

PR #337, opened 2021-12-30
NET-614: StreamID nominal type

harbu:
## Summary
Add new [nominal type](https://betterprogramming.pub/nominal-typescript-eee36e9432d2), aka branded string, `StreamID` to protocol package and use throughout codebase wherever stream ids are referred to.

## Nominal types and `StreamID`
Nominal typing isn't supported by TypeScript in an official capacity for the time being. In order to get such behavior from the compiler, the following pattern / hack in used.
```ts
export type StreamID = string & { readonly __brand: 'streamId' } // Nominal typing
```
The property `__brand` isn't actually used or assigned in runtime code. Instead it is there as marker to make the compiler raise an error in case you attempt to assign a `string` (or anything else incompatible) directly to a `StreamID`. Notice that instances of type `StreamID` are effectively `string`(s) in runtime code and can be used and passed as such anywhere a plain string would be expected.

Nominal types themselves don't incur a runtime cost as they only introduce additional compile-time checks, hence they can be considered a "zero-cost abstraction". 

## Motivation
There are at least two benefits to using such nominal types instead of just plain strings. 

If we have something like `publish(streamId: StreamID, content: string)` and then attempt to call it with the arguments in the wrong order, e.g. `publish('{ "hello": "world" }', streamId)`, the compiler will throw an error due to first argument `'{ "hello": "world" }'` not being assignable to `StreamID`. Thus we get a stricter level of type checking from the compiler.

Secondly, we can enforce and validate things when converting a string into a `StreamID`. In this PR, I introduce the method `toStreamID(s: string, ...): StreamID`, which is the primary way of receiving instances of type `StreamID`. This method will validate a few things about its string parameter, throwing an error if they don't hold, and also make sure the address part of the passed stream id  is lower-cased. The advantage of doing all this, is that anytime you see a `StreamID` type in the code, you can rely on the before-mentioned validation rules to have been already checked. Thus you don't need to re-validate (or re-lower-case) things! 

PS. The benefits of nominal typing are very similar to those of [value objects](https://en.wikipedia.org/wiki/Value_object) in Java, however it seems that value objects don't fit in so neatly with JavaScript.

## Usage
- When writing new code, favor `StreamID` type over plain strings when referring to a stream id.
    - Exception: public-facing streamr-client API methods should still use plain strings. Reason: We don't want to force our users to call `toStreamID()`, instead they should be able to pass just strings as is for a nicer API experience. Therefore, one of the first things the client should probably due upon receiving such a plain string is to call `toStreamID()` to convert the user passed string into a `StreamID` and then in internal code deal exclusively only with `StreamID`(s).
- Use `toStreamID(streamIdOrPathAsStr: string, baseAddress?: string): StreamID` to convert a string into a `StreamID`. It also supports path-only stream format when given 2nd argument `baseAddress`.
    - Exception: in the rare case where (a) we are in a performance critical section of the code and (b) `toStreamID()` would be called repeatedly way too often and (c) you can be 100% sure that the string(s) in question conforms to the `StreamID` validation rules, you can do an explicit cast `str as StreamID`.
    - Best practice: avoid having to call `toStreamID` again and again for the same string, ideally it should only be called once for a given string. The best way to achieve this is to use `StreamID` type everywhere and avoid cases where a `StreamID` would be downcasted back to a plain string.
    
## Future considerations
- Consider making `SPID` into a nominal string type. This would harmonize the behavior to be similar to that of `StreamID`, and one big advantage is that we could use `SPID` itself as a key in es6 map without having to deal with the whole `toKey()` / `fromKey()` situation. Another advantage is avoiding instantiating extra objects.
- Adding more nominal types to represent other domain concepts, e.g. EthereumAddress? The idea is to avoid [primitive obsession](https://wiki.c2.com/?PrimitiveObsession), one of the main advantages is type safety and ensuring validating happens once.

## TODO
- Make tests pass


juslesan (approved):

packages/client/src/NodeRegistry.ts
```diff
@@ -94,7 +95,7 @@ export class NodeRegistry {
     // Read from the NodeRegistry or StreamStorageRegistry contract
     // --------------------------------------------------------------------------------------------
 
-    async isStreamStoredInStorageNodeFromContract(streamId: string, nodeAddress: string): Promise<boolean> {
+    async isStreamStoredInStorageNodeFromContract(streamId: StreamID, nodeAddress: string): Promise<boolean> {
```
teogeb: Is this a part of client's  API? If it is , it should handle `streamdId` as `string`, otherwise we could mark it as private.

packages/client/src/NodeRegistry.ts
```diff
@@ -139,20 +140,22 @@ export class NodeRegistry {
     }
 
     async addStreamToStorageNode(streamId: string, nodeAddress: string): Promise<void> {
```
teogeb, line 143: If we want to be super clear about the API, we could rename `streamId` variable to `streamIdOrPath` in the public methods. And then the refined variable could be called streamId instead of `id` (e.g. at line 143).

Alternatively we could state somewhere in the documentation that every time we refer to `streamId`, the user can also provide a path. But then we'd need to verify that documentation statement somehow, so that remains valid also when new methods are added etc.

packages/client/src/Subscription.ts
```diff
@@ -2,12 +2,12 @@
  * The client.subscribe() return value.
  * Primary interface for consuming StreamMessages.
  */
-import { SPID, SPIDKeyShape } from 'streamr-client-protocol'
+import { SPID, SPIDKeyShape, StreamID } from 'streamr-client-protocol'
 import MessageStream, { MessageStreamOptions, MessageStreamOnMessage } from './MessageStream'
 import SubscriptionSession from './SubscriptionSession'
 
 export type SubscriptionOptions = {
```
teogeb: Could remove this type if it is not used anywhere

packages/protocol/src/utils/TimestampUtil.ts
```diff
@@ -10,6 +10,4 @@ export default function parse(millisOrString: number | string): number | never {
     }
 
     throw new Error(`Invalid timestamp: ${millisOrString}`)
-}
-
-export { parse }
```
teogeb, line 15: How the `parse` function should be visible for the users of `streamr-client-protocol` library? As a `TimestampUtil` function? (see export/import in `utils/index.ts`). Maybe removing this export doesn't do any regression, but it could make sense if this method would be available e.g. as `TimestampUtil.parse`.

packages/client/src/NodeRegistry.ts
harbu: We will do this change in NET-605

packages/client/src/NodeRegistry.ts
harbu: I'm going to assume for now that this is part of the public API (and is intended to be part of it) and thus make this work with `string`.

packages/client/src/Subscription.ts
harbu: removed

packages/protocol/src/utils/TimestampUtil.ts
harbu: removed file since isn't used anywhere

teogeb (approved):

PR #340, opened 2022-01-04
refactor: mv KeyServer to test-utils

harbu:
- Remove code duplication by moving key server to `streamr-test-utils` and using same implementation in client and broker tests
- Properly shutdown key server in broker tests


packages/test-utils/src/utils.ts
```diff
@@ -270,3 +273,39 @@ export const toReadableStream = (...args: unknown[]): Readable => {
...
+    private readonly server: http.Server
+
+    constructor() {
+        const app = express()
+        app.use(cors())
+
+        // try avoid sequential tests starting from same address
+        let c = (Math.round((Math.random() * 1000)) % 1000)
+        app.get('/key', (req, res) => {
+            c = ((c + 1) % 1000)
```
samt1803: this will also use 0 as a key. but that privatekey is invalid, since to get to the address at one point you need to devide through the key, and then it would devide through 0.
So 0 needs to be avoided.

samt1803 (changes requested):
only 0 as privatekey needs to be avoided, rest looks good

packages/test-utils/src/utils.ts
harbu: Right, I copied the client keyserver implementation here, looks like the broker version was slightly different but probably correct:

```ts
 const app = express()
    global.__StreamrKeyserver = new KeyServer()
    let c = 0
    app.get('/key', (_req, res) => {
        c += 1
        const hexString = c.toString(16)
        const privkey = '0x' + hexString.padStart(64, '0')
        log('key endpoint called, returning key ' + privkey)
        if (c > 1000 ) { c = 0 }
        res.send(privkey)
    })
    app.listen(45454)
```

harbu: I'll update to use the above logic instead

harbu: Good catch btw! :+1:

harbu: @samt1803 applied your changes

PR #344, opened 2022-01-07
Net-625: multichain config

juslesan:
Enable configurability for between multiple chains in different parts of the client


samt1803 (changes requested):
The steamregistry and the storagenoderegistry must always reside on the same chain, since the StreamStorageRegistry contract uses them both.
So the streamRegistryChainRPC and storageNodeRegistryChainRPC could be merged, for example a registriesChainRPC..

samt1803 (approved):
looks good to me!

samt1803 (changes requested):
You might need to make sure that the broker config now also reflects this change, this might be because of that: https://github.com/streamr-dev/network-monorepo/runs/4739561071?check_suite_focus=true#step:6:73

samt1803 (approved):
👍

PR #346, opened 2022-01-11
Fix broker ci

juslesan:
Fixed requirement to use the core-api in the broker metrics tests

### DIFF
--- a/.github/workflows/broker.yml
+++ b/.github/workflows/broker.yml
@@ -58,7 +58,7 @@ jobs:
       - name: Start Streamr Docker Stack
         uses: streamr-dev/streamr-docker-dev-action@v1.0.0-alpha.3
         with:
-          services-to-start: "cassandra init-keyspace nginx parity-sidechain-node0 graph-deploy-streamregistry-subgraph parity-node0"
+          services-to-start: "cassandra init-keyspace nginx parity-sidechain-node0 graph-deploy-streamregistry-subgraph parity-node0 core-api"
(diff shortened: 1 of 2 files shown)


.github/workflows/broker.yml
```diff
@@ -58,7 +58,7 @@ jobs:
       - name: Start Streamr Docker Stack
         uses: streamr-dev/streamr-docker-dev-action@v1.0.0-alpha.3
         with:
-          services-to-start: "cassandra init-keyspace nginx parity-sidechain-node0 graph-deploy-streamregistry-subgraph parity-node0"
+          services-to-start: "cassandra init-keyspace nginx parity-sidechain-node0 graph-deploy-streamregistry-subgraph parity-node0 core-api"
```
harbu: Hey I don't think we need `core-api` just for `(await client2.getUserInfo()).username)` below? Just use a different key?

harbu: https://github.com/streamr-dev/network-monorepo/blob/144926f485b04370a8de5bad6be9236c9ab166bd/packages/broker/src/plugins/metrics/node/MetricsPublisher.ts#L31-L33 <- can we do this without core api like here?

juslesan: This was fixed in https://github.com/streamr-dev/network-monorepo/pull/346/commits/e1782889a0ab3876b385b94fa289944bf8263753

harbu (approved):

PR #350, opened 2022-01-12
[NET-643]: feat: set client defaultconfig to use registries on Polygon mainnet

samt1803:
set clients defaultconfig for stream and node registries to Polygon mainnet addresses:
```
"nodeRegistry": "0x080F34fec2bc33928999Ea9e39ADc798bEF3E0d6",            
"chainlinkOracle": "0x36BF71D0ba2e449fc14f9C4cF51468948E4ED27D",            
"enscache": "0x870528c1aDe8f5eB4676AA2d15FC0B034E276A1A",            
"streamRegistry": "0x0D483E10612F327FC11965Fc82E90dC19b141641",            
"streamStorageRegistry": "0xe8e2660CeDf2a59C917a5ED05B72df4146b58399"
```

and theGraph endpoint to the official hosted theGraph server:
https://api.thegraph.com/subgraphs/name/streamr-dev/streams

in ConfigBase


samt1803:
I tested creating and deleting a stream through the client and it worked.

samt1803:
Ah wait, let me also change the release version to alpha 6.0

samt1803:
Not sure why CI is having problems with the new version now, npm i works locally for the monorepo..

packages/network/src/connection/WebRtcConnection.ts
```diff
@@ -51,7 +51,7 @@ export function isOffering(myId: PeerId, theirId: PeerId): boolean {
 
 function offeringHash(idPair: string): number {
     const buffer = crypto.createHash('md5').update(idPair).digest()
-    return buffer.readInt32LE()
+    return buffer.readInt32LE(0)
```
teogeb: Does this modification fix something? It seems that the same default `0` would be used if not specified: https://www.geeksforgeeks.org/node-js-buffer-readint32le-method/

harbu: There was problem with code compilation without it so I added it in 🤷

teogeb (approved):

PR #351, opened 2022-01-12
NET-644: remove v1 serializers from protocol

harbu:


juslesan (approved):

juslesan:
Seems that there are some assert checks in tests in the network that break with this change

PR #356, opened 2022-01-13
ETH-207: Create stream with ENS name

teogeb:
Add ENS support to `client.createStream()`


samt1803:
> Add ENS support to `client.createStream()`
> 
> ## TODO
> * Why the happy-path test (`StreamEndpoint.test.ts:98`) fails? Are we using the up-to-date contract?

The contract that is deployed in the docker-dev env and the jsons for interacting with it in the client still have that method with an old name ('syncEnsAndCreateStream' instead of 'createStreamWithENS')

I will update this probably today, then merging that branch into this should fix this. -> [ETH-213 ](https://linear.app/streamr/issue/ETH-213/update-docker-dev-env-with-new-proxified-contracts)

packages/client/test/integration/StreamEndpoints.test.ts
```diff
@@ -83,8 +83,43 @@ function TestStreamEndpoints(getName: () => string, delay: number) {
...
+
+        it('key-exchange format', async () => {
+            const streamId = 'SYSTEM/keyexchange/0x0000000000000000000000000000000000000000'
+            await expect(async () => client.createStream({ id: streamId })).rejects.toThrow(`stream id "${streamId}" not valid`)
+        })
+
+        describe('ENS', () => {
+
+            it('domain owned by user', async () => {
+                const streamId = 'testdomain1.eth/foobar'
```
harbu: Will this test fail if run multiple times due to `testdomain1.eth/foobar` already being present in docker smart contract? If so maybe some random `uuid()` needs to be added as part of the path.

harbu (changes requested):

packages/client/test/integration/StreamEndpoints.test.ts
teogeb: Good catch, fixed

harbu (approved):

PR #360, opened 2022-01-18
NET-598: Search streams

teogeb:
## Client

Replaced the `client.listStreams` method with new `client.searchStreams` method. The new method does a freetext search over the metadata field. The old method used only the `name` field.

The functionality of query options `max`, `offset` and `order` have not changed. Removed `orderBy` option: if the `order` option is specified, the ordering is always done by the `id` field. (In theory we could also sort the result set by `metadata`, `permissions` or `permissions` field, but those fields are internal JSONs data structures.)

## CLI-tools

Replaced the `stream list` command with new `stream search` command. The new command has different arguments and it doesn't support most of the filters which were available for `stream list` command.

## Other

- Removed `client.getStreamByName` method which was implemented by calling `client.listStreams`. All name-related methods will be removed in NET-632.
- Simplified `StreamEndpoints.test.ts` test wrapper


teogeb:
Note that the id field will be removed in https://linear.app/streamr/issue/ETH-214/streamregistry-contract-do-not-store-id-in-the-metadata-json

packages/client/src/StreamEndpoints.ts
```diff
@@ -27,18 +27,10 @@ import { StreamRegistry } from './StreamRegistry'
 import { NodeRegistry } from './NodeRegistry'
 import { StreamIDBuilder } from './StreamIDBuilder'
 
-export interface StreamListQuery {
-    name?: string
-    uiChannel?: boolean
-    noConfig?: boolean
-    search?: string
-    sortBy?: string
+export interface SearchStreamsOptions {
```
harbu: Should this interface definition be moved to StreamRegistry.ts?

harbu (approved):

packages/client/src/StreamEndpoints.ts
teogeb: Moved

PR #361, opened 2022-01-18
NET-647: StreamPartID nominal type

harbu:
## Summary
- Replace `SPID` with `StreamPartID` throughout code base.
- Harmonize public API of client to use type `StreamDefinition` anytime a stream part(ition) is expected.

### Type `StreamPartID`
Type `StreamPartID` represents a `"${streamId}#${streamPartition}"` tuple and is a branded string similar to `StreamID` (See #337). This means that it can directly be used wherever a string can (importantly in `Set` and `Map`), and that compile-time checks are in place to verify that a plain string cannot be directly assigned where a `StreamPartID` would be required. This means anytime you see a `StreamPartID` in the code, you can be sure that it has already been validated for you.

The two primary ways to instantiate a `StreamPartID` are
- from a streamId and partition `toStreamPartID(toStreamID('test.eth/foo/bar'), 10)`,
- from a string `StreamPartIDUtils.parse('test.eth/foo/bar#10')`.

### `StreamPartIDUtils` - collection of utilities
Class `StreamPartIDUtils` contains utility static methods for dealing with the `StreamPartID` type. The most important ones are `getStreamID` and `getStreamPartition` for extracting those elements from a `StreamPartID`.

### Type `StreamDefinition`
Type `StreamDefinition` is exported by the client package and it specifies the accepted argument formats wherever a stream part(ition) should be specified by the user of the package. The accepted formats are enumerated in what follows.

- Plain streamId as string, e.g., `0xc0ffee254729296a45a3885639AC7E10F9d54979/foo/bar`
- Stream part id as string, e.g., `0xc0ffee254729296a45a3885639AC7E10F9d54979/foo/bar#66`
- Object with `id` property, e.g. `{ "id": "test.eth/foo/bar" }`
- Object with `id` property and `partition` property, e.g. `{ "id": "test.eth/foo/bar", partition: 66 }`
- Object with `stream` property, e.g. `{ "stream": "test.eth/foo/bar" }`
- Object with `stream` property and `partition` property, e.g. `{ "stream": "test.eth/foo/bar", partition: 66 }`
- Object with `streamId` property, e.g. `{ "streamId": "test.eth/foo/bar" }`
- Object with `streamId` property and `partition` property, e.g. `{ "streamId": "test.eth/foo/bar", partition: 66 }`

When a `partition` is not specified by the user, a default interpretation will be made on their behalf. Most of the time this will be equivalent to setting `partition = 0`. The exception to this is when publishing a message, in which the absence of partition will be interpreted as a random partition. Also note that path-only stream ids are supported in all the above cases.

Why so many formats? We went through the public API and README of the client and noticed that there were a multitude of ways to specify a stream part(ition) depending on what method was being used. Some formats would work in some cases whilst not in others. Here we attempted to gather all the major formats and make them work in all use cases.

### Class `StreamIDBuilder`
This class was already introduced in #343. In this pull request, we added new methods to support `StreamDefinition`. The most important method being `toStreamPartID` which converts `StreamDefinition`(s) into `StreamPartID`(s). This method is heavily relied on in the client when receiving a `StreamDefinition` through the public API. It is similar in nature to `StreamIDBuilder#toStreamID` and its design principles, as described in #343.

## Design principles
- Going forward, in internal code we should always use `StreamID` and `StreamPartID` to represent those concepts.
- The public API of the client package should be the only code surface were we accept type `StreamDefinition`. This should be converted ASAP into a `StreamPartID` in order for it to be passed to the internal code base and to make sure any validation happens sooner rather than later (fail-fast principle).
- The above point applies to type `streamIdOrPath: string` in client as well. It should be convered to a `StreamID` ASAP. See #343 for more details.
- When designing something new for the public API of the client: use `streamIdOrPath: string` when partition is not relevant. Otherwise use `StreamDefinition`.

## API changes
- Class `StreamrStream` can no longer be directly passed as argument to those public API methods that used to support it.
  -  The fix is rather easy `func(streamrStream)` => `func(streamrStream.id)` and you are set.
- Property `streamPartition` is not supported in public API methods when giving a `StreamDefinition`. It is unclear to the authors of this PR whether such support ever existed in the first place?
  - The fix is to rename `streamPartition` => `partition`. 
- Some public API methods are effectively less strict than before with regards to the input they accept. This is due to harmonizing the entire public API to use `StreamDefinition`.

## Other changes
- Small change to the content of Node->Tracker status message: renamed the `stream` field to `streamPart` in `Status` (`network/src/identifiers.ts`)
- To avoid ambiguity, names of methods / functions that deal with stream part(itions) are re-named to reflect that if they already weren't.
- Methods `getSubscriptions` and `count` of class `Subscriber` are now `async` due to the need to resolve path-only stream ids.

## Future considerations
- Remove asynchronous Ethereum address resolution: if the Ethereum address of the logged in client user could be fetched synchronously (e.g. by caching?), most of `StreamIDBuilder` class could be probably binned.
- Abundance of `StreamDefinition` formats: could we retire most of the object formats and keep one?
- Interpreting absence of partition: what should the default partition be? It seems to be use-case dependent. See https://linear.app/streamr/issue/NET-344/harmonize-semantics-of-using-streamid-without-partition for good discussion on topic.


packages/network/src/identifiers.ts
```diff
@@ -13,15 +13,15 @@ export interface Location {
     city?: string
 }
 
-export interface StreamStatus {
-    id: string
+export interface StreamPartStatus {
+    id: StreamID
```
harbu, line 17: I'm not sure we can ensure this type in the use case where the tracker receives a `StatusMessage`.

packages/network/src/logic/tracker/Tracker.ts
```diff
@@ -138,12 +137,12 @@ export class Tracker extends EventEmitter {
...
 
-        const spidKey = SPID.toKey(stream.id, stream.partition)
+        const streamPartId = toStreamPartID(streamPart.id, streamPart.partition)
 
         // update topology
-        this.createTopology(spidKey)
-        this.updateNodeOnStream(source, stream)
-        this.formAndSendInstructions(source, spidKey)
+        this.createTopology(streamPartId)
+        this.updateNodeOnStream(source, streamPart)
```
harbu: I think the 2nd argument may be wrong here? Wonder what happened.

packages/network/test/unit/DisconnectionManager.test.ts
```diff
@@ -13,7 +13,7 @@ describe(DisconnectionManager, () => {
     function setUpManager(disconnectionDelayInMs: number, cleanUpIntervalInMs: number): void {
         manager = new DisconnectionManager({
             getAllNodes,
-            hasSharedStreams,
+            hasSharedStreamParts: hasSharedStreams,
```
harbu: Maybe just rename var `hasSharedStreams` to be `hasSharedStreamParts`

packages/network/src/logic/tracker/Tracker.ts
teogeb: It's correct, but I'll improve the readability

packages/network/src/identifiers.ts
teogeb: When we build a validation (#255), we can validate this field

teogeb (approved):

juslesan (approved):

PR #367, opened 2022-01-21
Fix for supporting the currently deployed DU 2.2 contracts

jtakalai:


hpihkala:
A few quick comments @jtakalai:
- Note that CI is failing due to some TS build errors
- Is it now the case that the fix which queries the sidechain DU address from the factory instead of computing it locally is now only in v5, not here? Can it be cherry picked from there?
- Best to create tickets for TODOs instead of leaving them in the comments

jtakalai:
> * Note that CI is failing due to some TS build errors

fixed

> * Is it now the case that the fix which queries the sidechain DU address from the factory instead of computing it locally is now only in v5, not here? Can it be cherry picked from there?

They're in #371. This, that and #370 should be merged to main anyway before next publish if possible, at least before 6.0. This PR specifically fixes the issue that is blocking BACK-200 by preventing https://github.com/streamr-dev/core-api/pull/1150 tests from passing. After this PR is in core-api/rest-e2e-tests, that problem should go away.

> * Best to create tickets for TODOs instead of leaving them in the comments

Both would be good I think? Why not point out the debt in code as well? Tickets may get buried and then maybe the intervention point is lost in code changes.

hpihkala:
Sounds reasonable! 👍🏻

packages/client/src/dataunion/Contracts.ts
```diff
@@ -21,6 +22,7 @@ function validateAddress(name: string, address: EthereumAddress) {
     }
 }
 
+/* TODO: remove this class and inline its functions into DataUnion.ts */
```
harbu, line 25: If you decide to make tickets about these issues (as discussed in the PR) it would be useful to refer to them in the comments, e.g.

```ts
/* TODO: ETH-1234, remove this class and inline its functions into DataUnion.ts */
```

harbu (commented):
From code style perspective looks good to me. I'm not that familiar with DU and it's current progress so can't really review this from a substantive angle.


The tests in main branch are green and (known) flaky ones have been disabled.  So you might want to merge from main again. Once you get tests passing here feel free to request re-review.

hpihkala (approved):

PR #368, opened 2022-01-21
Net-364: no longer encrypt streams with public subscribe permissions

juslesan:
- Automatically encrypt non-public streams and vice versa.
- Remove option for required encryption for streams


packages/client/src/Encrypt.ts
```diff
@@ -47,14 +47,19 @@ export default class PublisherEncryption implements Stoppable {
         }
 
         const streamId = streamMessage.getStreamId()
+
+        const isPublic = await this.streamEndpoints.isPublic(streamId)
+        if (isPublic) {
```
harbu: Stop signal after `await`
```suggestion
        if (isPublic || this.isStopped) {
```

harbu (approved):
lgtm :+1:

PR #369, opened 2022-01-21
NET-663: fix hanging resources in broker tests

harbu:
- Add some missing stop / destroy / cleanup calls in broker tests `afterEach` / `afterAll` blocks. 
- Remove unnecessary checks for `this.streamrClient === undefined` in broker plugins

### DIFF
--- a/packages/broker/test/integration/plugins/storage/StorageNode.test.ts
+++ b/packages/broker/test/integration/plugins/storage/StorageNode.test.ts
@@ -22,7 +22,7 @@ describe('StorageNode', () => {
 
     beforeAll(async () => {
         storageNodeAccount = new Wallet(await fetchPrivateKeyWithGas())
-        storageNode = await startStorageNode(storageNodeAccount.privateKey, httpPort1, trackerPort)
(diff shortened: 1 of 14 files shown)


juslesan (approved):

packages/broker/test/integration/plugins/storage/StorageNode.test.ts
```diff
@@ -8,7 +8,6 @@ import {
 import { Broker } from "../../../../src/broker"
 import StreamrClient from 'streamr-client'
 
-const httpPort1 = 12501
```
juslesan, line 11: Integration Tests seem to be failing because this was removed

packages/broker/test/integration/plugins/storage/StorageNode.test.ts
```diff
@@ -8,7 +8,6 @@ import {
...
 const trackerPort = 12503
 
 describe('StorageNode', () => {
```
teogeb, line 13: We could convert this test to a unit test (or remove, as it actually tests the logic of `StreamrClient` node id calculation from a private key)

teogeb (approved):

PR #370, opened 2022-01-21
Rename safeGetDataUnion -> getDataUnion

jtakalai:
remove safeGetDataUnion

we should only have safe functions by default...


harbu (commented):
> we should only have safe functions by default...

Good point :D 

LGTM; feel free to merge once tests go green

hpihkala (approved):

PR #371, opened 2022-01-21
Data Union fixes ported from v5

jtakalai:


packages/client/src/dataunion/Contracts.ts
```diff
@@ -56,60 +56,82 @@ export default class Contracts {
      * NOTE: if template address is not given, calculation only works for the newest currently deployed factory,
      *       i.e. can be used for "future deployments" but not necessarily old deployments
      * This can be used when deploying, but not for getDataUnion
-     * TODO: rename this to calculateDataUnionMainnetAddress, see v5 branch, around PR #285
      */
-    getDataUnionMainnetAddress(dataUnionName: string, deployerAddress: EthereumAddress) {
+    calculateDataUnionMainnetAddress(dataUnionName: string, deployerAddress: EthereumAddress, templateMainnetAddress?: EthereumAddress) {
```
harbu: It can be useful to be explicit about the return type when declaring a method / function, especially if it is a public one. It serves as documentation to a reader (I know the return type gets inferred but sometimes it can be unruly what gets generated) and it also prevents a certain type of errors I have seen where the author has ended up with a return type they probably didn't intend.

This is a bit artificial, but something similar in nature that I encountered a while ago:
```ts
function positiveAdd(a: number, b:number) {
        if (a < 0 || b < 0) { return }
        return a+ b
    }
 ```
 Quickly glancing, you may expect the return type to be `number` but it is `number | undefined`. By being explicit about the return type
 ```ts
 function positiveAdd(a: number, b: number): number {
    if (a < 0 || b < 0) { return }
    return a + b
}
 ```
 you get compilation error 
 ```
 TS2322: Type 'undefined' is not assignable to type 'number'.
 ```

File: packages/client/test/integration/dataunion/calculate.test.ts [hunk 2]
```diff
@@ -24,14 +26,17 @@ describe('DataUnion calculate', () => {
         const adminClient = new StreamrClient(clientOptions as any)
 
         const dataUnionName = 'test-' + Date.now()
-        // eslint-disable-next-line no-underscore-dangle
-        const dataUnionPredicted = adminClient._getDataUnionFromName({ dataUnionName, deployerAddress: adminWalletMainnet.address })
+
+        // @ts-expect-error
```
harbu: A good idea is to add a short description after `@ts-expect-error` to indicate to the future reader why it exists in the first place.

E.g.
```
// @ts-expect-error accessing private method
```

harbu (approved):
LGTM; merge at your convenience when tests are green

PR #372, opened 2022-01-21
skip known flaky tests

harbu:
- Disable known flaky tests for now with Jest's `skip` keyword. They are listed under [NET-620](https://linear.app/streamr/issue/NET-620/flaky-tests) and have their associated tickets' priority increased.
- With respect to [NET-666](https://linear.app/streamr/issue/NET-666/the-mystery-of-key) and the weird behavior of private key `000000000000000000000000000000000000000000000000000000000000000a`, make it so that the `KeyServer` never returns it as a result when fetching a private key.

### DIFF
--- a/packages/client/test/utils.ts
+++ b/packages/client/test/utils.ts
@@ -165,15 +165,20 @@ export const createRelativeTestStreamId = (module: NodeModule, suffix?: string) 
     return counterId(`/test/${randomTestRunId}/${getTestName(module)}${(suffix !== undefined) ? '-' + suffix : ''}`, '-')
 }
 
-// eslint-disable-next-line no-undef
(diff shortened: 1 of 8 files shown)


packages/client/test/utils.ts
```diff
@@ -171,7 +171,11 @@ export const createTestStream = async (streamrClient: StreamrClient, module: Nod
         id: createRelativeTestStreamId(module),
         ...props
     })
-    await until(async () => { return streamrClient.streamExistsOnTheGraph(stream.id) }, 20000, 500)
+    await until(
+        async () => { return streamrClient.streamExistsOnTheGraph(stream.id) },
+        30000,
```
harbu: try increasing?
```suggestion
        45000,
```

PR #376, opened 2022-01-25
NET-671: Export `ResendSubscription`

mondoreale:
Regular `Subscription` is already exported but it's different than `ResendSubscription`. I need both ​for my current `streamr-client-react` work.


mondoreale:
@teogeb can you have a quick look? It's a tiny change. ;)

harbu (approved):
Discussed with @mondoreale whether this should be exposed as public or not. I guess it should, I was going to recommend using the event `resent` as described in the project readme https://github.com/streamr-dev/network-monorepo/tree/main/packages/client#subscription-options

```
const sub = await client.subscribe(options)
sub.on('resent', () => {
    console.log('All caught up and received all requested historical messages! Now switching to real time!')
})
```

But after a cursory search through the client code, it seems like this event doesn't get emitted anymore? (It still might but I might be missing something).

Anyway looks like using `ResendSubscription#onResent` is the primary way to learn whether a resend is done, thus it should be public.

PR #377, opened 2022-01-25
NET-632: Remove stream.name

teogeb:
Remove the deprecated `Stream.name` field. The `id` and `description` fields should be used instead.

Removed also `EnvStressTest.test.js`, which was a disabled flaky stress test. We can't currently create streams for a single user concurrently in OCR. It would be possible to modify the test to use different private keys for each stream, but then it would mostly test the contract transaction handling. If we need a stress test, we should maybe implement a stress test for the contract instead (in the contract repository).


harbu (approved):
good stuff, merge when CI green

PR #378, opened 2022-01-25
NET-672: Skip the unguarded test-breaking code

mondoreale:
This little function is used in other areas of the code and make things explode when it's actually necessary. Here it behaves more like a environment precondition. So I drop it.

@timoxley, lmk if I'm missing something.


harbu (approved):
Wait has this been in the code on purpose or a copy-paste mistake?

Anyways prolly better to remove.

mondoreale:
I'll merge it then. Shouldn't hurt.

PR #380, opened 2022-01-26
client: validate publish parameters

teogeb:
Now `client.publish` throws an error if it is called with `partition` and `partitionKey`. Earlier `partition` was ignored with both were provided (`PublishPipeline:117`).

Fixed the error message of `FailedToPublishError`. Earlier a stream definition was stringified like this: `"Failed to publish to stream [object Object]"`. Also simplified the constructor parameters and fields of the class.

Added unit test for `Publisher` class and a unit test helper class `FakeStreamRegistry`.


harbu (approved):
LGTM; merge when tests green

PR #385, opened 2022-01-27
feat(cli-tool): add `streamr whoami` command

harbu:
Add new command `streamr wallet whoami` (inspired by Unix's counterpart) to display Ethereum address of current account.


packages/cli-tools/bin/streamr.ts
```diff
@@ -9,4 +9,5 @@ program
     .command('stream', 'stream subcommands')
     .command('storage-node', 'storage node subcommands')
     .command('mock-data', 'mock data subcommands')
+    .command('whoami', 'displays your public address')
```
teogeb: The commands are currently grouped into two level subcommand in object+verb form (`stream create`, `mock-data generate`, `storage-node list`...)

Do we want keep following that pattern ? If we do, we could rename this to e.g. `account whoami`. 

Or if the optimal grouping is not yet known, maybe we can move this command later if needed.

teogeb (approved):

PR #387, opened 2022-01-27
Added network-specific overrides to StreamRegistry

jtakalai:
set the gas price on Polygon to (currently rather high) 50gwei to see if that helps an urgent gas pricing issue

### DIFF
--- a/packages/client/src/ConfigBase.ts
+++ b/packages/client/src/ConfigBase.ts
@@ -165,6 +165,16 @@ export const STREAM_CLIENT_DEFAULTS: StrictStreamrClientConfig = {
         templateMainnetAddress: '0x67352e3F7dBA907aF877020aE7E9450C0029C70c',
         templateSidechainAddress: '0xaCF9e8134047eDc671162D9404BF63a587435bAa',
     },
+
+    ethereumNetworks: [
+        {
+            chainId: 137,
+            overrides: {
+                gasPrice: '50000000000', // TODO: implement the "estimate x 1.5" logic if possible
+            }
+        }
+    ],
+
     cache: {
         maxSize: 10000,
         maxAge: 30 * 60 * 1000, // 30 minutes
--- a/packages/client/src/StreamRegistry.ts
+++ b/packages/client/src/StreamRegistry.ts
@@ -80,7 +80,7 @@ export class StreamRegistry implements Context {
     streamRegistryContractReadonly: StreamRegistryContract
     chainProvider: Provider
     chainSigner?: Signer
-    defaultOverrides?: Overrides
+    defaultOverrides: Overrides
 
     constructor(
         context: Context,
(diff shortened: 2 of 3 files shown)


packages/client/src/StreamRegistry.ts
```diff
@@ -181,7 +186,7 @@ export class StreamRegistry implements Context {
         await this.connectToStreamRegistryContract()
         let tx
         if (StreamIDUtils.isENSAddress(domain)) {
-            tx = await this.streamRegistryContract!.createStreamWithENS(domain, path, JSON.stringify(normalizedProperties))
+            tx = await this.streamRegistryContract!.createStreamWithENS(domain, path, JSON.stringify(normalizedProperties), ethersOverrides)
```
harbu, line 189: Is there a way to pass the `ethersOverrides` to the `streamRegistryContract` at once (e.g. constructor or setter) without having to repeat?

packages/client/src/StreamRegistry.ts
```diff
@@ -94,6 +95,8 @@ export class StreamRegistry implements Context {
         this.chainProvider = this.ethereum.getStreamRegistryChainProvider()
         this.streamRegistryContractReadonly = new Contract(this.config.streamRegistryChainAddress,
             StreamRegistryArtifact, this.chainProvider) as StreamRegistryContract
+        const chainId = this.config.streamRegistryChainRPC?.chainId
+        this.defaultOverrides = this.config.ethereumNetworks?.find((network) => network.chainId === chainId)?.overrides
```
harbu: Do we need a check here in case `chainId` were undefined for whatever reason.
```suggestion
        if (chainId !== undefined) {
            this.defaultOverrides = this.config.ethereumNetworks?.find((network) => network.chainId === chainId)?.overrides
        }
```

packages/client/src/ConfigBase.ts
```diff
@@ -165,6 +165,16 @@ export const STREAM_CLIENT_DEFAULTS: StrictStreamrClientConfig = {
         templateMainnetAddress: '0x67352e3F7dBA907aF877020aE7E9450C0029C70c',
         templateSidechainAddress: '0xaCF9e8134047eDc671162D9404BF63a587435bAa',
     },
+
+    ethereumNetworks: [
+        {
+            chainId: 137,
```
harbu, line 171: Is there a reason to do it indirectly like this, instead of just having `overrides` property reside under config property `streamRegistryChainRPC`?

packages/client/src/StreamRegistry.ts
jtakalai: I wrote it like that so that it could be overridden inside the function call too by adding it to the function signature if needed later...

packages/client/src/ConfigBase.ts
jtakalai: I could put it there as well. Was thinking why not start writing stuff into a bit more "multi chain" structure, a bit like ETH-184 or https://hardhat.org/config/#networks-configuration and that array was just because it seemed simpler in terms of TypeScript types.

packages/client/src/StreamRegistry.ts
jtakalai: I guess the next line would still give out an undefined?

packages/client/src/ConfigBase.ts
jtakalai: I've changed it into network names as keys instead of array in #371

PR #391, opened 2022-01-28
NET-539: storage node assignment synchronization

harbu:
## Changes in a nutshell
- Add `blockNumber` to NodeRegistry's  `EtherumStorageEvent` and to the result of method `getStoredStreamsOf`, allowing us to determine which blockchain block the event or result applies to.
- Use this information in broker's storage plugin to synchronize events and polls appropriately. 
- Deprecate storage assignment event stream listening (we use smart contract event listeners to achieve the same) and thus broker's configuration setting `storageConfig.streamrAddress`.

## In detail

The client changes are straightforward. In the case of `EtherumStorageEvent`, we dig up the `blockNumber` directly from the ethers library callback function argument. For method `getStoredStreamsOf`, we modify the GraphQL query to also retrieve the latest block number with the result. In addition to these changes, some minor fixes to dependency injection to avoid circular dependency hassles that came up.

Broker has more major changes. Firstly, the core data structure and procedure for managing state is encapsulated in class `SetMembershipSynchronizer`. It is generalized to solve the problem of set membership over discrete time given full state updates and patches. In our case, the "elements" of the set are effectively `StreamPartID`(s) and we use The Graph polling for full state updates and blockchain-based events for patches.

Secondly, two separate data sources have been extracted from `StoragePlugin`. Everything related to polling full state data is in class `StoragePoller`, while event-based data handling can be found under class `StorageEventListener`. What responsibilities are left in `StoragePlugin` are (a) converting stream ids to stream parts, (b) taking care of storage cluster membership checks w.r.t. stream parts, and (c) instantiating the `StoragePoller` and `StorageEventListener` and connecting them with `SetMembershipSynchronizer`. Effectively `StoragePlugin` orchestrates the whole process of storage node assignments.


packages/broker/src/plugins/storage/SetMembershipSynchronizer.ts
```diff
@@ -0,0 +1,116 @@
...
+        return {
+            added: setDifference(this.computedState, previousComputedState),
+            removed: setDifference(previousComputedState, this.computedState)
+        }
+    }
+
+    /**
+     * Ingest a patch update
+     * @param elements the set of elements relevant to this patch
+     * @param operation indicates whether this is an assignment or a removal
```
teogeb:
```suggestion
     * @param operation indicates whether this is an addition or a removal
```

packages/broker/src/plugins/storage/SetMembershipSynchronizer.ts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+/**
+ * Represents a set that evolves over (discrete) time. The set is updated in
+ * two distinct ways:
+ *      1. Full state updates that define the set at a given time step.
+ *      2. Patch updates that add or remove a subset at a given time step.
+ */
+export class SetMembershipSynchronizer<E extends string> {
+    private lastState: ReadonlySet<E> = new Set<E>()
```
teogeb: The `lastState` field is actually not needed.

packages/broker/src/plugins/storage/StorageConfig.ts
```diff
@@ -1,288 +1,108 @@
...
+ *
+ *  Event-based assignments are great for picking up on changes quickly.
+ *  However, there is a risk of not receiving updates due to, e.g. connectivity
+ *  issues. Therefore, if the real-time system fails, polling acts as a sort-of
+ *  backup system.
+ */
 export class StorageConfig {
+    private readonly listener: StorageConfigListener
+    private readonly synchronizer = new SetMembershipSynchronizer<StreamPartID>()
+    private readonly clusterId: string
```
teogeb: The `clusterId` field is redundant, not needed outside of the constructor.

packages/broker/src/plugins/storage/SetMembershipSynchronizer.ts
```diff
@@ -0,0 +1,116 @@
...
+    getState(): ReadonlySet<E> {
+        return this.computedState
+    }
+
+    /**
+     * Ingest a full state update at a given sequence number
+     * @param elements the set of _all_ elements at this sequence number
+     * @param sequenceNo the sequence number for this state
+     */
+    ingestState(elements: Set<E>, sequenceNo: number): Diff<E> {
```
teogeb: Rename to `ingestSnapshot`? Then we'd have three clearly different words: `patch`, `snapshot` and `state`.

packages/broker/src/plugins/storage/StoragePoller.ts
```diff
@@ -0,0 +1,63 @@
...
+const logger = new Logger(module)
+
+/**
+ * Polls full state of storage node assignments on an interval.
+ */
+export class StoragePoller {
+    private readonly clusterId: string
+    private readonly pollInterval: number
+    private readonly streamrClient: StreamrClient
+    private readonly handleResult: (streams: Stream[], block: number) => void
```
teogeb: Rename e.g. to `onNewSnapshot`?

packages/broker/src/plugins/storage/StorageEventListener.ts
```diff
@@ -0,0 +1,46 @@
...
+const logger = new Logger(module)
+
+/**
+ * Hooks up to `StreamrClient#registerStorageEventListener` to learn about
+ * stream assignment and removal events in real-time.
+ */
+export class StorageEventListener {
+    private readonly clusterId: string
+    private readonly streamrClient: StreamrClient
+    private readonly handleEvent: (stream: Stream, type: 'added' | 'removed', block: number) => void
```
teogeb: Rename to `onEvent`?

packages/broker/src/plugins/storage/StoragePoller.ts
```diff
@@ -0,0 +1,63 @@
...
+        handleResult: (streams: Stream[], block: number) => void
+    ) {
+        this.clusterId = clusterId
+        this.pollInterval = pollInterval
+        this.streamrClient = streamrClient
+        this.handleResult = handleResult
+    }
+
+    async start(): Promise<void> {
+        const schedulePoll = async () => {
```
teogeb: Could use `scheduleAtInterval` helper method from `src/helpers/scheduler.ts`

packages/broker/test/unit/plugins/storage/StorageConfig.test.ts
```diff
@@ -1,146 +1,178 @@
...
+function makeStubStream(streamId: string): Stream {
+    return {
+        id: toStreamID(streamId),
+        partitions: PARTITION_COUNT_LOOKUP[streamId]
+    } as Stream
+}
+
+describe(StorageConfig, () => {
+    let getStoredStreamsOf: jest.Mock<Promise<{ streams: Stream[], blockNumber: number }>, [nodeAddress: string]>
+    let emitStorageAssignmentEvent: ((event: EthereumStorageEvent) => any) | undefined
```
teogeb: Rename to `storageEventListener` (as already in`StorageEventListener.test.ts`)

packages/broker/test/unit/plugins/storage/StorageEventListener.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    let storageEventListener: ((event: EthereumStorageEvent) => any) | undefined
+    let handleEvent: jest.Mock<void, [stream: Stream, type: 'added' | 'removed', block: number]>
+    let listener: StorageEventListener
+
+    beforeEach(() => {
+        stubClient = {
+            async getStream() {
+                return {
+                    id: 'streamId',
+                    partitions: 0
```
teogeb: The `partitions` should be > 0. 

(Maybe good to use a value > 1, as multi-partition situation is not the most typical situation and therefore worth checking. Not very important in this as we don't have any partition processing here.)

packages/client/src/Encrypt.ts
```diff
@@ -15,7 +15,7 @@ export default class PublisherEncryption implements Stoppable {
 
     constructor(
         private streamEndpoints: StreamEndpointsCached,
-        private keyExchange: PublisherKeyExchange,
+        @inject(delay(() => PublisherKeyExchange)) private keyExchange: PublisherKeyExchange,
```
teogeb, line 18: Does this fix something related to this PR?

teogeb (commented):
Great improvement 👍

packages/broker/src/plugins/storage/SetMembershipSynchronizer.ts
harbu: Removed.

packages/broker/src/plugins/storage/StorageConfig.ts
harbu: Removed

packages/broker/src/plugins/storage/SetMembershipSynchronizer.ts
harbu: Good name. Updated.

packages/broker/src/plugins/storage/StoragePoller.ts
harbu: Renamed

packages/broker/src/plugins/storage/StorageEventListener.ts
harbu: Renamed

packages/broker/src/plugins/storage/StoragePoller.ts
harbu: fixed

packages/broker/test/unit/plugins/storage/StorageConfig.test.ts
harbu: fix

packages/broker/test/unit/plugins/storage/StorageEventListener.test.ts
harbu: Fixed

packages/client/src/Encrypt.ts
harbu: Yeah the import paths have changed as a result of these changes, thereby causing circular dependencies.

teogeb (approved):

PR #393, opened 2022-01-28
NET-682: fix: pix possible issue with a stream being returned before it was cr…

samt1803:
…eated and synched

### DIFF
--- a/packages/client/src/StreamRegistry.ts
+++ b/packages/client/src/StreamRegistry.ts
@@ -194,17 +194,17 @@ export class StreamRegistry implements Context {
                 know what the actual error was. (Most likely it has anything to do with timeout
                 -> we don't use the error from until(), but throw an explicit error instead.)
             */
-            try {
-                await until(async () => { return this.streamExistsOnTheGraph(streamId) }, 20000, 500)
-            } catch (e) {
-                throw new Error(`unable to create stream "${streamId}"`)
-            }
         } else {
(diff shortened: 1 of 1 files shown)


harbu (approved):
If I understood correctly, the change here is that we poll the graph now also in non-ens domain cases. LGTM; @teogeb do you want to have a look as well, since you may be aware of any subtleties as I'm not as familiar with this part of the code.

teogeb (approved):
Would we like to add a similar check to `updateStream()`? 

That way we'd guarantee that if you use client to write something to the the contract registry, the operation is completed once it's propagated to The Graph. The definition of "completeness" would be consistent in that sense.

harbu:
> That way we'd guarantee that if you use client to write something to the the contract registry, the operation is completed once it's propagated to The Graph. The definition of "completeness" would be consistent in that sense.

I wonder if it would be useful in general to abide by a principle in the client that whenever we do an operation / transaction, we consider it "completed" only once it shows up in The Graph?

samt1803:
There is this ticket to make all transactions configurable (or global config) to wait for being mined or wait for graph-sync. 
https://linear.app/streamr/issue/NET-606/configurable-wait-parameters-for-transaction-generating-call-on-client
To mitigate the user creating this error the default could be the graph-synced state.
Or we could just always wait for theGraph to minimize errors.

harbu:
Yeah let's think about solving this in general in another ticket. I think this one is good to go, so feel free to merge at your convenience!

PR #396, opened 2022-01-31
Fix ethereum networks in schema

jtakalai:
also refactored the structure a bit to look more like https://linear.app/streamr/issue/ETH-184/config-package

### DIFF
--- a/packages/client/src/StreamRegistry.ts
+++ b/packages/client/src/StreamRegistry.ts
@@ -95,8 +95,7 @@ export class StreamRegistry implements Context {
         this.chainProvider = this.ethereum.getStreamRegistryChainProvider()
         this.streamRegistryContractReadonly = new Contract(this.config.streamRegistryChainAddress,
             StreamRegistryArtifact, this.chainProvider) as StreamRegistryContract
-        const chainId = this.config.streamRegistryChainRPC?.chainId
-        this.defaultOverrides = this.config.ethereumNetworks?.find((network) => network.chainId === chainId)?.overrides ?? {}
+        this.defaultOverrides = this.config.ethereumNetworks?.polygon?.overrides ?? {}
     }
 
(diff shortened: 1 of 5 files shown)


packages/client/src/StreamRegistry.ts
```diff
@@ -95,8 +95,7 @@ export class StreamRegistry implements Context {
         this.chainProvider = this.ethereum.getStreamRegistryChainProvider()
         this.streamRegistryContractReadonly = new Contract(this.config.streamRegistryChainAddress,
             StreamRegistryArtifact, this.chainProvider) as StreamRegistryContract
-        const chainId = this.config.streamRegistryChainRPC?.chainId
-        this.defaultOverrides = this.config.ethereumNetworks?.find((network) => network.chainId === chainId)?.overrides ?? {}
+        this.defaultOverrides = this.config.ethereumNetworks?.polygon?.overrides ?? {}
```
harbu: If you look at the file `ConfingBase.ts`, you will see the following entry:
```ts
streamRegistryChainRPC: {
        url: 'https://polygon-rpc.com',
        chainId: 137
}
```
Thus we leave it open to configuration whether it should use polygon network (or some other) network to deal with stream registries.

So I don't think we should hardcode `polygon` name here. I'm not sure what the intended configuration design here is, but you should probably lookup `streamRegistryChainRPC.chainId` and use that to filter the correct `etherumNetworks` entry?

harbu (changes requested):

packages/client/src/StreamRegistry.ts
jtakalai: You're right it looks very much off inside that file. The idea is definitely to add an indirection there, say, streamRegistryChain would specify `polygon` as the chain to use for stream registry. I could add that indirection right now, too.

PR #397, opened 2022-01-31
ETH-215: Search stream by id and permissions

teogeb:
The `client.searchStreams` can now search over the id field and stream permissions using the following API:

```ts
searchStreams(term: string | undefined, permissionFilter: SearchStreamsPermissionFilter | undefined): AsyncGenerator<Stream>

export interface SearchStreamsPermissionFilter {
    user: EthereumAddress
    allOf?: StreamPermission[]
    anyOf?: StreamPermission[]
    allowPublic: boolean
}
```

The new return type is `AsyncGenerator` instead of `Promise<Stream[]>`. 

Changed also return types of `getAllStreams`, `getStreamPublishers`, and `getStreamSubscribers` to `AsyncGenerators` as paging is now done automatically.

Updated CLI-tool command `stream search` to use the new search method.

### DIFF
--- a/packages/cli-tools/README.md
+++ b/packages/cli-tools/README.md
@@ -67,11 +67,43 @@ Useful for generating test data to be published to a stream with `publish`, e.g.
 streamr mock-data generate | streamr stream publish <streamId> --private-key <key>
 ```
 
-### list
-Fetch a list of streams that are accessible to the user authenticated by the private key
+### search
+Query a list of streams by a search term and/or permissions. E.g.:
 ```
-streamr stream list --private-key <key>
+streamr stream search foobar --user 0x1234567890123456789012345678901234567890
 ```
+
+#### Search term
+A search term query searchers over the stream id field. E.g:
+```
+streamr stream search foobar
+```
+It could find these streams:
+```
+0x1234567890123456789012345678901234567890/abc/foobar/1
+foobar.eth/lorem-ipsum
+```
+
+#### Permission
+A permission query searches over stream permissions. You can either query by direct permissions (which are explicitly granted to a user), or by all permissions (including public permissions, which apply to all users).
+
+E.g. all streams where a user has some direct permission:
+```
+streamr stream search --user 0x1234567890123456789012345678901234567890
+```
+All streams accessible by a user:
+```
+streamr stream search --user 0x1234567890123456789012345678901234567890 --public
+```
+
+The argument of the `--user` option can be omitted. In that case, it defaults to the authenticated user (specified by `--private-key`).
+
+It is also possible to filter by specific permissions by using `--all` and `--any`. E.g. if you want to find the streams you can subscribe to:
+```
+streamr stream search --user --public --all subscribe --private-key <key>
+```
+
+If more than one permission is needed, specify the permissions in a comma-separated list (e.g. `--all subscribe,publish`). It returns streams where _all_ listed permissions are granted. If just _any_ of the permissions is required, use `--any` instead of `--all`. Please prefer `--all` to `--any` when possible as it has better query performance.
 
 ### show
 Show detailed information about a specific stream
--- a/packages/cli-tools/bin/streamr-stream-search.ts
+++ b/packages/cli-tools/bin/streamr-stream-search.ts
@@ -1,14 +1,54 @@
 #!/usr/bin/env node
 import '../src/logLevel'
-import StreamrClient from 'streamr-client'
+import StreamrClient, { SearchStreamsPermissionFilter, StreamPermission } from 'streamr-client'
 import { createClientCommand } from '../src/command'
+import { Option } from 'commander'
+import { getPermission, PERMISSIONS } from '../src/permission'
+import { getOptionType, OptionType } from '../src/common'
 
-createClientCommand(async (client: StreamrClient, term: string) => {
-    const streams = client.searchStreams(term)
+const createPermissionFilter = async (
+    user: string | boolean | undefined,
+    allowPublic: boolean | undefined,
+    allOf: StreamPermission[] | undefined,
+    anyOf: StreamPermission[] | undefined,
+    client: StreamrClient
+): Promise<SearchStreamsPermissionFilter| undefined> => {
+    if (user !== undefined) {
+        return {
+            user: (getOptionType(user) === OptionType.ARGUMENT) ? user as string : await client.getAddress(),
+            allowPublic: allowPublic ?? false,
+            allOf,
+            anyOf
+        }
+    } else if ((allowPublic !== undefined) || (allOf !== undefined) || (anyOf !== undefined)) {
+        console.error('specify a user with "--user" when using "--public", "--all" or "--any"')
+        process.exit(1)
+    }
+}
+
+const createPermissionListOption = (id: string) => {
+    return new Option(`--${id} <permissions>`, 'comma-separated list of permissions')
+        .choices(Array.from(PERMISSIONS.keys()))
+        .argParser((value: string) => value.split(',').map((id) => getPermission(id)))
+}
+
+createClientCommand(async (client: StreamrClient, term: string, options: any ) => {
+    const permissionFilter = await createPermissionFilter(
+        options.user,
+        options.public,
+        options.all,
+        options.any,
+        client
+    )
+    const streams = client.searchStreams(term, permissionFilter)
     for await (const stream of streams) {
         console.log(stream.id)
     }
 })
-    .arguments('<term>')
+    .arguments('[term]')
     .description('search streams')
+    .option('--user [user]', 'a stream must have permissions for the given user, defaults to the authenticated user')
+    .option('--public', 'the permission can be implicit (a public permission to the stream)')
+    .addOption(createPermissionListOption('all'))
+    .addOption(createPermissionListOption('any'))
     .parseAsync()
--- a/packages/cli-tools/bin/streamr-stream-search.ts
+++ b/packages/cli-tools/bin/streamr-stream-search.ts
@@ -32,7 +32,7 @@ const createPermissionListOption = (id: string) => {
         .argParser((value: string) => value.split(',').map((id) => getPermission(id)))
 }
 
-createClientCommand(async (client: StreamrClient, term: string, options: any ) => {
+createClientCommand(async (client: StreamrClient, term: string | undefined, options: any ) => {
     const permissionFilter = await createPermissionFilter(
         options.user,
         options.public,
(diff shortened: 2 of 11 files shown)


packages/cli-tools/README.md
```diff
@@ -67,12 +67,44 @@ Useful for generating test data to be published to a stream with `publish`, e.g.
...
+```
+
+The argument of the `--user` option can be omitted. In that case, it defaults to the authenticated user (specified by `--private-key`).
+
+It is also possible to filter by specific permissions by using `--all` and `--any`. E.g. if you want to find the streams you can subscribe to:
+```
+streamr stream search --user --public --all subscribe --private-key <key>
+```
+
+If more than one permission is needed, specify the permissions in a comma-separated list (e.g. `--all subscribe,publish`). It returns streams where _all_ listed permissions are granted. If just _any_ of the permissions is required, use `--any` instead of `--all`. Please prefer `--all` to `--any` when possible as it has better query performance.
```
harbu, line 106: Good documentation! :+1:

packages/cli-tools/bin/streamr-stream-search.ts
```diff
@@ -1,17 +1,54 @@
...
+
+createClientCommand(async (client: StreamrClient, term: string, options: any ) => {
+    const permissionFilter = await createPermissionFilter(
+        options.user,
+        options.public,
+        options.all,
+        options.any,
+        client
+    )
+    const streams = client.searchStreams(term, permissionFilter)
```
harbu, line 43: Could the `term` be optional? If so, should its type be `string | undefined`?

packages/cli-tools/src/permission.ts
```diff
@@ -20,6 +20,14 @@ export const PERMISSIONS = new Map<string,StreamPermission>([
     ['grant', StreamPermission.GRANT]
 ])
 
+export const getPermission = (id: string): StreamPermission => {
```
harbu:
```suggestion
export const getPermission = (id: string): StreamPermission | never => {
```

packages/client/src/utils/GraphQLClient.ts
```diff
@@ -0,0 +1,72 @@
...
+                accept: '*/*',
+            },
+            body: gqlQuery
+        })
+        const resText = await res.text()
+        let resJson
+        try {
+            resJson = JSON.parse(resText)
+        } catch {
+            throw new Error(`GraphQL query failed with "${resText}", check that your theGraphUrl="${this.config.theGraphUrl}" is correct`)
```
harbu, line 35: This is actually a very good approach for getting descriptive error messages (with enough context) as opposed to just doing `await res.json()`.

packages/client/src/utils/GraphQLClient.ts
```diff
@@ -0,0 +1,72 @@
...
+        pageSize = 1000
+    ): AsyncGenerator<T, void, undefined> {
+        let lastResultSet: T[] | undefined
+        do {
+            const lastId = (lastResultSet !== undefined) ? lastResultSet[lastResultSet.length - 1].id : ''
+            const query = createQuery(lastId, pageSize)
+            // eslint-disable-next-line no-await-in-loop
+            const response = await this.sendQuery(query)
+            const rootKey = Object.keys(response)[0] // there is a always a one root level property, e.g. "streams" or "permissions"
+            const items: T[] = (response as any)[rootKey] as T[]
```
harbu, line 59: Method is a bit tricky type-wise, but yeah.. we don't really have any compile-time info to work with here...

packages/client/src/searchStreams.ts
```diff
@@ -0,0 +1,123 @@
...
+    user: EthereumAddress
+    /*
+     * If possible, prefer allOf to anyOf because the query performance is better
+     */
+    allOf?: StreamPermission[]
+    anyOf?: StreamPermission[]
+    allowPublic: boolean
+}
+
+export type SearchStreamsQueryItem = {
```
harbu: I find the name a bit confusing, isn't this part of the result set of the query? I thought this interface somehow relates to building the query. What about something like `SearchStreamsResultItem`?

packages/client/src/searchStreams.ts
```diff
@@ -0,0 +1,123 @@
...
+    permissionFilter: SearchStreamsPermissionFilter | undefined,
+    graphQLClient: GraphQLClient,
+): AsyncGenerator<SearchStreamsQueryItem> {
+    const backendResults = graphQLClient.fetchPaginatedResults<SearchStreamsQueryItem>(
+        (lastId: string, pageSize: number) => buildQuery(term, permissionFilter, lastId, pageSize)
+    )
+    /*
+     * There can be orphaned permission entities if a stream is deleted (currently
+     * we don't remove the assigned permissions, see ETH-222)
+     * TODO remove the filtering when ETH-222 has been implemented
```
harbu, line 35: Good clarification :+1:

packages/client/src/searchStreams.ts
```diff
@@ -0,0 +1,123 @@
...
+     */
+    const withoutOrphaned = filter(backendResults, (p) => p.stream !== null)
+    /*
+     * As we query via permissions entity, any stream can appear multiple times (once per
+     * permission user) if we don't do have exactly one userAddress in the GraphQL query.
+     * That is the case if no permission filter is defined at all, or if permission.allowPublic
+     * is true (then it appears twice: once for the user, and once for the public address).
+     */
```
harbu, line 43: :+1:

packages/client/src/searchStreams.ts
```diff
@@ -0,0 +1,123 @@
...
+         * entity. Therefore we may receive an assignment which doesn't have any permissions granted.
+         * Similar situation may happen also when some expirable permission (subscribe or publish)
+         * expire as the expiration doesn't trigger the removal of the Permission entity.
+         * -> To filter out these empty assignments, we define a fallback value for anyOf filter
+         */
+        const anyOf = permissionFilter.anyOf ?? Object.values(StreamPermission) as StreamPermission[]
+        yield* filter(withoutDuplicates, (item: SearchStreamsQueryItem) => {
+            const actual = StreamRegistry.getPermissionsFromChainPermissions(item)
+            return anyOf.some((p) => actual.includes(p))
+        })
```
harbu: I'm gonna need some help understanding this part.

packages/client/src/searchStreams.ts
```diff
@@ -0,0 +1,123 @@
...
+    } else {
+        yield* withoutDuplicates
+    }
+}
+
+/*
+ * Note that we query the results via permissions entity even if there is no permission filter
```
harbu, line 67: :+1:

packages/client/src/searchStreams.ts
```diff
@@ -0,0 +1,123 @@
...
+ * ID type is not a string)
+ */
+const buildQuery = (
+    term: string | undefined,
+    permissionFilter: SearchStreamsPermissionFilter | undefined,
+    lastId: string,
+    pageSize: number
+): string => {
+    const variables: Record<string, any> = {
+        stream_contains: term,
```
harbu, line 80: What does `stream_contains` actually search over?

packages/client/src/utils/GraphQLClient.ts
```diff
@@ -0,0 +1,72 @@
...
+                throw new Error('GraphQL query failed: ' + JSON.stringify(resJson.errors.map((e: any) => e.message)))
+            } else {
+                throw new Error('GraphQL query failed')
+            }
+        }
+        return resJson.data
+    }
+
+    async* fetchPaginatedResults<T extends { id: string }>(
+        createQuery: (lastId: string, pageSize: number) => string,
```
harbu, line 49: Would it be worth adding a clarification to this callback function that `lastId` will be the empty string on the first query?

packages/client/src/searchStreams.ts
```diff
@@ -0,0 +1,123 @@
...
+    }
+    const query = `
+        query (
+            $stream_contains: String,
+            $userAddress_in: [Bytes!]
+            $canEdit: Boolean
+            $canDelete: Boolean
+            $publishExpiration_gt: BigInt
+            $subscribeExpiration_gt: BigInt
+            $canGrant: Boolean
```
harbu, line 105: I'm assuming GraphQL is "smart" here and we don't need to remove these from the query string in case `permissionFilter === undefined`?

packages/client/test/integration/SearchStreams.test.ts
```diff
@@ -0,0 +1,201 @@
...
+
+    let client: StreamrClient
+    let streamWithoutPermission: Stream
+    let streamWithUserPermission: Stream
+    let streamWithPublicPermission: Stream
+    let streamWithUserAndPublicPermission: Stream
+    let streamWithGrantedAndRevokedPermission: Stream
+    const searcher = Wallet.createRandom()
+
+    const waitUntilStreamsExistOnTheGraph = async (streams: Stream[]) => {
```
harbu, line 21: I'm guessing this is no longer needed after #393 is merged to main.

harbu (commented):
Let's go thru a few points via call, also wondering if we should have a unit test for `GraphQLClient`, especially w.r.t. its paging functionality?

packages/client/src/StreamRegistry.ts
```diff
@@ -480,6 +448,14 @@ export class StreamRegistry implements Context {
 
     /* eslint-disable padding-line-between-statements */
     private static getPermissionsAssignment(permissionResult: PermissionQueryResult): PermissionsAssignment {
+        return {
+            address: permissionResult.userAddress,
+            permissions: StreamRegistry.getPermissionsFromChainPermissions(permissionResult)
```
harbu: Maybe name should be `convert`?

packages/client/src/searchStreams.ts
harbu: Good stuff, maybe an example could make it crystal clear.

packages/client/src/utils/GraphQLClient.ts
harbu: As this class is The Graph specific, the empty value actually makes sense in all queries. So it is most likely fine as it is.

packages/client/test/integration/SearchStreams.test.ts
harbu: Never mind, leave it in as a follow up PR to #393 will change the behavior not rely on the graph.

harbu:
> Let's go thru a few points via call, also wondering if we should have a unit test for `GraphQLClient`, especially w.r.t. its paging functionality?

So writing a test for the paging will be a laborious task. I suggest we leave it be as is for now, and come back if it fails at least once.

harbu (approved):
LGTM; feel free to merge after agreed upon changes

PR #398, opened 2022-01-31
NET-615: Client: remove duplicate id field from Stream class

teogeb:
Removed the legacy `streamId` field from `Stream` class. The `id` should be used instead of that.


harbu (approved):
LGTM; but I think there may be more places where name `streamId` is still relied on. Let's see what tests say.

PR #401, opened 2022-02-01
ETH-224: feat: reduced reliance on theGraph during stream creation.

samt1803:
It will check directly onchain if a stream exists, and during creation without ens it will just wait for the create transaction to be mined.

### DIFF
--- a/packages/client/src/StreamRegistry.ts
+++ b/packages/client/src/StreamRegistry.ts
@@ -189,8 +189,8 @@ export class StreamRegistry implements Context {
             /*
                 The call to createStreamWithENS delegates the ENS ownership check, and therefore the
                 call doesn't fail e.g. if the user doesn't own the ENS name. To see whether the stream
-                creation succeeeds, we need to poll The Graph. If the polling timeouts, we don'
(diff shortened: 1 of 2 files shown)


packages/client/src/StreamRegistry.ts
```diff
@@ -193,15 +193,16 @@ export class StreamRegistry implements Context {
                 know what the actual error was. (Most likely it has anything to do with timeout
```
harbu: Let's keep comment up-to-date. Maybe something like

```ts
/*
The call to createStreamWithENS delegates the ENS ownership check, and therefore the
call doesn't fail e.g. if the user doesn't own the ENS name. To see whether the stream
creation succeeeds, we need to poll the chain for stream existence. If the polling timeouts, we don't
know what the actual error was. (Most likely it has nothing to do with timeout
-> we don't use the error from until(), but throw an explicit error instead.)
*/
```

samt1803: Right! I didn't read that comment and will change it.

harbu (approved):
LGTM

PR #402, opened 2022-02-01
[NET-678] Client: drop support for explicit RSA key pair for group key exchange

teogeb:
We don't currently support that users could provide explicit RSA key pairs for group key exchange. That feature is controlled by `keyExchange` configuration option. Removed that option, and some other unused config options.

### Group keys:
- `keyExchange` 
- `subscriberGroupKeys`
- `publisherGroupKeys`
- `publisherStoreKeyHistory`

Also removed a constructor of `Encryption` class as it was not used in production code. `KeyExchangeSubscriber` always calls it without arguments. The constructor took a `privateKey` + `publicKey` object as parameter, so we could have called it with a `keyExchange` object. (But to be compatible with the documentation, we should have called the constructor, and also handled the case of `null` keyExchange.)

### Publish

- `maxPublishQueueSize`
- `publishWithSignature`
- `publishAutoDisconnectDelay`

Also added type annotations for `verifySignatures`.


harbu (approved):
lgtm :+1:

PR #409, opened 2022-02-02
Client storage node config options

teogeb:
Rename client config option `nodeRegistryChainAddress` -> `storageNodeRegistryChainAddress`

Remove deprecated client config option `storageNodeRegistry` (was not used in the client)

Rename client class `NodeRegistry` to `StorageNodeRegistry`

Rename `StorageNodeRegistry` methods:
-  `setNode` -> `createOrUpdateNodeInStorageNodeRegistry`
- `remove` -> `removeNodeFromStorageNodeRegistry`

Remove `storageNode` config option from `MetricsPlugin` (was not used by the plugin).


packages/client/src/getStorageNodeRegistryFromContract.ts
```diff
@@ -1,40 +0,0 @@
-import { JsonRpcProvider } from '@ethersproject/providers'
-import { Contract } from '@ethersproject/contracts'
-import { createStorageNodeRegistry, StorageNodeInfo, StorageNodeRegistry } from 'streamr-client-protocol'
```
harbu, line 3: Now that this file is deleted, I wonder if we are using `StorageNodeRegistry` from protocol anywhere (else than tests)? Maybe it can be removed as well...

harbu (approved):

packages/client/src/getStorageNodeRegistryFromContract.ts
harbu: never mind :D

PR #410, opened 2022-02-02
cleanup CI jobs

harbu:
- Make client integration tests run without Core API
    - Everything that requires Core API should be maybe run with data union tests from now on
- Move `LoginEndpoints.test.ts`, `Session.test.ts` and `authFetch.test.ts` tests to the dataUnion directory (disables the test, as all DU tests are currently disabled). These functionalities are used in DU-related operations. DU tests have a dependency to Core-API, while other clients tests don't have that
- Adjust  `parity-node0 `/ `parity-sidechain-node0` dependencies
- Remove `loginWithApiKey` and `loginWithUsernamePassword` methods and functionality
- Harmonize GitHub Actions workflow naming


harbu (commented):
Approve

PR #411, opened 2022-02-02
Rework client-layer proxy publishing

juslesan:
This PR fixes a problem where the client will default to using regular stream connections if the client starts publishing before the proxy connection is formed.

- Add ability to add and remove event listeners for node events related to proxy connections
- When calling `await setPublishProxy()` on the client it will now wait for such an event from the NetworkNode before resolving.

### DIFF
--- a/packages/client/src/BrubeckNode.ts
+++ b/packages/client/src/BrubeckNode.ts
@@ -9,8 +9,6 @@ import { Config } from './Config'
 import { StreamMessage, StreamPartID } from 'streamr-client-protocol'
 import { DestroySignal } from './DestroySignal'
 import Ethereum from './Ethereum'
-import { node } from 'webpack'
-import { str } from 'ajv'
 
 /**
  * Wrap a network node.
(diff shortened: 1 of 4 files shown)


packages/client/src/BrubeckNode.ts
```diff
@@ -170,12 +170,28 @@ export default class BrubeckNode implements Context {
...
+                    this.cachedNode!.addPurePublishingAcceptedListener((node, stream) => {
+                        if (node === nodeId && stream === streamPartId) {
+                            resolve()
+                        }
+                    })
+                    this.cachedNode!.addPurePublishingRejectedListener((node, stream) => {
+                        if (node === nodeId && stream === streamPartId) {
+                            reject()
+                        }
+                    })
```
harbu: To me it seems like this check should be included in the method `joinStreamPartAsPurePublisher`, hiding the implementation details from the "user" of network (i.e. client).

packages/client/src/BrubeckNode.ts
```diff
@@ -170,12 +170,28 @@ export default class BrubeckNode implements Context {
...
+                }),
+                this.cachedNode!.joinStreamPartAsPurePublisher(streamPartId, nodeId)
+            ])
+
         } finally {
             this.debug('openProxyConnectionOnStream << %o', streamPartId, nodeId)
+            this.cachedNode!.removePurePublishingAcceptedListener(() => {})
+            this.cachedNode!.removePurePublishingRejectedListener(() => {})
```
harbu: I believe you need to pass in the same function references as you passed to the `on` listeners.

E.g.
```ts
const handlerFn = (node, stream) => {
    if (node === nodeId && stream === streamPartId) {
        resolve()
    }
}
this.cachedNode!.addPurePublishingAcceptedListener(handlerFn)
...
this.cachedNode!.removePurePublishingAcceptedListener(handlerFn)
```

harbu (changes requested):

packages/client/src/BrubeckNode.ts
juslesan: Sure, I'll move it to the NetworkNode

harbu (approved):
LGTM; feel free to land when CI green

PR #413, opened 2022-02-02
client tests: simplify dev environment storage node handling

teogeb:
Modified the configuration of the storage node which is available when running the stack in `streamer-docker-dev `environment. 

The address is now configured as `DOCKER_DEV_STORAGE_NODE` in` ConfigTest.ts`. Other aspects of the storage node configs (url and private key) are not needed in tests. 

Removed also some unnecessary client instances which were created for storage node initialization.


harbu (approved):
lgtm

PR #414, opened 2022-02-02
broker: remove /api/v1 prefix from Metrics plugin /volume endpoint

teogeb:
Changed the url of Metrics plugin: `/api/v1/volume` -> `/volume`. Now the url mapping is consistent with other plugin endpoints (https://github.com/streamr-dev/network-monorepo/pull/408).


harbu (approved):
LGTM; let's merge _after_ streamr-docker-dev change

PR #415, opened 2022-02-03
Lazy start up of KeyServer

harbu:
Defer starting up `KeyServer` to the first time a private key with gas is request from it. The benefit being that the `KeyServer` doesn't have to be started if the test cases being run don't require it. Often we will want to run some unit test repeatedly when developing a new feature or such. It makes the turn around time faster not to keep starting  / stopping the `KeyServer` for nothing.

Also rename the methods to get private keys from the server as `fetchPrivateKeyWithGas`, clearly revealing the intention that the private key fetched should indeed be one with gas.

### DIFF
--- a/packages/broker/jest.config.js
+++ b/packages/broker/jest.config.js
@@ -19,7 +19,7 @@ module.exports = {
     // The directory where Jest should output its coverage files
     coverageDirectory: 'coverage',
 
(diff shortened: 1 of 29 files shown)


juslesan (approved):
LGTM

PR #416, opened 2022-02-03
Client: add debug-level logging for transactions

harbu:
Every time a transaction in `StorageNodeRegistry` or `StreamRegistry` is submitted to the blockchain, log debug-level information about it. Require `DEBUG=Streamr*` environment variable set.

E.g. output
```
Streamr:waitForTx transaction submitted { tx=0x2d20e01ac7065e18f80890f51ad2b6457a228670288438d76410312a9c747105, to=0x6cC...f222, nonce=60, gasLimit=210493, gasPrice=1100000000 } +0ms
Streamr:waitForTx transaction confirmed { tx=0x2d20e01ac7065e18f80890f51ad2b6457a228670288438d76410312a9c747105, block=36802, confirmations=2, gasUsed=207622, events=["StreamCreated","PermissionUpdated"] } +4s
```


samt1803 (approved):
LGTM!

teogeb (approved):
Good improvement 👍 Later we could also enhance the logging by including the name of the contract method. The `waitForTx` could see it automatically e.g. if we use something like this:

```js
const waitForTx = async (fn, ...params) => {
    console.log(`executing contract function ${fn.name}`)
    const tx = await fn.apply(undefined, params)
    console.log(`tx.hash=${tx.hash}`)
    // ...
    const receipt = tx.wait()
    // ...
    return receipt
}

const fakeContractFunction = async (s, n) => {
    console.log(`the fake method received params s=${s}, n=${n}`)
    return Promise.resolve({ hash: 'fake-hash', wait: () => 'fake-receipt' })
}

const main = async () => {
    const receipt = await waitForTx(fakeContractFunction, 'abc', 123)
    console.log(`receipt=${receipt}`)
}
main()
```

(we could investigate if TypeScript supports this well enough)

harbu:
> Good improvement +1 Later we could also enhance the logging by including the name of the contract method. The `waitForTx` could see it automatically e.g. if we use something like this:
> 
> ```js
> const waitForTx = async (fn, ...params) => {
>     console.log(`executing contract function ${fn.name}`)
>     const tx = await fn.apply(undefined, params)
>     console.log(`tx.hash=${tx.hash}`)
>     // ...
>     const receipt = tx.wait()
>     // ...
>     return receipt
> }
> 
> const fakeContractFunction = async (s, n) => {
>     console.log(`the fake method received params s=${s}, n=${n}`)
>     return Promise.resolve({ hash: 'fake-hash', wait: () => 'fake-receipt' })
> }
> 
> const main = async () => {
>     const receipt = await waitForTx(fakeContractFunction, 'abc', 123)
>     console.log(`receipt=${receipt}`)
> }
> main()
> ```
> 
> (we could investigate if TypeScript supports this well enough)

Good idea, should we try out in a follow up PR? (I want to make sure we spend enough time doing the types correctly and making sure it prints the right function name.)

PR #420, opened 2022-02-03
Enable browser-realtime tests

juslesan:
After testing the browser-realtime tests locally and CI there appears to be no flakiness in the client browser real-time tests


harbu (approved):
lgtm; let's merge to main

PR #425, opened 2022-02-04
split client integration tests into parallel jobs

harbu:
A quick ad-hoc attempt at alleviating the slow client integration tests.

Basically I divided the client integration tests (excluding dataunion-related) into five separate sub-folders. Then we configure Github Actions to run each sub-folder as a separate job to gain advantage of parallelization. 

I believe this might give us faster tests in the short term, but we need to figure out whether we can make the client tests faster by, for example, reducing smart contract interaction and waiting.


harbu:
Looks like this approach allows us to get < 20min run time, if we could get docker images used by `streamr-docker-dev` cached, we could get that probably even lower.

harbu:
Closing this PR. A better approach to solving slow tests is approached in this PR https://github.com/streamr-dev/network-monorepo/pull/432 instead.

PR #426, opened 2022-02-04
Log client transactions and give context for errors

teogeb:
Improved contract transaction logging and error. 

It is implemented by wrapping a `Contract`'s functions with delegator functions. The properties of the contract object are read-only -> it is not possible to modify the existing function definitions. (Also the functions are implemented as anonymous functions, and therefore it is not possible to get the function name just by querying  `function.name`).

### New features

If a transaction method execution (call or send) throws an `Exception`, the message contains the method name in the error message. E.g.:
```
Error: Error in contract call "streamRegistry.createStream"
  ...
  reason: Error: could not detect network (event="noNetwork", code=NETWORK_ERROR, version=providers/5.5.2)
  ...
```

Transaction log contains the name of the contract method. E.g.:
```
transaction submitted { method=streamStorageRegistry.addStorageNode, ... }
transaction confirmed { method=streamStorageRegistry.addStorageNode, ... }
```

The Ethereum address is printed as a human-readable name if known. E.g.: 
```
transaction submitted { ..., to=DOCKER_DEV_STREAM_STORAGE_REGISTRY, ... }
```

There are some human-readable addresses configured in `network/NameDirectory.ts`. It helps us to identify the services in Streamr docker dev environment. In other environments (and if the human-readable name is not known) the address will be printed as shortened Ethereum address. E.g.:
```
transaction submitted { ... to=0x123456, ... }
```

All transaction method executions are logged. E.g.:
```
execute streamRegistry.getStreamMetadata
```

### Future improvements
- We could use these logging + error handling improvements in all contract handling? (e.g. DataUnion contract and tracker registry)
- We could add all docker-dev addresses to `NameDirectory` if we find the human-readable names useful. Maybe that makes sense when the list of addresses are easily available in some always up-to-date JSON.


samt1803 (approved):
LGTM

packages/client/src/utils/contract.ts
```diff
@@ -0,0 +1,123 @@
...
+                (receipt.events || []).map((e) => e.event)
+            )
+        }
+    }
+}
+
+const withErrorHandling = async <T>(
+    execute: () => Promise<T>,
+    methodName: string
+): Promise<T> | never => {
```
harbu:
```suggestion
): Promise<T> => {
```

packages/client/src/utils/contract.ts
```diff
@@ -0,0 +1,123 @@
...
+
+const createWrappedContractMethod = (
+    originalMethod: (...args: any) => Promise<any>,
+    methodName: string,
+    logger: ContractLogger,
+) => {
+    return async (...args: any) => {
+        logger.onMethodExecute(methodName)
+        const returnValue = await withErrorHandling(() => originalMethod(...args), methodName)
+        if (isTransaction(returnValue)) {
```
harbu, line 76: Are there some specific circumstances under which the returned value is not a transaction?

harbu (approved):
Do we need to remove the individual `waitForTx` calls from classes `StorageNodeRegistry` and `StreamRegistry` now that the wrapper takes care of the transformation?

packages/client/src/utils/contract.ts
teogeb: The value is on not a transaction is this is a read-only call for querying information from the contract.

harbu: Ah of course, makes sense

teogeb:
> Do we need to remove the individual `waitForTx` calls from classes `StorageNodeRegistry` and `StreamRegistry` now that the wrapper takes care of the transformation?

We can't remove the `waitForTx` calls we want that the wrapped and non-wrapped usages are interchangeable.

For users of the contract, the method calls appear 100% compatible to the original ones. E.g calling e.g. `streamRegistry.createStream()` returned Promise<ContractTransaction>, and now it returns `Promise<{ wait: ... }>` . To convert `ContractTransaction` to `ContractReceipt`, the wait should be called by some component (e.g. the via `waitForTx` helper).

PR #429, opened 2022-02-04
NET-641

juslesan:
The problem in the Group Key Persistence is not related to the functionality itself but a network layer issue. The problem is solved by waiting for neighbors on the subscriber before publishing.

Other changes:

- Re-enabled skipped test `subscriber persists group key with resend last` which is no longer flaky

Related Issues?:

- Client's `subscription.collect(n)` resolves even if n messages are not received.


harbu:
@juslesan is this what you are looking for https://github.com/streamr-dev/network-monorepo/blob/938bb71a271ddff69f485db93577e5c9b877fc2e/packages/client/src/Subscription.ts#L31-L33?

juslesan:
> @juslesan is this what you are looking for
> 
> https://github.com/streamr-dev/network-monorepo/blob/938bb71a271ddff69f485db93577e5c9b877fc2e/packages/client/src/Subscription.ts#L31-L33
> 
> ?

This seems to work yes.

However, it feels a bit off to call` await subscribe()` and not have it actually await for the subscribe operation to fully resolve.

harbu (approved):

PR #432, opened 2022-02-07
Client fake test environment

teogeb:
Added fake versions of `BrubeckNode` and blockchain registry facades. We can use these classes in tests to run the test in a fake environment. In the fake environment all stream-related smart contract calls (`StreamRegistry` and `StorageNodeRegistry`) and messaging between nodes are emulated. 

Using a fake environment makes some tests are significantly faster. In this PR we applied the fake environment for two slow tests: 
- `Encryption.test.ts`: speed-up 420s -> 57s
- `GroupKeyPersistence.test.ts`: speed-up 225s -> 9s

In those two tests the functionality under the test is not related to smart contracts or the messaging between the nodes (the tests send messages, but there is nothing special about the messaging -> no need to test that aspect here).

We can make separate PRs for applying the fake environment for other tests. We should not apply the fake environment for all integration tests (e.g. if the main focus of a test is messaging and/or smart contracts), but maybe it is applicable for some of the tests.

### Implementation

We apply the fake environment to a test by creating `StreamrClient` instances via custom ClientFactory. We get the factory by calling `createClientFactory` (`fakeEnvironment/clientFactory.ts`). It uses dependency injection to inject some objects (mostly singletons). We can create multiple clients from each clientFactory: all instances from a factory share the same fake environment. For each test case we create a new factory and therefore get a clean fake environment that is independent of other test cases.

The tests don't need significant modification when we convert a "real" test to a fake test. Usually we can just change the way we make the client instances: e.g. call `clientFactory.createClient()` instead of `getCreateClient().createClient()` (which is helper method in `test-utils/utils.ts`). In the fake environment we can authenticate a client with any private key, there is no need to use `fetchPrivateKeyWithGas()`.

There are four direct replacements for `StreamrClient` dependencies:
- `FakeBrubeckNode` emulates `BrubeckNode`
- `FakeStreamRegistry` emulates `StreamRegistry`
- `FakeStorageNodeRegistry` emulates `StorageNodeRegistry`
- `FakeRest` emulates `Rest`

These fake classes implement a subset of the functionality of "real" classes. For some methods the methods may throw a "not implemented" exception if we have not yet implemented the functionality in the fake class. We need implementations only for those methods which are actually used in the test.

Each class implements the interface of the corresponding class and therefore ensures that the method definitions are kept in sync between the two implementations.

There are also two helper classes:
- `FakeStorageNode` which emulates the storage node functionality
- `ActiveNodes` which keeps track of `FakeBrubeckNode` instances (contains all nodes created for a fake environment)

`FakeStorageNodeRegistry` creates one storage node with docker-dev environment node address (`0xde1112…`). There is currently no implementation for `createOrUpdateNodeInStorageNodeRegistry`, but we can add that later if custom storage nodes are needed.

### Small changes for the core StreamClient internals

Small refactoring was needed to allow the faking of some classes:
- `Resends.fetchStream` moved to `Rest`
- `BrubeckNode.start` returns `Promise<void>` (instead of a promise which leaks a `NetworkNode`)

Also the return type of `BrubeckNod.getNode()` changed from `NetworkNode` to a new interface called `NetworkNodeStub`. It is the same object as previously, but by annotating the instance with the interface we limit the public API of the client. Most of `NetworkNode` methods/fields are internal implementation and there is no need to expose those to users of the client. Alternatively we could mark some methods/fields of `NetworkNode` as `@internal.`/`private` Or maybe we could provide the functionality of `NetworkNode` methods as `StreamrClient` methods.

### Other small bugfixes/refactoring

These are not strictly related to the fake implementation/refactoring:
- `FailedToPublishError` error message tweaked so that it doesn't try to make a JSON from a full Stream object (if it is provided as streamDefinition parameter). A full object may contain circular references and therefore stringifying it can fail.
- Changed field visibility in `StreamRegistry` and `StorageNodeRegistry`, removed redundant read-only contract in `StorageNodeRegistry`
- Fixed return type of `StreamEndpoints.getStreamLast`
- `test/utils.ts` moved to `test/test-utils/utils.ts`
- added `Multimap` helper class to test utils

### Future improvements

Do we want to as a separate test/end-to-end` and move some tests there? In that setup there would be three tests types:
- Unit tests to test one class directly by calling the class methods.
- Integration tests to test some larger functionality by executing methods via `StreamrClient` (usually interaction between nodes and/or smart contracts). Usually testing multiple edge cases related to that functionality.
- End-to-end tests to test that all Streamr components (especially calls to blockchain registries and Core-API etc.) work well together and we set up the usages of our external services correctly. Usually focusing on happy paths, but also full coverage for the smart contract usage.


packages/client/src/StreamIDBuilder.ts
```diff
@@ -37,6 +37,15 @@ function parseRawDefinition(definition: StreamDefinition): [string, number | und
     }
 }
 
+export const streamDefinitionToString = (definition: StreamDefinition) => {
```
harbu: return type

harbu: Method name could reveal that it is intended to produce just a description for e.g. logging purposes

packages/client/test/FakeBrubeckNode.ts
```diff
@@ -0,0 +1,81 @@
+import debug from 'debug'
+import { pull } from 'lodash'
+import { EthereumAddress, StreamMessage, StreamPartID } from 'streamr-client-protocol'
+import { DestroySignal } from '../src/DestroySignal'
+import { FakeBrubeckNodeRegistry } from './FakeBrubeckNodeRegistry'
+
+const log = debug('Streamr:FakeBrubeckNode')
+
+export class FakeBrubeckNode {
```
harbu:
```suggestion
export class FakeBrubeckNode implements BrubeckNode {
```

packages/client/test/FakeBrubeckNode.ts
```diff
@@ -0,0 +1,81 @@
...
+        log(`Created${name ? ' ' + name : ''}: ${address}`)
+    }
+
+    getAddress() {
+        return this.address
+    }
+
+    // eslint-disable-next-line class-methods-use-this
+    startNode() {
+    }
```
harbu: If I remember correctly this should delegate to `getNode`.

packages/client/test/FakeBrubeckNodeRegistry.ts
```diff
@@ -0,0 +1,30 @@
+import { EthereumAddress } from 'streamr-client-protocol'
+import { Lifecycle, scoped } from 'tsyringe'
+import { FakeBrubeckNode } from './FakeBrubeckNode'
+
+@scoped(Lifecycle.ContainerScoped)
+export class FakeBrubeckNodeRegistry {
```
harbu: This doesn't correspond to any existing class does it? If not I find the naming confusing, firstly the `Fake` part (there is no real that that is replaced?) and `Registry` part (this isn't a smart contract thing?)

Was this the singleton thing? Maybe something like `ActiveNodesSingleton` etc.?

packages/client/test/FakeStorageNode.ts
```diff
@@ -0,0 +1,46 @@
...
+    async getLast(streamPartId: StreamPartID, count: number): Promise<StreamMessage[]> {
+        const messages = this.streamPartMessages.get(streamPartId)
+        if (messages !== undefined) {
+            const firstIndex = Math.max(messages.length - count, 0)
+            const lastIndex = Math.min(firstIndex + count, messages.length - 1)
+            const result: StreamMessage[] = []
+            for (let i = firstIndex; i <= lastIndex; i++) {
+                result.push(messages[i])
+            }
+            return result
```
harbu: `slice`

packages/client/test/FakeStorageNodeRegistry.ts
```diff
@@ -0,0 +1,83 @@
...
+    ) {
+        this.streamIdBuilder = streamIdBuilder
+        this.fakeBrubeckNodeRegistry = fakeBrubeckNodeRegistry
+        this.fakeBrubeckNodeRegistry.addNode(new FakeStorageNode(DOCKER_DEV_STORAGE_NODE, fakeBrubeckNodeRegistry, 'storage'))
+    }
+
+    private async hasAssignment(streamIdOrPath: string, nodeAddress: string): Promise<boolean> {
+        const normalizedNodeAddress = nodeAddress.toLowerCase()
+        const streamId = await this.streamIdBuilder.toStreamID(streamIdOrPath)
+        const assignment = this.assignments.get(streamId)
```
harbu:
```suggestion
        const assignments = this.getStorageNodesOf(streamIdOrPath)
```

packages/client/test/FakeStorageNodeRegistry.ts
```diff
@@ -0,0 +1,83 @@
...
+    async getRandomStorageNodeFor(streamPartId: StreamPartID): Promise<FakeStorageNode> {
+        const nodeAddresses = await this.getStorageNodesOf(StreamPartIDUtils.getStreamID(streamPartId))
+        if (nodeAddresses.length > 0) {
+            const chosenAddress = nodeAddresses[Math.floor(Math.random() * nodeAddresses.length)]
+            const storageNode = this.fakeBrubeckNodeRegistry.getNode(chosenAddress)
+            if (storageNode !== undefined) {
+                return storageNode as FakeStorageNode
+                // eslint-disable-next-line no-else-return
+            } else {
+                throw new Error('no storage node available: ' + chosenAddress)
```
harbu:
```suggestion
                throw new Error('no storage node online: ' + chosenAddress)
```

harbu (commented):
lgtm; few things
- Group test fakes into a directory
- Return types for methods

packages/client/test/FakeBrubeckNode.ts
harbu: or `throw new Error('not implemented')`

packages/client/test/FakeBrubeckNode.ts
teogeb: Done

packages/client/test/FakeBrubeckNode.ts
teogeb: It's a no-op in fake environment (the real BrubeckNode startNode now returns Promise<void>)

packages/client/test/FakeBrubeckNodeRegistry.ts
teogeb: Renamed

packages/client/test/FakeStorageNode.ts
teogeb: Fixed

packages/client/test/FakeStorageNodeRegistry.ts
teogeb: Fixed

packages/client/test/FakeStorageNodeRegistry.ts
teogeb: Fixed

packages/client/src/StreamIDBuilder.ts
teogeb: Fixed

packages/client/test/test-utils/utils.ts
```diff
@@ -592,3 +592,64 @@ export async function until(condition: MaybeAsync<() => boolean>, timeOutMs = 10
...
+ * The values is an array
+ * -> when we query the data, we get it back in the same order
+ * -> an array may contain duplicates, if same value is added multiple times
+ *    (we could implement a Multiset class if we need a different kind of duplication handling)
+ *
+ * TODO: Move this class to a streamr-utils package when we create that? Also implement some
+ * unit tests if this is not just a test helper class.
+ */
+export class Multimap<K, V> {
+    private values: Map<K, V[]> = new Map()
```
```suggestion
    private readonly values: Map<K, V[]> = new Map()
```

teogeb: Fixed

packages/client/src/StreamIDBuilder.ts
```diff
@@ -37,6 +37,25 @@ function parseRawDefinition(definition: StreamDefinition): [string, number | und
     }
 }
 
+/*
+ * Use this e.g. for internal logging purposes where any kind of string representation is fine.
+ *
+ * Note that sometimes it is better to write this object to a log than just write e.g. JSON
+ * representation of a stream definition. The stream definition object may be any instance which
+ * implements the StreamDefinition interface (e.g a Stream object), and we don't want to write
```
harbu, line 45: Hmmm.. I recall the `Stream` object not being compatible with `StreamDefinition`?

packages/client/test/integration/GroupKeyPersistence.test.ts
```diff
@@ -70,14 +75,14 @@ describe('Group Key Persistence', () => {
                 // subscriber will need to ask new publisher
                 // for group keys, which the new publisher will have to read from
                 // persistence
-                await stream.addToStorageNode(DOCKER_DEV_STORAGE_NODE)
+                await stream.addToStorageNode(DOCKER_DEV_STORAGE_NODE) // tgtest tama on ylimaarainen (storage node on jo asetettu)
```
harbu: rm comment

packages/client/test/test-utils/fake/FakeBrubeckNode.ts
```diff
@@ -0,0 +1,146 @@
...
+        throw new Error('not implemented')
+    }
+
+    // eslint-disable-next-line class-methods-use-this
+    getNeighbors(): ReadonlyArray<string> {
+        throw new Error('not implemented')
+    }
+
+    getNeighborsForStreamPart(streamPartId: StreamPartID): ReadonlyArray<string> {
+        const allNodes = this.node.activeNodes.getNodes()
```
harbu, line 69: I guess the assumption here is that all nodes are neighbors of each other within the same stream part?

packages/client/test/test-utils/fake/fakeEnvironment.ts
```diff
@@ -0,0 +1,71 @@
...
+export interface ClientFactory {
+    createClient: (opts: any) => StreamrClient
+}
+
+export const createClientFactory = (): ClientFactory => {
+    const mockContainer = container.createChildContainer()
+    mockContainer.registerSingleton(StreamRegistry, FakeStreamRegistry as any)
+    mockContainer.registerSingleton(StorageNodeRegistry, FakeStorageNodeRegistry as any)
+    mockContainer.registerSingleton(ActiveNodes, ActiveNodes as any)
+    mockContainer.registerSingleton(Rest, FakeRest as any)
```
harbu, line 27: I'm thinking you should be able to assign these without the `any`?

packages/client/test/test-utils/fake/fakeEnvironment.ts
```diff
@@ -0,0 +1,71 @@
...
+    const ethereumAddressCache = createEthereumAddressCache()
+    mockContainer.register(BrubeckNode, { useFactory: (c: DependencyContainer) => {
+        /*
+         * We need to use a DI factory the register the BrubeckNode, because config-related
```
harbu:
```suggestion
         * We need to use a DI factory to register the BrubeckNode, because config-related
```

harbu (approved):

packages/client/src/StreamIDBuilder.ts
harbu: Never mind, I seem to be mistaken here.

PR #433, opened 2022-02-07
Default second argument of calculate DU address to streamrClient itself

jtakalai:
Here the signature changes from sync to async, not sure if the improvement is worth it. Maybe @tumppi can judge, since he's sort of end-user?


tumppi (approved):
Works for us either way. Synchronous version is a bit easier for us currently but doesn't really matter that much. Please let me know if this gets merged because we need to change frontend code accordingly.

harbu (approved):

PR #434, opened 2022-02-07
Update restUrl config

teogeb:
Change `/api/v1` -> `/api/v2`


harbu (commented):
Let's merge when tests go green, unless for whatever reason, `main` was also green. In which case we should think thru how on earth that is possible.

PR #435, opened 2022-02-08
Client: configurable timeouts and retry intervals

harbu:
Add timeouts and retry intervals as configuration options in `StrictStreamrClientConfig`. Increase timeout of polling of The Graph in production (10 seconds was not enough)


packages/client/src/ConfigBase.ts
```diff
@@ -22,6 +22,22 @@ export type CacheConfig = {
     maxAge: number
 }
 
+export type TimeoutsConfig = {
```
harbu: Don't make part of public API, keep as internal info in streamr-client.

packages/client/src/ConfigBase.ts
```diff
@@ -78,6 +94,7 @@ export type StrictStreamrClientConfig = {
     ensCacheChainAddress: EthereumAddress,
     dataUnion: DataUnionConfig
     cache: CacheConfig,
+    timeouts: TimeoutsConfig
```
harbu: Could `@internal` work here?

packages/client/src/utils/HttpFetcher.ts
```diff
@@ -0,0 +1,27 @@
+import { inject, Lifecycle, scoped } from 'tsyringe'
+import { Debugger } from 'debug'
+import { Context } from './Context'
+import { Config, StrictStreamrClientConfig } from '../Config'
+import { instanceId } from './index'
+import fetch, { Response } from 'node-fetch'
+
+@scoped(Lifecycle.ContainerScoped)
+export class HttpFetcher {
```
harbu, line 9: Replace usages with `Rest#get` and rm this class?

harbu (commented):
Teo's review points

packages/client/src/ConfigBase.ts
harbu: Works

packages/client/src/ConfigBase.ts
harbu: Works

packages/client/src/utils/HttpFetcher.ts
harbu: Actually looking at `Rest` more closely, it is design to handle session expiration with Core API requests. So not really general enough for this use case.

teogeb (approved):
LGTM. Good to have the timeouts configurable. Let's restructure the configs in another PR (as the `_` suffix suggest, this is just a temporary location for the timeout settings),

PR #436, opened 2022-02-09
Net-697: Allow configuring multiple chain RPC-endpoints for get queries

juslesan:
Make it possible to configure multiple chain RPC endpoints for redundancy in get queries to smart contracts. This improves the reliability and availability of the system as a whole as a single blockchain node's downtime cannot take down streams.


jtakalai (approved):
Interesting idea. Note that `ethers` comes with `FallbackProvider` that basically tries one at a time and goes forward on the list if it gets an error from the current one. But that's different from the redundancy approach in this PR which is to race the providers at the same time.

packages/client/src/Ethereum.ts
```diff
@@ -66,17 +67,17 @@ type EthereumNetworkConfig = {
 }
 
 export abstract class EthereumConfig {
-    abstract dataUnionBinanceWithdrawalChainRPC: ConnectionInfo & { chainId?: number, name?: string }
+    abstract dataUnionBinanceWithdrawalChainRPCs: ChainConnectionInfo[]
```
harbu: Looks like the `chainId` of the first element of this array is only ever used and rest are ignored. So should the config be like
```ts
{
    chainId?: number,
    name?: string,
    rpcs: ConnectionInfo[]
}
```
And same for rest of the RPCs settings?

packages/client/src/Ethereum.ts
```diff
@@ -130,14 +131,14 @@ class StreamrEthereum {
                 return metamaskSigner
             }
             this._getDataUnionChainSigner = async () => {
-                if (!ethereumConfig.dataUnionChainRPC || !ethereumConfig.dataUnionChainRPC.chainId) {
+                if (!ethereumConfig.dataUnionChainRPCs || !ethereumConfig.dataUnionChainRPCs[0].chainId) {
```
harbu: `!ethereumConfig.dataUnionChainRPCs[0].chainId` will evaluate to true if the value of `chainId` is zero. Don't know if that would be a valid value here ever but an explicit `undefined` check may be more appropriate (i.e. `ethereumConfig.dataUnionChainRPCs[0].chainId === undefined`

harbu (commented):
Yeah `Promise.any` approach is cool

harbu (approved):

PR #438, opened 2022-02-09
 Update packages (attempt no 2)

harbu:


packages/broker/src/helpers/PayloadFormat.ts
```diff
@@ -48,7 +48,7 @@ const parsePayloadJson = (payload: string) => {
     }
     try {
         return JSON.parse(payload)
-    } catch (e) {
+    } catch (e: any) {
```
timoxley, line 51: IMO we should probably either:
* switch the `useUnknownInCatchVariables` config off OR
* narrow the type before usage, something like this:

```
if (err instanceof Error) {
   throw new Error(`Payload is not a JSON string: ${e.message}`)
} else {
 throw err
}
```
Explicitly using `any` seems like worst of both, more explicit typing, but no improved type safety.

----

Deciding what should go in the `else` is another question though. Rethrow the non-error? Wrap in a `new Error(err)` and thus coerce the non-error value into a string and thus potentially losing information or breaking something that expects a non-error error, but also possibly adding new information about the current stack that would be lost if we just rethrow? Perhaps `new Error('weird error', { cause: err })` using the new error.cause https://v8.dev/features/error-cause

PR #440, opened 2022-02-09
Net-699

juslesan:
Add Network-layer functions:
 `subscribeAndWaitForJoin(streamPartId, timeout)`
  `WaitForJoinAndPublish(streamPartId, timeout)`


packages/network/src/logic/node/Node.ts
```diff
@@ -59,13 +61,15 @@ export interface Node {
...
+    on(event: Event.JOIN_COMPLETED, listener: (streamPartId: StreamPartID, numOfNeighbors: number) => void): this
+    on(event: Event.JOIN_FAILED, listener: (streamPartId: StreamPartID, error: string) => void): this
 }
 
 export class Node extends EventEmitter {
     /** @internal */
     public readonly peerInfo: PeerInfo
     protected readonly nodeToNode: NodeToNode
-    private readonly nodeConnectTimeout: number
+    protected readonly nodeConnectTimeout: number
```
harbu:
```suggestion
    private readonly nodeConnectTimeout: number
```

packages/network/src/logic/tracker/InstructionCounter.ts
```diff
@@ -3,6 +3,8 @@ import { Status } from '../../identifiers'
 import { NodeId } from '../node/Node'
 
 export const COUNTER_UNSUBSCRIBE = -1
+// Used by the tracker to signal to nodes that they are alone in the topology
+export const COUNTER_LONE_NODE = -2
```
harbu, line 7: Is there a reason we signal this via counter and not just an empty neighbor list?

packages/client/test/test-utils/fake/FakeBrubeckNode.ts
```diff
@@ -38,6 +38,17 @@ class FakeNetworkNodeStub implements NetworkNodeStub {
         this.subsciptions.delete(streamPartId)
     }
 
+    async subscribeAndWaitForJoin(streamPartId: StreamPartID, _timeout?: number): Promise<number> {
+        this.subsciptions.add(streamPartId)
+        return 1
```
harbu: There's probably a way to pull the number of other nodes in that stream part. @teogeb do you know?

packages/client/test/test-utils/fake/FakeBrubeckNode.ts
```diff
@@ -38,6 +38,17 @@ class FakeNetworkNodeStub implements NetworkNodeStub {
...
+    }
+
+    async waitForJoinAndPublish(msg: StreamMessage, _timeout?: number): Promise<number> {
+        this.subsciptions.add(msg.getStreamPartID())
+        this.publish(msg)
+        return 1
```
harbu: same as above

packages/network/src/logic/tracker/Tracker.ts
```diff
@@ -232,15 +232,29 @@ export class Tracker extends EventEmitter {
...
         if (this.overlayPerStreamPart[streamPartId]) {
             const instructions = this.overlayPerStreamPart[streamPartId].formInstructions(node, forceGenerate)
-            Object.entries(instructions).forEach(async ([nodeId, newNeighbors]) => {
-                const counterValue = this.instructionCounter.setOrIncrement(nodeId, streamPartId)
-                await this.instructionSender.addInstruction({
-                    nodeId,
+
+            // Send empty instruction if and only if the node is alone in the topology
+            if (this.overlayPerStreamPart[streamPartId].hasNode(node)
+                && this.overlayPerStreamPart[streamPartId].getNumberOfNodes() === 1
```
harbu, line 238: Is this check here just in case or have there been observed cases of no neighbors being assigned after `formInstructions` call even in the presence of other nodes?

Edit: Now that I think of it, this is a possible outcome at least if node degree parameter was set to be 1. Let's say the topology already contained 2 nodes that were connected to each other. If a third node now joined, it would have no neighbors to connect to (as we wouldn't cut the existing connection in this case.) But I still wonder if this can happen with reasonable node degrees, e.g. `d=4`?

packages/network/src/logic/tracker/Tracker.ts
```diff
@@ -232,15 +232,29 @@ export class Tracker extends EventEmitter {
...
+                && Object.keys(instructions).length === 0) {
+                this.instructionSender.bypassBuffer({
```
harbu: Why do we need to bypass the buffer? Could use an explanation, as it is not immediately obvious.

packages/network/test/integration/wait-for-join-subscribe-and-publish.test.ts
```diff
@@ -0,0 +1,107 @@
...
+import { MessageID, StreamMessage, toStreamID, toStreamPartID } from 'streamr-client-protocol'
+import { waitForEvent } from 'streamr-test-utils'
+
+import { createNetworkNode, startTracker } from '../../src/composition'
+import { Event as NodeEvent } from "../../src/logic/node/Node"
+
+/**
+ * This test verifies that on receiving a duplicate message, it is not re-emitted to the node's subscribers.
+ */
+describe('subscribe and wait for the node to join the stream', () => {
```
harbu, line 12: Good stuff :+1:

harbu (commented):
lgtm; I had a few questions of things that are unclear to me

packages/client/test/test-utils/fake/FakeBrubeckNode.ts
teogeb: Check the logic in in `publish()`:
```
this.node.activeNodes.getNodes()
            .forEach(async (n) => {
                const networkNode = await n.getNode()
                if (networkNode.subsciptions.has(msg.getStreamPartID())) {
```

If we reuse these lines in two places, let's put these lines into a helper method used by `publish` and this `subscribeAndWaitForJoin`.

packages/network/src/logic/tracker/Tracker.ts
juslesan: There's a chance that this is not required anymore I'll test it out.

packages/network/src/logic/tracker/Tracker.ts
juslesan: There is some test flakiness if the check is not there, so it's there as a useful safeguard.

packages/network/src/logic/tracker/InstructionCounter.ts
juslesan: Sending and processing empty lists on the nodes' side causes some unintended consequences, especially around "unsubscribe" events.

packages/network/src/logic/tracker/Tracker.ts
juslesan: Removed the bypass functionality as it was not necessary

timoxley (approved):
looks good to me

PR #442, opened 2022-02-10
[NET-698]: Config wizard: only write non-default values

teogeb:
Config wizard creates a minimal configuration file as it doesn't write a values if that value is a default value of the property.

Also changed some defaults, most importantly network `stunUrls`. @juslesan Is it ok that we have the `stunUrl` default has been changed in `network`-repo, affecting all use cases of network package?

### Details 
- We don't write `network.id`, because by default we derive that information from the private key. Note: now when an explicit id is not set, the network id will use session id suffix. Trackers and miner plugins should be compatible with that.
- We don't write `network.name`. The intention is that all Brubeck nodes won't be miners and therefore the name could be misleading. The name is used to give name to the MetricsContext (`BrubeckNode.ts:67`), but it is maybe ok that we change that? Note that network.id now includes a sessionId and therefore it is not the same in two consecutive runs. Therefore is some use case needs to have deterministic name for the metrics context, the user should set an explicit network.id. (If we don't change the functionality, e.g. by removing the sessionId suffix.)
- Updated the defaults of `brubeckMiner` plugin to reflect what was earlier defined in config wizard.
- Client restUrl doesn't have an `/` character at the of the url definition. Maybe fixes something, or is just more accurate value?

### DIFF
--- a/packages/broker/src/plugins/brubeckMiner/config.schema.json
+++ b/packages/broker/src/plugins/brubeckMiner/config.schema.json
@@ -9,11 +9,22 @@
             "type": "array",
             "items": { "type": "string" },
             "uniqueItems": true,
-            "default": ["streamr.eth/brubeck/brubeck-rewards"]
+            "default": [
+                "streamr.eth/brubeck-mainnet/rewards/001ae9",
+                "streamr.eth/brubeck-mainnet/rewards/b82a43",
+                "streamr.eth/brubeck-mainnet/rewards/d72f1a",
+                "streamr.eth/brubeck-mainnet/rewards/d1aed2",
+                "streamr.eth/brubeck-mainnet/rewards/14aptk",
+                "streamr.eth/brubeck-mainnet/rewards/b2ab22",
+                "streamr.eth/brubeck-mainnet/rewards/7cd249",
+                "streamr.eth/brubeck-mainnet/rewards/af81e2",
+                "streamr.eth/brubeck-mainnet/rewards/52ada6",
+                "streamr.eth/brubeck-mainnet/rewards/cbab52"
+            ]
         },
         "claimServerUrl": {
             "type": "string",
-            "default": "https://brubeck-rewards.streamr.network"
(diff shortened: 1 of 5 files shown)


packages/broker/src/plugins/brubeckMiner/config.schema.json
```diff
@@ -25,7 +36,7 @@
                 "string",
                 "null"
             ],
-            "default": "stun.streamr.network:5349"
+            "default": "stun.sipgate.net"
```
harbu, line 39: I'm assuming this is the one used by the nat-type detection plugin?

packages/network/src/createNetworkNode.ts
```diff
@@ -39,7 +39,7 @@ export const createNetworkNode = ({
     rttUpdateTimeout,
     webrtcDatachannelBufferThresholdLow,
     webrtcDatachannelBufferThresholdHigh,
-    stunUrls = ['stun:stun.l.google.com:19302'],
+    stunUrls = ['stun:stun.streamr.network:5349'],
```
harbu, line 42: This is up and running, yeah?

harbu (approved):
lgtm

packages/broker/src/plugins/brubeckMiner/config.schema.json
teogeb: Yes. `streamr.network:5349` doesn't support that yet.

juslesan: Slight correction, the library uses an out-dated version of STUN and `stun.sipgate.net` still runs using that version. We could implement our own NAT-type detection algorithm for the newer STUN versions.

juslesan (approved):

packages/network/src/createNetworkNode.ts
teogeb: Yes, @juslesan confirmed

PR #443, opened 2022-02-10
NET-611: fix and enable browser-resend tests

juslesan:
- Enable browser-resend tests in CI.
- The test failed because the storage node did not receive and store published messages in time for the resend request.


.github/workflows/client-build.yml
```diff
@@ -112,8 +112,7 @@ jobs:
 
   browser-resend:
     # skip to get test suite green, not sure why failing.
-    if: ${{ false }}
-    name: Test Browser Resend
+    name: Test Browser Realtime
```
harbu: Should this be:
```suggestion
    name: Test Browser Resend
```

Also comment above should be removed.

packages/client/test/browser/resend.js
```diff
@@ -55,6 +57,10 @@ describe('StreamrClient Resend', () => {
             .verify.containsText('#result', '{"msg":9}')
             .assert.containsText('#result', '[{"msg":0},{"msg":1},{"msg":2},{"msg":3},{"msg":4},{"msg":5},{"msg":6},{"msg":7},{"msg":8},{"msg":9}]')
             .assert.not.elementPresent('.error')
+
+        // Wait for published messages to arrive at the storage node
+        await wait(2000)
```
harbu: Polling the storage node here would make for a more reliable solution w.rt. time and flakyness.

Do you have direct access to `streamrClient` here? Then something like could be used:
```ts
await streamrClient.waitForStorage(lastStreamMessage)
```

If not, is it possible to push the resend button multiple times in this test until a condition is true?

harbu (changes requested):

packages/client/test/browser/resend.js
juslesan: It should be possible to wait for this, I'll give it a try

juslesan: This seems a bit too workload heavy to implement given the current test setup as opposed to just just waiting. It would be good to consider replacing the client browser-tests with jest mockups similarly to the network-layer.

juslesan: Turned out to be much easier than I initially anticipated

harbu (approved):

PR #446, opened 2022-02-11
[NET-595]: Migrate Testnet3 broker config

teogeb:
Automatically migrate a Testnet3 config to the new format used in Brubeck mainnet.

The migration drops many properties, either because the new value is already the default in the config schema or because the value was Testnet3-specific. See `migration.ts:convertTestnet3ToV1()` method for details.

### Schema

There is a new `$schema` property in the config file, which identifies the version of the config. For migrated configs, and new configs created with Config Wizard, the reference is `http://schema.streamr.com/config-1.schema.json`

That value is JSON-schema compatible URL. In the migration we use the filename part of the URL to investigate
the config file version. We don't fetch the content of the schema file at all. The benefit of using schema  instead of just a version number is that it provides more information to any external tools, e.g. IDEs. If we upload the schema file to some public server, an IDE can provide validation and autocompletion if a user edits a configuration file manually.

We aren't expected to change the schema format very often. Therefore the uploading of the schema file could be just a manual task. It is also not a critical task: the migration works fine even we aren't able to upload the file for some reason. 

### File location

User can specify the config file location as a command line parameter:
- If it specified, we create the migrated file to that location and copy a backup file to the same directory.
- If it is not specified, we check the current default location `~/.streamr/config/default.json` and the legacy default location `~/.streamr/broker-config.json`. If the migration has not been done yet, the file is very likely found from the legacy location, and there is no file in the current default location. In that situation we create the migrated file to the current default location. We also "archive" the legacy file by renaming it. The new `broker-config.json.backup` name communicates that the legacy file is no longer used. (The renaming is the last step of the migration, and therefore we don't do that if an earlier step fails.)

### Testing

There is an extensive unit test for the migration logic, but no tests for the file handling (reading the config file, creating a backup, writing a new file). We could add a test if we implement some kind of file system emulation (e.g. by using [mock-fs](https://www.npmjs.com/package/mock-fs). Maybe not worth doing as this it is relatively easy to test this manually and we don't expect that the implementation is modified very often?


packages/broker/src/config/migration.ts
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * Some new features were added between Testnet3 and Brubeck mainnet. The migration assumes
+ * that a source config doesn't contain settings for those features. Therefore it ignores:
+ * - network.webrtcDisallowPrivateAddresses
+ * - subscriber plugin
+ */
+const convertTestnet3ToV1 = (source: any): Config => {
+    const DEFAULT_NAME = 'miner-node'
+    const target: any = {
+        $schema: formSchemaUrl(CURRENT_CONFIGURATION_VERSION)
```
harbu: Maybe explicit set to `1`
```suggestion
        $schema: formSchemaUrl(1)
```

packages/broker/src/config/migration.ts
```diff
@@ -0,0 +1,155 @@
...
+            // eslint-disable-next-line max-len
+            throw new Error(`Config file not found in the default location. You can run "streamr-broker-init" to generate a config file interactively, or specify the config file as argument: "streamr-broker path-to-config/file.json"`)
+        }
+    }
+    let content = JSON.parse(fs.readFileSync(fileName, 'utf8'))
+    if (needsMigration(content)) {
+        const backupFile = formBackupFileName(fileName)
+        // eslint-disable-next-line no-console
+        console.log(`Migrating config ${fileName}, saving backup to ${backupFile}`)
+        fs.renameSync(fileName, backupFile)
```
harbu: I would maybe copy file at this point in case the program crashes for some reason before finishing the migration. Teo: what about file permissions, are they copied?

harbu (commented):
If user doesn't specify config file explicitly, and (thus implict) config file needs migration, let's create new file in new default directory, backup is actually not needed in this case.

packages/broker/src/config/migration.ts
teogeb: Fixed

packages/broker/src/config/migration.ts
teogeb: Fixed. The permissions are copied ok. Also, there were no need to explicitly set the file permissions of the new file as we just write rewrite the content instead of creating a new file.

packages/broker/src/config/migration.ts
```diff
@@ -135,21 +135,44 @@ const formBackupFileName = (originalFileName: string) => {
...
+        const legacyTargetFile = getLegacyDefaultFile()
+        fileName = [defaultTargetFile, legacyTargetFile].find((file) => fs.existsSync(file))
         if (fileName === undefined) {
             // eslint-disable-next-line max-len
             throw new Error(`Config file not found in the default location. You can run "streamr-broker-init" to generate a config file interactively, or specify the config file as argument: "streamr-broker path-to-config/file.json"`)
         }
+        if (fileName === legacyTargetFile) {
+            /*
+             * User has not specified the config file location in the command line and we found
+             * the file from the legacy default location. We'll write the migrated the file to current
```
harbu: `We'll write the migrated the file `

harbu (approved):

PR #449, opened 2022-02-14
client: subscribe type annotations, remove deprecated resend parameter handling

teogeb:
Simplified subscription-related API by marking some methods private/internal and removing some deprecated methods. The `subscribe` method and `Subscription` return types have up-to-date type annotations and exports.

### subscribe()

Dropped deprecated style for calling a `subscribe()` where the resend data was not inside `resend` block. Now subscribe which contains a resend request, must have a form of:
```
client.subscribe({
    streamId: '...'.  // or other streamDefinition style, maybe including partition
    resend: {
       last: 123
    }
}
```

Earlier also this format was supported:
```
client.subscribe({
    streamId: '...'.  // or other streamDefinition style, maybe including partition
    last: 123
}
```

If resend block is given, the return type is `ResendSubcription` (which has `onResent()` method). Otherwise it is a normal subscription. Simplified also the method annotation of `onResent`: it is just a method which takes `() => void` as a callback.

The `ResendOptions` type now represents one of the three resend types (last, from, range). Earlier there were a separate type for those, and `ResendOptions` was a wrapper which contained that information in `resend` block.

### resend()

The `client.resend()` now takes three parameters: `streamDefinition`, `resendOptions` and a `onMessage` callback. Earlier the stream definition and resend options were in the same object, which contained a required `resend` block.

### Subscription object

Some methods in `Subscription` (and in the super class `MessageStream`) are annotated as internal. Subcription's `cancel()` method was removed, use subscription.unsubscribe instead.

### Implementation

The `subscribe()` moved from `ResendSubscribe` to `StreamrClient`. That makes it easy to provide the type annotations, and it is more logical place as the most common use case of `subscribe()` doesn't have anything to do with resends. The movement doesn't change affect the public API. The overridden methods are ordered so that an IDE can pick both forms of` subscribe()` (first the specific resend case and then more generic case).

### Imports

Relevant classes are exported explicitly in `index-exports.ts`. Alternatively we could have exported all types for each file in `src/subscribe` directory, if we'd marked a lot of methods/types as `@internal`. To make exporting easier `Subscription` and `MessageStreams` now use named export instead of default export.


harbu (approved):
lgtm

PR #450, opened 2022-02-16
[NET-708]: Permission API

teogeb:
Simplified and unified the permission handling public API. Replaced all previous permission methods with a new set methods. There are no significant new features added or removed.

This PR makes it easy to add support for [ETH-228](https://linear.app/streamr/issue/ETH-228/streamregistry-bulk-set-permissions-cleanup). That ticket adds new contract methods which allow us to update multiple permissions  in one transaction. In this PR we change the permission methods to support array of assignments, and therefore we don't need to change the public API when integrate ETH-228 (it adds features but can be backwards-compatible to this new API).  

### New API

The current methods are:
```
hasPermission(query: PermissionQuery): Promise<boolean>
getPermissions(streamIdOrPath: string): Promise<PermissionAssignment[]> {
grantPermissions(streamIdOrPath: string, ...assignments: PermissionAssignment[]): Promise<void>
revokePermissions(streamIdOrPath: string, ...assignments: PermissionAssignment[]): Promise<void>
setPermissions(streamIdOrPath: string, ...assignments: PermissionAssignment[]): Promise<void>
```

- with `hasPermission` method we can query the existence of a single permission, either a user permission or a public permission (we support permission inheritance for user permission queries with `allowPublic` flag)
- with `getPermissions` we can query all permissions of a stream (an array containing an item for each user, and one for public permissions)
- with `grantPermissions`, `revokePermissions` and `setPermissions` we can modify the existing permissions (the `setPermissions` overwrites any existing permissions, and it is therefore more efficient but also "dangerous")

Note that all update methods (grant, revoke and set) take the assignment list in rest parameters syntax. Therefore users can easily provide one or multiple assignment as an argument.

There are no separate methods to query own permissions. We can fetch that information by calling `hasPermission` with `client.getAddress()`. Or get own permissions by that address from `getPermissions()` result.

### Usage from Stream 

There are delegate methods in `Stream`, which don't take the `streamIdParameter` but are otherwise equal. The `setPermissions` is left out from `Stream` because we want to encourage to users to use `grantPermissions` and `revokePermissions` instead of `setPermissions` (as it overrides the existing permissions and is in that sense more "dangerous").

### Other fixes

Fixed also cache handling related to permission modifications. Earlier the `StreamEndpointsCached` was cleared for a stream if there was a permission modification made by calling a `Stream` methods (e.g.. `stream.grantPermission(...)`) but not if that same modification was made by calling a `StreamrClient` method (e.g.. `client.grantPermission(streamId, …)`. Now the cache is cleared in both situations.


packages/cli-tools/bin/streamr-stream-grant-permission.ts
```diff
@@ -4,7 +4,7 @@ import { Stream, StreamPermission } from 'streamr-client'
 import { runModifyPermissionsCommand } from '../src/permission'
 
 runModifyPermissionsCommand(
-    (stream: Stream, permission: StreamPermission, target: string) => stream.grantUserPermission(permission, target),
-    (stream: Stream, permission: StreamPermission) => stream.grantPublicPermission(permission),
+    (stream: Stream, permission: StreamPermission, target: string) => stream.grantPermissions({ permissions: [permission], user: target }),
+    (stream: Stream, permission: StreamPermission) => stream.grantPermissions({ permissions: [permission], public: true }),
```
harbu: Could do this in one `grantPermissions` call?

packages/cli-tools/bin/streamr-stream-revoke-permission.ts
```diff
@@ -4,7 +4,7 @@ import { Stream, StreamPermission } from 'streamr-client'
 import { runModifyPermissionsCommand } from '../src/permission'
 
 runModifyPermissionsCommand(
-    (stream: Stream, permission: StreamPermission, target: string) => stream.revokeUserPermission(permission, target),
-    (stream: Stream, permission: StreamPermission) => stream.revokePublicPermission(permission),
+    (stream: Stream, permission: StreamPermission, target: string) => stream.revokePermissions({ permissions: [permission], user: target }),
```
harbu: Could do this is one `revokePermissions` call?

packages/client/README.md
```diff
@@ -347,9 +347,9 @@ All the below functions return a Promise which gets resolved with the result.
...
-| hasPermission(operation, user)            | Returns a permission object, or null if no such permission was found. Valid `operation` values for streams are: stream_get, stream_edit, stream_delete, stream_publish, stream_subscribe, and stream_share. `user` is the username of a user, or null for public permissions. |
-| grantPermission(operation, user)          | Grants the permission to do `operation` to `user`, which are defined as above.                                                                                                                                                                                                |
-| revokePermission(permissionId)            | Revokes a permission identified by its `id`.                                                                                                                                                                                                                                  |
+| hasPermissions()                          | |
+| grantPermissions(              )          | |
+| revokePermissions()                       | |
```
harbu, line 352: Let's write these down as code snippets.

packages/client/test/integration/Permissions.test.ts
```diff
@@ -0,0 +1,181 @@
...
+    beforeAll(async () => {
+        const wallet = new Wallet(await fetchPrivateKeyWithGas())
+        otherUser = randomEthereumAddress()
+        client = new StreamrClient({
+            ...clientOptions,
+            auth: {
+                privateKey: wallet.privateKey,
+            }
+        })
+    })
```
harbu, line 28: Add cleanup (even thought its not strictly needed here since connections aren't formed)
```ts
afterAll(async () => {
    await client?.destroy()
})
```

packages/client/test/integration/Permissions.test.ts
```diff
@@ -0,0 +1,181 @@
...
+                allowPublic: false
+            })).toBe(false)
+            expect(await stream.hasPermission({
+                permission: StreamPermission.EDIT,
+                user: otherUser,
+                allowPublic: false
+            })).toBe(false)
+        })
+
+        it('inherited permissions', async () => {
```
harbu:
```suggestion
        it('public permissions', async () => {
```

packages/client/test/integration/Permissions.test.ts
```diff
@@ -0,0 +1,181 @@
...
+            await stream.grantPermissions({
+                public: true,
+                permissions: [StreamPermission.PUBLISH, StreamPermission.SUBSCRIBE],
+            })
+            expect(await stream.hasPermission({
+                permission: StreamPermission.PUBLISH,
+                user: otherUser,
+                allowPublic: true
```
harbu, line 80: Could test case where `allowPublic` set to `false`. (The method should return `false`)

packages/client/test/integration/Permissions.test.ts
```diff
@@ -0,0 +1,181 @@
...
+                StreamPermission.SUBSCRIBE,
+                StreamPermission.GRANT
+            ]
+        }, {
+            public: true,
+            permissions: [StreamPermission.PUBLISH]
+        }])
+    })
+
+    it('no permissions for other users', async () => {
```
harbu:
```suggestion
    it('no permissions initially for other users', async () => {
```

packages/client/test/integration/Permissions.test.ts
```diff
@@ -0,0 +1,181 @@
...
+        await stream.revokePermissions({
+            public: true,
+            permissions: [StreamPermission.SUBSCRIBE]
+        })
+    })
+
+    it('set permissions', async () => {
+        const user1 = randomEthereumAddress()
+        const user2 = randomEthereumAddress()
+        await client.setPermissions(stream.id, {
```
harbu, line 168: This would be more interesting if there were some permissions granted before.

harbu (commented):
lgtm see comments

packages/cli-tools/bin/streamr-stream-grant-permission.ts
teogeb: Fixed

packages/client/test/integration/Permissions.test.ts
teogeb: Added

packages/client/test/integration/Permissions.test.ts
teogeb: Fixed

packages/client/test/integration/Permissions.test.ts
teogeb: Fixed

packages/client/README.md
teogeb: Added snippets

harbu (approved):

PR #453, opened 2022-02-17
NET-683, client uses on-chain Tracker Registry by default

juslesan:
- Client now gets the list of trackers from the tracker registry smart contract.


packages/client/src/BrubeckNode.ts
```diff
@@ -213,4 +223,17 @@ export default class BrubeckNode implements Context {
             this.debug('closeProxyConnectionOnStream << %o', streamPartId, nodeId)
         }
     }
+
+    async getTrackerList(): Promise<SmartContractRecord[]> {
```
teogeb: This method is used only in` Broker.start()`. I'd remove the method and log just the smart contract address in `Broker.start()`. In Broker startup that gives us enough information. The list of trackers is not relevant to most of end-users, so there is maybe not too many other usages for this method.

(In case we want to keep the method, we could consider to query the data from `NetworkNode`. It is a long queue of methods (`TrackerManager.trackerRegistry.getAllTrackers()`) but that way we'd avoid storing the same information twice.)

packages/client/src/BrubeckNode.ts
```diff
@@ -38,12 +39,13 @@ export default class BrubeckNode implements Context {
     debug
     private startNodeCalled = false
     private startNodeComplete = false
+    private trackers: SmartContractRecord[] = []
```
teogeb: If we change/remove `getTrackerList()` (see the previous comment), we can remove this field as it is then only used in `initNode()`.

packages/client/src/BrubeckNode.ts
```diff
@@ -62,6 +64,14 @@ export default class BrubeckNode implements Context {
         const { options } = this
         let { id } = options
 
+        if ((options.trackers as TrackerRegistrySmartContract).contractAddress) {
```
teogeb: Could extract this to a method called `getNormalizedNetworkOptions()` which would not modify `this.options` but return a modified clone? That way we could keep `this.options` as an immutable variable.

packages/client/src/getTrackerRegistryFromContract.ts
```diff
@@ -1,12 +1,17 @@
-import { JsonRpcProvider } from '@ethersproject/providers'
+import { getDefaultProvider, JsonRpcProvider } from '@ethersproject/providers'
 import { Contract } from '@ethersproject/contracts'
 import { createTrackerRegistry, SmartContractRecord, TrackerRegistry } from 'streamr-client-protocol'
 import * as trackerRegistryConfig from '../contracts/TrackerRegistry.json'
 
 type ProviderConnectionInfo = ConstructorParameters<typeof JsonRpcProvider>[0]
 
-async function fetchTrackers(contractAddress: string, jsonRpcProvider: ProviderConnectionInfo) {
```
teogeb, line 8: Would it make sense to add a new configuration option to the client, which would define a `JsonRpcProvider` for tracker fetching? Or is it better that we use the default provider in this case?

teogeb (changes requested):
Good improvement 👍  Could maybe simplify this if we want to remove the `getTrackerList()` method.

packages/client/src/getTrackerRegistryFromContract.ts
juslesan: There is a config option to set a custom rpc config for the provider, see `trackers`-field in the client config.

packages/client/src/BrubeckNode.ts
```diff
@@ -55,13 +56,24 @@ export default class BrubeckNode implements Context {
         this.destroySignal.assertNotDestroyed(this)
     }
 
+    private async getNormalizedNetworkOptions(): Promise<NetworkNodeOptions> {
```
teogeb, line 59: private async getNormalizedNetworkOptions(): Promise<NetworkNodeOptions> {
        if ((this.options.trackers as TrackerRegistrySmartContract).contractAddress) {
            const trackerRegistry = await getTrackerRegistryFromContract((this.options.trackers as TrackerRegistrySmartContract))
            return {
                ...this.options,
                trackers: trackerRegistry.getAllTrackers()
            }
        }
        return this.options as NetworkNodeOptions
    }

packages/client/src/getTrackerRegistryFromContract.ts
```diff
@@ -1,12 +1,16 @@
...
-type ProviderConnectionInfo = ConstructorParameters<typeof JsonRpcProvider>[0]
-
-async function fetchTrackers(contractAddress: string, jsonRpcProvider: ProviderConnectionInfo) {
-    const provider = new JsonRpcProvider(jsonRpcProvider)
+async function fetchTrackers(contractAddress: string, jsonRpcProvider?: ConnectionInfo) {
+    let provider
+    if (jsonRpcProvider) {
+        provider = new JsonRpcProvider(jsonRpcProvider)
+    } else {
+        provider = getDefaultProvider()
```
teogeb: Possibly related error message:
```
========= NOTICE =========
Request-Rate Exceeded  (this message will not be repeated)

The default API keys for each service are provided as a highly-throttled,
community resource for low-traffic projects and early prototyping.

While your application will continue to function, we highly recommended
signing up for your own API keys to improve performance, increase your
request rate/limit and enable other perks, such as metrics and advanced APIs.

For more details: https://docs.ethers.io/api-keys/
==========================
```

packages/client/src/BrubeckNode.ts
```diff
@@ -55,13 +56,29 @@ export default class BrubeckNode implements Context {
...
+    }
+
     private async initNode() {
         this.assertNotDestroyed()
         if (this.cachedNode) { return this.cachedNode }
 
         const { options } = this
         let { id } = options
 
+        const networkOptions = await this.getNormalizedNetworkOptions()
```
teogeb: Could do this right before `createNetworkNode()` call to make the flow super readable. Then it is also possible remove lines 54 & 67 from `BrubeckNode.test.ts`.

teogeb (approved):
👍

PR #454, opened 2022-02-17
fix: field collision between StreamrClientBase and Session

teogeb:
Annotated some fields of `Session.ts` as private to fix the field collision between `StreamrClientBase` and `Session`. 

Both classes had these fields, and the types are not equal:
- `options` (`StrictBrubeckClientConfig` vs. `AuthConfig`)
- `loginEndpoints` (`LoginEndpoints` vs a function)

Before the fix there was an error in `node_modules/streamr-client/types/src/StreamrClient.d.ts` when the client is used in an external project:

```
Interface 'StreamrClient' cannot simultaneously extend types 'StreamrClientBase' and 'Methods<Session>'.
  Named property 'loginEndpoints' of types 'StreamrClientBase' and 'Methods<Session>' are not identical.ts(2320)
Interface 'StreamrClient' cannot simultaneously extend types 'StreamrClientBase' and 'Methods<Session>'.
  Named property 'options' of types 'StreamrClientBase' and 'Methods<Session>' are not identical.ts(2320)
```

### DIFF
--- a/packages/client/src/Session.ts
+++ b/packages/client/src/Session.ts
@@ -21,12 +21,12 @@ enum State {
 
 @scoped(Lifecycle.ContainerScoped)
 export default class Session extends EventEmitter {
-    state: State
-    sessionTokenPromise?: Promise<string>
+    private state: State
+    private sessionTokenPromise?: Promise<string>
 
     constructor(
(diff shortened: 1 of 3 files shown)


harbu (approved):
lgtm

PR #455, opened 2022-02-17
[NET-711]: Support multiple streams in setPermissions

teogeb:
Modified `client.setPermissions()` to support multiple streams. The method signature was changed, and therefore this is not a backwards-compatible change.

Current API:
```ts
setPermissions(streams:
    {
        streamId: string,
        assignments: PermissionAssignment[]
    }[]
): Promise<void>
```

Previous API:
```ts
setPermissions(
    streamIdOrPath: string,
    ...assignments: PermissionAssignment[]
): Promise<void>
```

### DIFF
--- a/packages/client/README.md
+++ b/packages/client/README.md
@@ -624,10 +624,10 @@ await stream.revokePermissions({
 })
 ```
 
-There is also a `client.setPermissions`. You can use it to set an exact set of permissions for multiple streams. Note that if there are existing permissions for the same users in a stream, the previous permissions are overwritten:
+There is also method `client.setPermissions`. You can use it to set an exact set of permissions for one or more streams. Note that if there are existing permissions for the same users in a stream, the previous permissions are overwritten:
 
 ```js
-await client.setPermissions([{
+await client.setPermissions({
     streamId,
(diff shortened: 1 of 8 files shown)


packages/client/README.md
```diff
@@ -624,19 +624,24 @@ await stream.revokePermissions({
...
-    user: '0x1111111111111111111111111111111111111111',
-    permissions: [StreamPermission.EDIT]
-}, {
-    user: '0x2222222222222222222222222222222222222222',
-    permissions: [StreamPermission.GRANT]
-}, {
-    public: true,
-    permissions: [StreamPermission.PUBLISH, StreamPermission.SUBSCRIBE]
-})
+There is also a `client.setPermissions`. You can use it to set an exact set of permissions for multiple streams. Note that if there are existing permissions for the same users in a stream, the previous permissions are overwritten:
```
harbu:
```suggestion
There is also method `client.setPermissions`. You can use it to set an exact set of permissions for one more more streams. Note that if there are existing permissions for the same users in a stream, the previous permissions are overwritten:
```

harbu (approved):

PR #458, opened 2022-02-21
client: Optimize tests

teogeb:
Optimized some client tests to use partitioned streams. That way we don't need need to create multiple streams in a single test. 

Speed-up:
- `Subscriber.test.ts`: 152s -> 39s
- `StreamrClient.test.ts`: 126s -> 26s (24s+2s)

In the updated `Subscriber.test.ts` and `StreamrClient.test.ts` we create one stream with 100 partitions. Each test case then "consumes" one partition and executes the test operations for that partition. Added some helper functions to make it easy to manage the test partitions (`createPartitionedTestStream((`, `createStreamPartIterator()` in `utils.ts`).

Subscribing and publishing don't need independent streams in these tests, because each subscribe/publish operation targets a single stream partition. Any operations to other partitions should not interfere the functionality. In that sense these modifications actually add some test coverage for subscriber/publisher implementations.

Extracted also some partitioning-related test cases from `StreamrClient.test.ts` to separate test (`Partitions.test.ts`). That tests uses the fake environment as partition computation/selection doesn't need real streams. 

### Future improvements

We can use this partitioning style also in some other tests. It is not applicable for all tests, since many functionalities have some stream-level aspects, e.g. permissions and storage node.


packages/client/test/integration/Partitions.test.ts
```diff
@@ -0,0 +1,180 @@
...
+            const published3 = await publishTestStreamMessages(1, {
+                partitionKey: String(i),
+            })
+            expect(published3).toHaveLength(1)
+            const foundPartition = published3[0].getStreamPartition()
+            foundPartitions[foundPartition] = true
+            if (foundPartitions.every((v) => v)) {
+                break
+            }
+        }
```
harbu, line 113: I know you've prolly copied this test code from elsewhere, but I can't resist to point out a few points (maybe for later refactor):

1. There seems to be a common pattern with these tests that could be extracted (i.e. filling up array).
2. Invocations of `client.publisher.pipeline.messageCreator.streamPartitioner.clear()` feel a bit iffy. I'd almost expose this as a public function or find a clever way around without having to rely on private (e.g. creating new client?)

packages/client/test/integration/StreamrClient.test.ts
```diff
@@ -54,30 +61,10 @@ describeRepeats('StreamrClient', () => {
...
-            ...opts,
-        })
-        await until(async () => { return client.streamExistsOnTheGraph(s.id) }, 100000, 1000)
-        expect(s.id).toBeTruthy()
-        expect(s.requireSignedData).toBe(requireSignedData)
-        return s
-    }
-
     beforeEach(async () => {
         client = await createClient()
```
harbu: Do we actually need client with gas here?

packages/client/test/test-utils/utils.ts
```diff
@@ -662,3 +663,35 @@ export class Multimap<K, V> {
...
 }
+
+export const createPartitionedTestStream = async (): Promise<Stream> => {
+    const client = new StreamrClient({
+        ...clientOptions,
+        auth: {
+            privateKey: await fetchPrivateKeyWithGas()
+        }
+    })
+    const stream = await createTestStream(client, module, {
```
teogeb, line 675: Maybe the `module` should be a parameter to `createPartitionedTestStream() `so IDs reflect the test name

harbu (approved):

packages/client/test/integration/Partitions.test.ts
teogeb: Yes, just modified the existing test and did not refactor it. Agree that we could make the test more readable/maintainable.

packages/client/test/integration/StreamrClient.test.ts
teogeb: Fixed

PR #459, opened 2022-02-21
Fix SSR

mondoreale:
In this PR I make sure we don't run any build-time SSR-unsafe code.

_(SSR is a node/browser hybrid environment, i.e. we can't say it's node just by looking at the undefined `window`.)_


mondoreale:
@harbu does it need further explanation or are we good here?

harbu (approved):

PR #465, opened 2022-02-21
client: merge Config.ts and ConfigBase.ts

teogeb:
Most config-related types and methods are now in `Config.ts`. The types are exported explicitly
and some default exports are converted to named exports.

Inline  `BrubeckClientConfig` helper type: now `network` and `debug` configs are included in `StreamrClientConfig`
similarly to other options. The `BrubeckConfig` helper method is no longer needed as `createStrictConfig`
(previosly named as `ClientConfig`) handles also network- and debug-related config initialization.

Remove redundant `test/integration/config.js` as all tests now use `ConfigTest` directly.

Rename `BrubeckConfigInjection` to `ConfigInjectionToken` and all client classes use it directly (instead of using an alias name `Config`).

The config parsing is now done explicitly by calling `createStrictConfig()` from `StreamrClient` constructor. Previously it was done inside the `initContainer()` call.

### User-visible changes

Because `BrubeckClientConfig` is inlined, that type is not visible to users anymore. Also `StrictBrubeckClientConfig` and `DEFAULTS` (i.e. `BRUBECK_CLIENT_DEFAULTS`) are not available.

The root level config`StreamrClientConfig` was aliased to `StreamrClientOptions`. Removed the redundant alias name `StreamrClientOptions`, all users should use `StreamrClientConfig` instead.

Remove config types for deprecated authentication (`APIKeyAuthConfig` and `UsernamePasswordAuthConfig`).

Remove `auth.privateKey` from `ConfigTest`. That object defines the Docker test environment, and we support multiple users in that environment. Each test should preferrably define an explicit (and maybe also separate) test user.

### Other

Fix chain-RPC config usage in some tests.


packages/client/src/Config.ts
```diff
@@ -1,99 +1,302 @@
...
-}
+export const STREAM_CLIENT_DEFAULTS: StrictStreamrClientConfig = {
+    auth: {},
 
-export * from './ConfigBase'
+    // Streamr Core options
+    restUrl: 'https://streamr.network/api/v2',
+    theGraphUrl: 'https://api.thegraph.com/subgraphs/name/streamr-dev/streams',
+    streamrNodeAddress: '0xf3E5A65851C3779f468c9EcB32E6f25D9D68601a',
+    // storageNodeAddressDev = new StorageNode('0xde1112f631486CfC759A50196853011528bC5FA0', '')
```
harbu, line 132: Rm deprecated comment?

harbu (approved):

harbu:
Are there API changes btw? New exports?

PR #466, opened 2022-02-22
client [NET-630]: API documentation

teogeb:
Defined the `StreamrClient` public API by annotating many methods/fields as internal methods. Some method/classes were annotated to the `Important` category to make it easier to focus to the most important API entitities.

The API document can be generated by running `npm run docs`.

Before this PR the `StreamrClient` class had 144 visible methods. That is now reduced to 56 public methods. In total ~200 methods/fields have been hidden.

Currently the public API is defined by excluding internal methods: that is, all exported classes are part of public API if we don't explicitly add the `@internal` annotation. 

That "exclude the internals"  style was good for this PR as the goal was to define the current public API. But in the future we may want to swap the opt-out style to opt-in. There are less public methods than private method, and therefore it is easier to manage the public methods than the private methods. The opt-in style is also very straightforward when we add new methods (typically we add just some public methods and many private methods), and it is less prone to errors.

### Other user-visible changes

- Removed some deprecated/obsolete methods/types:
  - `client.disconnect()` (deprecated) -> use `client.destroy()`
  - `unsubscribeAll()` -> use `unsubscribe()` without parameters
  - `getStreamFromContract` -> use `getStream()`
  - `APIKeyAuthConfig` (deprecated)
  - `UsernamePasswordAuthConfig` (deprecated)
- Some new helper types are exported: e.g. `AuthConfig`, `EthereumAddress`
- Typo fixed: `unregisterStorageEventListeners`

### Refactoring

- `Session` doesn't extend `EventEmitter` (so that we don't need to exclude `EventEmitter` internals manually)
- In `index-exports.ts` some classes are listed explicitly instead of `export * from ...` (so that we can control some visibility without @internal annotations)
-  Some exports were moved from class files to `index-exports.ts`
- `StreamrClientOptions` is a direct import, not an alias to `BrubeckClientConfig`
- `EthereumConfig` is a type, not an abstract class (for consistency)

### Future improvements

- We could unify some method names, e.g. `register*Listener()` vs `onMessage`/`onError()`/....
- When we generate the API doc (`npm run docs`), we get a list of classes which are referenced in the API docs but are not yet included in the docs. There are e.g. some methods from `network` and `protocol` packages. Maybe we should export most of those?
- Some methods are listed in the API docs as "properties" (e.g. `connect` and `getNode`), because we implement those as properties. For end-users that is an implementation detail, which we could hide.
- Some method return types could be simplified (e.g. `client.destroy()`), see `BrubeckNode.ts:159` for possible implementation.
- The "hierarchy view" of `StreamrClient` class shows a lot of implementation details. We should hide it, if possible.


README.md
```diff
@@ -163,6 +163,12 @@ git push origin client/vX.Y.Z
 npm run build-production
 cd dist
 npm publish
+
+# Generate & upload API docs (if a major/minor version update)
```
harbu, line 167: 👍

packages/broker/src/broker.ts
```diff
@@ -56,7 +56,7 @@ export const createBroker = async (config: Config): Promise<Broker> => {
                 httpServer = await startHttpServer(httpServerRoutes, config.httpServer, apiAuthenticator)
             }
 
-            const nodeId = await streamrClient.getNodeId()
+            const nodeId = (await streamrClient.getNode()).getNodeId()
```
harbu, line 59: Hmm I thought the intention is that we move more to calls being delegated to node via client?

packages/broker/test/utils.ts
```diff
@@ -163,6 +163,7 @@ export const createTestStream = async (
         id,
         ...props
     })
+    // @ts-expect-error internal method (let's use a parameter to createStream when we have NET-606)
```
harbu: Now that NET-606 has been redefined, will this actually apply after all?

harbu (commented):
LGTM but I'm wonder about the amount of `// @ts-expect-error` in tests? If we made things `@internal` instead of `private` can we get rid some of them? (They skip compilation so a bit meh)

harbu (approved):

PR #467, opened 2022-02-22
fix: onResent

teogeb:
Fix the `onResent` method signature in `ResendSubscription `. It needs to expose the `Signal` or we should add a way to add+remove a listener. Needed e.g. in React front-end.


harbu (approved):
lgtm; merge when tests green

PR #468, opened 2022-02-22
client: update() requires a  parameter

teogeb:
Method `stream.update()` now requires a parameter `props`


packages/client/src/Stream.ts
```diff
@@ -108,15 +108,19 @@ class StreamrStream implements StreamMetadata {
     /**
      * Persist stream metadata updates.
      */
-    async update() {
+    async update(props: Omit<StreamProperties, 'id'>) {
+        for (const key of Object.keys(props)) {
+            // @ts-expect-error
+            this[key] = props[key]
+        }
```
harbu: Since updating the stream can fail due to smart contract issues, I think we should update the object itself only after we know the the transaction went thru successfully?

There should also be a test case for this.

harbu (changes requested):
Also CHANGELOG entry

packages/client/src/Stream.ts
teogeb: Fixed and added a test

harbu (approved):
lgtm

harbu (approved):

PR #469, opened 2022-02-22
test: maybe fix flaky test for test-utils.wait()

teogeb:
**Eric's writing:**
I'm gonna go on a gut feeling here and replace this to work with `Date.now()` instead of `performance.now()`. My recollection is that it was like that in the past and was not flaky. I have no idea why taking the time in a more accurate manner causes this, but yeah, if this test stays flaky let's re-look into it.

### DIFF
--- a/packages/test-utils/test/utils.test.ts
+++ b/packages/test-utils/test/utils.test.ts
@@ -160,13 +160,10 @@ describe(waitForCondition, () => {
 })
 
 describe(wait, () => {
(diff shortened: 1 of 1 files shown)


harbu:
I'm gonna go on a gut feeling here and replace this to work with `Date.now()` instead of `performance.now()`. My recollection is that before that change this was not flaky. May be wrong here ofc, but let's test it out.

harbu (approved):

PR #470, opened 2022-02-22
NET-717: client: do not crash on validation

harbu:
Basically just treats the message as in valid if JSON RPC is down.

### DIFF
--- a/packages/client/src/ConfigBase.ts
+++ b/packages/client/src/ConfigBase.ts
@@ -186,8 +186,8 @@ export const STREAM_CLIENT_DEFAULTS: StrictStreamrClientConfig = {
         }
     },
     cache: {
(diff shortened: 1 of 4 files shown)


packages/client/src/ConfigBase.ts
```diff
@@ -186,8 +186,8 @@ export const STREAM_CLIENT_DEFAULTS: StrictStreamrClientConfig = {
         }
     },
     cache: {
-        maxSize: 10000,
-        maxAge: 30 * 60 * 1000, // 30 minutes
+        maxSize: 100,
+        maxAge: 5000, // 30 minutes
```
harbu: Ignore this one for now, just for being able to reproduce issue.

teogeb (approved):

PR #476, opened 2022-02-24
feat(client): Add client.subscribeAll.

timoxley:
As above


packages/client/src/subscribe/Subscriber.ts
```diff
@@ -35,6 +38,58 @@ export default class Subscriber implements Context {
         return this.subscribeTo(streamPartId, onMessage)
     }
 
+    /**
+     * Subscribe to all partitions for stream.
+     */
+    async subscribeAll<T>(streamId: StreamID, onMessage?: SubscriptionOnMessage<T>): Promise<MessageStream<T>> {
```
teogeb, line 45: Will the user be able to unsubscribe this subscription by calling `sub.unsubcribe()`? 

Maybe this could return a `Subscription` instead of `MessageStream`, e.g. if we convert `Subscription` to an interface?

packages/client/src/subscribe/Subscriber.ts
```diff
@@ -116,7 +171,7 @@ export default class Subscriber implements Context {
     /**
      * Count all subscriptions.
      */
-    private countAll() {
```
teogeb, line 120: Could be annotated as `@internal` and rename to `getSubscriptionCount()`

teogeb (commented):
Great improvement 👍  One open question about unsubscribing

packages/client/test/integration/SubscribeAll.test.ts
```diff
@@ -0,0 +1,155 @@
...
+    })
+
+    beforeEach(async () => {
+        // eslint-disable-next-line require-atomic-updates
+        client = await createClient()
+        M = client.subscriber
+        client.debug('connecting before test >>')
+        await Promise.all([
+            client.connect(),
+        ])
```
harbu:
```suggestion
        await client.connect()
```

packages/client/test/integration/SubscribeAll.test.ts
```diff
@@ -0,0 +1,155 @@
...
+                client.unsubscribeAll()
+            }
+        })
+        const eachPartition = Array(PARTITIONS).fill(0).map((_v, streamPartition) => streamPartition)
+        const pubs = await Promise.all(eachPartition.map((streamPartition) => {
+            return publishTestMessages(NUM_MESSAGES, { partitionKey: streamPartition })
+        }))
+        const publishedMsgs = pubs.flat()
+        expect(publishedMsgs.length).toBe(PARTITIONS * NUM_MESSAGES)
+        await sub.onFinally()
```
harbu: What's the function of `sub.onFinally()` here?

packages/client/test/integration/SubscribeAll.test.ts
```diff
@@ -0,0 +1,155 @@
...
+            // eslint-disable-next-line no-await-in-loop
+            await client.unsubscribe({ streamId: stream.id, partition: p })
+            if (p === eachPartition.at(-1)) {
+                // should have ended after last partition unsubbed
+                expect(onFinallyCalled).toHaveBeenCalledTimes(1)
+            } else {
+                expect(onFinallyCalled).toHaveBeenCalledTimes(0)
+            }
+        }
+
```
harbu: Simpler?
```suggestion
        for (const p of eachPartition) {
            expect(onFinallyCalled).toHaveBeenCalledTimes(0)
            // eslint-disable-next-line no-await-in-loop
            await client.unsubscribe({ streamId: stream.id, partition: p })
        }
        expect(onFinallyCalled).toHaveBeenCalledTimes(1)

```

harbu (approved):

harbu:
Let's add entry to changelog

harbu (commented):
- Let's add entry to changelog under "Added" section.
- What about unsubscribing

packages/client/test/integration/SubscribeAll.test.ts
timoxley: Sure, nice one.

packages/client/test/integration/SubscribeAll.test.ts
timoxley: waits for subscription to finish, but doesn't tell it to finish.

packages/client/src/subscribe/Subscriber.ts
timoxley: Moved `unsubscribe` onto the `MessageStream`.

packages/client/test/integration/SubscribeAll.test.ts
timoxley: Done

packages/client/src/subscribe/MessageStream.ts
```diff
@@ -100,4 +100,51 @@ export class MessageStream<
...
+    async unsubscribe() {
+        this.end()
+        await this.return()
+    }
+}
+
+export async function pullManyToOne<T>(context: Context, inputStreams: MessageStream<T>[], onMessage?: MessageStreamOnMessage<T>) {
+    if (inputStreams.length === 1) {
+        return inputStreams[0]
+    }
```
harbu: What about connecting up message handler `onMessage` in single item case?

packages/client/src/subscribe/MessageStream.ts
```diff
@@ -100,4 +100,51 @@ export class MessageStream<
...
         return super.forEachBefore(fn) as MessageStream<T, InType, OutType>
     }
+
+    async unsubscribe() {
+        this.end()
+        await this.return()
+    }
+}
+
+export async function pullManyToOne<T>(context: Context, inputStreams: MessageStream<T>[], onMessage?: MessageStreamOnMessage<T>) {
```
harbu: Not obvious what the return type would be here. Should it be explicit?

packages/client/src/subscribe/MessageStream.ts
```diff
@@ -100,4 +100,51 @@ export class MessageStream<
         // this method override just fixes the output type to be MessageStream rather than Pipeline
         return super.forEachBefore(fn) as MessageStream<T, InType, OutType>
     }
+
+    async unsubscribe() {
```
harbu: Not obvious what the return type would be here. Should it be explicit?

packages/client/src/subscribe/Resends.ts
```diff
@@ -102,6 +104,28 @@ export default class Resend implements Context {
...
+     * Resend for all partitions of a stream.
+     */
+    async resendAll<T>(streamId: StreamID, options: ResendOptions, onMessage?: MessageStreamOnMessage<T>): Promise<MessageStream<T>> {
+        const { partitions } = await this.streamEndpoints.getStream(streamId)
+        if (partitions === 1) {
+            // nothing interesting to do, treat as regular subscription
+            return this.resend<T>(streamId, options, onMessage)
+        }
+
+        const eachPartition = Array(partitions).fill(0).map((_, streamPartition) => streamPartition)
```
harbu: I've noticed this same need come up many times in the code. The need to create a "range" array from `n` to `m`. Wonder if we should extract it as a utility function?

packages/client/src/subscribe/Resends.ts
```diff
@@ -102,6 +104,28 @@ export default class Resend implements Context {
...
+        // create resend for each partition
+        const subs = await Promise.all(eachPartition.map(async (streamPartition) => {
+            return this.resend<T>({
+                streamId,
+                partition: streamPartition,
+            }, options)
+        }))
+
+        return pullManyToOne(this, subs, onMessage)
```
harbu, line 126: nice 👍

packages/client/test/end-to-end/ResendAll.test.ts
```diff
@@ -0,0 +1,121 @@
...
+            stream,
+            timeout: WAIT_FOR_STORAGE_TIMEOUT,
+        })
+    }, WAIT_FOR_STORAGE_TIMEOUT * 2)
+
+    afterAll(async () => {
+        await client?.destroy()
+    })
+
+    describe('no data', () => {
```
harbu, line 63: Maybe a comment here to let the future reader know that order of tests is important.

packages/client/test/end-to-end/SubscribeAll.test.ts
```diff
@@ -0,0 +1,149 @@
...
+        const sub = await client.subscribeAll(stream.id, (msg) => {
+            subMsgs.push(msg)
+            if (subMsgs.length === MAX_MESSAGES) {
+                client.unsubscribe()
+            }
+        })
+        const eachPartition = Array(PARTITIONS).fill(0).map((_v, streamPartition) => streamPartition)
+        const pubs = await Promise.all(eachPartition.map((streamPartition) => {
+            return publishTestMessages(NUM_MESSAGES, { partitionKey: streamPartition })
+        }))
```
harbu: I noticed this same (similar?) segment is repeated (also present in ResendAll test?), extract into helper function?

harbu:
Hey @timoxley ,

Sorry to go back and forth on this so much, I was wondering if it would be a good idea to make methods `subscribe` and `resend` work directly in this way instead of first releasing them under the different method names `subscribeAll ` and `resendAll`.

My thinking is that if people start building things on top of those new method names, we probably want to get rid of them later and just replace them with `subscribe` and `resend`? So are they forced to make code changes soon after anyways?

EDIT: Yeah, looks like we're up for major version bump anyway soonish: https://github.com/streamr-dev/network-monorepo/blob/main/packages/client/CHANGELOG.md. So dunno, what do you think, would it just make sense to make subscribe / resend work so that they do all partitions by default at one go?

packages/client/src/subscribe/Subscriber.ts
teogeb: What would be the role of `Subscription` class when this PR is in `main`? Will also the subscribe() method returns `MessageStream` and then we remove the `Subscription` class?

timoxley:
@harbu no problem, but I actually think we should just keep them as-is. If partitioning is used primarily to reduce load, and the default subscribe/resend subscribes to all partitions, then you may end up being sent too much data.

packages/client/src/subscribe/MessageStream.ts
timoxley: ah you're right

packages/client/test/end-to-end/SubscribeAll.test.ts
timoxley: Switched to https://lodash.com/docs/4.17.15#range

packages/client/test/end-to-end/ResendAll.test.ts
timoxley: sure, done.

packages/client/src/subscribe/Resends.ts
timoxley: Switched to https://lodash.com/docs/4.17.15#range

packages/client/src/subscribe/MessageStream.ts
timoxley: Added.

harbu (approved):

PR #479, opened 2022-02-25
fix: fixed error, invalid streams are ignored now

samt1803:


packages/client/src/StreamRegistry.ts
```diff
@@ -258,7 +258,16 @@ export class StreamRegistry implements Context {
     searchStreams(term: string | undefined, permissionFilter: SearchStreamsPermissionFilter | undefined): AsyncGenerator<Stream> {
         this.debug('Search streams term=%s permissions=%j', term, permissionFilter)
         return map(
-            fetchSearchStreamsResultFromTheGraph(term, permissionFilter, this.graphQLClient),
+            filter(fetchSearchStreamsResultFromTheGraph(term, permissionFilter, this.graphQLClient),
+                (item: SearchStreamsResultItem) => {
+                    try {
+                        Stream.parsePropertiesFromMetadata(item.stream.metadata)
```
teogeb, line 264: Maybe we don't need to parse the JSON twice (here, and in `parseStream()`). Let's optimize that later.

samt1803:
the parsing of metadata is done twice now, this could be optimised in the future

teogeb (approved):

PR #485, opened 2022-02-28
Remove client.getStreamPartsByStorageNode() and update related cli-tools command

teogeb:
Remove de-facto internal method `client.getStreamPartsByStorageNode()`.  If we need to query streams of a storage node, we can use `client.getStoredStreamsOf()`. (It returns `Stream` objects instead of `StreamPartID`s).

Rename cli-tools command `storage-node list-stream-parts` to `storage-node list-streams` and change the output format.


packages/cli-tools/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Rename command `storage-node list-stream-parts` to `storage-node list-streams` and change the output format
```
harbu: Just thought about this, it's more like the command "replaced" the old one than "renamed" since "rename" refers to the functionality staying the same (it's not since it now lists stream instead of stream parts).

```suggestion
- Replace command `storage-node list-stream-parts` with `storage-node list-streams` and change the output format
```

harbu (approved):
lgtm; see comment about changelog

PR #486, opened 2022-02-28
refactor: Annotate Ethereum address parameters

teogeb:
Annotated `string` parameters which are actually Ethereum addresses with `EthereumAddress` type.


packages/broker/src/plugins/metrics/node/MetricsPublisher.ts
```diff
@@ -13,11 +14,11 @@ const logger = new Logger(module)
 
 export class MetricsPublisher {
 
-    private readonly nodeAddress: string
+    private readonly nodeAddress: EthereumAddress
     private readonly client: StreamrClient
     private readonly streamIdPrefix: string
 
-    constructor(nodeAddress: string, client: StreamrClient, streamIdPrefix: string) {
+    constructor(nodeAddress: EthereumAddress, client: StreamrClient, streamIdPrefix: string) {
```
harbu: Ah. I think the naming of the variable here is confusing, I think what actually gets passed in is a "address#sessionId# style string.

harbu (changes requested):

PR #490, opened 2022-03-01
Net-728: proxy subscriptions

juslesan:
Rework proxy publishing to allow proxy subscribing as well.

New protocol enum ProxyConnection that has two types, 'publish' and 'subscribe'.

```javascript
async openProxyConnections(streamDefinition: StreamDefinition, nodeId: string, direction: ProxyDirection)
async closeProxyConnections(streamDefinition: StreamDefinition, nodeId: string, direction: ProxyDirection)
```


packages/client/src/BrubeckNode.ts
```diff
@@ -236,4 +236,26 @@ export default class BrubeckNode implements Context {
             this.debug('closeProxyConnectionOnStream << %o', streamPartId, nodeId)
         }
     }
+
+    async openSubscribeProxyConnectionOnStreamPart(streamPartId: StreamPartID, nodeId: string): Promise<void> {
+        try {
+            if (!this.cachedNode || !this.startNodeComplete) {
```
timoxley: I'm thinking this `!this.cachedNode || !this.startNodeComplete` should probably become a method e.g. 
```ts
isStarting(): boolean {
  return !this.cachedNode || !this.startNodeComplete
}
```
Low priority though.

packages/client/src/subscribe/Subscriber.ts
```diff
@@ -185,6 +187,30 @@ export default class Subscriber implements Context {
...
+        ])
+    }
+
+    async removeSubscribeProxies(streamDefinition: StreamDefinition, nodeIds: string[]): Promise<void> {
+        const streamPartId = await this.streamIdBuilder.toStreamPartID(streamDefinition)
+        await Promise.allSettled([
+            ...nodeIds.map(async (nodeId) => this.node.closeSubscribeProxyConnectionOnStreamPart(streamPartId, nodeId))
+        ])
+    }
+
```
timoxley: You know I feel like this stuff should probably go into its own file e.g. `subscribe/SubscriberProxy.ts`

e.g.
```ts
class SubscriberProxy {
    constructor(
        context: Context,
        private node: BrubeckNode,
        @inject(StreamIDBuilder) private streamIdBuilder: StreamIDBuilder,
    ) {
    
    }
    
    async setSubscribeProxy(streamDefinition: StreamDefinition, nodeId: string): Promise<void> {
        const streamPartId = await this.streamIdBuilder.toStreamPartID(streamDefinition)
        await this.node.openSubscribeProxyConnectionOnStreamPart(streamPartId, nodeId)
    }
    // etc
}
```
Your call though.

packages/network/src/logic/node/Node.ts
```diff
@@ -255,11 +265,61 @@ export class Node extends EventEmitter {
...
+        await Promise.all([
+            new Promise<void>((resolve, reject) => {
+                resolveHandler = (node: string, stream: StreamPartID) => {
+                    if (node === contactNodeId && stream === streamPartId) {
+                        resolve()
+                    }
+                }
+                rejectHandler = (node: string, stream: StreamPartID) => {
+                    if (node === contactNodeId && stream === streamPartId) {
+                        reject(`Joining stream as pure publisher failed on contact-node ${contactNodeId} for stream ${streamPartId}`)
```
timoxley: Should reject with an `Error` rather than a `string` so there's a stack trace.

packages/network/src/logic/node/Node.ts
```diff
@@ -255,11 +265,61 @@ export class Node extends EventEmitter {
...
+        await Promise.all([
+            new Promise<void>((resolve, reject) => {
+                resolveHandler = (node: string, stream: StreamPartID) => {
+                    if (node === contactNodeId && stream === streamPartId) {
+                        resolve()
+                    }
+                }
+                rejectHandler = (node: string, stream: StreamPartID) => {
+                    if (node === contactNodeId && stream === streamPartId) {
+                        reject(`Joining stream as pure subscriber failed on contact-node ${contactNodeId} for stream ${streamPartId}`)
```
timoxley:
```js
reject(new Error(`Joining stream as pure subscriber …`))
```

packages/network/src/logic/node/Node.ts
```diff
@@ -255,11 +265,61 @@ export class Node extends EventEmitter {
...
+                    }
+                }
+                this.on(Event.SUBSCRIBE_STREAM_ACCEPTED, resolveHandler)
+                this.on(Event.SUBSCRIBE_STREAM_REJECTED, rejectHandler)
+            }),
+            this.proxyStreamConnectionManager.openOnewayStreamConnection(streamPartId, contactNodeId, Direction.IN)
+        ]).finally(() => {
```
timoxley, line 282: does this need some kind of system for removing these listeners in response to other events e.g. when node is disconnected or stream is unsubscribed?

juslesan: That's a good point, currently they get cleaned up by time outs at the latest. It would be useful to come up with specific scenarios where cleaning up the listeners fast would be required.

Currently if a proxy node gets disconnected, the ProxyStreamConnectionManager will attempt to reconnect to the node until `closeOnewayStreamConnection` is called. However, if the first attempt to create the proxy-connection fails `openOnewayStreamConnection` should throw and the listener is cleaned in the `finally` block.

In terms of a client unsubscribing in the current design, the client should explicitly tell its network-node to remove the proxy connection if it is no longer needed.

packages/client/src/subscribe/ProxySubscriber.ts
```diff
@@ -0,0 +1,37 @@
...
+        ])
+    }
+
+    async removeSubscribeProxies(streamDefinition: StreamDefinition, nodeIds: string[]): Promise<void> {
+        const streamPartId = await this.streamIdBuilder.toStreamPartID(streamDefinition)
+        await Promise.allSettled([
+            ...nodeIds.map(async (nodeId) => this.node.closeSubscribeProxyConnectionOnStreamPart(streamPartId, nodeId))
+        ])
+    }
+}
```
timoxley: Yeah this is neat. 👍

timoxley (approved):

packages/client/src/BrubeckNode.ts
```diff
@@ -201,23 +201,23 @@ export default class BrubeckNode implements Context {
...
             this.debug('publishToNode << %o', streamMessage.getMessageID())
         }
     }
 
     async openPublishProxyConnectionOnStreamPart(streamPartId: StreamPartID, nodeId: string): Promise<void> {
         try {
-            if (!this.cachedNode || !this.startNodeComplete) {
+            if (this.isStarting()) {
                 await this.startNodeTask()
             }
```
harbu: Correct me if I'm wrong @timoxley , but wouldn't invocation to `this.getNode()` (which aliases to `this.startNodeTask` which is surrounded by `pOnce`) achieve the same result?

So something like:
```ts
const node = await this.getNode() // starts node if necessary
await node.joinStreamPartAsPurePublisher(streamPartId, nodeId)
```

packages/client/src/BrubeckNode.ts
```diff
@@ -228,12 +228,39 @@ export default class BrubeckNode implements Context {
...
+            if (this.isStarting()) {
+                return
+            }
+            await this.cachedNode!.leavePureSubscribingStreamPart(streamPartId, nodeId)
+        } finally {
+            this.debug('closeProxyConnectionOnStream << %o', streamPartId, nodeId)
+        }
+    }
+
+    private isStarting(): boolean {
```
harbu, line 244: Name is a bit confusing cause it seems to refer to something actively starting when it returns true. Would prefer something that indicates that node needs starting, e.g.
```suggestion
    private nodeRequiresStarting(): boolean {
```

or flipping the logic 
```suggestion
    private hasNodeStarted(): boolean {
```

packages/client/src/publish/ProxyPublisher.ts
```diff
@@ -0,0 +1,39 @@
...
+    async removePublishProxy(streamDefinition: StreamDefinition, nodeId: string): Promise<void> {
+        const streamPartId = await this.streamIdBuilder.toStreamPartID(streamDefinition)
+        await this.node.closePublishProxyConnectionOnStreamPart(streamPartId, nodeId)
+    }
+
+    async setPublishProxies(streamDefinition: StreamDefinition, nodeIds: string[]): Promise<void> {
+        const streamPartId = await this.streamIdBuilder.toStreamPartID(streamDefinition)
+        await Promise.allSettled([
+            ...nodeIds.map((nodeId) => this.node.openPublishProxyConnectionOnStreamPart(streamPartId, nodeId))
+        ])
```
harbu: `Array.prototype.map()` already returns an array so you shouldn't have to spread?
```suggestion
        await Promise.allSettled(
            nodeIds.map((nodeId) => this.node.openPublishProxyConnectionOnStreamPart(streamPartId, nodeId))
        ])
```

packages/client/src/publish/ProxyPublisher.ts
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    async removePublishProxies(streamDefinition: StreamDefinition, nodeIds: string[]): Promise<void> {
+        const streamPartId = await this.streamIdBuilder.toStreamPartID(streamDefinition)
+        await Promise.allSettled([
+            ...nodeIds.map(async (nodeId) => this.node.closePublishProxyConnectionOnStreamPart(streamPartId, nodeId))
+        ])
```
harbu:
```suggestion
        await Promise.allSettled(
            nodeIds.map(async (nodeId) => this.node.closePublishProxyConnectionOnStreamPart(streamPartId, nodeId))
        )
```

packages/client/src/subscribe/ProxySubscriber.ts
```diff
@@ -0,0 +1,38 @@
...
+    async removeSubscribeProxy(streamDefinition: StreamDefinition, nodeId: string): Promise<void> {
+        const streamPartId = await this.streamIdBuilder.toStreamPartID(streamDefinition)
+        await this.node.closeSubscribeProxyConnectionOnStreamPart(streamPartId, nodeId)
+    }
+
+    async setSubscribeProxies(streamDefinition: StreamDefinition, nodeIds: string[]): Promise<void> {
+        const streamPartId = await this.streamIdBuilder.toStreamPartID(streamDefinition)
+        await Promise.allSettled([
+            ...nodeIds.map((nodeId) => this.node.openSubscribeProxyConnectionOnStreamPart(streamPartId, nodeId))
+        ])
```
harbu:
```suggestion
        await Promise.allSettled(
            nodeIds.map((nodeId) => this.node.openSubscribeProxyConnectionOnStreamPart(streamPartId, nodeId))
        )
```

packages/client/src/subscribe/ProxySubscriber.ts
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    async removeSubscribeProxies(streamDefinition: StreamDefinition, nodeIds: string[]): Promise<void> {
+        const streamPartId = await this.streamIdBuilder.toStreamPartID(streamDefinition)
+        await Promise.allSettled([
+            ...nodeIds.map(async (nodeId) => this.node.closeSubscribeProxyConnectionOnStreamPart(streamPartId, nodeId))
+        ])
```
harbu:
```suggestion
        await Promise.allSettled(
            nodeIds.map(async (nodeId) => this.node.closeSubscribeProxyConnectionOnStreamPart(streamPartId, nodeId))
        )
```

packages/client/test/end-to-end/ProxyPublishSubscribe.test.ts
```diff
@@ -132,29 +132,110 @@ describe('PubSub with proxy connections', () => {
...
+            msg: 'hellow'
+        })
+        await proxyClient1.publish(stream, {
+            msg: 'hellow'
+        })
+        await proxyClient1.publish(stream, {
+            msg: 'hellow'
+        })
+        await wait(2500)
+        expect(receivedMessages.length).toEqual(3)
```
harbu:
```suggestion
        await until(() => receivedMessages.length >= 3 )
        expect(receivedMessages.length).toEqual(3)
```

packages/client/test/end-to-end/ProxyPublishSubscribe.test.ts
```diff
@@ -132,29 +132,110 @@ describe('PubSub with proxy connections', () => {
...
             // @ts-expect-error private
             .streamPartManager.isSetUp(toStreamPartID(stream.id, 0)))
             .toEqual(false)
 
     }, 15000)
+
+    it('Subscribe only connections work', async () => {
+        const receivedMessages: any[] = []
+        await proxyClient1.subscribe(stream)
+        await wait(2000)
```
harbu: Extract constant

```
const SUBSCRIBE_WAIT_TIME = 2000
```

```suggestion
        await wait(SUBSCRIBE_WAIT_TIME)
```

packages/client/test/end-to-end/ProxyPublishSubscribe.test.ts
```diff
@@ -132,29 +132,110 @@ describe('PubSub with proxy connections', () => {
...
+
+        expect((await onewayClient.getNode())
+            // @ts-expect-error private
+            .streamPartManager.hasInOnlyConnection(toStreamPartID(stream.id, 0), proxyNodeId1))
+            .toEqual(true)
+    }, 15000)
+
+    it('removing proxy subscribing node works', async () => {
+        await proxyClient2.subscribe(stream)
+        await wait(4000)
```
harbu:
```suggestion
        await wait(SUBSCRIBE_WAIT_TIME)
```

packages/client/test/end-to-end/ProxyPublishSubscribe.test.ts
```diff
@@ -132,29 +132,110 @@ describe('PubSub with proxy connections', () => {
...
+        await wait(2000)
+
+        await onewayClient.setSubscribeProxy(stream, proxyNodeId1)
+        await onewayClient.subscribe(stream, (msg) => {
+            receivedMessages.push(msg)
+        })
+        expect((await onewayClient.getNode())
+            // @ts-expect-error private
+            .streamPartManager.hasInOnlyConnection(toStreamPartID(stream.id, 0), proxyNodeId1))
+            .toEqual(true)
```
harbu: Shouldn't be digging into `network` packages internal in client test imo. What about using
```ts
(await onewayClient.getNode()).isProxiedSubscription()
```
?

Or if that isn't enough, just go ahead and create some public helper methods on `Node` / `NetworkNode` classes and use those.

packages/client/test/end-to-end/ProxyPublishSubscribe.test.ts
```diff
@@ -132,29 +132,110 @@ describe('PubSub with proxy connections', () => {
...
+        expect((await onewayClient.getNode())
+            // @ts-expect-error private
+            .streamPartManager.isSetUp(toStreamPartID(stream.id, 0)))
+            .toEqual(false)
+    }, 15000)
+
+    it('setSubscribeProxies, removeSubscribeProxies', async () => {
+        await proxyClient1.subscribe(stream)
+        await proxyClient2.subscribe(stream)
+        await wait(2000)
```
harbu:
```suggestion
        await wait(SUBSCRIBE_WAIT_TIME)
```

packages/network/src/logic/node/NetworkNode.ts
```diff
@@ -16,39 +16,30 @@ export class NetworkNode extends Node {
         this.extraMetadata = metadata
     }
 
-    publish(streamMessage: StreamMessage): void {
+    publish(streamMessage: StreamMessage): void | never {
+        const streamPartId = streamMessage.getStreamPartID()
+        if (this.isProxiedSubscription(streamPartId)) {
+            throw new Error(`Cannot publish to ${streamPartId} as subscribe only connections have been set`)
```
harbu:
```suggestion
            throw new Error(`Cannot publish to ${streamPartId} as subscribe-only connections have been set`)
```

packages/network/src/logic/node/NetworkNode.ts
```diff
@@ -16,39 +16,30 @@ export class NetworkNode extends Node {
...
-                }
-                this.on(Event.PUBLISH_STREAM_ACCEPTED, resolveHandler)
-                this.on(Event.PUBLISH_STREAM_REJECTED, rejectHandler)
-            }),
-            this.openOutgoingStreamConnection(streamPartId, contactNodeId)
-        ]).finally(() => {
-            this.off(Event.PUBLISH_STREAM_ACCEPTED, resolveHandler)
-            this.off(Event.PUBLISH_STREAM_REJECTED, rejectHandler)
-        })
+        await this.openOutgoingStreamConnection(streamPartId, contactNodeId)
```
harbu: Should the naming be coherent between outbound / inbound?:
- openOutgoingStreamConnection vs. openInboundStreamConnection
- closeOutgoingStreamConnection vs. closeInboundStreamConnection

packages/network/src/logic/node/Node.ts
```diff
@@ -31,7 +31,9 @@ export enum Event {
     NODE_SUBSCRIBED = 'streamr:node:subscribed-successfully',
     NODE_UNSUBSCRIBED = 'streamr:node:node-unsubscribed',
     PUBLISH_STREAM_ACCEPTED = 'streamr:node:publish-stream-accepted',
-    PUBLISH_STREAM_REJECTED = 'streamr:node:node-stream-rejected',
+    PUBLISH_STREAM_REJECTED = 'streamr:node:publish-stream-rejected',
+    SUBSCRIBE_STREAM_ACCEPTED = 'streamr:node:subscribe-stream-accepted',
+    SUBSCRIBE_STREAM_REJECTED = 'streamr:node:subscribe-stream-rejected',
```
harbu: A bit of a general thought here.

There seems to be multiple naming conventions for the same concept:
- PublishStream
- OutgoingStreamConnection
- PublishProxy
- pure publisher 
- publish-only connection?
- one-way stream connection (with Direction.OUT)?

(Similar for Subscribe-side)

Is this intentional and does it follow some logic? Or would it be preferable to keep to one name?

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
```diff
@@ -27,9 +29,15 @@ enum State {
     RENEGOTIATING
 }
 
+export enum Direction {
+    OUT = 'outgoing',
```
harbu:
```suggestion
    OUT = 'outbound',
```

packages/network/src/logic/node/ProxyStreamConnectionManager.ts
```diff
@@ -208,6 +253,13 @@ export class ProxyStreamConnectionManager {
         }
     }
 
+    isSubscribeOnlyStream(streamPartId: StreamPartID): boolean {
+        if (this.connections.get(streamPartId) && [...this.connections.values()].length > 0) {
```
harbu:
```suggestion
        if (this.connections.get(streamPartId) && [...this.connections.get(streamPartId)!.values()].length > 0) {
```
?

packages/network/test/integration/oneway-stream-connections.test.ts
```diff
@@ -0,0 +1,303 @@
...
+import { Event as NodeEvent } from '../../src/logic/node/Node'
+import { TrackerInfo } from '../../src/identifiers'
+
+const { StreamMessage, MessageID } = MessageLayer
+
+const defaultStreamPartId = StreamPartIDUtils.parse('stream-0#0')
+
+/**
+ * This test verifies that on receiving a duplicate message, it is not re-emitted to the node's subscribers.
+ */
```
harbu: rm comment?

packages/network/test/integration/oneway-stream-connections.test.ts
```diff
@@ -0,0 +1,303 @@
...
+        await onewayNode.start()
+    })
+
+    afterEach(async () => {
+        await tracker.stop()
+        await Promise.all([
+            onewayNode.stop(),
+            contactNode.stop(),
+            contactNode2.stop(),
+        ])
```
harbu:
```suggestion
        await Promise.allSettled([
            tracker?.stop(),
            onewayNode?.stop(),
            contactNode?.stop(),
            contactNode2?.stop(),
        ])
```

packages/network/test/integration/oneway-stream-connections.test.ts
```diff
@@ -0,0 +1,303 @@
...
+                new Error('Joining stream as pure publisher failed on contact-node non-contact-node for stream stream-0#0')
+            )
+    })
+
+    it('Cannot open publish only stream connection to a node without an existing subscription to the given stream', async () => {
+        await expect(onewayNode.joinStreamPartAsPurePublisher(defaultStreamPartId, 'non-contact-node'))
+            .rejects
+            .toMatchObject(
+                new Error('Joining stream as pure publisher failed on contact-node non-contact-node for stream stream-0#0')
+            )
```
harbu: Same body as previous test case?

packages/network/test/integration/oneway-stream-connections.test.ts
```diff
@@ -0,0 +1,303 @@
...
+            waitForEvent(contactNode, NodeEvent.MESSAGE_RECEIVED),
+            waitForEvent(contactNode2, NodeEvent.MESSAGE_RECEIVED),
+            onewayNode.publish(new StreamMessage({
+                messageId: new MessageID(toStreamID('stream-0'), 0, 120, 0, 'publisher', 'session'),
+                content: {
+                    hello: 'world'
+                },
+            }))
+        ])
+    }, 20000)
```
harbu: Make this slightly higher than the  `waitForEvent` timeout. This way if the `waitForEvent` timeouts, jest will clearly tell us the line number:
``` suggestion
    }, 20100)
```

harbu (changes requested):

packages/network/src/logic/node/Node.ts
juslesan: Yeah this definitely confusing at the moment so I'll unify the naming a bit.

packages/client/src/BrubeckNode.ts
```diff
@@ -27,6 +27,9 @@ export interface NetworkNodeStub {
     getRtt: (nodeId: string) => number|undefined
     setExtraMetadata: (metadata: Record<string, unknown>) => void
     getMetricsContext: () => MetricsContext
+    isStreamSetUp: (streamPartId: StreamPartID) => boolean
```
teogeb:
```suggestion
    hasStreamPart: (streamPartId: StreamPartID) => boolean
```

packages/client/src/publish/ProxyPublisher.ts
```diff
@@ -0,0 +1,39 @@
...
+        const streamPartId = await this.streamIdBuilder.toStreamPartID(streamDefinition)
+        await this.node.openPublishProxyConnectionOnStreamPart(streamPartId, nodeId)
+    }
+
+    async removePublishProxy(streamDefinition: StreamDefinition, nodeId: string): Promise<void> {
+        const streamPartId = await this.streamIdBuilder.toStreamPartID(streamDefinition)
+        await this.node.closePublishProxyConnectionOnStreamPart(streamPartId, nodeId)
+    }
+
+    async setPublishProxies(streamDefinition: StreamDefinition, nodeIds: string[]): Promise<void> {
```
teogeb: To keep the public API simple, we could remove the singular form of methods (`setPublishProxy` and `removePublishProxy`) and keep only the plural forms. It is trivial to provide a single-item list as a `nodeIds` argument. We did similar change in client with permission-related methods.

packages/client/src/publish/ProxyPublisher.ts
```diff
@@ -0,0 +1,39 @@
...
+        const streamPartId = await this.streamIdBuilder.toStreamPartID(streamDefinition)
+        await Promise.allSettled(
+            nodeIds.map((nodeId) => this.node.openPublishProxyConnectionOnStreamPart(streamPartId, nodeId))
+        )
+    }
+
+    async removePublishProxies(streamDefinition: StreamDefinition, nodeIds: string[]): Promise<void> {
```
teogeb: Would it make sense if the method names would be a pair of `addPublishProxy`+`removePublishProxy`? Or `openPublishProxy`+`closePublishProxy`, as the method names are in `BrubeckNode`.

packages/network/src/logic/node/NetworkNode.ts
```diff
@@ -16,37 +16,28 @@ export class NetworkNode extends Node {
         this.extraMetadata = metadata
     }
 
-    publish(streamMessage: StreamMessage): void {
+    publish(streamMessage: StreamMessage): void | never {
+        const streamPartId = streamMessage.getStreamPartID()
+        if (this.isProxiedSubscription(streamPartId)) {
```
teogeb: Should we have similar check in `subscribe()` (and `subscribeAndWaitForJoinOperation()`)?

packages/network/src/logic/node/Node.ts
```diff
@@ -408,4 +472,8 @@ export class Node extends EventEmitter {
     emitJoinFailed(streamPartId: StreamPartID, error: string): void {
         this.emit(streamPartId, error)
     }
+
+    isProxiedSubscription(streamPartId: StreamPartID): boolean {
+        return this.proxyStreamConnectionManager.isSubscribeOnlyStream(streamPartId)
```
teogeb: Should this return true also for publisher-only streams?

juslesan: No, it is exclusively used for proxy subscriptions

packages/network/src/logic/node/NetworkNode.ts
juslesan: I'll this as it's good for clarity

packages/client/src/publish/ProxyPublisher.ts
juslesan: I personally prefer set/remove as the key word here. It feels stricter than add and more appropriate than open/close as the NetworkNode will attempt to maintain a proxy-connection after the initial attempt succeeds

packages/client/src/publish/ProxyPublisher.ts
```diff
@@ -0,0 +1,30 @@
+import { StreamDefinition } from '../types'
+import { StreamIDBuilder } from '../StreamIDBuilder'
+import { inject, Lifecycle, scoped } from 'tsyringe'
+import BrubeckNode from '../BrubeckNode'
+import { Context } from '../utils/Context'
+import { ProxyDirection } from 'streamr-client-protocol'
+
+@scoped(Lifecycle.ContainerScoped)
+export default class ProxyPublisher {
```
teogeb: We could combine `ProxyPublisher` and `ProxySubscriber` to a single class? Actually the methods in these classes are identical, so it would make it easy to read+maintain if the combined class would just have methods:
```
async setSubscribeProxies(streamDefinition: StreamDefinition, nodeIds: NodeId[], direction: ProxyDirection)
async removeSubscribeProxies(streamDefinition: StreamDefinition, nodeIds: NodeId[], direction: ProxyDirection)
```

packages/client/src/publish/ProxyPublisher.ts
```diff
@@ -0,0 +1,30 @@
...
+    constructor(
+        context: Context,
+        private node: BrubeckNode,
+        @inject(StreamIDBuilder) private streamIdBuilder: StreamIDBuilder,
+    ) {
+    }
+
+    async setPublishProxies(streamDefinition: StreamDefinition, nodeIds: string[]): Promise<void> {
```
teogeb: If user gives a `streamDefinition` without a partition, we will set proxies for all partitions of that stream? (Is that feature part of this PR, or maybe next PR?)

packages/client/src/BrubeckNode.ts
```diff
@@ -201,39 +203,44 @@ export default class BrubeckNode implements Context {
...
 
-            return this.cachedNode.publish(streamMessage)
+            return this.cachedNode!.publish(streamMessage)
         } finally {
             this.debug('publishToNode << %o', streamMessage.getMessageID())
         }
     }
 
-    async openPublishProxyConnectionOnStreamPart(streamPartId: StreamPartID, nodeId: string): Promise<void> {
+    async openProxyConnection(streamPartId: StreamPartID, nodeId: string, direction: ProxyDirection): Promise<void> {
```
teogeb, line 221: Could annotate this method with @internal as end-user should use this via `ProxyPublishSubscribe`. Also `closeProxyConnection`.

packages/client/src/ProxyPublishSubscribe.ts
```diff
@@ -0,0 +1,30 @@
...
+import { StreamIDBuilder } from './StreamIDBuilder'
+import { inject, Lifecycle, scoped } from 'tsyringe'
+import BrubeckNode from './BrubeckNode'
+import { Context } from './utils/Context'
+import { ProxyDirection } from 'streamr-client-protocol'
+
+@scoped(Lifecycle.ContainerScoped)
+export default class ProxyPublishSubscribe {
+    constructor(
+        context: Context,
```
teogeb: Unused parameter

teogeb (approved):
LGTM 👍

harbu (approved):

PR #491, opened 2022-03-01
refactor: divide tests to integration and end-to-end tests

teogeb:


harbu:
The idea is that we want to move a lot of the end-to-end tests to eventually be in unit test form, as unit tests are much faster to run. 

Integration tests (with fake environment) should be thought of as a temporary bridge to get us to the above goal.

harbu (approved):

PR #492, opened 2022-03-01
NET-735: broker build issues with npm 8.5.x

kare:
This PR currently changes only broker, but not client. Client uses npm 8.5.2 at the moment.


harbu:
> Client uses npm 8.5.2 at the moment.

Any specific reason?

Dockerfile
```diff
@@ -1,12 +1,10 @@
...
-	# explicitly use npm v8
-	npm install -g npm@8.4.1 --no-audit
+RUN npm config set \
+	unsafe-perm=true \
+	python="$(which python3)"
 COPY . .
-RUN npm config set python "$(which python3)" && npm run bootstrap-pkg -- streamr-broker
-
-RUN npm run prune-pkg -- streamr-broker
+RUN npm run bootstrap-pkg -- streamr-broker && npm run prune-pkg -- streamr-broker
```
harbu, line 7: Is there an advantage of combining these into one line since we are doing this in a "build" image from which we copy stuff over later?

harbu (approved):

Dockerfile
kare: In general it's better to have one layer. If that line is split into two commands/layers then depending on the output of the commands the layers may or may not differ depending on source code changes between command runs.

We could combine those commands into a single `package.json` script and just call one command in `Dockerfile`.

kare:
> > Client uses npm 8.5.2 at the moment.
> 
> Any specific reason?

It uses `npm install -g npm@8` in the workflow. I'll remove those commands from all CI workflows.

Dockerfile
harbu: But aren't we using the multi-stage pattern here https://docs.docker.com/develop/develop-images/multistage-build/ exactly so we don't have to condense steps into a couple giant ones?

kare: Good point. Multistage build is partly for not installing build toolchain into production container. Layer caching rules apply.

https://github.com/hadolint/hadolint/wiki/DL3059
> Please review Docker's layer caching. In general you want layers first which don't change often, so they can be cached.

I guess the result in this case is one or two non-cacheable layers (layers that change often). So it's not a big deal either way.

PR #499, opened 2022-03-02
Add PR template

timoxley:


timoxley:
Merging for now so we have an example to look at.

PR #500, opened 2022-03-03
fix(broker): Use streamr user inside Docker container instead of root

kare:
Broker doesn't need root permissions to run.

BREAKING CHANGE: Changes file system path internal to Docker container
from `/root` to `/home/streamr`. Affects Docker volume mappings when
starting/running container.

## Checklist

- [ ] Updated Changelog
- [x] Updated Documentation https://github.com/streamr-dev/core-frontend/pull/1337/files

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 FROM node:16.14-bullseye as build
-WORKDIR /usr/src/monorepo
+WORKDIR /usr/src/network
 RUN npm config set \
 	unsafe-perm=true \
 	python="$(which python3)"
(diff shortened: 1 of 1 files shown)


timoxley (approved):
👍

timoxley:
> Has passing tests that demonstrate this change works

Perhaps we *could* add some kind of simple CI test for stuff like this?

kare:
> > Has passing tests that demonstrate this change works
> 
> Perhaps we _could_ add some kind of simple CI test for stuff like this?

Do you mean something like this? https://github.com/hadolint/hadolint/wiki/DL3002
Or actually exercising and running the container?

PR #502, opened 2022-03-04
client [NET-606]: Wait for The Graph to sync when necessary 

teogeb:
If we write to `StreamRegistry` or `StorageNodeRegistry` via client methods, all subsequent reads from The Graph provide up-to-date responses automatically. 

Before this PR there was an internal method `streamExistsOnTheGraph` to do the poll the stream creation, but not similar methods for permissions etc. That kind of manual polling is no longer needed as `SynchronizedGraphQLClient` polls the state of The Graph.

Typical scenario for the synchronization:
1.  First we write info to the blockchain, e.g. call `grantPermissions()`:  
    - extract the block number from the transaction receipt (notified by `ObservableContract`'s event emitter)
    - update  `SynchronizedGraphQLClient`'s `requiredBlockNumber`
2. After that, we query the related data, e.g. call `getPermissions()`:
    - start polling the indexing state from The Graph
    - wait until  The Graph metadata that it has indexed the required block (`blockNumber`>=`requiredBlockNumber`)
    - execute the query and return the data to the caller

If we have not written anything to the blockchain via client methods, The Graph polling is not needed and `SynchronizedGraphQLClient` executes the queries immediately.

We manage the waiting by using `Gate` instances. Each queried `blockNumber` corresponds a gate, and the poller opens a gate when it sees that The Graph has indexed the waited block. When we add gates, we manually start the poller if needed. It automatically stops when there are no more work to do.

## Open questions

- Do we need the functionality where users can query The Graph without waiting the sync? It is easy to add but would make the API more complex. Maybe we could implement that later, if needed?

## Other

- Add `jest-extended` library (needed in `Permissions.test.ts`)
- Fix CI setup for `test-integration`

## Future improvements

- There is an implementation of `wait()` method in` SynchronizedGraphQLClient.ts`. It would be good to import that method a library (or alterntively move it to `utils.ts` and import from there). We have an implementation for the method in`streamr-test-utils`, but maybe we shouldn't include that library as a non-test dependency? The `streamr-test-utils` has dependency e.g. to  the `express` library, which is not needed in `StreamrClient` and can increase the bundle size unnecessarily.

## Checklist


packages/client/src/utils/SynchronizedGraphQLClient.ts
```diff
@@ -0,0 +1,177 @@
...
+        debug: Debugger
+    ) {
+        this.getCurrentBlockNumber = getCurrentBlockNumber
+        this.pollTimeout = pollTimeout
+        this.pollRetryInterval = pollRetryInterval
+        this.debug = debug
+    }
+
+    async waitUntiIndexed(blockNumber: number): Promise<void> {
+        this.debug(`wait until indexed ${blockNumber}`)
```
harbu:
```suggestion
        this.debug(`wait until The Graph is synchronized to block ${blockNumber}`)
```

packages/client/src/utils/SynchronizedGraphQLClient.ts
```diff
@@ -0,0 +1,177 @@
...
+        const gate = this.getOrCreateGate(blockNumber)
+        await Promise.race([
+            gate.check(),
+            timeout(
+                this.pollTimeout,
+                `timed out while waiting for The Graph index update for block ${blockNumber}`,
```
harbu:
```suggestion
                `timed out while waiting for The Graph to synchronized to block ${blockNumber}`,
```

packages/client/src/utils/SynchronizedGraphQLClient.ts
```diff
@@ -0,0 +1,177 @@
...
+                () => this.removeGate(blockNumber)
```
harbu: Possible race condition here in the presence of multiple same `blockNumber` invocation?

packages/client/src/utils/SynchronizedGraphQLClient.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    /* eslint-disable no-constant-condition, no-await-in-loop, padding-line-between-statements */
+    private async startPolling(): Promise<void> {
+        this.debug('start polling')
+        while (true) {
+            const newBlockNumber = await this.getCurrentBlockNumber()
+            if (newBlockNumber !== this.blockNumber) {
+                this.blockNumber = newBlockNumber
+                this.debug(`poll result: blockNumber=${this.blockNumber}`)
+                const gate = this.gates.get(this.blockNumber)
```
harbu: We should open all gates for which `gate.blockNumber <= this.blockNumber`.

packages/client/src/utils/SynchronizedGraphQLClient.ts
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    private removeGate(blockNumber: number): void {
+        this.gates.delete(blockNumber)
+    }
+
+    /* eslint-disable no-constant-condition, no-await-in-loop, padding-line-between-statements */
+    private async startPolling(): Promise<void> {
+        this.debug('start polling')
+        while (true) {
```
harbu: When we come back from `await` it could be the case that remaining gates were removed while we were `await`ing.
 
```suggestion
        while (this.gates.size === 0) {
```

packages/client/test/unit/SynchronizedGraphQLClient.test.ts
```diff
@@ -0,0 +1,177 @@
...
+    private blockNumber = 0
+    // eslint-disable-next-line no-undef
+    private timer: NodeJS.Timer | undefined
+
+    constructor(states: IndexState[]) {
+        this.states = states
+    }
+
+    getState(): IndexState {
+        return this.states.find((state) => state.blockNumber >= this.getBlockNumber())!
```
harbu: Are you sure about the blockNumber logic here? It seems as if in `this.blockNumber = 3` it would be returning a result with `{ blockNumber: 4, ... }` with the below `beforeEach` block?

packages/client/test/unit/SynchronizedGraphQLClient.test.ts
```diff
@@ -0,0 +1,177 @@
...
+const POLL_INTERVAL = 50
+const INDEXING_INTERVAL = 100
+const MOCK_QUERY = 'mock-query'
+
+interface IndexState {
+    blockNumber: number
+    queryResult: any
+}
+
+class FakeIndex {
```
harbu:
```suggestion
class EmulatedTheGraphIndex {
```

harbu (changes requested):
looking good, see comments

harbu (approved):
lgtm :medal_sports:

PR #503, opened 2022-03-07
refactor: Simplify resend queries in Storage.ts

teogeb:
Refactor the Cassandra query logic in `Storage.ts` by combining `from` and `range` requests to one generic `range` request. The `from` request are now handled as a `range` request which have `toTimestamp` and `toSequenceNumber` set to `MAX_TIMESTAMP_VALUE` and `MAX_SEQUENCE_NUMBER_VALUE`.

To support that change, the Cassandra queries are generated more declaratively. `Storage.fetchRange` supports all possible combinations of `publisherId`+`msgChainId` arguments: e.g. a query with `publisherId` but without `msgChainId`. 

## Other

Optimize the typical case of range/from request, where sequence numbers are not used. In that case we need to execute only one Cassandra query instead of 3 queries.

## Future improvements

- In DataQueryEndpoints.ts the `range` request supports only two combinations of `publisherId`+`msgChainId:` there must be either both of those arguments, or none of those. The from request can have `publisherId`, but not `msgChainId`.  Is there a need to keep this restrictions? https://github.com/streamr-dev/network-monorepo/blob/1f43f6f01e4c09e8a3fff04fe6a82c3512d5f390/packages/broker/src/plugins/storage/DataQueryEndpoints.ts#L208


harbu (approved):
LGTM :medal_sports: , but let's do the sanity performance check we discussed on call before landing

PR #504, opened 2022-03-08
refactor: Search streams error handling

teogeb:
Implement error handling in `StreamRegistry.searchStreams` using `GeneratorUtils.map`'s `onError` callback . This is simpler and more efficient, as we don't need to parse the `metadata` JSON twice.


samt1803 (approved):
LGTM

PR #510, opened 2022-03-14
NET-745: trackers counters bug

harbu:
Fixed bug with @teogeb  in which a node unsubscribing from a stream part would cause all its counters to reset on tracker. Now it will only reset the counter corresponding to the unsubscribed stream part.

Also changed `NetworkNode#getNeighborsForStreamPart` to not throw error if given stream part does not exist. Instead it will return an empty array. I think this reasonable behavior for a public API. 


juslesan (approved):
LGTM

I wonder if there are any remaining states where the node would ignore the trackers instructions, ie. a tracker restarts and the nodes ignore all instructions for a while. Perhaps there should be some syncing logic for such scenarios if they exist.

PR #522, opened 2022-03-24
Add logging for browsers that do not support webrtc connectionState API

juslesan:
Add logging for webrtc browser libraries that do not support the connectionState API

### DIFF
--- a/packages/network/src/connection/WebRtcEndpoint.ts
+++ b/packages/network/src/connection/WebRtcEndpoint.ts
@@ -135,17 +135,25 @@ export class WebRtcEndpoint extends EventEmitter implements IWebRtcEndpoint {
         this.statusReportTimer = setInterval(() => {
             const connectedPeerIds = []
             const pendingPeerIds = []
-            for (const peerId of Object.keys(this.connections)) {
+            const undefinedStates = []
+            const connections = Object.keys(this.connections)
+            for (const peerId of connections) {
                 const lastState = this.connections[peerId].getLastState()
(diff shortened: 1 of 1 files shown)


fonty1 (approved):
👍

packages/network/src/connection/WebRtcEndpoint.ts
```diff
@@ -135,17 +135,25 @@ export class WebRtcEndpoint extends EventEmitter implements IWebRtcEndpoint {
...
                 }
             }
-            const suffix = (pendingPeerIds.length > 0) ? ', still trying to connect: %s' : ''
-            this.logger.info(`Successfully connected to %d peers (%s)${suffix}`,
-                connectedPeerIds.length, getPeerNameList(connectedPeerIds), getPeerNameList(pendingPeerIds))
+            if (connections.length > 0 && connections.length === undefinedStates.length) {
+                this.logger.warn('Cannot determine webrtc datachannel connection states')
+            } else {
+                const suffix = (pendingPeerIds.length > 0) ? ', still trying to connect: %s' : ''
+                this.logger.info(`Successfully connected to %d peers (%s)${suffix}`,
```
fonty1, line 154: Can it still get here with 0 peers? The log, "Successfully connected to 0 peers" sounds a bit off. It would be better without 'successfully'. I.e. "Connected to 0 peers".. "Connected to 4 peers" etc.

juslesan: Yeah that makes sense, I'll make the change

PR #523, opened 2022-03-24
NET-749: fix issue where resend abruptly ends 

harbu:
Fixed issue in which storage node resend would abruptly end before sending all messages to HTTP client.


teogeb (approved):

harbu:
Fixed issue in which storage node resend would abruptly end before sending all messages to HTTP client.

PR #531, opened 2022-03-29
client: Storage node assignment event listeners

teogeb:
Refactor storage node assignment event listeners to use EventEmitter style:
-  add a listener with `on('addToStorageNode')` / `on('removeFromStorageNode')` instead of `registerStorageEventListeners()`
- remove a listener with `off('addToStorageNode')` / `off('removeFromStorageNode')` instead of `unregisterStorageEventListeners()`

The `unRegisterStorageEventListeners()` removed all listeners. The `off('...', listener)` call always removes a specific listener.

Previously there was only one event, which had a field to indicate whether the event is an addition or a removal. Now it is possible to listen either or both of the events.

## Implementation

We get storage node events from the StorageNode contract by adding a listener to that contract. Adding a listener starts a poller which queries the blockchain state in 4 second intervals. 

To avoid unnecessary polling, we add a contract listener only when `StreamrClient` needs the event information. That is, when `client.on('addToStorageNode')` or `client.on('removeFromStorageNode)` is called.  

To support that kind of dynamic listener binding, there is an observable `StreamrClientEventEmitter` which emits an event whenever new listeners are added to a `StreamrClient` instance. The `initEventGateway` helper method is used to do the actual binding (binding the listener in `start` callback, and unbinding it in `stop` callback).

## Future improvements

- refactor `onResent` to use EventEmitter style
- automatically remove listeners when `destroy()` is called (e.g. by using `destroySignal.onDestroy()`)
- use generics in `EventEmitter3` instead of `strict-event-emitter-types` in client

## Checklist

- [x] Updated Changelog

### DIFF
--- a/packages/broker/test/unit/plugins/storage/StorageConfig.test.ts
+++ b/packages/broker/test/unit/plugins/storage/StorageConfig.test.ts
@@ -1,5 +1,5 @@
 import { StorageConfig } from '../../../../src/plugins/storage/StorageConfig'
-import { StorageNodeAssignmentEvent, Stream, StreamrClient } from 'streamr-client'
+import { StorageNodeAssignmentEvent, Stream, StreamrClient, StreamrClientEvents } from 'streamr-client'
 import { EthereumAddress, StreamPartID, StreamPartIDUtils, toStreamID} from 'streamr-client-protocol'
 import { wait } from 'streamr-test-utils'
 
@@ -22,28 +22,25 @@ function makeStubStream(streamId: string): Stream {
 
 describe(StorageConfig, () => {
     let getStoredStreams: jest.Mock<Promise<{ streams: Stream[], blockNumber: number }>, [nodeAddress: EthereumAddress]>
-    let storageEventListener: ((event: StorageNodeAssignmentEvent) => any) | undefined
-    let stubClient: Pick<StreamrClient, 'getStream'
-        | 'getStoredStreams'
-        | 'registerStorageEventListener'
-        | 'unregisterStorageEventListeners' >
+    let storageEventListeners: Map<keyof StreamrClientEvents, ((event: StorageNodeAssignmentEvent) => any)>
+    let stubClient: Pick<StreamrClient, 'getStream' | 'getStoredStreams' | 'on' | 'off' >
     let onStreamPartAdded: jest.Mock<void, [StreamPartID]>
     let onStreamPartRemoved: jest.Mock<void, [StreamPartID]>
     let storageConfig: StorageConfig
 
     beforeEach(async () => {
         getStoredStreams = jest.fn()
-        storageEventListener = undefined
+        storageEventListeners = new Map()
         stubClient = {
             getStoredStreams,
             async getStream(streamIdOrPath: string) {
                 return makeStubStream(streamIdOrPath, )
             },
-            async registerStorageEventListener(cb: (event: StorageNodeAssignmentEvent) => any) {
-                storageEventListener = cb
+            on(eventName: keyof StreamrClientEvents, listener: any) {
+                storageEventListeners.set(eventName, listener)
             },
-            unregisterStorageEventListeners: async () => {
-                storageEventListener = undefined
+            off(eventName: keyof StreamrClientEvents) {
+                storageEventListeners.delete(eventName)
             }
         }
         onStreamPartAdded = jest.fn()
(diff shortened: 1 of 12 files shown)


packages/client/src/StreamrClient.ts
```diff
@@ -96,6 +98,7 @@ class StreamrClientBase implements Context {
         Plugin(this, this.dataunions)
         Plugin(this, this.streamRegistry)
         Plugin(this, this.storageNodeRegistry)
+        Plugin(this, this.eventEmitter)
```
harbu: Needed?

packages/client/src/StreamrClient.ts
```diff
@@ -170,13 +173,27 @@ class StreamrClientBase implements Context {
...
+    }
 }
 
 /**
  * @internal
  */
 export function initContainer(config: StrictStreamrClientConfig, parentContainer = rootContainer) {
     const c = parentContainer.createChildContainer()
+    const eventEmitter = new StreamrClientEventEmitter()
+    c.register(EventEmitterInjectionToken, { useValue: eventEmitter } as any)
```
harbu: Construct implicitly?

packages/client/src/events.ts
```diff
@@ -0,0 +1,80 @@
...
+interface ObservableEventEmitterEvents {
+    addEventListener: (eventName: keyof StreamrClientEvents) => void
+    removeEventListener: (eventName: keyof StreamrClientEvents) => void
+}
+
+/*
+ * An observable EventEmitter: emits an addEventListener/removeEventListener event when a listener
+ * is added or removed
+ */
+export class StreamrClientEventEmitter<C = any> extends EventEmitter3<StreamrClientEvents & ObservableEventEmitterEvents> {
```
harbu: Does this class have to extend `EventEmitter3`?

packages/client/src/events.ts
```diff
@@ -0,0 +1,80 @@
...
+        this.emitListenerEvent('addEventListener', event)
+        return this
+    }
+
+    once<T extends EventNames<StreamrClientEvents & ObservableEventEmitterEvents>>(
+        event: T, fn: EventListener<StreamrClientEvents & ObservableEventEmitterEvents, T>, context?: C
+    ) {
+        const wrappedFn: any = (...args: any[]) => {
+            (fn as any).apply(context, args)
+            this.emitListenerEvent('removeEventListener', event)
```
harbu: I think it may be more clear if the "meta"-events (so to say) wouldn't be events at all but instead implemented as callback functions.

harbu: Alternatively have "meta"-event be emitted and listened on a separate eventemitter instance.

packages/client/src/events.ts
harbu: Maybe name `ObservableEventEmitter`?

packages/client/src/events.ts
```diff
@@ -0,0 +1,80 @@
...
+    private emitListenerEvent(
+        eventName: keyof ObservableEventEmitterEvents, sourceEvent: keyof StreamrClientEvents | keyof ObservableEventEmitterEvents
+    ) {
+        if ((sourceEvent !== 'addEventListener') && (sourceEvent !== 'removeEventListener')) {
+            this.emit(eventName, sourceEvent)
+        }
+    }
+}
+
+export const initEventGateway = <L extends (...args: any[]) => void>(
```
harbu: Maybe a small description of what the function does?

packages/broker/test/unit/plugins/storage/StorageConfig.test.ts
```diff
@@ -22,28 +22,25 @@ function makeStubStream(streamId: string): Stream {
 
 describe(StorageConfig, () => {
     let getStoredStreams: jest.Mock<Promise<{ streams: Stream[], blockNumber: number }>, [nodeAddress: EthereumAddress]>
-    let storageEventListener: ((event: StorageNodeAssignmentEvent) => any) | undefined
-    let stubClient: Pick<StreamrClient, 'getStream'
-        | 'getStoredStreams'
-        | 'registerStorageEventListener'
-        | 'unregisterStorageEventListeners' >
+    let storageEventListeners: Map<keyof StreamrClientEvents, ((event: StorageNodeAssignmentEvent) => any)>
```
harbu:
```suggestion
    let storageEventListeners: Map<keyof StreamrClientEvents, ((event: StorageNodeAssignmentEvent) => void)>
```
?

harbu (approved):

PR #541, opened 2022-04-01
refactor(client): waitForStorage() uses resend()

teogeb:
The `waitForStorage` method uses `resend()` instead of `getStreamLast()`. The functionalities of those methods are almost identical (`resend` has better validation as uses `StreamMessage` objects instead of `StreamMessageAsObject`). 

Removed the obsolete `getStreamLast()` method and `StreamMessageAsObject` interface.

The `waitForStorage` moved from `Publisher` to `Resend` (and updated some DI injections use `delay()` to support this move). The only changes in the implementation are:
- getting the messages from `resend()` instead of `getStreamLast()` (lines 269-297)
- we don't check for stop-state in the while loop (`Publisher` is a stateful instance but `Resend` is not)
  -  maybe no need to add statefulness? or stop-state check that to all resend methods?

Updated `FakeStorageNode` and `FakeStorageNode` to clone the messages before delivering those to the client, because message decryption process modifies message objects.

## Checklist


packages/client/test/integration/Resends.test.ts
```diff
@@ -0,0 +1,90 @@
...
+                messageId: new MessageID(stream.id, 0, Date.now(), 0, 'publisherId', 'msgChainId'),
+                content: {
+                    mockId
+                }
+            })
+            await client.publish(stream.id, msg.getParsedContent())
+            const collectedIds: Set<number> = new Set()
+            await expect(() => client.waitForStorage(msg, {
+                interval: 100,
+                timeout: 4000,
```
harbu: lower timeout

packages/client/test/integration/Resends.test.ts
```diff
@@ -0,0 +1,90 @@
...
+            await stream.addToStorageNode(DOCKER_DEV_STORAGE_NODE)
+            const mockId = Date.now()
+            const msg = new StreamMessage({
+                messageId: new MessageID(stream.id, 0, Date.now(), 0, 'publisherId', 'msgChainId'),
+                content: {
+                    mockId
+                }
+            })
+            await client.publish(stream.id, msg.getParsedContent())
+            const collectedIds: Set<number> = new Set()
```
harbu: `Set` is a bit weird here since we don't e.g. publish multiple messages, maybe just a singular value?

packages/client/test/integration/Resends.test.ts
```diff
@@ -0,0 +1,90 @@
...
+
+        it('no message', async () => {
+            await stream.addToStorageNode(DOCKER_DEV_STORAGE_NODE)
+            const msg = new StreamMessage({
+                messageId: new MessageID(stream.id, 0, Date.now(), 0, 'publisherId', 'msgChainId'),
+                content: {}
+            })
+            await expect(() => client.waitForStorage(msg, {
+                interval: 100,
+                timeout: 2000,
```
harbu: shorter?

packages/client/test/integration/Resends.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        })
+
+        it('no storage assigned', async () => {
+            const msg = new StreamMessage({
+                messageId: new MessageID(stream.id, 0, Date.now(), 0, 'publisherId', 'msgChainId'),
+                content: {}
+            })
+            await expect(() => client.waitForStorage(msg, {
+                interval: 100,
+                timeout: 2000,
```
harbu: shorter timeout?

packages/client/test/test-utils/fake/FakeBrubeckNode.ts
```diff
@@ -61,8 +61,10 @@ class FakeNetworkNodeStub implements NetworkNodeStub {
             .forEach(async (n) => {
                 const networkNode = await n.getNode()
                 if (networkNode.subscriptions.has(msg.getStreamPartID())) {
-                    const deserialized = StreamMessage.deserialize(serialized)
-                    networkNode.messageListeners.forEach((listener) => listener(deserialized))
+                    networkNode.messageListeners.forEach((listener) => {
+                        const deserialized = StreamMessage.deserialize(serialized)
+                        listener(deserialized)
```
harbu, line 67: Maybe a comment on why we `deserialize` per listener

harbu (commented):
LG; few minor changes here and there

harbu (approved):

PR #542, opened 2022-04-01
feat(client): Resends uses random node address

teogeb:
If there are multiple storage node, `resend()` fetches the data from a random node.


harbu (commented):
should we add a changelog.md entry?

harbu (approved):

PR #550, opened 2022-04-05
Net-752: different errors for stream registry queries

juslesan:
Failed queries to on-chain stream registries give more hints than "Stream not found:"

### DIFF
--- a/packages/client/src/StreamRegistry.ts
+++ b/packages/client/src/StreamRegistry.ts
@@ -49,6 +49,20 @@ type StreamPublisherOrSubscriberItem = {
     userAddress: EthereumAddress
 }
 
+const streamContractErrorProcessor = (err: any, streamId: StreamID, registry: string): never {
+    if (err.errors) {
+        if (err.errors.some((e: any) => {
+            return e.reason && e.reason.code === 'CALL_EXCEPTION'
+        })) {
+            throw new NotFoundError('Stream not found: id=' + streamId)
+        } else {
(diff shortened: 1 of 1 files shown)


packages/client/src/StreamRegistry.ts
```diff
@@ -465,4 +465,19 @@ export class StreamRegistry implements Context {
...
+                return e.reason && e.reason.code === 'CALL_EXCEPTION'
+            })) {
+                throw new NotFoundError('Stream not found: id=' + streamId)
+            } else {
+                throw new Error(`Could not reach the ${registry} Smart Contract: ${err.errors[0]}`)
+            }
+        } else {
+            throw new Error(err)
+        }
+    }
```
harbu: If you move this to be a normal function within the file you can get rid of the eslint-disable directive.

packages/client/src/StreamRegistry.ts
```diff
@@ -465,4 +465,19 @@ export class StreamRegistry implements Context {
             })
         ])
     }
+
+    // eslint-disable-next-line class-methods-use-this
+    private streamContractErrorProcessor(err: any, streamId: StreamID, registry: string): never {
+        if (err.errors) {
+            if (err.errors.some((e: any) => {
+                return e.reason && e.reason.code === 'CALL_EXCEPTION'
+            })) {
```
harbu:
```suggestion
            if (err.errors.some((e: any) => e.reason?.code === 'CALL_EXCEPTION')) {
```

harbu (approved):

PR #551, opened 2022-04-06
client: use shared monorepo eslint rules

harbu:
- Make client package use `eslint-config-monorepo` rules
- Make certain rules `warning` level for now (`@typescript-eslint/ban-ts-comment` and `@typescript-eslint/explicit-module-boundary-types`), there's a few slightly involved typing situations that need to be checked for these to get fixed

### DIFF
--- a/packages/client/src/StreamEndpoints.ts
+++ b/packages/client/src/StreamEndpoints.ts
@@ -88,7 +88,7 @@ export class StreamEndpoints implements Context {
         streamIdOrPath: string,
         data: any,
         requestOptions: any = {},
-        keepAlive: boolean = true
+        keepAlive = true
     ) {
         const streamId = await this.streamIdBuilder.toStreamID(streamIdOrPath)
         this.debug('publishHttp %o', {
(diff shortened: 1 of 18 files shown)


packages/client/src/encryption/GroupKey.ts
```diff
@@ -33,8 +33,6 @@ function GroupKeyObjectFromProps(data: GroupKeyProps | GroupKeyObject) {
     return data
 }
 
-export interface GroupKey extends GroupKeyObject {}
```
teogeb, line 36: Is this redundant?

packages/client/src/StreamEndpoints.ts
```diff
@@ -88,7 +88,7 @@ export class StreamEndpoints implements Context {
         streamIdOrPath: string,
         data: any,
         requestOptions: any = {},
-        keepAlive: boolean = true
```
teogeb: Maybe the eslint rules could allow this? It would make the type annotations consistent (always `variableName: type`, and optionally the default value)

packages/client/src/encryption/GroupKey.ts
harbu: yes

teogeb (approved):

packages/client/src/StreamEndpoints.ts
harbu: fixed, we added a new eslint rule with @teogeb and reverted changes related to this

PR #562, opened 2022-04-08
feat(broker): Console metrics plugin

teogeb:
Move console metrics feature from `metrics` plugin to a separate plugin.

The only functionality change is that it doesn't support `0` as a special interval value to disable the logging. As logging is the only feature of this plugin, it is trivial to disable the logging by removing the plugin.


packages/broker/src/plugins/consoleMetrics/ConsoleMetricsPlugin.ts
```diff
@@ -1,25 +1,28 @@
...
+import { Logger } from 'streamr-network'
 
 const logger = new Logger(module)
 
 function formatNumber(n: number) {
     return n < 10 ? n.toFixed(1) : Math.round(n)
 }
 
-export class ConsoleMetrics {
+export interface MetricsPluginConfig {
```
harbu:
```suggestion
export interface ConsleMetricsPluginConfig {
```

harbu (approved):

PR #567, opened 2022-04-11
feat(client): Update encryption key api

teogeb:
Add a public API method to update an encryption key:
```
client.updateEncryptionKey({
    streamId: '/foobar',
    distributionMethod: 'rotate'
})
```

Both `rotate` and `rekey` now support an explicit encryption key (earlier only `rotate` method supported it).

The `groupKeys` config option was renamed to `encryptionKeys`.

### Future improvements

- Rename all client classes/methods to use term `EncryptionKey` instead of `GroupKey` (it is a protocol-level term)
- Support path-format in `encryptionKeys` config option
- Fine-tune encryption key exports or annotations: e.g.  `encryptionKeys` config option uses `GroupKeysSerialized` type which we don't export (we could either export `GroupKeysSerialized`, `GroupKeyish` and `GroupKeyObject` or modify the `encryptionKeys` config option to use `GroupKey` objects) 

### Checklist

- [x] Updated Changelog
- [x] Updated Documentation

### DIFF
--- a/packages/client/src/Config.ts
+++ b/packages/client/src/Config.ts
@@ -141,7 +141,7 @@ export const STREAM_CLIENT_DEFAULTS: StrictStreamrClientConfig = {
 
     // Encryption options
     verifySignatures: 'auto',
-    groupKeys: {}, // {streamId: groupKey}
+    encryptionKeys: {}, // {streamId: groupKey}
 
     // Ethereum and Data Union related options
     // For ethers.js provider params, see https://docs.ethers.io/ethers.js/v5-beta/api-providers.html#provider
--- a/packages/client/src/Config.ts
+++ b/packages/client/src/Config.ts
@@ -141,7 +141,7 @@ export const STREAM_CLIENT_DEFAULTS: StrictStreamrClientConfig = {
 
     // Encryption options
     verifySignatures: 'auto',
-    encryptionKeys: {}, // {streamId: groupKey}
+    encryptionKeys: {},
 
     // Ethereum and Data Union related options
     // For ethers.js provider params, see https://docs.ethers.io/ethers.js/v5-beta/api-providers.html#provider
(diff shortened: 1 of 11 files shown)


packages/client/src/Config.ts
```diff
@@ -141,7 +141,7 @@ export const STREAM_CLIENT_DEFAULTS: StrictStreamrClientConfig = {
 
     // Encryption options
     verifySignatures: 'auto',
-    groupKeys: {}, // {streamId: groupKey}
+    encryptionKeys: {}, // {streamId: groupKey}
```
harbu: Maybe remove comment?
```suggestion
    encryptionKeys: {}, // {streamId: groupKey}
```

packages/client/src/StreamrClient.ts
```diff
@@ -144,6 +145,24 @@ class StreamrClientBase implements Context {
         return sub
     }
 
+    async updateEncryptionKey(opts: UpdateEncryptionKeyOptions): Promise<void> {
+        if (opts.streamId === undefined) {
+            throw new Error('streamId required')
+        }
+        const streamId = toStreamID(opts.streamId)
```
harbu: Use method `StreamIdBuilder#toStreamID` instead for path-only format support.

```suggestion
        const streamId = toStreamID(opts.streamId)
```

packages/client/src/config.schema.json
```diff
@@ -192,7 +192,7 @@
         "gapFillTimeout": {
             "type": "number"
         },
-        "groupKeys": {
+        "encryptionKeys": {
             "type": "object"
         },
```
harbu, line 197: Could be made more precise if you want.

harbu (approved):

harbu (commented):
look above

harbu (approved):

PR #569, opened 2022-04-11
NET-773: publisher validation issue

harbu:
Prevents the publishing of a message to the network if the publisher does not have publish permission on the stream.

### Changes
- Add an end-to-end test that verifies the above statement.
- Modify "class" `Defer` by replacing `isResolved` with `isSettled`, which also covers rejections.

## Checklist

### DIFF
--- a/packages/client/src/publish/PublishPipeline.ts
+++ b/packages/client/src/publish/PublishPipeline.ts
@@ -99,12 +99,12 @@ export class PublishPipeline implements Context, Stoppable {
     }
 
     private filterResolved = ([_streamMessage, defer]: PublishQueueOut): boolean => {
-        if (this.isStopped && !defer.isResolved()) {
+        if (this.isStopped && !defer.isSettled()) {
             defer.reject(new ContextError(this, 'Pipeline Stopped. Client probably disconnected'))
             return false
         }
 
-        return !defer.isResolved()
+        return !defer.isSettled()
     }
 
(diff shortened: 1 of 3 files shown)


packages/client/test/end-to-end/publisher-message-validation.test.ts
```diff
@@ -0,0 +1,46 @@
...
+                privateKey: await fetchPrivateKeyWithGas()
+            }
+        })
+        const stream = await createTestStream(subscriberClient, module)
+        streamId = stream.id
+    }, TIMEOUT)
+
+    it('publishing message with insufficient permissions prevents message from getting sent to network', async () => {
+        let subscriberReceivedMsgs = 0
+        const subscription = await subscriberClient.subscribe(streamId, () => {
```
teogeb: How about if there were two jest mock functions:
```
onMessage = jest.fn()
onError = jest.fn()
```

and in expect line we check that neither is called:
```
expect(onMessage).not.toHaveBeenCalled()
expect(onError).not.toHaveBeenCalled()
```

packages/client/src/publish/PublishPipeline.ts
```diff
@@ -99,12 +99,12 @@ export class PublishPipeline implements Context, Stoppable {
     }
 
     private filterResolved = ([_streamMessage, defer]: PublishQueueOut): boolean => {
```
teogeb: Would it make sense to rename this method to e.g. `filterNonSettled`?

harbu:
@teogeb addressed your PR comments, re-review when you get a chance

teogeb (approved):

PR #573, opened 2022-04-13
NET-774: subscriber shouldn't throw on invalid data

harbu:
Change behavior of subscriptions so that they don't throw unhandled promise rejection in case of invalid messages. Instead a default error handler will be invoked that simply `debug` logs the error.

## Checklist

- [x] Updated Changelog
- [x] Updated Documentation


packages/client/test/end-to-end/client-behaviour-on-invalid-messages.test.ts
```diff
@@ -0,0 +1,87 @@
...
+            })
+        } finally {
+            await creatorClient.destroy()
+        }
+    }, TIMEOUT)
+
+    beforeEach(async () => {
+        subscriberClient = await createClient({
+            auth: {
+                privateKey: await fetchPrivateKeyWithGas()
```
harbu:
```suggestion
                privateKey: fastPrivateKey()
```

teogeb (approved):

PR #574, opened 2022-04-14
fix(client): Publish encrypted messages

teogeb:
Removed a guard in `Encryption.ts` which caused message encryption to be skipped. The check was modified in https://github.com/streamr-dev/network-monorepo/commit/5d4e11e34a96761c7e7cdc776b559c23ec8f5e47, and it seems that the guard is no longer relevant as all private streams should be encrypted (i.e. stream.requireEncryptedData would be true, it that field still would be available). 

The encryption of public streams is checked separately in line 51:
```
const isPublic = await this.streamEndpoints.isPublic(streamId)
if (isPublic || this.isStopped) {
    return
}
```

## Mutability

When we decrypted message, we would mutate the `StreamMessage` so that its content went from encrypted to decrypted. If that message got sent to further nodes (which could happen if the the decryption process occurred before message propagation), it would go out to neighbors in cleartext. This would mean that upon receiving said message, signature validation would fail on neighbors. 

## Possible future improvements

- Any problems with `WeakSet` in SubscribePipeline class?
- Should we also clone `StreamMessage` before encrypting? It would be symmetrical to the decrypt case, and in a sense, less mutability in system.
- If we make `StreamMessage` to be immutable, the cloning can be removed

## Checklist

- [x] Updated Changelog

### DIFF
--- a/packages/broker/test/integration/plugins/storage/DataMetadataEndpoints.test.ts
+++ b/packages/broker/test/integration/plugins/storage/DataMetadataEndpoints.test.ts
@@ -99,7 +99,7 @@ describe('DataMetadataEndpoints', () => {
         const res = JSON.parse(json)
 
         expect(status).toEqual(200)
-        expect(res.totalBytes).toEqual(1443)
+        expect(res.totalBytes).toEqual(1775)
         expect(res.totalMessages).toEqual(4)
         expect(
             new Date(res.firstMessage).getTime()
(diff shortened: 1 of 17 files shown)


hpihkala:
This sounds like the correct fix.

With the old code, is it the case that the client would only encrypt messages if a group key was explicitly set, or there was a persisted group key from a previous session? (I guess in these cases `await this.keyExchange.hasAnyGroupKey(stream.id)` would return `true`)

PR #578, opened 2022-04-19
feat: add installation instructions for m1 macs

ptesavol:
* Added a new readme file containing instructions for running npm install on M1 Macs
* Linked the new readme file from the main README.md
* Added type-casting to a couple of places in MessageQueue.ts and OrderedMsgChain.ts in order to accommodate changes that have happened the latest versions of the types file for the 'heap' package
* Added to package.json the forcing of the executable flag on the 'husky' binary, which has apparently been dropped in some later version of husky

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Monorepo for Streamr Network packages.
 | NodeJS version `16.13.x` and NPM version `8.x` is required |
 | --- |
 
-Installation on m1 Macs requires additional steps, see [install-on-m1.md](/install-on-m1)
+Installation on an M1 Mac requires additional steps, see [install-on-m1.md](/install-on-m1)
 
 Uses [npm workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces) to manage monorepo.
 
--- a/README.md
+++ b/README.md
@@ -31,6 +31,8 @@ Monorepo for Streamr Network packages.
 | NodeJS version `16.13.x` and NPM version `8.x` is required |
 | --- |
 
+Installation on m1 Macs requires additional steps, see [install-on-m1.md](/install-on-m1)
+
 Uses [npm workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces) to manage monorepo.
 
 **Important:** Do not use `npm ci` or `npm install` directly in the sub-package directories.
--- a/install-on-m1.md
+++ b/install-on-m1.md
@@ -1,4 +1,4 @@
-#  Starting network-monorepo development from scratch on M1 Macs 
+#  Starting network-monorepo development from scratch on an M1 Mac
 
 These instructions were tested on 
 
--- a/install-on-m1.md
+++ b/install-on-m1.md
@@ -44,7 +44,7 @@ export OPENSSL_ROOT_DIR=/opt/homebrew/Cellar/openssl@1.1/1.1.1n/
 export OPENSSL_INCLUDE_DIR=/opt/homebrew/Cellar/openssl@1.1/1.1.1n/include
 ```
 
-* Install networ-monorepo from scratch: 
+* Install network-monorepo from scratch: 
 ```
 npm install
 ```
--- a/install-on-m1.md
+++ b/install-on-m1.md
@@ -16,7 +16,7 @@ The current (April 19th 2022) main branch of streamr-network-monorepo  does not 
 
 Running `npm install` also requires a couple of preparations, because 
 
-* Python 2 is not avaialable by default on macOS 12.3 (which prevents the package weak-napi from compiling)
+* Python 2 is not available by default on macOS 12.3 (which prevents the package weak-napi from compiling)
 * Package node-datachannel does not come with pre-compiled M1 binaries, and gets recompiled
 
 
--- a/install-on-m1.md
+++ b/install-on-m1.md
@@ -1,0 +1,50 @@
+#  Starting network-monorepo development from scratch on M1 Macs 
+
+These instructions were tested on 
+
+```
+macOS Monterey Version 12.3
+Apple M1 Max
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -3,7 +3,7 @@
     "private": true,
     "scripts": {
         "check": "npm run --workspaces --if-present check",
-        "prepare": "husky install && npm run versions",
+        "prepare": "chmod a+x node_modules/.bin/husky && husky install && npm run versions",
```
timoxley, line 6: @ptesavol This one seems weird, everything in `node_modules/.bin` *should* already be made executable by `npm`. I wonder if this was required due to a premature exit during `npm install`?

Also `node_modules/.bin` is automatically added to the `PATH` for npm run scripts, so should never be needed inside `package.json` scripts.

install-on-m1.md
```diff
@@ -0,0 +1,50 @@
...
+
+* Install the prerequsites for compiling node-datachannel:
+```
+brew install openssl
+brew install cmake
+export OPENSSL_ROOT_DIR=/opt/homebrew/Cellar/openssl@1.1/1.1.1n/  
+export OPENSSL_INCLUDE_DIR=/opt/homebrew/Cellar/openssl@1.1/1.1.1n/include
+```
+
+* Install networ-monorepo from scratch:
```
timoxley:
```suggestion
* Install network-monorepo from scratch: 
```

install-on-m1.md
```diff
@@ -0,0 +1,50 @@
...
+macOS Monterey Version 12.3
+Apple M1 Max
+nodejs v16.14.2
+npm 8.5.0
+```
+
+## Introduction
+
+The current (April 19th 2022) main branch of streamr-network-monorepo  does not install out-of-the-box on M1 macs using `npm ci`, but rather requires deleting the 
+`package-lock.json` file, and installing from scratch using `npm install`.
```
timoxley, line 15: Hm, we should probably just update the `package-lock` so the same one works for everyone.

packages/network/src/connection/MessageQueue.ts
```diff
@@ -81,11 +81,11 @@ export class MessageQueue<M> {
     }
 
     peek(): QueueItem<M> {
-        return this.heap.peek()
+        return this.heap.peek() as QueueItem<M>
     }
 
     pop(): QueueItem<M> {
-        return this.heap.pop()
+        return this.heap.pop() as QueueItem<M>
```
timoxley, line 88: I wonder why these are now needed, perhaps a TypeScript update changed something.

install-on-m1.md
```diff
@@ -0,0 +1,50 @@
...
+```
+
+## Introduction
+
+The current (April 19th 2022) main branch of streamr-network-monorepo  does not install out-of-the-box on M1 macs using `npm ci`, but rather requires deleting the 
+`package-lock.json` file, and installing from scratch using `npm install`. 
+
+Running `npm install` also requires a couple of preparations, because 
+
+* Python 2 is not avaialable by default on macOS 12.3 (which prevents the package weak-napi from compiling)
```
timoxley:
```suggestion
* Python 2 is not available by default on macOS 12.3 (which prevents the package weak-napi from compiling)
```

README.md
```diff
@@ -31,6 +31,8 @@ Monorepo for Streamr Network packages.
 | NodeJS version `16.13.x` and NPM version `8.x` is required |
 | --- |
 
+Installation on m1 Macs requires additional steps, see [install-on-m1.md](/install-on-m1)
```
timoxley:
```suggestion
Installation on an M1 Mac requires additional steps, see [install-on-m1.md](/install-on-m1)
```

M1 vs m1 for consistency.

timoxley: Note also that this link will *probably* break when viewed via https://www.npmjs.com/package/streamr-client

If the link is an absolute url it'd work in both places, but you'd want to prematurely link it to the `main` tree.  

e.g. https://github.com/streamr-dev/network-monorepo/tree/main/network-monorepo/install-on-m1.md

Can't be the only people with this problem though perhaps npm has some automatic magic specifically for transforming github links to work on npm.com, or perhaps there's some other tool/service for doing a buildtime transform or live redirect.

install-on-m1.md
```diff
@@ -0,0 +1,50 @@
+#  Starting network-monorepo development from scratch on M1 Macs
```
timoxley:
```suggestion
#  Starting network-monorepo development from scratch on an M1 Mac
```

Singular probably better?

install-on-m1.md
```diff
@@ -0,0 +1,50 @@
...
+```
+macOS Monterey Version 12.3
+Apple M1 Max
+nodejs v16.14.2
+npm 8.5.0
+```
+
+## Introduction
+
+The current (April 19th 2022) main branch of streamr-network-monorepo  does not install out-of-the-box on M1 macs using `npm ci`, but rather requires deleting the
```
timoxley:
```suggestion
The current (April 19th 2022) main branch of streamr-network-monorepo does not install out-of-the-box on an M1 Mac using `npm ci`, but rather requires deleting the 
```

timoxley (commented):
A few comments but otherwise lgtm, nice work!

packages/network/src/connection/MessageQueue.ts
ptesavol: Probably a change to the typescript typings of the heap package

install-on-m1.md
ptesavol: I agree agree

package.json
ptesavol: No idea where this comes from. Could be the bash->zsh change that took place?

ptesavol:
Implemented as part of net-763

PR #591, opened 2022-04-25
feat(client): Remove state metrics

teogeb:
Remove some metrics from consoleMetrics plugin:
- `WebRtcEndpoint.messageQueueSize`
- `broker/cassandra.totalBatches`
- `broker/cassandra.meanBatchAge`

These metrics represent a snapshot state of a node. That kind of state data could be queried e.g. via `info` plugin in the future.


harbu (approved):
Let's make sure to write down somewhere which metrics got removed. I'd like to see these be part of the "state" metrics later.

PR #597, opened 2022-04-29
NET-786: Improve metrics data

teogeb:
### New MetricContext

New `MetricsContext` which allow us to collect and report metrics. Reports are generated by `metricsContext.createReportProducer()`, which produces a report e.g. once a minute and delivers it to a listener by calling a callback function.

A report contains one aggregated number for each metric, if the metric recorded some data during the reporting period. In some cases the aggregated number can have value (usually `0`) even if there is no recorded data. There are four built-in metrics:
- `CountMetric`: sum of recorded values (e.g. failed connection count)
- `AverageMetric`: average of recorded values (e.g. average latency)
- `LevelMetric`: similar to `AverageMetric`, but contains data even if there are no recorded values during the report period (e.g. average count of currently active connections)
- `RateMetric`: sum of recorded values divided by seconds (e.g. download speed).

### Usage in Broker

`NodeMetrics` uses separate producers for `sec`, `min`, `hour` and `day` periods. Therefore there is no need to aggregate e.g. `hour` data from `min` samples. 

Earlier the aggregated samples were calculated from `sec` samples by averaging each metric data. There is no longer mismatch in `WebRtcEndpoint.failedConnection` metric (which represents the total count of failed connection attempts) as it is not averaged for `sec`, `min`, `hour` and `day` periods.

Note that a report always contains a data for a full period. E.g if a `hour` producer is started at `01:30:00`, the first report is produced at `03:00:00` and it contains data from `02:00:00` to `03:00:00`.

### Changes in metrics data generation

Adjusted the aggregation logic of these metrics:
- `WebRtcEndpoint.connections`: is an average level of active connections during the reporting period instead of a snapshot count at the reporting period end
- `node.latency`: uses standard average calculation instead of exponential moving average

### Future improvements

In a separate PR we'll simplify `Sample.ts` and `ConsoleMetricsPlugin.ts` to publish/print all available metrics, not just some subset of metrics.

Related to that separate PR we also unify the naming of some metrics:
- ids can be more descriptive: e.g. `connections` is a average connection level, and `failedConnection` is a total count (and it should be in plural)
- we can use dot-notation in the namespaces to get better structure for the report output
- we can use `broker.plugin...` prefix for all plugin metrics

## Checklist


packages/network/src/helpers/Metric.ts
```diff
@@ -0,0 +1,234 @@
...
+    protected abstract onRecord(value: number): void
+
+    abstract getAggregatedValue(): number | undefined
+}
+
+export type MetricsDefinition = Record<string,Metric>
+
+interface MetricEvents {
+    record: (value: number) => void
+}
```
harbu: I'd move these above the class declaration above.

packages/network/src/helpers/Metric.ts
```diff
@@ -0,0 +1,234 @@
+import EventEmitter from 'eventemitter3'
+import { set } from 'lodash'
+import { scheduleAtFixedRate } from './scheduler'
+
+export abstract class Sampler {
+
+    protected metric: Metric
```
harbu:
```suggestion
    protected readonly metric: Metric
```

packages/network/src/helpers/Metric.ts
```diff
@@ -0,0 +1,234 @@
...
+    private listener: any
```
harbu:
```suggestion
    private readonly listener: any
```

packages/network/src/helpers/Metric.ts
```diff
@@ -0,0 +1,234 @@
...
+
+interface MetricEvents {
+    record: (value: number) => void
+}
+
+export class Metric {
+
+    private latestValue: number | undefined
+    private eventEmitter: EventEmitter<MetricEvents> = new EventEmitter()
+    private samplerFactory: (metric: Metric) => Sampler
```
harbu:
```suggestion
    private readonly samplerFactory: (metric: Metric) => Sampler
```

packages/network/src/helpers/Metric.ts
```diff
@@ -0,0 +1,234 @@
...
+export type MetricsDefinition = Record<string,Metric>
+
+interface MetricEvents {
+    record: (value: number) => void
+}
+
+export class Metric {
+
+    private latestValue: number | undefined
+    private eventEmitter: EventEmitter<MetricEvents> = new EventEmitter()
```
harbu:
```suggestion
    private readonly eventEmitter: EventEmitter<MetricEvents> = new EventEmitter()
```

packages/network/src/helpers/Metric.ts
```diff
@@ -0,0 +1,234 @@
...
+    }
+}
+
+/*
+ * Average of all records within a sampling period. If the sampling data is
+ * continuous, you may want to use LevelSampler instead.
+ *
+ * E.g. average latency
+ */
+class AverageSampler extends CountSampler {
```
harbu: I'd prefer it to keep track of `sum` itself and not inherit. Maybe just simpler from inheritance analysis viewpoint.
```suggestion
class AverageSampler extends CountSampler {
```

packages/network/src/helpers/Metric.ts
```diff
@@ -0,0 +1,234 @@
...
+        super((m) => new LevelSampler(m), initialValue)
+    }
+}
+
+/*
+ * Sum of records divided by seconds.
+ *
+ * E.g. download speed (bytes per second)
+ */
+class RateSampler extends CountSampler {
```
harbu: Same point as above

harbu (approved):

PR #603, opened 2022-05-03
NET-785: Dynamic metrics

teogeb:
All available metrics are sent to the metrics streams (if `metrics` plugin enabled) and/or printed to console (if `consoleMetrics` plugin enabled). Neither of the plugins picks the metrics explicitly, but instead just includes the whole report to as the published/printed content.

Earlier the metrics plugins needed to know about other plugins (the `storage` plugin) and what metric data they generate.

### Output

The metrics plugin output has two new fields about a node (`sendMessagesPerSecond`, `receiveBytesPerSecond`) and five new fields about a storage node (`readMessagesPerSecond`, `writeMessagesPerSecond`, `resend*`)

If storage plugin is enabled, this full output for both plugins is:
```
{
    node: {
        sendMessagesPerSecond
        sendBytesPerSecond
        receiveMessagesPerSecond
        receiveBytesPerSecond
        connectionAverageCount
        connectionTotalFailureCount
        publishMessagesPerSecond
        publishBytesPerSecond
        latencyAverageMs
    },
    broker: {
        plugin: {
            storage: {
                readMessagesPerSecond
                readBytesPerSecond
                writeMessagesPerSecond
                writeBytesPerSecond
                resendLastQueriesPerSecond
                resendFromQueriesPerSecond
                resendRangeQueriesPerSecond
            }
        }
    }
}
```


packages/broker/src/plugins/consoleMetrics/ConsoleMetricsPlugin.ts
```diff
@@ -24,68 +25,17 @@ export class ConsoleMetricsPlugin extends Plugin<ConsoleMetricsPluginConfig> {
     async start(): Promise<void> {
         const metricsContext = (await (this.streamrClient!.getNode())).getMetricsContext()
         this.producer = metricsContext.createReportProducer((report: MetricsReport) => {
-            this.printReport(report)
+            // omit timestamp info as that is printed by the logger
+            const data = omit(report, 'period')
```
harbu, line 29: Does this function handle gracefully the case in which `period` would not be present in `report` (for whatever reason)?

harbu (approved):

packages/broker/src/plugins/consoleMetrics/ConsoleMetricsPlugin.ts
teogeb: Yes

PR #605, opened 2022-05-03
network: update node-datachannel 0.3.1

teogeb:


juslesan (commented):
It seems that the network's integration tests are hanging. Would be good to check that it's not caused by the update before merging.

PR #610, opened 2022-05-05
NET-778: use @streamr/config package

harbu:
Use smart contract -related configuration values from package `@streamr/config` as default values for client. This reduces the amount of places we have to maintain hard-coded addresses.  

Related PR: https://github.com/streamr-dev/network-contracts/pull/263

### DIFF
--- a/packages/client/src/Config.ts
+++ b/packages/client/src/Config.ts
@@ -101,28 +101,24 @@ export type StreamrClientConfig = Partial<Omit<StrictStreamrClientConfig, 'netwo
 
 export const STREAMR_STORAGE_NODE_GERMANY = '0x31546eEA76F2B2b3C5cC06B1c93601dc35c9D916'
 
-const info = Chains.load('production')
-
-const mainChainRpcs = info.ethereum.rpcEndpoints.map(({ url }) => ({
-    url,
-    timeout: toNumber(process.env.TEST_TIMEOUT) ?? 30 * 1000
-}))
-
-const sideChainRpcs = info.polygon.rpcEndpoints.map(({ url }) => ({
-    url,
-    timeout: toNumber(process.env.TEST_TIMEOUT) ?? 30 * 1000
-}))
+const chainConfig = Chains.load('production')
 
 const mainChainConfig = {
     name: 'ethereum',
-    chainId: info.ethereum.id,
-    rpcs: mainChainRpcs
+    chainId: chainConfig.ethereum.id,
+    rpcs: chainConfig.ethereum.rpcEndpoints.map(({ url }) => ({
+        url,
+        timeout: toNumber(process.env.TEST_TIMEOUT) ?? 30 * 1000
+    }))
 }
 
 const sideChainConfig = {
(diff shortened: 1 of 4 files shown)


packages/client/src/Config.ts
```diff
@@ -99,6 +100,28 @@ export type StreamrClientConfig = Partial<Omit<StrictStreamrClientConfig, 'netwo
...
 export const STREAMR_STORAGE_NODE_GERMANY = '0x31546eEA76F2B2b3C5cC06B1c93601dc35c9D916'
 
+const chainConfig = Chains.load('production')
+
+const DEFAULT_RPC_TIMEOUT = 120 * 1000
+
+const mainChainConfig = {
+    name: 'ethereum',
+    chainId: chainConfig.ethereum.id,
+    rpcs: chainConfig.ethereum.rpcEndpoints.map(({ url }) => ({
```
teogeb, line 110: Create an object with all properties, and override timeout of that is not set

packages/client/src/ConfigTest.ts
```diff
@@ -1,24 +1,39 @@
...
 function toNumber(value: any): number | undefined {
     return (value !== undefined) ? Number(value) : undefined
 }
 
+const chainConfig = Chains.load('development')
+
+const DEFAULT_RPC_TIMEOUT = 30 * 1000
+
+const mainChainConfig = {
+    name: 'dev_ethereum',
```
teogeb, line 12: Would it make sense to prefix all test config names with `"dev_"`? Then it would be get the name from the `chainConfig` property

packages/client/src/Config.ts
```diff
@@ -99,6 +100,28 @@ export type StreamrClientConfig = Partial<Omit<StrictStreamrClientConfig, 'netwo
 
 export const STREAMR_STORAGE_NODE_GERMANY = '0x31546eEA76F2B2b3C5cC06B1c93601dc35c9D916'
 
+const chainConfig = Chains.load('production')
+
+const DEFAULT_RPC_TIMEOUT = 120 * 1000
```
teogeb, line 105: Timeouts to the config package? Ideally we wouldn't need to do any transformations here?

packages/client/src/ConfigTest.ts
```diff
@@ -1,24 +1,39 @@
...
     name: 'streamr',
-    chainId: 8995,
-    rpcs: [{
-        url: process.env.SIDECHAIN_URL || `http://${process.env.STREAMR_DOCKER_DEV_HOST || '10.200.10.1'}:8546`,
-        timeout: toNumber(process.env.TEST_TIMEOUT) ?? 30 * 1000,
-    }]
+    chainId: chainConfig.streamr.id,
+    rpcs: chainConfig.streamr.rpcEndpoints.map(({ url }) => ({
+        url,
+        timeout: toNumber(process.env.TEST_TIMEOUT) ?? DEFAULT_RPC_TIMEOUT
```
teogeb, line 25: Do we need this special case about timeouts?

PR #618, opened 2022-05-06
NET-812: Configurable metrics periods

teogeb:
Report periods for `metrics` plugin can be defined in the config file. Each period definition is a pair of `duration` + `streamId`.

The previously hardcoded periods are now used as defaults (`config.schema.json`).

### DIFF
--- a/packages/broker/test/unit/configMigration.test.ts
+++ b/packages/broker/test/unit/configMigration.test.ts
@@ -241,6 +241,7 @@ describe('Config migration', () => {
             city: null
         }
         source.plugins.metrics.consoleAndPM2IntervalInSeconds = 123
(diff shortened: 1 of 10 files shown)


packages/broker/test/unit/configMigration.test.ts
```diff
@@ -277,4 +277,77 @@ describe('Config migration', () => {
         const source = {}
         expect(() => createMigratedConfig(source)).toThrow('Unable to migrate the config')
     })
+
+    describe('from v1 to v2', () => {
```
harbu, line 281: 👍

harbu (approved):
nice stuff! 👍

PR #622, opened 2022-05-09
update to node-datachannel 0.3.2

teogeb:
Update `network` to use `node-datachannel` v0.3.2.

Also fix cleanup logic when using node-datachannel: when all `NodeWebRtcEndpoints` stop, we clean up the global `nodeDataChannel` library object.


packages/network/src/connection/webrtc/BrowserWebRtcConnection.ts
```diff
@@ -3,16 +3,16 @@ import { Logger } from "../../helpers/Logger"
 import { NameDirectory } from "../../NameDirectory"
 import { WebRtcConnectionFactory } from "./WebRtcEndpoint"
 
-const BrowserWebRtcConnectionFactory: WebRtcConnectionFactory = Object.freeze({
+export const webRtcConnectionFactory = new class implements WebRtcConnectionFactory {
+    activeWebRtcEndpointCount = 0
```
harbu: No need for counter here
```suggestion
```

harbu (changes requested):

PR #631, opened 2022-05-10
feat(client): Publish metrics

teogeb:
Client publishes metrics to the network at regular intervals. The functionality is similar to Broker's `metrics` plugin. 

Functionality differences:
- Metrics publishing starts only after the client publishes or subscribes to some stream.
- Partition key is calculated from the Ethereum address of the node instead of the full node id. This way we can deterministically derive the partition e.g. from a node private key.

There is a new `metrics` client configuration option to disable metrics publishing or to define custom duration/streamId for reports.

### Internal events

To support the observing of client's publish and subscribe actions, new internal events have been added. Those events are for internal communication only and there is currently no need to expose the events to end-users. 

- `publish` event is emitted when an end-user publishes some content
- `subscribe` event is emitted when an end-user starts a subscription

Each of those event is emitted only after the underlying promise is resolved (the observed operation is "completed").

### Broker

Removed `metrics` plugin. The configuration is migrated automatically: the metrics publishing will be enabled if a config file contained metrics plugin and `nodeMetrics` was not explicitly disabled.

Fixed also `testnet3` migration to support explicitly disabled `nodeMetrics`.

### Future improvements

- We should maybe stop publishing the metrics data if the node is idle for a long period of time?

- We could move the calculation of `publishBytesPerSecond` and `publishMessagesPerSecond` from `Node.ts` to the client. Those metrics could calculate only messages, which are explictly published by the client. Now the reports include data also about metrics JSONs. 

### Checklist

- [x] Updated Changelog
- [x] Updated Documentation


packages/client/README.md
```diff
@@ -128,6 +129,8 @@ You can also create an anonymous client instance that can interact with public s
 const streamr = new StreamrClient()
 ```
 
+By default, the `StreamrClient` publishes metrics to the network at regular intervals. See [Metrics publishing](#metrics-publishing).
```
harbu:
```suggestion
By default, the `StreamrClient` publishes telemetry metrics to the network at regular intervals. See [Metrics publishing](#metrics-publishing).
```

packages/client/src/BrubeckNode.ts
```diff
@@ -31,6 +31,11 @@ export interface NetworkNodeStub {
     hasProxyConnection: (streamPartId: StreamPartID, contactNodeId: string, direction: ProxyDirection) => boolean
 }
 
+export const getEthereumAddressFromNodeId = (nodeId: string) => {
```
harbu:
```suggestion
export const getEthereumAddressFromNodeId = (nodeId: string): string => {
```

packages/client/src/Config.ts
```diff
@@ -62,6 +62,15 @@ export type DebugConfig = {
     inspectOpts: InspectOptions
 }
 
+export interface MetricsPeriodConfig {
+    streamId: string,
+    duration: number
```
harbu:
```suggestion
    streamId: string
    duration: number
```

packages/client/src/MetricsPublisher.ts
```diff
@@ -0,0 +1,96 @@
...
+        @inject(ConfigInjectionToken.Root) rootConfig: StrictStreamrClientConfig
+    ) {
+        this.publisher = publisher
+        this.brubeckNode = brubeckNode
+        this.eventEmitter = eventEmitter
+        this.destroySignal = destroySignal
+        this.periodConfigs = getPeriodConfig(rootConfig)
+    }
+
+    init() {
```
harbu: Was there a reason not to do this stuff in constructor?
```suggestion
    init(): void {
```

packages/client/src/MetricsPublisher.ts
```diff
@@ -0,0 +1,96 @@
...
+        if (this.periodConfigs.length > 0) {
+            this.eventEmitter.on('publish', () => this.ensureStarted())
+            this.eventEmitter.on('subscribe', () => this.ensureStarted())
+            this.destroySignal.onDestroy(() => this.stop())
+        }
+    }
+
+    async ensureStarted(): Promise<void> {
```
harbu:
```suggestion
    private async ensureStarted(): Promise<void> {
```

packages/client/src/StreamrClient.ts
```diff
@@ -93,6 +96,22 @@ class StreamrClientBase implements Context {
...
+     * @category Important
+     */
+    async publish<T>(
+        streamDefinition: StreamDefinition,
+        content: T,
+        timestamp: string | number | Date = Date.now(),
+        partitionKey?: string | number
+    ): Promise<StreamMessage<T>> {
+        const result = await this.publisher.publish(streamDefinition, content, timestamp, partitionKey)
+        this.eventEmitter.emit('publish', undefined)
```
harbu, line 111: Not possible to simplify?
```suggestion
        this.eventEmitter.emit('publish')
```

packages/client/src/StreamrClient.ts
```diff
@@ -106,15 +125,18 @@ class StreamrClientBase implements Context {
...
+        let result
         if (options.resend !== undefined) {
-            return this.resendSubscribe(options, options.resend, onMessage)
+            result = await this.resendSubscribe(options, options.resend, onMessage)
+        } else {
+            result = await this.subscriber.subscribe(options, onMessage)
         }
-
-        return this.subscriber.subscribe(options, onMessage)
+        this.eventEmitter.emit('subscribe', undefined)
```
harbu, line 136: same question as above
```suggestion
        this.eventEmitter.emit('subscribe')
```

packages/client/test/end-to-end/client-behaviour-on-invalid-messages.test.ts
```diff
@@ -76,7 +76,7 @@ describe('client behaviour on invalid message', () => {
             throw new Error('should not get here')
         })
         networkNode = await createNetworkNode({
-            ...ConfigTest.network,
+            ...ConfigTest.network as any,
```
harbu, line 80: ?

packages/client/test/end-to-end/publish-subscribe.test.ts
```diff
@@ -13,7 +13,7 @@ const ENCRYPTED_MESSSAGE_FORMAT = /^[0-9A-Fa-f]+$/
 
 async function startNetworkNodeAndListenForAtLeastOneMessage(streamId: StreamID): Promise<unknown[]> {
     const networkNode = await createNetworkNode({
-        ...ConfigTest.network,
+        ...ConfigTest.network as any,
```
harbu, line 17: ?

packages/client/src/StreamrClient.ts
teogeb: Necessary to compliation

PR #634, opened 2022-05-10
Improve broker readmes

hpihkala:
Quick contribution to broker readmes (md changes only):

- Clean up `README.md` by making it more user-oriented and splitting dev-related information to a new `develop.md` file
- Clean up `plugins.md`, with focus on consistency and describing how to pass the API key to each plugin

### DIFF
--- a/packages/broker/README.md
+++ b/packages/broker/README.md
@@ -6,20 +6,13 @@
 
 # streamr-broker
 
-Main executable for running a broker node in Streamr Network.
-
-The broker node extends the minimal network node provided by the
-[streamr-network](https://github.com/streamr-dev/network) library with
-- client-facing support for foreign protocols (e.g. HTTP, MQTT) via plugins
-- support for long-term persistence of data using Apache Cassandra.
+The Broker node is your application's access point to data streams in the Streamr Network. The Broker node is also used for mining and staking.
 
 ## Table of Contents
 - [Install](#install)
-- [Configure](#configure)
+- [Plugins](#plugins)
 - [Run](#run)
 - [Develop](#develop)
-- [Release](#release)
-- [Misc](#misc)
 
 ## Install
 | NodeJS version `16.13.x` and NPM version `8.x` is required |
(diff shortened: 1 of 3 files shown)


packages/broker/README.md
```diff
@@ -30,96 +23,25 @@ To install streamr-broker:
...
-If instead you want to run a broker node by itself without Docker, follow the steps below.
 
 First install the package
 ```
 npm install -g streamr-broker
 ```
-Optionally, create a configuration file with interactive tool:
+Create a configuration file with interactive tool:
 ```
 broker-init
```
teogeb: Not related to this PR, but seems that the names of the executables are out-of-date. The correct are `streamr-broker-init` and `streamr-broker` (line 42, and `develop.md` line 13) https://github.com/streamr-dev/network-monorepo/blob/main/packages/broker/package.json#L11

packages/broker/develop.md
```diff
@@ -0,0 +1,82 @@
...
+
+    npm run test
+
+We use [eslint](https://github.com/eslint/eslint) for code formatting:
+
+    npm run eslint
+
+Code coverage:
+
+    npm run coverage
```
teogeb: There is no `coverage` script in package.json

packages/broker/develop.md
```diff
@@ -0,0 +1,82 @@
...
+... or adjust debugging to desired level:
+
+    LOG_LEVEL=[debug|info|warn|error]
+
+To disable all logs
+
+    NOLOG=true
+
+
+## Release
```
teogeb, line 47: The up-to-date release procedure is described in monorepo root README.md: https://github.com/streamr-dev/network-monorepo/blob/main/README.md

packages/broker/develop.md
```diff
@@ -0,0 +1,82 @@
...
+
+```bash
+openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -days 36500 -nodes -subj "/CN=localhost"
+```
+
+### Deleting expired data from Storage node
+To delete expired data from storage node run
+
+```
+broker <configFile> --deleteExpired
```
teogeb: This is under command `delete-expired-data` in package.json: https://github.com/streamr-dev/network-monorepo/blob/b679cd0efaa7a91204870ca6e1e2bee4946ae364/packages/broker/package.json#L14

teogeb (approved):
LGTM, some comments about out-of-date executables/scripts

packages/broker/plugins.md
```diff
@@ -1,17 +1,34 @@
 # Plugins
 
+The Broker ships with a number of plugins that add functionality or APIs.
+
 ## Table of Contents
-- [Dependencies](#dependencies)
+- [Authentication](#authentication)
 - [Websocket](#websocket)
 - [MQTT](#mqtt)
 - [PublishHttp](#publishhttp)
```
fonty1: I guess it's a bit late now, but can we call the plugin `HTTP` rather than `PublishHttp` or is there good reason behind the name?

fonty1 (approved):

packages/broker/plugins.md
hpihkala: I guess the reason for the name is that it only supports publish (subscribing over HTTP is non-trivial, as HTTP is a request-response protocol). That being said, I don't see a problem in renaming it to `http` and just saying in the docs that it only supports publish for now.

I did a quick cleanup: In `plugins.md` I used "HTTP" instead of `publishHttp` where possible, however changing the actual plugin name needs to be left for [another ticket](https://linear.app/streamr/issue/NET-820/rename-publishhttp-plugin-to-http).

Then there's the interesting topic of how to actually enable subscribing over HTTP, and is that even necessary. There's [a ticket](https://linear.app/streamr/issue/NET-395/how-to-subscribe-over-http) about it, but it needs some research and hasn't been prioritized very high for now, as there are easy alternatives to subscribing.

PR #639, opened 2022-05-11
NET-804: receipt layer (poc)

harbu:
### Bucket cleanup logic

- How long to store buckets for, when to cleanup, possible persistence of them for later use?
- If messages arrive later (how late?) to an old bucket, do we consider this a new bucket to collect statistics for, or do we revive the original bucket and keep updating it?
 - What to do about buckets of disconnected nodes?
 - **Current implementation**: store buckets in memory until they get closed, after which they are removed from memory.

### Empty bucket handling
- It is possible that no messages occur during a span of time covering one or more buckets.
- Buckets are now tied to specific message chains. How do we determine which message chains to emit empty buckets for? Or do we have some sort of meta-bucket covering multiple message chains?
- **Current implementation**: empty buckets are not implemented.

### Node identity and `sessionId`
- Nodes usually run under a node id of format `ethereumAddress#sessionId` but receipt signing works with Ethereum addresses and private keys.
- Multiple nodes with same ethereum address but different sessionId, how to handle?
- A node can crash and re-join, it's `sessionId` will change.
- **Current implementation**: we treat all nodes as separate, even if they have same Ethereum address (as long as the `sessionId` is different of course). Signature validation and signing logic knows how to strip out the `sessionId`.

### Signing of messages
- Network layer remains unaware of Ethereum addresses, private keys, and hence the exact format of node ids.
- Instead we provide an interface for signature generation / validation that the client implements with Ethereum-specific choices.
- **Current implementation**: pretty much done.

### Misc missing things still
- Message validation missing on `protocol` layer
- Unit test for higher-level classes
- End-to-end test
- Tests that verify  for possible "malicious" activity
 
 ### Terminology
 - Window vs. Bucket
 - Closed Bucket
 

- [ ] Updated Changelog
- [ ] Updated Documentation


packages/network/src/logic/Node.ts
```diff
@@ -349,6 +383,27 @@ export class Node extends EventEmitter {
         return propagationTargets
     }
 
+    protected inspect(nodeId: NodeId, inspectionTarget: NodeId): Promise<Receipt[]> {
```
harbu, line 386: This is WIP, maybe comment out?

packages/network/src/logic/Node.ts
```diff
@@ -107,6 +119,27 @@ export class Node extends EventEmitter {
         }
         this.metricsContext.addMetrics('node', this.metrics)
 
+        if (opts.signers !== undefined) {
+            this.receiptStore = new ReceiptStore(this.peerInfo.peerId)
+            this.receiptRequester = new ReceiptRequester({
```
harbu, line 124: The receipt request could be conditionally enabled; namely if we are a miner node in a bounty. Actually might to need add logic that only certain streams are active for receipt requesting. Those would be the streams that we are getting paid to broker (i.e. part of bounty contract).

packages/network/src/logic/receipts/ReceiptRequester.ts
```diff
@@ -0,0 +1,145 @@
...
+                // TODO: cut connection?
+                return
+            }
+            if (sender !== this.myNodeId) {
+                logger.warn('unexpected receipt from %s (the claim directed to %s does not concern me)',
+                    source, sender)
+                // TODO: cut connection?
+                return
+            }
+            if (!this.signers.receipt.validate(receipt)) {
```
harbu, line 84: Does this also validate the claim itself? I.e. is the if-check after this one necessary?

packages/network/src/logic/receipts/ReceiptStore.ts
```diff
@@ -0,0 +1,34 @@
...
+import { Receipt } from 'streamr-client-protocol'
+
+function store(key: string, receipt: Receipt, map: Map<string, Receipt[]>): void {
+    if (!map.has(key)) {
+        map.set(key, [])
+    }
+    map.get(key)!.push(receipt)
+}
+
+export class ReceiptStore {
```
harbu, line 11: Do we want to persist these on the disk? If so, who's responsible, network, client, or broker?

packages/network/src/logic/receipts/attachInspectResponder.ts
```diff
@@ -0,0 +1,28 @@
+import { NodeToTracker } from '../../protocol/NodeToTracker'
+import { ReceiptStore } from './ReceiptStore'
+import { Receipt } from 'streamr-client-protocol'
+import { Event as NodeToTrackerEvent } from '../../protocol/NodeToTracker'
+import { PeerInfo } from '../../connection/PeerInfo'
+
+export function attachInspectResponder({ myPeerInfo, receiptStore, nodeToTracker }: {
```
harbu, line 7: WIP

packages/network/src/protocol/NodeToTracker.ts
```diff
@@ -143,10 +147,55 @@ export class NodeToTracker extends EventEmitter {
         return requestId
     }
 
+    async sendInspectRequestMessage(
```
harbu, line 150: wip

packages/network/src/protocol/NodeToTracker.ts
```diff
@@ -143,10 +147,55 @@ export class NodeToTracker extends EventEmitter {
...
+            targetNode,
+            subType: RelayMessageSubType.INSPECT_REQUEST,
+            data: {
+                inspectionTarget
+            }
+        }))
+        return requestId
+    }
+
+    async sendInspectResponsePart(
```
harbu, line 169: wip

packages/network/test/unit/ReceiptRequester.test.ts
```diff
@@ -0,0 +1,212 @@
...
+            ),
+            content: {
+                str
+            },
+            signature: 'signature'
+        })
+    })
+}
+
+describe(ReceiptRequester, () => {
```
harbu, line 52: WIP; could use `jest-mock-extended` for clearer mocking

harbu:
Closing PR as we are most likely not going with this approach. Will not deleted branch because we could need the stuff form here for future reference.

PR #642, opened 2022-05-11
Net 763 move composition inside dht node

ptesavol:
- Moved composition to happen inside DHTNode class
- Simplified tests by introducing a Simulator class


juslesan (approved):
LGTM

PR #643, opened 2022-05-12
test(broker): Config wizard and migration

teogeb:
Enhanced config-related tests.

`ConfigWizard.test.ts`:
- checks that a generated config file can be used to create a Broker instance
- checks that a generated config file doesn't need migration
- added test for "no plugins" user flow

`configMigration.test.ts`:
- uses `pluginRegistry` to check validity of plugins and plugin configs
- checks that migrated config can't have unknown plugins
- checks that migrated config can't have invalid plugin configs

Added also type annotations for `ConfigWizard` answers.

### DIFF
--- a/packages/broker/test/unit/configMigration.test.ts
+++ b/packages/broker/test/unit/configMigration.test.ts
@@ -2,10 +2,8 @@ import { cloneDeep, merge } from 'lodash'
 import { validateConfig as validateClientConfig } from 'streamr-client'
 import { createMigratedConfig, CURRENT_CONFIGURATION_VERSION, formSchemaUrl, needsMigration } from '../../src/config/migration'
 import BROKER_CONFIG_SCHEMA from '../../src/config/config.schema.json'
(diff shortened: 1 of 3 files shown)


packages/broker/test/unit/configMigration.test.ts
```diff
@@ -130,35 +128,36 @@ const configWizardFull = {
...
+            name: pluginName,
+            streamrClient: undefined,
+            apiAuthenticator: undefined,
+            brokerConfig: config
+        })
+        const schema = plugin.getConfigSchema()
         if (schema !== undefined) {
-            validateConfig(pluginConfig, schema, name)
+            validateConfig(pluginConfig, schema, pluginName)
         }
```
harbu: This is already maybe checked by the `createPlugin` call above?

harbu (approved):

packages/broker/test/unit/configMigration.test.ts
teogeb: Fixed

PR #652, opened 2022-05-17
test: enable PushBuffer tests

teogeb:
Not flaky in local environment. How about in CI environment?


harbu:
Worth keeping around or close?

PR #655, opened 2022-05-17
NET-835: increase default webrtc ping-pong from 5sec to 30sec

harbu:


juslesan (approved):
LGTM

Noting here that there's a chance for this change to be detrimental for long lasting idle connections between nodes behind NATs. For some NAT types the 5 second interval may have acted an unintended tool that maintains connectivity via continuous UDP punching. For now, this is just a hypothesis.

PR #663, opened 2022-05-19
refactor(client): Fix eslint warnings

teogeb:
### Fix/disable eslint warnings

- Annotate method return types
- Write descriptions for @ts-expect-error annotations
- Disable some `explicit-module-boundary-types`warnings where parameter of type `any` is passed between modules
- Remove `prefer-destructuring` eslint rule

### Other refactoring

- Remove obsolete `StreamRegistryCached#getStreamValidationInfo` method
- Remove obsolete version handling in `Signer.test.ts`
- Convert `IteratorTest` to TypeScript
- Code deduplication: 
  - remove duplicate `until()` function
  - remove `sleep()` function, use `wait()` instead

### Future improvements

- `SignalAndListen` could have better generic types
- `Encryption.test.ts `defines `groupKeys` in an argument of `subscribe()` call: we should update the test or implement support for that parameter


packages/client/src/Rest.ts
```diff
@@ -72,7 +73,8 @@ export class Rest implements Context {
         )
     }
 
-    post<T extends object>(urlParts: UrlParts, body: any, options: FetchOptions) {
+    // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+    post<T extends object>(urlParts: UrlParts, body: any, options: FetchOptions): Promise<T> {
```
harbu:
```suggestion
    post<T extends object>(urlParts: UrlParts, body: object, options: FetchOptions): Promise<T> {
```
Maybe we could use https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html#object-type?

packages/client/src/Rest.ts
```diff
@@ -97,7 +99,8 @@ export class Rest implements Context {
         })
     }
 
-    put<T extends object>(urlParts: UrlParts, body: any, options: FetchOptions) {
+    // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+    put<T extends object>(urlParts: UrlParts, body: any, options: FetchOptions): Promise<T> {
```
harbu:
```suggestion
    put<T extends object>(urlParts: UrlParts, body: object, options: FetchOptions): Promise<T> {
```

packages/client/src/Stream.ts
```diff
@@ -265,7 +264,8 @@ class StreamrStream implements StreamMetadata {
         return this._streamRegistry.revokePermissions(this.id, ...assignments)
     }
 
-    [Symbol.for('nodejs.util.inspect.custom')](depth: number, options: any) {
+    // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+    [Symbol.for('nodejs.util.inspect.custom')](depth: number, options: any): string {
```
harbu:
```suggestion
    [Symbol.for('nodejs.util.inspect.custom')](depth: number, options: util.InspectOptions): string {
```

packages/client/src/StreamIDBuilder.ts
```diff
@@ -49,6 +49,7 @@ function parseRawDefinition(definition: StreamDefinition): [string, number | und
  * */
 export const formStreamDefinitionDescription = (definition: StreamDefinition): string => {
     const raw = parseRawDefinition(definition)
+    // eslint-disable-next-line prefer-destructuring
```
harbu: we should just disable this rule... whether destructuring or classic variable assignment is more readable is a context-dependent choice imo

packages/client/src/encryption/BrowserPersistentStore.ts
```diff
@@ -15,22 +15,22 @@ export default class BrowserPersistentStore implements PersistentStore<string, s
...
-    async has(key: string) {
+    async has(key: string): Promise<boolean> {
         const val = await this.get(key)
         return val == null
     }
 
-    async get(key: string) {
+    async get(key: string): Promise<any> {
         return get(key, this.store)
     }
```
harbu: Do we know if the type returned is always a string here?
```suggestion
    async get(key: string): Promise<string> {
        return get(key, this.store)
    }
```

packages/client/src/encryption/EncryptionUtil.ts
```diff
@@ -139,7 +139,7 @@ class EncryptionUtilBase {
      * message content and returns null.
      */
 
-    static decryptStreamMessage(streamMessage: StreamMessage, groupKey: GroupKey) {
+    static decryptStreamMessage(streamMessage: StreamMessage, groupKey: GroupKey): any {
```
harbu:
```suggestion
    static decryptStreamMessage(streamMessage: StreamMessage, groupKey: GroupKey): null | never {
```

packages/client/src/encryption/GroupKey.ts
```diff
@@ -4,6 +4,7 @@ import { uuid } from '../utils'
 import { inspect } from '../utils/log'
 
 class InvalidGroupKeyError extends ValidationError {
+    // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
     constructor(message: string, public groupKey?: any) {
```
harbu:
```suggestion
    constructor(message: string, public groupKey?: GroupKeyish) {
```

packages/client/src/publish/MessageChain.ts
```diff
@@ -17,7 +17,7 @@ export type MessageChainOptions = {
     msgChainId?: string
 }
 
-export function getCachedMessageChain(cacheConfig: CacheConfig) {
+export function getCachedMessageChain(cacheConfig: CacheConfig): any {
```
harbu:
```suggestion
export function getCachedMessageChain(cacheConfig: CacheConfig):
    ((streamPartId: StreamPartID, msgChainOptions: MessageChainOptions) => MessageChain) & {
    clear: () => void,
    clearMatching: (matchFn: (key: string) => boolean) => void
} {
```

packages/client/src/publish/PublishPipeline.ts
```diff
@@ -33,7 +33,8 @@ export class FailedToPublishError extends Error {
         }
     }
 
-    [Symbol.for('nodejs.util.inspect.custom')](depth: number, options: any) {
+    // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+    [Symbol.for('nodejs.util.inspect.custom')](depth: number, options: any): string {
```
harbu:
```suggestion
    [Symbol.for('nodejs.util.inspect.custom')](depth: number, options: util.InspectOptions): string {
```

packages/client/src/utils/Scaffold.ts
```diff
@@ -30,7 +30,7 @@ export function Scaffold(
     sequence: Step[] = [],
     _checkFn: (() => Promise<boolean>) | (() => boolean),
     { id = '', onError, onDone, onChange }: ScaffoldOptions = {}
-) {
+): any {
```
harbu: We should prolly think together what the type should be here.

packages/client/src/utils/Signal.ts
```diff
@@ -13,6 +13,10 @@ export enum TRIGGER_TYPE {
     PARALLEL = 'PARALLEL',
 }
 
+// TODO better type definition
+type ListenFn<T extends any[]> = (cb?: (arg: T[0]) => any) => any
+type SignalAndListen<T extends any[]> = ListenFn<T> & Methods<Signal<T>>
```
harbu, line 18: Let's think thru these together

packages/client/src/utils/index.ts
```diff
@@ -270,6 +272,7 @@ class DeferredWrapper<T> {
 
 export type Deferred<T> = ReturnType<DeferredWrapper<T>['wrap']>
 
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
 export function Defer<T>(executor: (...args: Parameters<Promise<T>['then']>) => void = noop) {
```
harbu: The inferred type for this is very very long, let's see if we can make smaller?

packages/client/src/utils/index.ts
```diff
@@ -357,7 +360,8 @@ export function Defer<T>(executor: (...args: Parameters<Promise<T>['then']>) =>
  */
 type LimitFn = ReturnType<typeof pLimit>
 
-export function LimitAsyncFnByKey<KeyType>(limit = 1) {
+// TODO better return type?
+export function LimitAsyncFnByKey<KeyType>(limit = 1): any {
```
harbu: Let's see

packages/client/test/test-utils/utils.ts
```diff
@@ -61,7 +62,8 @@ export async function fetchPrivateKeyWithGas(): Promise<string> {
 
 const TEST_REPEATS = (process.env.TEST_REPEATS) ? parseInt(process.env.TEST_REPEATS, 10) : 1
 
-export function describeRepeats(msg: any, fn: any, describeFn = describe) {
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export function describeRepeats(msg: any, fn: any, describeFn = describe): void {
```
```suggestion
export function describeRepeats(msg: string, fn: any, describeFn = describe): void {
```

packages/client/test/test-utils/utils.ts
```diff
@@ -78,7 +80,7 @@ describeRepeats.only = (msg: any, fn: any) => {
     describeRepeats(msg, fn, describe.only)
 }
 
-export async function collect(iterator: any, fn: MaybeAsync<(item: any) => void> = async () => {}) {
+export async function collect(iterator: AsyncIterable<any>, fn: MaybeAsync<(item: any) => void> = async () => {}): Promise<any[]> {
```
harbu: With types:
```ts
export async function collect<T>(
    iterator: AsyncIterable<StreamMessage<T>>,
    fn: MaybeAsync<(item: {
        msg: StreamMessage<T>,
        iterator: AsyncIterable<StreamMessage<T>>,
        received: T[]
    }) => void> = async () => {}
): Promise<T[]> {
    const received: T[] = []
    for await (const msg of iterator) {
        received.push(msg.getParsedContent())
        await fn({
            msg, iterator, received,
        })
    }

    return received
}
```

packages/client/test/test-utils/utils.ts
```diff
@@ -159,20 +157,23 @@ const getTestName = (module: NodeModule) => {
 const randomTestRunId = process.pid != null ? process.pid : crypto.randomBytes(4).toString('hex')
 
 // eslint-disable-next-line no-undef
```
harbu: eslint-disable directive can be removed

packages/client/test/test-utils/utils.ts
```diff
@@ -159,20 +157,23 @@ const getTestName = (module: NodeModule) => {
...
-export const createRelativeTestStreamId = (module: NodeModule, suffix?: string) => {
+export const createRelativeTestStreamId = (module: NodeModule, suffix?: string): any => {
```
```suggestion
export const createRelativeTestStreamId = (module: NodeModule, suffix?: string): string => {
```

packages/client/test/test-utils/utils.ts
```diff
@@ -159,20 +157,23 @@ const getTestName = (module: NodeModule) => {
...
     return counterId(`/test/${randomTestRunId}/${getTestName(module)}${(suffix !== undefined) ? '-' + suffix : ''}`, '-')
 }
 
 // eslint-disable-next-line no-undef
```
harbu: eslint-disable directive can be removed

packages/client/test/test-utils/utils.ts
```diff
@@ -159,20 +157,23 @@ const getTestName = (module: NodeModule) => {
...
         ...props
     })
     return stream
 }
 
-export const getCreateClient = (defaultOpts = {}, defaultParentContainer?: DependencyContainer) => {
+export const getCreateClient = (
+    defaultOpts = {}, 
+    defaultParentContainer?: DependencyContainer
+): (opts?: any, parentContainer?: DependencyContainer) => Promise<StreamrClient> => {
```
```suggestion
): (opts?: StreamrClientConfig, parentContainer?: DependencyContainer) => Promise<StreamrClient> => {
```

packages/client/test/test-utils/utils.ts
```diff
@@ -233,7 +234,8 @@ export class LeaksDetector {
 
     private counter = CounterId(this.id, { maxPrefixes: 1024 })
 
-    add(name: string, obj: any) {
+    // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+    add(name: string, obj: any): void {
```
harbu, line 243: Class should probably implement a generic type to capture values' type.

packages/client/test/test-utils/utils.ts
```diff
@@ -525,15 +530,15 @@ export function getPublishTestMessages(
...
+): (maxMessages?: number, opts?: PublishTestMessageOptions) => Promise<unknown[]> {
     const publishTestStreamMessages = getPublishTestStreamMessages(client, streamDefinition, defaultOpts)
     return async (maxMessages: number = 5, opts: PublishTestMessageOptions = {}) => {
         const streamMessages = await publishTestStreamMessages(maxMessages, opts)
         return streamMessages.map((s) => s.getParsedContent())
     }
 }
 
-export function getWaitForStorage(client: StreamrClient, defaultOpts = {}) {
+export function getWaitForStorage(client: StreamrClient, defaultOpts = {}): (lastPublished: StreamMessage, opts?: any) => Promise<void> {
```
```suggestion
export function getWaitForStorage(client: StreamrClient, defaultOpts = {}): (lastPublished: StreamMessage, opts?: {
        interval?: number
        timeout?: number
        count?: number
        messageMatchFn?: (msgTarget: StreamMessage, msgGot: StreamMessage) => boolean
    }) => Promise<void> {
```

packages/client/src/Rest.ts
teogeb: FIxed

packages/client/src/Rest.ts
teogeb: Fixed

packages/client/src/StreamIDBuilder.ts
teogeb: 👍

packages/client/src/encryption/EncryptionUtil.ts
teogeb: The documentation seems to be out of date:

```
Decrypts the serialized content of 'streamMessage' with 'groupKey'. If the resulting plaintext is the concatenation
of a new group key and a message content, sets the content of 'streamMessage' with that message content and returns the key. If the resulting plaintext is only a message content, sets the content of 'streamMessage' with that
message content and returns null.
```

We don't seem to need the return value at all. Maybe this could return `void | never` (that it, it throws an `Error` or returns nothing)?

packages/client/src/utils/Scaffold.ts
teogeb: 👍

packages/client/test/test-utils/utils.ts
teogeb: It is just an test utility

harbu (approved):

PR #675, opened 2022-05-24
client [NET-332]: MessageMetadata.msgChainId field

teogeb:
New `msgChainId` field in `MessageMetadata`. This allows `client.publish` and `stream.publish` methods to get an explicit `msgChainId` as a parameter.

- [x] Updated Changelog

### DIFF
--- a/packages/client/CHANGELOG.md
+++ b/packages/client/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Method signatures of `client.publish` and `stream.publish` have changed: optional `metadata` is given an object instead of positional arguments
+  - new metadata fields: `sequenceNumber` and `msgChainId`
 - Method `getStorageNodesOf()` renamed to `getStorageNodes()`
 - Method `getStoredStreamsOf()` renamed to `getStoredStreams()`
 - Method `isStreamStoredInStorageNode()` renamed to `isStoredStream()`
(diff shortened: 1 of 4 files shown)


packages/client/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Method signatures of `client.publish` and `stream.publish` have changed: optional `metadata` is given an object instead of positional arguments
+  - new metadata fields: `sequenceNumber` and `msgChainId`
```
harbu, line 19: The field `sequenceNumber` wasn't already included before?

harbu, line 19: The field `sequenceNumber` wasn't already included before?

packages/client/test/unit/Publisher.test.ts
```diff
@@ -79,7 +76,7 @@ describe('Publisher', () => {
                 sequenceNumber: 0,
                 streamId: STREAM_ID,
                 streamPartition: DEFAULT_PARTITION,
-                timestamp: TIMESTAMP
+                timestamp: expect.any(Number)
```
harbu: I would the make the assertion stricter, e.g. store `Date.now()` before calling `client.publish` and then use `expect.toBeBetween(storedDate, Date.now() + 1)`.

https://github.com/jest-community/jest-extended#tobebetweenstartdate-enddate

harbu (commented):
Should we test that when using multiple message chains, the `prevMessageRef` are set correctly and things don't get confused? Or is it obvious from the code such would not happen.

packages/client/CHANGELOG.md
teogeb: The feature was implemented in https://github.com/streamr-dev/network-monorepo/pull/674, but was not included in changelog of that PR

harbu (approved):

harbu:
> Should we test that when using multiple message chains, the `prevMessageRef` are set correctly and things don't get confused? Or is it obvious from the code such would not happen.

Reading the code it would seem that in fact it should work correctly.

PR #679, opened 2022-05-25
broker [NET-839]:  Publisher plugins set msgChainId

teogeb:
Each publisher plugin sets `msgChainId` when it publishes messages via `StreamrClient`. For stateful plugins (`mqtt`, `websocket`), the `msgChainId` is connection-specific. For `http` plugin the `msgChain` is generated at the startup of the plugin.

### Implementation

`mqtt` plugin uses Aedes's `client.id` field as `msgChainId`. The value to that field is read from an incoming connection  packet. The spec required that there can't be two clients connected with same clientId simultanously. Users of MQTT client libraries can possibly set a custom value to that field, and in that situation the `client.id` wouldn't be globally unique. If a custom value is not used, the value is typically very unique:
- spec: "3.1.3.1 Client Identifier" spec in http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html 
- implementation is Aedes: https://github.com/moscajs/aedes/blob/e3e50ded602e7d69bdf8ba15ee07ce3527e004f1/lib/handlers/connect.js#L78
- e.g. `clientId` option settable in https://github.com/mqttjs/MQTT.js client library, but defaults to a unique identifier

### Future improvements

- `http` plugin could use cookies to emulate "connections", and assign connection-specific `msgChainId` values that way (if needed in the future)

### Checklist

### DIFF
--- a/packages/broker/test/integration/multiple-publisher-plugins.test.ts
+++ b/packages/broker/test/integration/multiple-publisher-plugins.test.ts
@@ -5,7 +5,7 @@ import { StreamPermission } from 'streamr-client'
 import { Tracker } from '@streamr/network-tracker'
 import { Broker } from '../../src/broker'
 import { startBroker, createClient, createTestStream, fetchPrivateKeyWithGas, startTestTracker, Queue } from '../utils'
-import { fastPrivateKey, wait, waitForCondition, waitForEvent } from 'streamr-test-utils'
+import { fastPrivateKey, waitForCondition, waitForEvent } from 'streamr-test-utils'
 import { range, sample } from 'lodash'
 
 const MESSAGE_COUNT = 120
@@ -102,7 +102,7 @@ describe('multiple publisher plugins', () => {
         const client = await createClient(tracker, privateKey)
         const stream = await createTestStream(client, module)
         streamId = stream.id
-        stream.grantPermissions({
+        await stream.grantPermissions({
             permissions: [StreamPermission.SUBSCRIBE],
             public: true
         })
@@ -110,7 +110,7 @@ describe('multiple publisher plugins', () => {
     })
 
     afterAll(async () => {
-        await tracker.stop()
+        await tracker?.stop()
     })
 
     beforeEach(async () => {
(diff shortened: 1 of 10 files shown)


harbu:
> mqtt plugin uses client.id field as msgChainId. Users of mqtt client libraries can possibly set a value to that field, but typically it is very unique:

Just pointing out that there is a very very edge case situation that would occur if multiple publishing MQTT clients join with the same _clientId_. Then they would publish effectively to the same message chain if they publish to the same stream part. However, it could be argued that this would be correct behaviour in a sense since they are joining with the same id!

packages/broker/test/integration/multiple-publisher-plugins.test.ts
```diff
@@ -0,0 +1,189 @@
...
+        streamId = stream.id
+        stream.grantPermissions({
+            permissions: [StreamPermission.SUBSCRIBE],
+            public: true
+        })
+        await client.destroy()
+    })
+
+    afterAll(async () => {
+        await tracker.stop()
```
harbu:
```suggestion
        await tracker?.stop()
```

packages/broker/test/integration/multiple-publisher-plugins.test.ts
```diff
@@ -0,0 +1,189 @@
...
+
+    it('subscribe by StreamrClient', async () => {
+
+        const receivedMessages: Queue<object> = new Queue()
+        const subscriber = await createClient(tracker, fastPrivateKey())
+        await subscriber.subscribe(streamId, (message: object) => {
+            receivedMessages.push(message)
+        })
+        // TODO: better waiting
+        await wait(3000)
```
harbu: Let's figure out a better waiting strategy.

We can either use `NetworkNodeStub#getNeighborsForStreamPart` or we could consult tracker on what nodes are in which stream partition.

harbu (changes requested):

harbu (approved):

PR #680, opened 2022-05-25
fix: crashing proxy reconnect

juslesan:
Fixes a fatal issue with proxy reconnections caused by a race condition.


harbu (approved):

harbu:
NVM, wrong PR my comment here :D

PR #682, opened 2022-05-25
client: re-order messages in subscribe pipeline

teogeb:
The original motivation of this ticket was to find out whether the 2nd message ordering phase in `SubscribePipeline` is necessary or not. It was unclear just from reading code whether it can be removed or not (however, strong guess was that it could).

The key question became whether `OrderingUtil` provides messages always in order, even in the presence of gaps (fillable and unfillable), duplicate messages, and messages arriving in wrong order in real-time.

To test this key question an integration test was written `test/integration/OrderingUtil.test.ts` in _protocol_ package. The finding was that indeed, `OrderingUtil` takes care of the order. Therefore the test seems to "prove" that re-ordering is not necessary and can be removed from `PublishPipeline`.

A secondary finding was that the gap filling didn't attempt to fill all gaps in situations where previous gaps had failed `maxGapRequests`. This was probably a shortcut that was added back in the day to avoid `OrderingUtil` getting "stuck". It is @harbu opinion that the `OrderingUtil` should always attempt to gap fill, but the default parameters should maybe be lowered to avoid the experience of `OrderingUtil` getting "stuck". (It doesn't actually get stuck, but in the presence of multiple gaps, it tries so aggressively to fill the gaps that it will take very long per gap, therefore seemingly becoming "stuck".)

Regarding the secondary finding, gaps not being filled appropriately, it will be dealt with in a separate ticket. Meanwhile the test will be marked skipped.


packages/protocol/test/integration/OrderingUtil.test.ts
```diff
@@ -0,0 +1,169 @@
...
+    publisherId: string
+    timestamp: number
+    delivery: Delivery
+}
+
+function duplicateElements<T>(arr: readonly T[], numOfDuplicates: number): T[] {
+    const newArr = Array.from(arr)
+    for (let i = 0; i < numOfDuplicates; ++i) {
+        newArr.push(arr[i % arr.length])
+    }
```
harbu:
```suggestion
    for (let i = 0; i < numOfDuplicates; ++i) {
        newArr.push(arr[Math.floor(Math.random() * arr.length)])
    }
```

packages/protocol/test/integration/OrderingUtil.test.ts
```diff
@@ -0,0 +1,169 @@
...
+        const totalUnfillableGaps = PUBLISHER_IDS.reduce((sum, publisherId) => (
+            sum + calculateNumberOfUnfillableGaps(groundTruthMessages[publisherId])
+        ), 0)
+
+        const inOrderHandler = (msg: StreamMessage) => {
+            actual[msg.getPublisherId()].push(msg.getTimestamp())
+        }
+
+        const gapHandler = async (from: MessageRef, to: MessageRef, publisherId: string) => {
+            //console.info(`gapHandler: ${from.serialize()} -> ${to.serialize()} (${publisherId})`)
```
harbu:
```suggestion
```

PR #687, opened 2022-05-31
broker [NET-842]: Publish to non-existent stream

teogeb:
Improved error handling in `websocket` and `mqtt` plugins. Publishing to a non-existent stream caused Broker to throw an unhandled error. 

We close the `websocket` connection if an error occurs. For `mqtt` we only log a warning to Broker's console.

### Open questions

- Should we close the connection when a publish error occurs? Should `websocket` and `mqtt` plugin behave similarly?

### DIFF
--- a/packages/broker/src/plugins/mqtt/Bridge.ts
+++ b/packages/broker/src/plugins/mqtt/Bridge.ts
@@ -52,7 +52,7 @@ export class Bridge implements MqttServerListener {
             })
             this.publishMessageChains.add(createMessageChainKey(publishedMessage))
         } catch (err: any) {
-            logger.warn('Unable to publish')
+            logger.warn('Unable to publish, reason: %s', err)
         }
     }
 
--- a/packages/broker/src/plugins/mqtt/Bridge.ts
+++ b/packages/broker/src/plugins/mqtt/Bridge.ts
@@ -45,11 +45,15 @@ export class Bridge implements MqttServerListener {
             return
         }
         const { content, metadata } = message
-        const publishedMessage = await this.streamrClient.publish(this.getStreamId(topic), content, {
-            timestamp: metadata.timestamp,
-            msgChainId: clientId
-        })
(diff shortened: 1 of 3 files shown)


packages/broker/src/plugins/mqtt/Bridge.ts
```diff
@@ -45,11 +45,15 @@ export class Bridge implements MqttServerListener {
...
-        })
-        this.publishMessageChains.add(createMessageChainKey(publishedMessage))
+        try {
+            const publishedMessage = await this.streamrClient.publish(this.getStreamId(topic), content, {
+                timestamp: metadata.timestamp,
+                msgChainId: clientId
+            })
+            this.publishMessageChains.add(createMessageChainKey(publishedMessage))
+        } catch (err: any) {
+            logger.warn('Unable to publish')
```
harbu: Should we provide the reason in the error message?
```suggestion
            logger.warn('Unable to publish, reason: %s', err)
```

packages/broker/test/integration/createMessagingPluginTest.ts
```diff
@@ -93,24 +94,39 @@ export const createMessagingPluginTest = <T>(
...
+        it('publish to non-existent stream', async () => {
+            const streamId = 'non-existent-stream'
+            pluginClient = await api.createClient('publish', streamId, MOCK_API_KEY)
+            await api.publish(MOCK_MESSAGE, streamId, pluginClient)
+            // Wait for some time so that the plugin can handle the publish request (api.publish() 
+            // resolves immediately e.g. in websocket plugin test as websocket.send() doesn't 
+            // return a promise). 
+            // If the api.publish call causes the plugin to throw an unhandled error, jest catches 
+            // the error and this test fails. There should be "Unable to publish" warning in the 
+            // Broker log, but this test can't verify it.
```
harbu, line 128: :+1:

harbu (approved):

PR #690, opened 2022-06-02
refactor tracker message relay

harbu:
Refactor tracker message relay code in preparation for adding relaying of receipt queries and results in https://github.com/streamr-dev/network-monorepo/pull/639.

Goal is to clearly separate message relaying & WebRTC signalling as their own concepts in the code base whilst making non-functional changes to WebRTC signalling keeping the behavior intact. 

## Checklist


juslesan (approved):
LGTM

PR #692, opened 2022-06-02
Publish Utils

timoxley:
Tools for bumping versions. 
Needs proper `bin` scripts, but for now you can run it like so:

`npx ts-node --dir packages/monorepo-tools src/bumpVersions.ts`

Currently broken after merging latest `main` due to circular dependency between `network` and `network-tracker`. Should probably resolve that first.


harbu:
Will go with unified versioning, closing for now instead we change opinion.

PR #695, opened 2022-06-03
fix(network): gracefully handle message validation error

harbu:
## Checklist

- [ ] Updated Changelog
- [ ] Updated Documentation


packages/network/src/protocol/utils.ts
```diff
@@ -1,11 +1,7 @@
-export function decode<M, R>(serializedMessage: M, deserializeFn: (serializedMessage: M) => R): R | null | never {
+export function decode<R>(serializedMessage: string, deserializeFn: (serializedMessage: string) => R): R | null | never {
```
teogeb:
```suggestion
export function decode<R>(serializedMessage: string, deserializeFn: (serializedMessage: string) => R): R | null {
```

teogeb (approved):

PR #704, opened 2022-06-08
refactor(client): Remove `stop()` methods

teogeb:
- remove `stop()` methods which were not called by any component:
  - `KeyExchangeStream`
  - `GroupKeyStoreFactory`
  - `SubscriberKeyExchange`
  - `PublisherKeyExchange`
- remove obsolete `isStopped` field from classes (the value was never read):
  - `MessageCreator`
  - `Publisher`
- remove obsolete `Stoppable` interface
- remove methods from `Publisher` which started/stopped `PublisherKeyExchange`:
  - these methods were not part of public API and not called by other components
  - if we want to add public API methods to start/stop `PublisherKeyExchange`, we could use `PublisherKeyExchange#start` and `PublisherKeyExchange#stop` directly
  
 ### Future improvements
 
 - Do we want to start `PublisherKeyExchange` automatically if there are GroupKeys to be served to potential subscribers? Or do we want to have a public API method to start the the service? (NET-855)
 - If we need, we could add a new helper class (e.g. `TaskManager`), which could stop all pending promises when `client.destroy()` is called. (NET-856)

### DIFF
--- a/packages/client/src/encryption/GroupKeyStoreFactory.ts
+++ b/packages/client/src/encryption/GroupKeyStoreFactory.ts
@@ -88,11 +88,4 @@ export class GroupKeyStoreFactory implements Context {
         const store = await this.getStore(streamId)
         return store.rekey(newKey)
     }
-
-    async stop(): Promise<void> {
-        this.getStore.clear()
-        const { cleanupFns } = this
-        this.cleanupFns = []
-        await Promise.allSettled(cleanupFns)
-    }
 }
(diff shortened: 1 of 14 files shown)


packages/client/src/encryption/GroupKeyStoreFactory.ts
```diff
@@ -88,11 +88,4 @@ export class GroupKeyStoreFactory implements Context {
         const store = await this.getStore(streamId)
         return store.rekey(newKey)
     }
-
-    async stop(): Promise<void> {
-        this.getStore.clear()
-        const { cleanupFns } = this
-        this.cleanupFns = []
-        await Promise.allSettled(cleanupFns)
-    }
```
harbu: Keep method but make sure it actually get called (e.g. in client.destroy())

harbu (changes requested):

packages/client/src/encryption/GroupKeyStoreFactory.ts
teogeb: Fixed

harbu (approved):

PR #706, opened 2022-06-08
refactor(client): RsaKeyPair

teogeb:
**TODO broswer test fails, maybe needs longer key (and must by dividable by 8 bits, i.e. min 592)**

Extracted `RsaKeyPair` class from `EncryptionUtil`:
- `EncryptionUtil` contains static utility methods for `RSA` and `AES` encryption/decryption
- `RsaKeyPair` is a data class for a RSA key pair (a public key and a private key)

Modified `index-exports.ts` to export only the custom `Error` classes.  The `EncryptionUtil` is no longer exported as it should not be used by end-users.

Divided `Encryption.test.ts` to three tests:
- `EncryptionUtil.test.ts`
- `RsaKeyPair.test.ts`
- `GroupKey.test.ts`

Earlier `Encryption.test.ts` run the tests also in an emulated browser environment. It is not clear whether that emulation worked correctly. Now we run all tests only in the normal Node environment. The browser tests (`npm run test-browser-realtime` and `npm run test-browser-resend`) give us basic test coverage for the encryption functionality as encrypted streams are used in some browser tests. If we need more coverage, we could maybe run encryption-related unit tests in a real browser environment instead of a simulated one.

### DIFF
--- a/packages/client/test/unit/GroupKey.test.ts
+++ b/packages/client/test/unit/GroupKey.test.ts
@@ -1,0 +1,25 @@
+import crypto from 'crypto'
+import { GroupKey } from '../../src/encryption/GroupKey'
+
+describe('GroupKey.validate', () => {
+
+    describe('validate', () => {
+        it('throws if key is the wrong size', () => {
+            expect(() => {
+                GroupKey.validate(GroupKey.from(['test', crypto.randomBytes(16)]))
+            }).toThrow('size')
+        })
+    
+        it('throws if key is not a buffer', () => {
+            expect(() => {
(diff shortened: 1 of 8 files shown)


packages/client/test/unit/GroupKey.test.ts
```diff
@@ -0,0 +1,25 @@
+import crypto from 'crypto'
+import { GroupKey } from '../../src/encryption/GroupKey'
+
+describe('GroupKey.validate', () => {
```
harbu:
```suggestion
describe('GroupKey', () => {
```

harbu (approved):

PR #708, opened 2022-06-09
refactor(protocol): KeyExchangeStream partition 

teogeb:
Changed `StreamIDUtils.formKeyExchangeStreamID` to return `StreamPartID` instead of `StreamPart`. 

The method is used by `StreamrClient`. Therefore when the client publishes/subscribes to a key exchange stream, it uses `0` partition explicitly . The functionality doesn't change as publish and subscribe currently use `0` partition by default if no partition is specified. But after this PR, key exchange operations don't need to rely on that default behavior.

Also extracted key exchange stream id utility methods to a separate `KeyExchangeStreamIDUtils` utils. As one of the methods doesn't operate on `StreamID` but `StreamPartID`, it wouldn't be optimal to leave key exchange methods to 
 be part `StreamIDUtils`. This change also simplifies the basic usage of `StreamIDUtils`.

### DIFF
--- a/packages/protocol/src/utils/KeyExchangeStreamID.ts
+++ b/packages/protocol/src/utils/KeyExchangeStreamID.ts
@@ -4,8 +4,8 @@ import { EthereumAddress } from './types'
 
 export class KeyExchangeStreamIDUtils {
 
-    static STREAM_ID_PREFIX = 'SYSTEM/keyexchange/'
-    static PARTITION = 0
+    static readonly STREAM_ID_PREFIX = 'SYSTEM/keyexchange/'
+    static readonly PARTITION = 0
 
     static formStreamPartID(recipient: EthereumAddress): StreamPartID {
         const streamId = (KeyExchangeStreamIDUtils.STREAM_ID_PREFIX + recipient.toLowerCase()) as StreamID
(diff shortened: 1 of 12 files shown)


packages/protocol/src/utils/KeyExchangeStreamID.ts
```diff
@@ -0,0 +1,25 @@
+import { StreamID } from './StreamID'
+import { StreamPartID, toStreamPartID } from './StreamPartID'
+import { EthereumAddress } from './types'
+
+export class KeyExchangeStreamIDUtils {
+
+    static STREAM_ID_PREFIX = 'SYSTEM/keyexchange/'
+    static PARTITION = 0
```
harbu:
```suggestion
    readonly static STREAM_ID_PREFIX = 'SYSTEM/keyexchange/'
    readonly static PARTITION = 0
```

packages/protocol/src/utils/KeyExchangeStreamID.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    static formStreamPartID(recipient: EthereumAddress): StreamPartID {
+        const streamId = (KeyExchangeStreamIDUtils.STREAM_ID_PREFIX + recipient.toLowerCase()) as StreamID
+        return toStreamPartID(streamId, KeyExchangeStreamIDUtils.PARTITION)
+    }
+    
+    static isKeyExchangeStream(streamId: StreamID | string): boolean {
+        return streamId.startsWith(KeyExchangeStreamIDUtils.STREAM_ID_PREFIX)
```
harbu, line 16:
```suggestion
        return streamId.startsWith(KeyExchangeStreamIDUtils.STREAM_ID_PREFIX) // TODO: stricter validation, e.g. address at end?
```

harbu (approved):

PR #711, opened 2022-06-10
fix(client): Group key error response read

teogeb:
Fix reading of `GroupKeyErrorResponse`. 

A call to `subscriberKeyExchange.getGroupKey()` wasn't able to return the expected value (`undefined`), because:
- some response parsing was implemented only for the success response (`GroupKeyResponse`, not `GroupKeyErrorResponse`)
- `SubscriptionPipeline` threw an error on `GroupKeyResponse`
  - it should throw an error if we get `GroupKeyError` response during normal subscription, i.e. we are unable to fetch a group key for which this subscription saw a `groupKeyId` (**TODO**: let's have a test to prove that this happens)
  - but for key exchange stream, the response should be delivered to the subscribe (`KeyExchangeStream#request`)

Also added some test cases for `SubscriberKeyExchange`.

**TODO** it is maybe better that `subscriberKeyExchange.getGroupKey()`  throws on error. See `Decrypt.ts:52`


harbu (approved):

teogeb:
Closing this PR because we'll keep behavior that `subscriberKeyExchange.getGroupKey()` throws on error. The relevant tests added here: https://github.com/streamr-dev/network-monorepo/pull/712

The modifications made to `KeyExchangeStream.ts` and `SubscriberKeyExchange.ts` are maybe needed in the future if we refactor `KeyExchangeStream` to bypass `SubscribePipeline` (and use a network node directly instead). But currently we don't need to change those files are the lines are not reached in group key error use case.

PR #713, opened 2022-06-13
refactor(client): Remove message container

teogeb:
Add new options for `MessageCreator` so that it can create messages with custom `messageType` or `encryptionType`.  `KeyExchangeStream` and `PublisherKeyExchangeStream` use those options to create group key messages. The options are marked as `@internal`, because we don't need to expose this functionality to end-users.

Earlier `GroupKeyMessages` were created via `StreamMessageContainer`. That functionality is now removed as it is no longer needed.


packages/client/src/publish/MessageCreator.ts
```diff
@@ -88,17 +92,12 @@ export class MessageCreator implements IMessageCreator {
...
-                // e.g. transparent handling for StreamMessage<SomeClass> where SomeClass implements toStreamMessage & {de}serialization methods
-                ? (content.toStreamMessage(messageId, prevMsgRef || null)) as StreamMessage<any>
-                : new StreamMessage({
-                    messageId,
-                    prevMsgRef,
-                    content,
-                    ...opts
-                })
-
+            const streamMessage =new StreamMessage({
```
harbu:
```suggestion
            const streamMessage = new StreamMessage({
```

packages/client/src/encryption/PublisherKeyExchange.ts
```diff
@@ -106,7 +108,7 @@ export class PublisherKeyExchange implements Context {
                     errorMessage: error.message,
                     groupKeyIds
                 })
-                return await this.keyExchangeStream.response(subscriberId, response)
+                return this.keyExchangeStream.response(subscriberId, response)
```
harbu: I think `await` is necessary for the `catch`-clause to be called when the promise rejects.

harbu (approved):

PR #714, opened 2022-06-13
test(client): Anonymous publisher

teogeb:
Added a test case for anonymous publishing. 

Removed obsolete `MessageCreatorAnonymous` class and  `IMessageCreator` interface.

### DIFF
--- a/packages/client/test/integration/anonymous-client.test.ts
+++ b/packages/client/test/integration/anonymous-client.test.ts
@@ -1,0 +1,21 @@
+import { fastPrivateKey } from 'streamr-test-utils'
+import { StreamrClient } from '../../src/StreamrClient'
+import { createFakeContainer, DEFAULT_CLIENT_OPTIONS } from '../test-utils/fake/fakeEnvironment'
+import { createTestStream } from '../test-utils/utils'
+
+describe('anonymous client', () => {
+
+    it('fails to publish', async () => {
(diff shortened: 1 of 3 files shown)


packages/client/test/integration/anonymous-client.test.ts
```diff
@@ -0,0 +1,21 @@
...
+describe('anonymous client', () => {
+
+    it('fails to publish', async () => {
+        const dependencyContainer = createFakeContainer(undefined)
+        const owner = new StreamrClient({
+            auth: {
+                privateKey: fastPrivateKey()
+            },
+            ...DEFAULT_CLIENT_OPTIONS
+        }, dependencyContainer)
```
harbu, line 15: This will work fine, but what about the case of not passing any `privateKey` to client constructor?

teogeb: Then there would not be gas and therefore we couldn't create a new stream. Note that this `owner` is for test setup, the actual publishing is done by an anonymous client (line 18)

`const publisher = new StreamrClient(DEFAULT_CLIENT_OPTIONS, dependencyContainer)`

harbu: nvm read the code wrong...

harbu (approved):

PR #718, opened 2022-06-15
refactor(client): Divide utilities

teogeb:
Divide utility functions from utils/index.ts to separate fields:
- cache-related utilities to `caches.ts`
- promises-related utilities to `promises.ts`
- Defer class to `Defer.ts`
- renamed the file from `utils/index.ts` to `utils/utils.ts` (easier identification and navigation in IDE).

Other small refactorings:
- Some classes which were exported to components via `utils.ts` are now imported directly (`uuid`, `AggregatedError` and `Scaffold`).
- Removed obsolete `waitFor` function which was not used by any component. 

This PR contains no code changes.

### Future improvements

- a utility function `until` could be replacds with `waitForCondition` function from `streamr-test-utils` (when that is no longer a test-only dependency)


harbu (approved):
good stuff

PR #719, opened 2022-06-15
Signature handling with ethers library (poc)

teogeb:
Maybe we could use `@ethersproject/wallet` library directly instead of `SigningUtils`.

The signature is a standard Ethereum signature (https://github.com/streamr-dev/streamr-specs/blob/master/PROTOCOL.md), which we could easily produce by calling `Wallet#signMessage`:
```
const privateKey = '0x348ce564d427a3311b6536bbcff9390d69395b06ed6c486954e971d960fe8709'
const custom = SigningUtil.sign(payload, privateKey.substring(2).toString())  // 0x084b3a...
const ethers = await (new Wallet(privateKey).signMessage(payload))  // 0x084b3a...
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25984,6 +25984,7 @@
             "license": "Apache-2.0",
             "dependencies": {
                 "@ethersproject/hdnode": "^5.4.0",
+                "@ethersproject/wallet": "^5.5.0",
                 "debug": "^4.3.2",
                 "heap": "^0.2.6",
                 "secp256k1": "^4.0.2",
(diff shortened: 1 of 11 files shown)


harbu:
Check performance with removed benchmark test validate.test.ts https://github.com/streamr-dev/network-monorepo/pull/390

teogeb:
Added the test to this PR and documented the `SigningUtils`: https://github.com/streamr-dev/network-monorepo/pull/724

Closing this PR because we should keep using the custom implementation for performance reasons.

PR #724, opened 2022-06-16
test(protocol): Signing benchmark

teogeb:
Added a benchmark test which compares the performance of `SigningUtils` against ethers.js implementation. Also documented `SigningUtils`.

The test is based on these PRs:
- https://github.com/streamr-dev/streamr-client-protocol-js/pull/35
- https://github.com/streamr-dev/network-monorepo/pull/390

### DIFF
--- a/packages/protocol/src/utils/SigningUtil.ts
+++ b/packages/protocol/src/utils/SigningUtil.ts
@@ -24,6 +24,17 @@ function recoverPublicKey(signatureBuffer: Buffer, payloadBuffer: Buffer) {
     )
 }
 
+/**
+ * Creates and verifies Ethereum signatures. This is a faster implementation 
+ * of standard Ethereum signing, and therefore compatible with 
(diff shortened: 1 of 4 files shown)


packages/protocol/src/utils/SigningUtil.ts
```diff
@@ -24,6 +24,17 @@ function recoverPublicKey(signatureBuffer: Buffer, payloadBuffer: Buffer) {
     )
 }
 
+/**
+ * Creates and verifies Ethereum signatures. This is a faster implementation 
+ * of standard Ethereum signing, and therefore compatible with 
+ * e.g. ether.js's verifyMessage and signMessage functions.
```
harbu:
```suggestion
 * Creates and verifies standard Ethereum signatures. This is a faster implementation 
 * than found in ether.js library. It is compatible with 
 * e.g. ether.js's verifyMessage and signMessage functions.
```

harbu (approved):
good stuff

PR #726, opened 2022-06-16
refactor(client): Simplify publisher

teogeb:
Refactored `Publisher` class. 
- moved method `publishFromMetadata` to `test-utils/utils.ts` as it is only used in tests
- removed methods:
  - `publishFrom` and `collect`: not used by any component 
  - `collectMessages`: used in tests, but can use `collect` from `GeneratorUtils.ts` instead


harbu (approved):
👍

PR #727, opened 2022-06-16
test(client): Divide test utils

teogeb:
Moved `fetchPrivateKeyWithGas` from `client` to `test-utils`. Changed also Broker to use that method. Broker had almost identical method, which had longer timeouts (not sure if those are needed).

Extracted classes from `test-utils/utils.ts` to separate files:
- `LeaksDetector`
- `Multimap`

Extracted publish-related test utilities to a separate file, and `jest`-related test utilities to another.

Removed obsolete methods:
- `getTestSetTimeout`
- `getRandomClient`

### Future improvements

We have quite many methods which publish test messages to a test stream, see `test-utils/publish.ts`. Maybe we could unify the publishing logic in tests and remove some of the methods. (And then we don't maybe need a separate file for those methods anymore.)


harbu (approved):
lgtm!

PR #730, opened 2022-06-17
NET-603: @streamr/utils package

harbu:
### Utility package @streamr/utils 
Introduce new package `@streamr/utils`, which is a collection of shared common utilities used by sub-packages of monorepo (and perhaps other projects as well).

Aim is to collect useful utilities (functions, classes etc.) while keeping 3rd party library dependencies as light as possible.

### Utilities migrated
- Logging facilities `Logger`
- Method `withTimeout` to attach a promise to a timeout


harbu:
Continuing this on #746

PR #732, opened 2022-06-17
feat(client): [NET-865]: gap filling strategy option

harbu:
## Summary

Add the ability to control the gap fill strategy with config option `lighterGapFill`.

By default this is `true` to be in line with current behavior. However, if more pedantic gap filling is wanted by the user, it is now possible to enable it by setting `lighterGapFill` option to`false`.


packages/client/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add config option `lighterGapFill` (true by default) to control gap filling strategy
```
teogeb: Would it make sense if we'd configure this by enum: `gapFillStrategy: 'full'` or `gapFillStrategy: 'light'`? It would be easier to communicate if the strategies have exact names. 

The boolean logic is quite easy, but not as straightforward as it could be: i.e. if we want full gap fill, we need to set lighter to `false`.

packages/client/src/subscribe/ordering/OrderedMsgChain.ts
```diff
@@ -182,7 +183,8 @@ class OrderedMsgChain extends MsgChainEmitter {
         gapHandler: GapHandler,
         propagationTimeout = DEFAULT_PROPAGATION_TIMEOUT,
         resendTimeout = DEFAULT_RESEND_TIMEOUT,
-        maxGapRequests = MAX_GAP_REQUESTS
+        maxGapRequests = MAX_GAP_REQUESTS,
+        lighterGapFill = true
```
teogeb: All client defaults are in `config.schema.json`, so there is no need to duplicate the info here.

Should also remove other defaults, e.g.  DEFAULT_PROPAGATION_TIMEOUT (it is the `gapFillTimeout` config option) in some PR.

teogeb: Other defaults removed here: https://github.com/streamr-dev/network/pull/1333

harbu:
Addressed PR comments @teogeb, please re-review

packages/client/test/unit/OrderedMsgChain.test.ts
```diff
@@ -93,7 +93,7 @@ describe('OrderedMsgChain', () => {
             received.push(msg)
```
harbu, line 95: Change all these to `true`

packages/client/test/unit/OrderingUtil.test.ts
```diff
@@ -44,7 +44,7 @@ describe('OrderingUtil', () => {
             assert.deepStrictEqual(streamMessage.serialize(), msg.serialize())
             done()
         }
-        util = new OrderingUtil(handler, () => {}, DEFAULT_GAP_FILL_TIMEOUT, DEFAULT_RETRY_RESEND_AFTER, DEFAULT_MAX_GAP_REQUESTS)
+        util = new OrderingUtil(handler, () => {}, DEFAULT_GAP_FILL_TIMEOUT, DEFAULT_RETRY_RESEND_AFTER, DEFAULT_MAX_GAP_REQUESTS, false)
```
harbu: Change all these to `true`

packages/client/test/unit/OrderingUtil2.test.ts
```diff
@@ -103,8 +103,8 @@ function calculateNumberOfUnfillableGaps(messageInfosInOrder: MessageInfo[]): nu
     return gaps
 }
 
-describe.skip(OrderingUtil, () => {
-    it('randomized "worst-case" scenario with unavailable messages and gap fill needs', async () => {
+describe(OrderingUtil, () => {
+    it('randomized "worst-case" scenario with unavailable messages and gap fill needs (lightGapFill disabled)', async () => {
```
harbu:
```suggestion
    it('randomized "worst-case" scenario with unavailable messages and gap fill needs (using full gap filling strategy)', async () => {
```

packages/client/src/subscribe/ordering/OrderedMsgChain.ts
```diff
@@ -320,6 +323,11 @@ export class OrderedMsgChain {
                 }
             }
 
+            if (!this.dropSubsequentGapsOnFailedGapFill) {
```
teogeb, line 325: The `pop` method is called also when we receive data from realtime / resend? In that case we shouldn't set the counter to zero?

teogeb:
Replaced by https://github.com/streamr-dev/network/pull/1628

PR #741, opened 2022-06-23
test(client): Implement tests without monkey patching

teogeb:
Many tests interact with internal `PublishPipeline`. In this PR we setup test scenarios so that tests don't rely on internal implementation of `PublishPipeline`, but use public methods of `StreamrClient`/`Publisher`/etc. instead.

Refactored and moved tests cases to `unit/Publisher.test.ts`:
- all `integration/Partitions.test.ts`
- some `integration/Encryption.test.ts`

### TODO

- `StreamrClient.test.ts`  (uses private `PublishPipeline`)
- `EncryptionTest.test.ts`  (uses private `PublishPipeline`)
- `GapFill.test.ts`(uses explicit `monkeypatchMessageHandler`)
- other usages  of `getPublishTestStreamMessages`, or modify `getPublishTestStreamMessages` not to use internal `PublishPipeline`
- remove obsolete test utils from `test-utils/publish.ts` (if there are some methods which we don't use anymore)


teogeb:
Merged the changes to https://github.com/streamr-dev/network-monorepo/pull/762, closing this PR

PR #742, opened 2022-06-23
fix(client): Key exchange stream permissions

teogeb:
Updated `StreamRegistry` to handle queries about key exchange streams. 

There was already a handler `getStream`, and in this PR we added support for `isStreamPublisher`, `isStreamSubscriber`, `getStreamPublishers` and `getStreamSubscribers`.

- all users can publish to all key exchange streams (to send a `GroupKeyRequest` or a `GroupKeyResponse`)
- only recipient of a key exchange stream can publish to a key exchange stream (to receive a `GroupKeyRequest` or a `GroupKeyResponse`) 
- `getStreamPublishers` and `getStreamSubscribers` throw an error because it is impossible to return a valid response for `getStreamSubscribers` query (it should return all possible user addresses)
  - `getStreamPublishers` could return the only address, but as that kind of query is not relevant to make. Therefore it is possible to throw an error also in this case. Maybe this unified behavior between the two methods is better?


teogeb:
Closing this PR as it makes sense to implement https://github.com/streamr-dev/network-monorepo/pull/743 instead.

PR #743, opened 2022-06-27
refactor(client): Key exchange stream support in `StreamRegistry`

teogeb:
The `StreamRegistry` doesn't support key exchange streams. Removed an exception where we returned a simulated stream if `getStream` was used to query a metadata of a key exchange stream. 

In https://github.com/streamr-dev/network-monorepo/pull/708 we started to use explicit stream partition for key exchange streams. Therefore we don't need the exception as we don't call `StreamPartitioner` for key exchange streams anymore: https://github.com/streamr-dev/network-monorepo/blob/02a38664d944fac13e0d9a08aeb52c8ebb82e72d/packages/client/src/publish/StreamPartitioner.ts#L31

Added also a new test case to `PublisherKeyExchange.test.ts`.

### DIFF
--- a/packages/client/src/registry/StreamRegistry.ts
+++ b/packages/client/src/registry/StreamRegistry.ts
@@ -16,8 +16,7 @@ import {
     StreamID,
     EthereumAddress,
     StreamIDUtils, 
-    toStreamID,
-    KeyExchangeStreamIDUtils
+    toStreamID
 } from 'streamr-client-protocol'
 import { StreamIDBuilder } from '../StreamIDBuilder'
 import { omit } from 'lodash'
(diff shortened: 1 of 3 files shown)


packages/client/src/registry/StreamRegistry.ts
```diff
@@ -42,6 +41,13 @@ import {
 import { StreamRegistryCached } from './StreamRegistryCached'
 import { Authentication, AuthenticationInjectionToken } from '../Authentication'
 
+/* 
+ * On-chain registry of stream metadata and permissions.
+ *
+ * Does not support system streams (the key exchange stream) or legacy streams (streamIds 
+ * like '7wa7APtlTq6EC5iTCBy6dw')
```
harbu:
```suggestion
 * Does not support system streams (the key exchange stream)
```

harbu (approved):

PR #746, opened 2022-06-28
NET-603: extract @streamr/utils package

harbu:
### New utility package `@streamr/utils` 
Introduce new package `@streamr/utils`, which is a collection of shared common utilities used by sub-packages of monorepo (and perhaps other projects as well).

Aim is to collect useful utilities (functions, classes etc.) while keeping 3rd party library dependencies as light as possible.

### Utilities migrated
- Logging facilities `Logger`
    - LoggerBrowser has been removed, it is merged with `Logger` so there is only one class shared between Node.js and browser environments
- Function `withTimeout` to attach a promise to a timeout


packages/utils/test/Logger.test.ts
```diff
@@ -1,5 +1,5 @@
 import path from 'path'
-import { Logger } from "../../src/helpers/Logger"
+import { LoggerNode as Logger } from '../src/LoggerNode'
```
teogeb:
```suggestion
import { Logger } from '../src/Logger'
```

teogeb (approved):
LGTM 👍 One import to fix. Could mention in the PR description that node/browser-specific versions of Logger are now merged to one.

.github/workflows/utils.yml
```diff
@@ -0,0 +1,45 @@
...
+  push:
+    tags:
+      - '*'
+    branches:
+      - main
+  pull_request:
+    branches:
+      - '*'
+concurrency:
+  group: test-utils-build-${{ github.head_ref }}
```
teogeb: Group maybe without `test` prefix. Refers to `test-utils`?

teogeb (approved):

PR #747, opened 2022-06-28
build: remove npm prepare scripts

harbu:
Remove `prepare` script from each sub-packages' **`package.json`**.

Rationale being the undesired way it interacts with `npm ci` and `npm install` commands at monorepo root level. When you run either aforementioned command at root-level of monorepo (with the intention of installing node modules), if there is any compilation problem, all the installed `node_modules` directories will be cleared and the work undone. This is hardly helpful when trying to determine the root cause of the compilation issue...


harbu:
Same thing is accomplished with #736 so closing this.

PR #751, opened 2022-06-30
test(client): Resend historical data with existing key

teogeb:
Added a new resend integration test. 

In the test cases a subscriber does a resend and uses previously acquired group keys to decrypt the data.  In the simulated situation, the publisher is offline and therefore the subscriber can't get keys from it (all group key requests timeout).

Other changes needed to support the implementation:
- Group key requests can timeout: there is a new internal configuration field, which specifies maximum time to wait for a group key from the publisher.
- Some group key related methods extracted to new methods of `EncryptionUtil`
- A small enhancement in `FakeStorageNode#getRange`

Added also a DI config token for `TimeoutsConfig`. All classes which used the root token to get the timeout config are now using the new DI token. 

### Future improvements

The default value for group key request timeout is currently undefined. Therefore this PR doesn't change the functionality of group key requests. If we want that group key requests can timeout in production/test environment, we could specify a default value for the new setting.

All timeout configs are currently internal. Those configs are not documented and in `StreamrClientConfig` the config block has an underscore prefix to annotate that the configs are internal. We could promote the config block to be part of official configuration. That would require:
- Specify a good structure for timeout-related configs (maybe not a separate timeouts block, but e.g. the group key timeout could be inside encryption-related configs)
- Write documentation about the configs
- Add some test coverage for the config values
- Enable users to provide partial timeout configs: i.e. merge user config with defaults if only part of the config values are defined by the user

### DIFF
--- a/packages/client/src/encryption/EncryptionUtil.ts
+++ b/packages/client/src/encryption/EncryptionUtil.ts
@@ -75,7 +75,7 @@ export class EncryptionUtil {
         streamMessage.groupKeyId = groupKey.id
         streamMessage.serializedContent = this.encrypt(Buffer.from(streamMessage.getSerializedContent(), 'utf8'), groupKey)
         if (nextGroupKey) {
-            streamMessage.newGroupKey = new EncryptedGroupKey(nextGroupKey.id, this.encrypt(nextGroupKey.data, groupKey))
+            streamMessage.newGroupKey = EncryptionUtil.encryptGroupKey(nextGroupKey, groupKey)
         }
         streamMessage.parsedContent = undefined
         /* eslint-enable no-param-reassign */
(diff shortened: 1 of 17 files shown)


packages/client/src/encryption/EncryptionUtil.ts
```diff
@@ -108,16 +108,24 @@ export class EncryptionUtil {
...
+                streamMessage.newGroupKey = EncryptionUtil.decryptGroupKey(newGroupKey, groupKey)
             }
         } catch (err) {
             streamMessage.encryptionType = StreamMessage.ENCRYPTION_TYPES.AES
             throw new UnableToDecryptError('Could not decrypt new group key: ' + err.stack, streamMessage)
         }
         /* eslint-enable no-param-reassign */
     }
+
+    static encryptGroupKey(nextGroupKey: GroupKey, currentGroupKey: GroupKey): EncryptedGroupKey {
```
harbu, line 120: Could be made `private`?
```suggestion
    private static encryptGroupKey(nextGroupKey: GroupKey, currentGroupKey: GroupKey): EncryptedGroupKey {
```

packages/client/src/encryption/EncryptionUtil.ts
```diff
@@ -108,16 +108,24 @@ export class EncryptionUtil {
...
+        return new EncryptedGroupKey(nextGroupKey.id, this.encrypt(nextGroupKey.data, currentGroupKey))
+    }
+
+    static decryptGroupKey(newGroupKey: EncryptedGroupKey, currentGroupKey: GroupKey): GroupKey {
```
harbu, line 124: Could be made `private`?
```suggestion
    private static decryptGroupKey(newGroupKey: EncryptedGroupKey, currentGroupKey: GroupKey): GroupKey {
```

harbu (approved):

PR #754, opened 2022-07-01
cleanup root package json

harbu:
- Move husky (conventional commit enforcer) installation from `prepare` phase to its own script `install-git-hooks`.
- Don't display version matrix in `prepare` phase to reduce visual noise when installing. (You can still check it with `npm run versions`.)
-  Fix `streamr-client` versions in `broker` and `cli-tool` packages as`manypkg` was warning about version compatibility every time you ran certain NPM scripts.
- Remove `clean-package-locks` script from root **`package.json`**. We don't have package-lock.json files in sub-packages and removing the root one is trivial by hand anyways.
- Add missing `clean` script to `cli-tool` sub-package
- Remove `clean-cache` script. Felt it was a bit redundant given that the `clean` scripts of each sub-package also clear caches.
- Re-order scripts in root-level **`package.json`**.
- Remove `bootstrap-root` script. Seemed to serve a very marginal use case at best.
- In `versions` script run `manypkg` check before printing symlink versions
- Remove `manypkg` check from `eslint` script in root **`package.json`**
- Add new script `fix-package-jsons` to root **`package.json`**


package.json
```diff
@@ -2,20 +2,18 @@
     "name": "streamr-monorepo",
     "private": true,
     "scripts": {
```
teogeb, line 4: - The`eslint` command could run just eslint, not `manypkg check`? (include to be part of the `check` command maybe?)
- As we do re-ordering of scripts, we could e.g.:
  - put all mostly used script first: `build`, `eslint`, `check` 
  - then all installing command (`bootstrap` etc.)
  - then all cleanup command
  - then other commands

teogeb (approved):

package.json
harbu: Discussed and appropriate changes applied.

PR #756, opened 2022-07-04
refactor(client): Group key validation

teogeb:
Removed calls to `GroupKey#validate`. A `GroupKey` object is always in valid state, because the constructor calls `validate()` method internally. Therefore users of that object don't need to revalidate the object. The `GroupKey#validate` is now a private method.

### Future improvements

Is there a good reason to store the byte data of a `GroupKey` in two separate fields, which contain the same information in two different formats? 

Do we need `GroupKeyish` type?

### DIFF
--- a/packages/client/src/encryption/GroupKey.ts
+++ b/packages/client/src/encryption/GroupKey.ts
@@ -50,11 +50,11 @@ export class GroupKey {
     static InvalidGroupKeyError = InvalidGroupKeyError
 
     /** @internal */
-    id: string
+    readonly id: string
     /** @internal */
-    hex: string
+    readonly hex: string
     /** @internal */
-    data: Uint8Array
+    readonly data: Uint8Array
 
     constructor(groupKeyId: string, groupKeyBufferOrHexString: Uint8Array | string) {
         this.id = groupKeyId
--- a/packages/client/src/encryption/GroupKey.ts
+++ b/packages/client/src/encryption/GroupKey.ts
@@ -69,7 +69,7 @@ export class GroupKey {
         (this.constructor as typeof GroupKey).validate(this)
     }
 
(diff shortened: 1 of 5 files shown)


packages/client/src/encryption/GroupKey.ts
```diff
@@ -36,12 +36,48 @@ function GroupKeyObjectFromProps(data: GroupKeyProps | GroupKeyObject): GroupKey
...
     /** @internal */
     static InvalidGroupKeyError = InvalidGroupKeyError
 
-    static validate(maybeGroupKey: GroupKey): void | never {
+    /** @internal */
+    id: string
+    /** @internal */
+    hex: string
+    /** @internal */
+    data: Uint8Array
```
harbu: Regarding original code, could we make these readonly for clarity?
```suggestion
    readonly id: string
    /** @internal */
    readonly hex: string
    /** @internal */
    readonly data: Uint8Array
```

packages/client/src/encryption/GroupKey.ts
```diff
@@ -36,12 +36,48 @@ function GroupKeyObjectFromProps(data: GroupKeyProps | GroupKeyObject): GroupKey
...
+
+        if (typeof groupKeyBufferOrHexString === 'string') {
+            this.hex = groupKeyBufferOrHexString
+            this.data = Buffer.from(this.hex, 'hex')
+        } else {
+            this.data = groupKeyBufferOrHexString
+            this.hex = Buffer.from(this.data).toString('hex')
+        }
+
+        (this.constructor as typeof GroupKey).validate(this)
```
harbu: Regarding original code, some weird indirectness here?
```suggestion
        GroupKey.validate(this)
```

harbu (approved):

PR #757, opened 2022-07-04
refactor(client): `EncryptionUtil` names

teogeb:
Rename some methods in `EncryptionUtil` to be more specific about the encryption method:
- `encrypt` and `decrypt` -> `encryptWithAES` and `decryptWithAES`
- `encryptWithRSAPublicKey`, `decryptWithRSAPrivateKey` and `validateRSAPublicKey` now have `RSA` word in the names

The `encryptWithAES` and `decryptWithAES` methods are now public. The methods can be seen to be part of public API of that class, similarly to RSA-related methods. We also test the functionality of these method separately in unit tests.

Modified `encryptWithAES` and `decryptWithAES` method to get CipherKey as a parameter instead of GroupKey. This way there is a clear separation of abstraction level in the methods of EncryptionUtil: some methods operate on low-level encryption logic (buffers, cipher keys etc.), and other method operate on high-level classes like `StreamMessage` and `GroupKey`.

### Other refactoring

Renamed `RsaKeyPair` to `RSAKeyPair` so that we have consistent naming in RSA-related classes/methods.


packages/client/src/encryption/EncryptionUtil.ts
```diff
@@ -22,12 +22,12 @@ export class EncryptionUtil {
...
+    static encryptWithRSAPublicKey(plaintextBuffer: Uint8Array, publicKey: crypto.KeyLike, outputInHex: true): string
     // These overrides tell ts outputInHex returns string
-    static encryptWithPublicKey(plaintextBuffer: Uint8Array, publicKey: crypto.KeyLike): string
-    static encryptWithPublicKey(plaintextBuffer: Uint8Array, publicKey: crypto.KeyLike, outputInHex: false): Buffer
-    static encryptWithPublicKey(plaintextBuffer: Uint8Array, publicKey: crypto.KeyLike, outputInHex: boolean = false): string | Buffer {
-        this.validatePublicKey(publicKey)
+    static encryptWithRSAPublicKey(plaintextBuffer: Uint8Array, publicKey: crypto.KeyLike): string
+    static encryptWithRSAPublicKey(plaintextBuffer: Uint8Array, publicKey: crypto.KeyLike, outputInHex: false): Buffer
+    static encryptWithRSAPublicKey(plaintextBuffer: Uint8Array, publicKey: crypto.KeyLike, outputInHex: boolean = false): string | Buffer {
+        this.validateRSAPublicKey(publicKey)
```
harbu, line 30: I wonder if all these variants are in use?

packages/client/test/unit/RSAKeyPair.test.tss
```diff
@@ -0,0 +1,22 @@
+import { RSAKeyPair } from '../../src/encryption/RSAKeyPair'
+import { EncryptionUtil } from '../../src/encryption/EncryptionUtil'
+
+describe('RSAKeyPair', () => {
```
harbu: Rename file to remove extra s at end

harbu (approved):

packages/client/src/encryption/EncryptionUtil.ts
teogeb: We could maybe remove some of these in a separate PR

PR #762, opened 2022-07-04
refactor(client): Refactor `Publisher`

teogeb:
Refactored `Publisher` to use new `MessageFactory` class instead of `PublishPipeline`. Inlined some helper classes: `Encrypt`, `MessageCreator`, `Signer` and `StreamPartitioner`.

### Small changes

- In publish process we don't call `Validator#validate` as it is not possible to create invalid messages with the `MessageFactory` (user can't define a signature or give a parameter which would yield to an out-of-range), and `Publisher` checks from the contract that the publisher is valid
- `GroupKeyStore#useGroupKey` returns a `GroupKeySequence`
- `FailedToPublishError` (renamed to `PublishError`): metadata and error chain
- Some type annotations for `publisherId` (it is an `EthereumAddress`)

### Future improvements

- Divide `StreamrClient.test.ts` integration test


packages/client/src/publish/MessageFactory.ts
```diff
@@ -0,0 +1,105 @@
...
+    constructor(opts: MessageFactoryOptions) {
+        this.streamId = opts.streamId
+        this.partitionCount = opts.partitionCount
+        this.selectedDefaultPartition = random(opts.partitionCount - 1)
+        this.isPublicStream = opts.isPublicStream
+        this.publisherId = opts.publisherId
+        this.createSignature = opts.createSignature
+        this.useGroupKey = opts.useGroupKey
+        this.getStreamPartitionForKey = CacheFn((partitionKey: string | number) => {
+            return keyToArrayIndex(opts.partitionCount, partitionKey)
```
harbu: I'd like to do a faster benchmark of the md5 code path, is it really that expensive?

packages/client/src/publish/MessageFactory.ts
```diff
@@ -0,0 +1,105 @@
...
+    private readonly createSignature: (payload: string) => Promise<string>
+    private readonly useGroupKey: () => Promise<GroupKeySequence>
+    private readonly getStreamPartitionForKey: (partitionKey: string | number) => number
+    private readonly getMsgChain: (streamPartId: StreamPartID, publisherId: EthereumAddress, msgChainId?: string) => MessageChain
+
+    constructor(opts: MessageFactoryOptions) {
+        this.streamId = opts.streamId
+        this.partitionCount = opts.partitionCount
+        this.selectedDefaultPartition = random(opts.partitionCount - 1)
+        this.isPublicStream = opts.isPublicStream
```
harbu: We should not store these values here but instead rely on (cached) functions as the underlying facts can change in the smart contract between publishes.

packages/client/src/publish/MessageFactory.ts
```diff
@@ -0,0 +1,105 @@
...
+                throw new Error('Invalid combination of "partition" and "partitionKey"')
+            }
+        }
+        const partition = explicitPartition
+            ?? ((metadata.partitionKey !== undefined)
+                ? this.getStreamPartitionForKey(metadata.partitionKey!)
+                : this.selectedDefaultPartition)
+        const streamPartId = toStreamPartID(this.streamId, partition)
+
+        const chain = this.getMsgChain(streamPartId, this.publisherId, metadata?.msgChainId)
```
harbu:
```suggestion
        const chain = this.getMsgChain(streamPartId, this.publisherId, metadata.msgChainId)
```

packages/client/src/publish/MessageFactory.ts
```diff
@@ -0,0 +1,105 @@
...
+        this.getMsgChain = CacheFn((streamPartId: StreamPartID, publisherId: EthereumAddress, msgChainId?: string) => { // TODO would it ok to just use pMemoize (we don't have many chains)
+            return new MessageChain(streamPartId, publisherId, msgChainId)
+        }, {
+            cacheKey: ([streamPartId, publisherId, msgChainId]) => [streamPartId, publisherId, msgChainId ?? ''].join('|'),
+            ...opts.cacheConfig,
+            maxAge: Infinity
+        })
+    }
+
+    async createMessage<T>(
```
harbu, line 51: Inside this method we need to add the check that whether current user is valid publisher to this stream.

packages/client/src/publish/Publisher.ts
```diff
@@ -1,65 +1,171 @@
...
+        const timestamp = parseTimestamp(metadata)
+        /*
+         * There are some steps in the publish process which need to be done sequentially:
+         * - message chaining
+         * - consuming a group key from a queue
+         * 
+         * It is also good if messages are published to node in the same sequence, as that can avoid
+         * unnecessary gap fills (if a subscriber would receive messages m1, m2, m3 in order m1, m3, m2
+         * it would try to get m2 via a gap fill resend before it receives it normally).
+         *
```
harbu:
```suggestion
         * It is also good if messages are published to node in the same sequence (within a message chain), as that can avoid
         * unnecessary gap fills (if a subscriber would receive messages m1, m2, m3 in order m1, m3, m2
         * it would try to get m2 via a gap fill resend before it receives it normally).
         * 
```

packages/client/src/publish/MessageFactory.ts
teogeb: Fixed

packages/client/src/publish/MessageFactory.ts
teogeb: Fixed

packages/client/src/publish/MessageFactory.ts
teogeb: Fixed

packages/client/src/publish/Publisher.ts
teogeb: Fixed

packages/client/src/publish/MessageFactory.ts
teogeb: Did a benchmark: 1M partitionKey->partition queries using 100 separate partitionKeys for a stream which has 50 partitions.

It takes ~400ms to get results if we have caching, and ~4000ms if we don't have it. Its 10x difference, but on the other hand both operations are super-fast anyway.

Also this map partitionKey->partition is just a convenience feature which doesn't need to be super-fast.

```
    const getPartitionCount = async () => 50
    const cachedFn = CacheAsyncFn(async (partitionKey: string | number) => {
        return keyToArrayIndex(await getPartitionCount(), partitionKey)
    }, {
        ...STREAM_CLIENT_DEFAULTS.cache,
        cacheKey: ([partitionKey]) => partitionKey
    })
    const noncachedFn = async (partitionKey: string | number) => {
        return keyToArrayIndex(await getPartitionCount(), partitionKey)
    }
    const fns: any = {
        cachedFn,
        noncachedFn
    }

    const KEY_COUNT = 100
    const ITERATION_COUNT = 1000000
    const keys = range(KEY_COUNT).map(() => randomString(20))
    
    for (const fnName of Object.keys(fns)) {
        const fn = fns[fnName]
        const startTime = Date.now()
        let sum = 0
        for (let i = 0; i < ITERATION_COUNT; i++) {
            const k = keys[i % keys.length]
            const partition = await fn(k)
            sum += partition
        }
        console.log(fnName + '   iterations: ' + ITERATION_COUNT + ', sum:' + sum + ', took ' + (Date.now() - startTime) + 'ms')
    }
```

teogeb: -> Removed the caching

packages/client/src/publish/Publisher.ts
```diff
@@ -1,65 +1,156 @@
...
-        return this.pipeline.publish({
-            streamDefinition,
-            content,
-            timestamp: parseTimestamp(metadata),
-            partitionKey: metadata?.partitionKey,
-            msgChainId: metadata?.msgChainId,
-            messageType: metadata?.messageType,
-            encryptionType: metadata?.encryptionType
+    /* eslint-disable @typescript-eslint/no-shadow */
+    private async createMessageFactory(streamId: StreamID): Promise<MessageFactory> {
```
harbu: Could we move this definition down in the file after the public methods?

packages/client/src/publish/MessageFactory.ts
```diff
@@ -0,0 +1,119 @@
...
+    private readonly publisherId: EthereumAddress
+    private readonly streamId: StreamID
+    private defaultPartition: number | undefined
+    private readonly defaultMessageChainIds: Mapping<[partition: number], string>
+    private readonly messageChains: Mapping<[partition: number, msgChainId: string], MessageChain>
+    private readonly getPartitionCount: (streamId: StreamID) => Promise<number>
+    private readonly isPublicStream: (streamId: StreamID) => Promise<boolean>
+    private readonly isPublisher: (streamId: StreamID, publisherId: EthereumAddress) => Promise<boolean>
+    private readonly createSignature: (payload: string) => Promise<string>
+    private readonly useGroupKey: () => Promise<GroupKeySequence>
```
harbu, line 32: I like to organize these from "strictest" to least "strict", i.e. `private readonly` comes before `private` so you can quickly glance what references can change during the lifecycle of the instance of the class.

packages/client/src/publish/MessageFactory.ts
```diff
@@ -0,0 +1,119 @@
...
+        this.useGroupKey = opts.useGroupKey
+        this.defaultMessageChainIds = new Mapping(async (_partition: number) => {
+            return createRandomMsgChainId()
+        })
+        this.messageChains = new Mapping(async (partition: number, msgChainId: string) => {
+            return new MessageChain(toStreamPartID(this.streamId, partition), this.publisherId, msgChainId)
+        })
+    }
+
+    /* eslint-disable padding-line-between-statements */
```
harbu, line 50: Let's remove this rule from .eslintrc altogether?

packages/client/src/publish/MessageFactory.ts
```diff
@@ -0,0 +1,119 @@
...
+        this.publisherId = opts.publisherId
+        this.streamId = opts.streamId
+        this.getPartitionCount = opts.getPartitionCount
+        this.isPublicStream = opts.isPublicStream
+        this.isPublisher = opts.isPublisher
+        this.createSignature = opts.createSignature
+        this.useGroupKey = opts.useGroupKey
+        this.defaultMessageChainIds = new Mapping(async (_partition: number) => {
+            return createRandomMsgChainId()
+        })
```
harbu, line 44: I wonder if we could define `Mapping` so that it could also work without `async` valueFactory but instead non-async valueFactory?

packages/client/src/publish/MessageFactory.ts
```diff
@@ -0,0 +1,119 @@
...
+    async createMessage<T>(
+        content: T,
+        metadata: MessageMetadata & { timestamp: number },
+        explicitPartition?: number
+    ): Promise<StreamMessage<T>> {
+        const isPublisher = await this.isPublisher(this.streamId, this.publisherId)
+        if (!isPublisher) {
+            throw new Error(`${this.publisherId} is not a publisher on stream ${this.streamId}`)
+        }
+        const partitionCount = await this.getPartitionCount(this.streamId)
```
harbu: Maybe a line break here to visually separate publisher checking from figuring out partition

packages/client/src/publish/MessageFactory.ts
```diff
@@ -0,0 +1,119 @@
...
+            partition = explicitPartition
+        } else {
+            partition = (metadata.partitionKey !== undefined)
+                ? keyToArrayIndex(partitionCount, metadata.partitionKey)
+                : await this.getDefaultPartition(partitionCount)
+        }
+
+        const chain = await this.messageChains.get(
+            partition,
+            metadata.msgChainId ?? await this.defaultMessageChainIds.get(partition)
```
harbu, line 79: Again if we can find a way to make `Mapper` type support non-async valueFactories, we could get rid of the `await` here.

packages/client/src/publish/MessageFactory.ts
```diff
@@ -0,0 +1,119 @@
...
+            groupKeyId,
+            newGroupKey,
+            signatureType: StreamMessage.SIGNATURE_TYPES.ETH
+        })
+        message.signature = await this.createSignature(message.getPayloadToSign())
+
+        return message
+    }
+
+    private async getDefaultPartition(partitionCount: number): Promise<number> {
```
harbu: No need for `async`?

packages/client/src/utils/Mapping.ts
```diff
@@ -0,0 +1,34 @@
+/*
+ * A map data structure which lazily evaluates values. A factory function
+ * is called to create a value when when an item is queried for the first time.
+ * The map stores the value and any subsequent call to get() returns
+ * the same value.
+ */
```
harbu: I'd move this right before keyword `class`

packages/client/src/utils/Mapping.ts
```diff
@@ -0,0 +1,34 @@
...
+
+// an wrapper object is used so that we can store undefined values
```
harbu, line 1: Good clarification :+1:

harbu (commented):
Didn't read tests yet, here is review on non-test source code

packages/client/test/unit/Mapping.test.ts
```diff
@@ -0,0 +1,34 @@
...
+        const mapping = new Mapping(async (p1: string, p2: number) => `${p1}${p2}`)
+        expect(await mapping.get('foo', 1)).toBe('foo1')
+        expect(await mapping.get('bar', 2)).toBe('bar2')
+    })
+
+    it('memorize', async () => {
+        let evaluationIndex = 0
+        const mapping = new Mapping(async (_p: string) => {
+            const result = evaluationIndex
+            // eslint-disable-next-line no-plusplus
```
harbu, line 15: Let's remove this rule for good.

packages/client/test/unit/MessageFactory.test.ts
```diff
@@ -0,0 +1,220 @@
...
+            isPublisher: async () => false
+        })
+        return expect(async () => {
+            await messageFactory.createMessage(CONTENT, {
+                timestamp: TIMESTAMP
+            })
+        }).rejects.toThrow(/is not a publisher on stream/)
+    })
+
+    describe('partitions', () => {
```
harbu, line 121: Add two test cases
- negative partition
- partition > PARTITION_COUNT

packages/client/test/unit/MessageFactory.test.ts
```diff
@@ -0,0 +1,220 @@
...
+            expect(msg2.getPreviousMessageRef()).toBe(null)
+        })
+
+        it('explicit msgChainId', async () => {
+            const messageFactory = createMessageFactory()
+            const msg1 = await messageFactory.createMessage(CONTENT, { timestamp: TIMESTAMP, msgChainId: 'mock-id' })
+            const msg2 = await messageFactory.createMessage(CONTENT, { timestamp: TIMESTAMP })
+            const msg3 = await messageFactory.createMessage(CONTENT, { timestamp: TIMESTAMP, msgChainId: 'mock-id' })
+            expect(msg1.getMessageID().msgChainId).toBe('mock-id')
+            expect(msg2.getMessageID().msgChainId).not.toBe(msg1.getMessageID().msgChainId)
```
harbu:
```suggestion
            expect(msg2.getMessageID().msgChainId).not.toBe('mock-id')
```

packages/client/test/unit/MessageFactory.test.ts
```diff
@@ -0,0 +1,220 @@
...
+                    timestamp: TIMESTAMP
+                })
+                expect(msg.messageId.streamPartition).toBeLessThan(partitionCount)
+                // eslint-disable-next-line no-plusplus
+                partitionCount--
+            }
+        })
+    })
+
+    describe('message chains', () => {
```
harbu, line 183: Consider adding test case for the following publish two message with same msgChainId but different partitions. They should have the same `msgChainId` value but _not_ be linked (i.e. `prevMsgRef`  is null)

harbu (approved):

packages/client/test/unit/Mapping.test.ts
harbu: Addressed in https://github.com/streamr-dev/network-monorepo/pull/866

packages/client/src/publish/MessageFactory.ts
harbu: Addressed in https://github.com/streamr-dev/network-monorepo/pull/866

packages/client/src/publish/MessageFactory.ts
teogeb: Not trivial as `Mapping#get` needs an `async` keyword. We could implement that later if needed.

packages/client/src/publish/MessageFactory.ts
teogeb: Yes, that logic is almost followed here. But `defaultPartition` is just after `streamId` because those are about the same "topic", and helper accessor functions are after other fields.

PR #764, opened 2022-07-06
Move classes and functions to @streamr/utils

harbu:
- Add `@streamr/utils` dependency to `protocol`, `client`, and `cli-tools` package.
- Move `Multimap` to `@streamr/utils` package.
- Replace "custom" `withTimeout` in `streamr-client` with one from `@streamr/utils`.
    - Added ability to define custom error message context in `withTimeout`. 
- Move `scheduleAtInterval` and `scheduleAtFixedRate` to `@streamr/utils` package.
    - Refactor implementations by extracting shared common function
- Move `wait` to `@streamr/utils` package.


packages/client/src/utils/promises.ts
```diff
@@ -325,20 +325,3 @@ export async function until(
 
 // TODO import this from a library (e.g. streamr-test-utils if that is no longer a test-only dependency)
 export const wait = (ms: number): Promise<void> => new Promise((resolve) => setTimeout(resolve, ms))
```
teogeb: Remove also this `wait` function

teogeb (approved):
Great 👍 We could also remove the custom `until` function in `packages/client/src/utils/promises.ts:287` and use `waitForCondition` instead

PR #780, opened 2022-07-19
feat: implement last value cache

cehali:
## Checklist

- [ ] Updated Changelog
- [ ] Updated Documentation


harbu:
Hey, may I ask what are you trying to achieve here?

If I understand correctly you want to maintain a cache of the latest value of some stream? Is this something that could be achieved by writing a "cache wrapper" around the `client.subscribe` sort of like:

```ts
class LastValueCacher {
    private lastValue = ''
    
    constructor(client: StreamrClient, streamId: string) {
        client.subscribe(streamId, (msg) => {
            lastValue = msg
        })
    }
    
    getLastValue() {
        return lastValue
    }
}
```

cehali:
Hey, thanks for your comment and I believe we will go with your suggestion. We had a discussion regarding using Streamr within RedStone ecosystem and we come up with the idea of a cache with the last value which could be fetched by anyone. I implemented PoC just to avoid coming to you with empty hands 😄. We agreed that it would be easier to implement cache on our side and not mess with your codebase. So thank you once again for the comment, sorry for any trouble and I will close the PR.

PR #788, opened 2022-08-11
test(client): Publish without `PublishPipeline` internals

teogeb:
Refactored tests to publish messages with public API methods (`client.publish()`). Earlier some tests called and observed the internal `PublishPipeline` class instead.

As tests don't rely on internal implementation details, it will be easier to refactor and enhance the publish process in the future (https://github.com/streamr-dev/network-monorepo/pull/762)

### Refactoring

- The `getPublishTestStreamMessages` helper method is now used in most of the tests. It publishes messages via `client.publish()` and returns the published messages in an array. As the most of the tests streams are private, those messages are typically encrypted. Earlier that method observed the `PublishPipeline` and transformed the array to contain non-encrypted messages instead.
- In many tests we compare messages by analyzing if the signatures of received messages match to the signatures of published messages. We can't compare the full message objects as published messages are encrypted and received messages are decrypted.
- `Encryption.test.ts` and some other tests use values returned by `getPublishTestStreamMessages` instead of listening the internal `PublishPipeline` object.
- Redundant helper methods, options and interfaces in `test-utils/publish.ts` have been removed.

### Other changes 

- `StreamrClient.test.ts` and `Subscriber.tests.ts` now use encrypted streams.
- Some test duplication reduced: e.g. we don't check message encryption types in `Encryption.test.ts` as that logic is already verified in `Publisher.test.ts` and `PublisherKeyExchange.test.ts`.


juslesan (approved):
LGTM

PR #808, opened 2022-08-30
feat(client) Group key exchange using unicast and multicast

teogeb:
Requires `NetworkNode` and `Tracker` changes (in a separate PR)

- If we don't validate sender/receiver of GroupKeyRequest/GroupKeyResponse in protocol level, we must check at least that the requestor has subscribe permission before we send the response (is it good to check that the publisher has publish permissions, too, or is it redundant as we could think that anyone is allowed respond to the query -> if the node has the requested group key, it is ok to provide that to the requestor)

- TODO: maybe we should not requery the group key immediatelly if we didn't get it when processing the previous message (that operation timeouted)

- TODO: error responses
  - we could possibly also add error handling for situation where `GroupKeyRequest` can't be routed (e.g. publisher not connected to the tracker)

- TODO should we remove unicast/multicast listeners on client destroy?

- would it make sense to throttle group key exchanges (the request which we don't have yet received responses for)?


teogeb:
Not needed as key exchange uses in-stream delivery: https://github.com/streamr-dev/network-monorepo/pull/824

PR #813, opened 2022-09-01
feat(tracker): Unicast and multicast messages

teogeb:
Nodes can send  and receive unicast message and multicast messages to/from the network.

- new concept: `userId`, which is a prefix in nodeId
  - TODO `userId` is typically an Ethereum address, should we make it case-insensitive, or e.g. limit that nodeIds can only contain lowercase addresses? 

- should  `sendUnicastMessage`/`sendMulticastMessage` return void or Promise<void>? (publish returns void)

### TODO

- tests for `UnicastMessage` and `MulticastMessage`
- rename `addMessageListener` to `addBroadcastMessageListener` or `on('broadcastMessageReceive')`
  - and removeMessageListener
  - and if we use `on` method, then also unicast and multicast listener with that style
- should `parseUserIdFromNodeId` return lowercased string if we decide that userIds are case-insensitive?
- rename `NodeId` type to `NodeID` (and `TrackerId` and `UserId`)
- should `TrackerManager#nodeToTracker` should be private field?

### DIFF
--- a/packages/client/src/NetworkNodeFacade.ts
+++ b/packages/client/src/NetworkNodeFacade.ts
@@ -2,13 +2,13 @@
  * Wrap a network node.
  */
 import { inject, Lifecycle, scoped } from 'tsyringe'
-import { NetworkNodeOptions, createNetworkNode as _createNetworkNode, MetricsContext } from 'streamr-network'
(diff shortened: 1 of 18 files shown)


teogeb:
Not needed as key exchange uses in-stream delivery: https://github.com/streamr-dev/network-monorepo/pull/824

PR #824, opened 2022-09-13
feat: Key exchange in stream

teogeb:
Group key requests and responses are delivered within the stream and not in a separate key exchange stream (`SYSTEM/keyexchange/${publisherId}`).

This improves privacy as it is not clear where the publisher node is in a stream topology. Also there is less connection management as we don't need to form separate connections for key exchange streams, resulting in more stable key-exchange. But as the messages are now broadcast messages, we need to deliver significantly more messages. It has some implications to overall network performance due to added noise (nodes in topology will propagate group key messages that do not concern them).

### Version compatibility

This change is not 100% compatible with v6 clients when _private streams_ are used. If there are both v6 and v7 clients in the same stream topology:

- If a v7 client receives a message from a v6 client, the subscriber won't be able to decrypt the messages
  - v7 subscriber will try to request a key in-stream, but won't get a response -> timeouts
  - all v6 publishers will see the key request as an invalid message because in v6 key requests are expected to be delivered in key exchange streams -> `sub.onError` is triggered for the `Subscription`
- If a v6 client receives a message from a v7 client, the subscriber won't be able to decrypt the message
  - v6 subscriber will send a group key request to the key-exchange stream, but won't get a response -> timeouts or fully blocks the decryption process (there is a timeout in `6.1.0` but no in earlier v6 releases)
  - v7 publisher doesn't listen to key-exchange streams -> request is ignored
- If a v7 client receives a message from a v7 client, v6 subscribers in that same topology may see the request
  - v6 subscribers see the key request as an invalid message -> `sub.onError` is triggered for the `Subscription`

For public streams there are no incompatibilities between v6 and v7.

### Protocol

`GroupKeyRequest` and `GroupKeyResponse` have a new `recipient` field (the recipient of the request or the response, the opposite participant of the message publisher). The classes no longer have a `streamId` field as that info is contained in the enclosing `StreamMessage` already. 

This is a backwards-incompatible change, and we could increase the protocol version number. But in practice the new key requests are not delivered to v6 clients, and therefore it is ok to keep the current version number. This way we also allow v6 and v7 to communicate with each other via public streams.

### Broker

A storage node stores only messages of type `StreamMessageType.MESSAGE`, i.e. it skips group key requests and responses.

### Other changes

- Message validation is done in parallel for each message chain (in `SubscribePipeline`)
- Small changes to `Decrypt` lifecycle handling, e.g. removed an obsolete `stop()` call
- Fix in `BrowserPersistence#has`
- Fallback name in browser test
- `UnableToDecryptError` renamed to `DecryptError` (a semi-public class)

### Notes
- In `requestGroupKey` we could check if there is a pending request and not re-request it. But in practice it wouldn't add any benefit as it is very unlikely that we'd be requesting same group key twice. The decryption process blocks for each message chain and doesn't not process a next message before the current message (which could have the same group key).
- Before this PR we used `PublishPipeline`/`SubscribePipeline` to send and receive the key requests. Therefore we implicitly validated the `GroupKeyRequest` instances which we generated. Now we don't do that. With `validator.validate()` we could check that the client which creates the request is really a subscriber, but we already almost know that as we got the message from some encrypted stream (client subscribed or made a resend to that stream). Removed also a duplicate `isStreamSubscriber` check when parsing the group key request (as validating the `GroupKeyRequest` message already validated that fact)
- If v6 publishes to v6, the key-error response may not be delivered
  - if a key-error response is routed via v7, it sees the message as an invalid message and doesn't route it
  - _in practice_ that doesn't happen because v7 clients don't subscribe to key-exchange streams

### Future improvements
- `PublisherKeyExchange` (and maybe `SubscriberKeyExchange`) could send an event to `StreamrClient` if there is an error in key handler so that a possible problem is observable by end-users (currently it just log the error to the debug log)
- When a client instance is destroyed `Decrypt` could throw some specific error (e.g. "`StoppedError`") and not return the non-decrypted message (just internal refactoring as the non-decrypted message is not delivered to end-users anyway)
- We could maybe have a public method to start the group key responder (if we'd have that, we should use the public method in tests instead of `startPublisherKeyExchangeSubscription`)


packages/client/src/Config.ts
```diff
@@ -156,6 +162,10 @@ export const STREAM_CLIENT_DEFAULTS: StrictStreamrClientConfig = {
             gasPriceStrategy: (estimatedGasPrice: BigNumber) => estimatedGasPrice.add('10000000000'),
         }
     },
+    decryption: {
+        keyRequestTimeout: 30 * 1000,
+        maxKeyRequestsPerSecond: 999999 // TODO just a placeholder value, define a valid value by executing some benchmarks
```
harbu:
```suggestion
        maxKeyRequestsPerSecond: 20
```

harbu (approved):

PR #826, opened 2022-09-14
Groupkey exchange via proxies

juslesan:
This PR makes it possible for nodes utilizing the proxy pattern to propagate GroupKey exchange related control traffic.

As a new requirement the proxy node must be made aware of the "identity" of the proxied publishers and subscribers. The identity is used by the proxy to route GroupKey related traffic to its appropriate destination.

## Checklist


packages/protocol/src/protocol/control_layer/proxy_connection_request/ProxyConnectionRequest.ts
```diff
@@ -14,15 +14,17 @@ interface Options extends ControlMessageOptions {
...
     direction: ProxyDirection
+    identity: string
 }
 
 export default class ProxyConnectionRequest extends ControlMessage {
     senderId: string
     streamId: StreamID
     streamPartition: number
     direction: ProxyDirection
+    identity: string
```
harbu: Based on discussion let's rename to `userId`. Not necessarily the final name forever, but a good compromise for now.

packages/network/test/integration/proxy-groupkey-exchange.test.ts
```diff
@@ -0,0 +1,136 @@
...
+
+        await proxy.subscribeAndWaitForJoin(streamPartId)
+    })
+
+    afterEach(async () => {
+        await Promise.all([
+            tracker.stop(),
+            proxy.stop(),
+            publisher.stop(),
+            subscriber.stop()
```
harbu:
```suggestion
            tracker?.stop(),
            proxy?.stop(),
            publisher?.stop(),
            subscriber?.stop()
```

packages/network/src/logic/propagation/Propagation.ts
```diff
@@ -47,14 +47,14 @@ export class Propagation {
     /**
      * Node should invoke this when it learns about a new message
      */
-    feedUnseenMessage(message: StreamMessage, source: NodeId | null): void {
+    feedUnseenMessage(message: StreamMessage, source: NodeId | null, additionalTargets: NodeId[] = []): void {
```
harbu: Let's refactor this class so that all neighbors are given as 2nd parameter to this method, i.e.

```ts
    feedUnseenMessage(message: StreamMessage, targets: NodeId[], source: NodeId | null): void {
```

packages/network/src/logic/ProxyStreamConnectionManager.ts
```diff
@@ -99,7 +101,18 @@ export class ProxyStreamConnectionManager {
         return this.connections.get(streamPartId)!.get(nodeId)!
     }
 
-    async openProxyConnection(streamPartId: StreamPartID, targetNodeId: string, direction: ProxyDirection): Promise<void> {
+    public getAllConnectionNodeIds(streamPartId: StreamPartID, identity: string): NodeId[] {
```
harbu:
```suggestion
    public getNodeIdsForUserId(streamPartId: StreamPartID, userId: string): NodeId[] {
```

packages/network/src/logic/ProxyStreamConnectionManager.ts
```diff
@@ -35,6 +35,7 @@ interface ProxyConnection {
     state?: State
     reconnectionTimer?: NodeJS.Timeout
     direction: ProxyDirection
+    identity?: string
```
harbu: Make identity (i.e. userId) mandatory
```suggestion
    identity: string
```

packages/network/src/logic/Node.ts
```diff
@@ -308,8 +311,25 @@ export class Node extends EventEmitter {
...
+                } else if (GroupKeyResponse.is(streamMessage)) {
+                    const { recipient } = GroupKeyResponse.fromStreamMessage(streamMessage) as GroupKeyResponse
+                    groupKeyPropagationTargets = this.proxyStreamConnectionManager.getAllConnectionNodeIds(streamPartId, recipient)
+                }
+            } else if (
+                this.streamPartManager.isBehindProxy(streamMessage.getStreamPartID())
+                && this.proxyStreamConnectionManager.isProxiedStreamPart(streamMessage.getStreamPartID(), ProxyDirection.SUBSCRIBE)
+            ) {
+                groupKeyPropagationTargets = [...this.streamPartManager.getInboundNodesForStreamPart(streamPartId)]
+            }
```
harbu: Extract method and include the normal neighbors also as part of the `Propagation` refactor.

harbu (changes requested):

harbu (approved):

PR #828, opened 2022-09-15
refactor(client): Simplify `Persistence`

teogeb:
Removed obsolete methods from `Persistence` interface, and the implementations of the interface (`ServerPersistence` and `BrowserPersistence`)
- `delete`
- `clear`
- `size`

Removed also related obsolete methods from GroupKeyStore:
- `clear`
- `size`
- `isEmpty`

Added `GroupKeyStore#destroy` so that tests can remove the generated data in test teardown (it uses `ServerPersistence#clear`/`BrowserPersistence#clear` internally and closes the connection).

Refactored `Persistence.set` to return `Promise<void>` instead of `Promise<boolean>`

### DIFF
--- a/packages/client/test/unit/GroupKeyStore.test.ts
+++ b/packages/client/test/unit/GroupKeyStore.test.ts
@@ -27,7 +27,8 @@ describeRepeats('GroupKeyStore', () => {
 
     afterEach(async () => {
         if (!store) { return }
-        await store.clear()
+        // @ts-expect-error private
+        await store.persistence.destroy()
         // @ts-expect-error doesn't want us to unassign, but it's ok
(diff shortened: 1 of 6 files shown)


packages/client/test/unit/GroupKeyStore.test.ts
```diff
@@ -27,7 +27,8 @@ describeRepeats('GroupKeyStore', () => {
 
     afterEach(async () => {
         if (!store) { return }
-        await store.clear()
+        // @ts-expect-error private
+        await store.persistence.destroy()
```
harbu: I would prefer to avoid having to use `// @ts-expect-error` here.

harbu (approved):

PR #842, opened 2022-09-22
feat: enable using real amazon latencies in DHT simulations

ptesavol:
Enable using real latencies measured from Amazon experiments in DHT simulations. PeerDescriptor has new "region" field, that is used to point a node to an Amazon region in the simulations.


juslesan (approved):
LGTM

PR #852, opened 2022-09-30
refactor(client): Extract `GroupKeyQueue`

teogeb:
Extracted group key queuing logic to a separate class. 

Before this PR the logic was implemented in `GroupKeyStore`. That class had two separate roles: handling key persistence and managing the group key queue. The extracted `GroupKeyQueue` class uses `GroupKeyStore` to persist the generated class, but is otherwise an independent entity.

Moved all test cases from `GroupKeyStore.test.ts` to a new `GroupKeyQueue.test.ts`, and therefore renamed `GroupKeyQueue2.test.ts` to `GroupKeyQueue.test.ts`.

Other changes:
- `GroupKeyStore#add` returns `Promise<void>`
- inlined `GroupKeyStore#storeKey`


packages/client/src/publish/GroupKeyQueue.ts
```diff
@@ -0,0 +1,45 @@
+import { GroupKey } from '../encryption/GroupKey'
+import { GroupKeyStore } from '../encryption/GroupKeyStore'
+
+export class GroupKeyQueue {
+
+    private currentGroupKey: GroupKey | undefined // current key id if any
```
harbu: Is code comment up-to-date? Talking about `current key id` but here we have a `GroupKey`?

packages/client/src/publish/GroupKeyQueue.ts
```diff
@@ -0,0 +1,45 @@
...
+    private queuedGroupKey: GroupKey | undefined // a group key queued to be rotated into use after the call to useGroupKey
+    private store: GroupKeyStore
```
harbu:
```suggestion
    private readonly store: GroupKeyStore
```

packages/client/src/publish/Publisher.ts
```diff
@@ -24,13 +27,21 @@ const parseTimestamp = (metadata?: MessageMetadata): number => {
...
 
     constructor(
         context: Context,
-        @inject(delay(() => PublishPipeline)) private pipeline: PublishPipeline
+        @inject(delay(() => PublishPipeline)) private pipeline: PublishPipeline,
+        groupKeyStoreFactory: GroupKeyStoreFactory
     ) {
         this.id = instanceId(this)
         this.debug = context.debug.extend(this.id)
+        this.getGroupKeyQueue = pMemoize(async (streamId: StreamID) => {
```
harbu, line 39: Why memoized?

teogeb: For each `streamId` we should keep one group key queue. This way we get the same instance for each call.

packages/client/src/publish/GroupKeyQueue.ts
teogeb: Fixed

packages/client/src/publish/GroupKeyQueue.ts
teogeb: Fixed

harbu (approved):

PR #855, opened 2022-10-03
fix(client): Concurrency issue in group key storage

teogeb:
Fixed a concurrency issue when group keys are added in parallel for multiple streams. 

As multiple streams initialized a separate `ServerPersistence` instance, each instance called `ServerPersistence#init()` in parallel. Therefore some of the calls could fail with `SQLITE_ERROR: no such table: GroupKeys` error.

### Refactoring

To fix this issue `GroupKeyStore` uses one `Persistence` instance instead of multiple instances. We give an explicit streamId parameter for `Persistence#get` and `Persistence#set `calls.

There was no real need to have multiple instances as `ServerPersistence` already filtered its data with a `streamId` filter. `BrowserPersistence` implementation used stream-specific stores, and now it hides the store management internally.


packages/client/src/utils/persistence/BrowserPersistence.ts
```diff
@@ -1,22 +1,24 @@
...
 
 export default class BrowserPersistence implements Persistence<string, string> {
-    private store: UseStore
-    private dbName: string
+    private getStore: (streamId: StreamID) => UseStore
 
-    constructor({ clientId, streamId }: { clientId: string, streamId: StreamID }) {
-        this.dbName = `streamr-client::${clientId}::${streamId}`
-        this.store = createStore(this.dbName, 'GroupKeys')
+    constructor({ clientId }: { clientId: string }) {
```
harbu, line 9: Just a question, what is `clientId` in practice? Is it the public Ethereum address of the client?

harbu (approved):

packages/client/src/utils/persistence/BrowserPersistence.ts
teogeb: Yes, it is `authentication.getAddress()` (`GroupKeyStore:38`), not an optimal variable name

PR #863, opened 2022-10-05
EthereumAddress and ENSName as branded strings

harbu:
Turn `EthereumAddress` and `ENSName` types into branded strings. This leads to more type safety and a guarantee that anytime you see one of the types you are dealing with a normalized representation (i.e. lowercased).

We should prefer these branded strings whenever we are dealing with internal parts of the code base. Our public-facing APIs (in the client package) should take in normal strings instead but turn them into branded strings ASAP. 

#### Design Principle: avoid forcing client users to use `EthereumAddress` or `ENSName` 

A design goal of this PR is to avoid users of  _streamr-client_ public API of having to use these branded types. E.g. instead of

```ts
async clientPublicMethod(address: EthereumAddress) {
      this.delegatee.internalMethod(address)
}
```

we have 

```ts
async clientPublicMethod(address: string) {
      this.delegatee.internalMethod(toEthereumAddress(address))
}
```

Thus leading to a more simple to use API
```ts
clientPublicMethod(toEthereumAddress('0x....')) // vs
clientPublicMethod('0x....')
```


packages/broker/test/integration/plugins/storage/storage-lots-of-data.test.ts
```diff
@@ -29,7 +30,7 @@ describe('Storage: lots of data', () => {
                 bucketKeepAliveSeconds: 1
             }
         })
-        streamId = getTestName(module) + '/' + Date.now()
+        streamId = getTestName(module) + Date.now()
```
teogeb, line 33: Intentional removal?

packages/client/src/Ethereum.ts
```diff
@@ -19,10 +18,10 @@ export interface EthereumNetworkConfig {
 }
 
 export interface EthereumConfig {
-    streamRegistryChainAddress: EthereumAddress
-    streamStorageRegistryChainAddress: EthereumAddress
-    storageNodeRegistryChainAddress: EthereumAddress
-    ensCacheChainAddress: EthereumAddress
+    streamRegistryChainAddress: string
```
teogeb, line 21: It is good that we don't require end-users to provide the data in our internal format (the `EthereumAddress` object). On the other hand, this previous style was good for readability as the data type was clearly communicated with a type.

packages/client/test/test-utils/fake/FakeStreamRegistry.ts
```diff
@@ -139,7 +139,7 @@ export class FakeStreamRegistry implements Omit<Methods<StreamRegistry>, 'debug'
         return this.updatePermissions(
             streamIdOrPath,
             assignments,
-            (registryItem: StreamRegistryItem, target: string, permissions: StreamPermission[]) => {
+            (registryItem: StreamRegistryItem, target: EthereumAddress | 'public', permissions: StreamPermission[]) => {
```
teogeb: The `public` type is defined in `FakeChain#PublicPermissionTarget`, could use that

packages/broker/test/unit/helpers/generateMnemonicFromAddress.test.ts
```diff
@@ -1,12 +1,10 @@
-import { randomBytes } from 'crypto'
 import { generateMnemonicFromAddress } from '../../../src/helpers/generateMnemonicFromAddress'
```
teogeb, line 1: We can now simplify `generateMnemonicFromAddress` function in (`generateMnemonicFromAddress.ts`) as it gets EthereumAddress as a parameter

packages/test-utils/src/utils.ts
```diff
@@ -237,8 +237,8 @@ export function fastWallet(): Wallet {
     return new Wallet(fastPrivateKey())
 }
 
-export function randomEthereumAddress(): string {
-    return '0x' + crypto.randomBytes(20).toString('hex')
+export function randomEthereumAddress(): EthereumAddress {
+    return toEthereumAddress('0x' + crypto.randomBytes(20).toString('hex'))
```
teogeb, line 241: This could be a utility in `streamr/utils`? (and we could migrate `createEthereumAddress` from `client/test/test-utils/utils.ts` usages to use this method or the method which we move to `streamr/utils`)

teogeb (approved):
Looks good! 👍 We could now remove almost all remaining `toLowercase()` calls from client etc.?

packages/broker/test/integration/plugins/storage/storage-lots-of-data.test.ts
harbu: Yeah `getTestName(module)` isn't a valid Ethereum address.

packages/client/test/test-utils/fake/FakeStreamRegistry.ts
harbu: Feels a bit wrong to be importing something from a fake-class here, should the type be moved to somewhere else? What would be a good place?

packages/broker/test/unit/helpers/generateMnemonicFromAddress.test.ts
harbu: Fixed in afd8a3475

packages/test-utils/src/utils.ts
harbu: I would keep it in test-utils. Hard to imagine a use case of the top off my head where generating random Ethereum public addresses would be needed in non-test code?

teogeb: True. We could then just remove `createEthereumAddress` from `client/test/test-utils/utils.ts` and use this `randomEthereumAddress` method in `StorageNodeRegistry2.test.ts`

packages/client/test/test-utils/fake/FakeStreamRegistry.ts
teogeb: We'd importing from a fake class to another fake class

packages/broker/test/integration/plugins/storage/storage-lots-of-data.test.ts
teogeb: Ah ok, if it'd include `/` it couldn't be in legacy format (which is kind of is)

packages/client/test/test-utils/fake/FakeStreamRegistry.ts
harbu: Fixed in 80eaa349e0886e040cbe684383ab0f8af20cc04a

teogeb:
As we now prefer plain types in public API, ce could remove `GroupKeyId` and other ID types which are aliases to String from `index-exports.ts`. Could do that in a separate PR.

packages/test-utils/src/utils.ts
harbu: Fixed 4fcf53f6e62e156ed35e9d23d7e26ec71de21465

PR #866, opened 2022-10-06
remove extra rules from client .eslintrc

harbu:
Remove extra rules from client .eslintrc to harmonize it with rest of packages according to monorepo root eslintrc.


harbu:
Remove extra rules from client .eslintrc to harmonize it with rest of packages according to monorepo root eslintrc.

PR #867, opened 2022-10-06
UpdateEncryptionKeyOptions: rename `streamId` to `streamIdOrPath`

harbu:
Should it be named `streamIdOrPath` to be in line with rest of client's public API?

### DIFF
--- a/packages/client/src/StreamrClient.ts
+++ b/packages/client/src/StreamrClient.ts
@@ -103,10 +103,10 @@ export class StreamrClient implements Context {
     }
 
     async updateEncryptionKey(opts: UpdateEncryptionKeyOptions): Promise<void> {
-        if (opts.streamId === undefined) {
+        if (opts.streamIdOrPath === undefined) {
             throw new Error('streamId required')
         }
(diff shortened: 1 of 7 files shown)


teogeb (changes requested):
I'd suggest that we keep the current name. 

There are some places in `StreamrClient` where user can provide info about a stream in a configuration object like this. E.g. for `StreamRegistry#setPermissions`  we give an object of
```
{ 
    streamId,
    assignments
}
```

In these situations we very typically provide a full `streamId`. 

The naming is a tradeoff between readability and writability. I'd usually favor readability over writability. The fact that we support also the path format is usually not very important for end-users. It won't cause bugs if we don't mention that we support also extra formats.

If both of formats would be used equally often it would maybe make sense to have a longer and more noisy name. One option would be to rename the parameter to `stream`. It would indicate that there are multiple ways to reference to stream, e..g. by an id or a path.

(For method parameters we should keep using the longer `streamIdOrPath` name as that is doesn't add noise to the end-user codebases.)

harbu:
Alright, wasn't sure if this was intentional choice or just "forgotten". But since it is intentional choice no need to further look at this PR.

PR #875, opened 2022-10-10
refactor(client): Create signed message

teogeb:
Sign `StreamMessage` instances with new `createSignedMessage` utility function.

Also related refactoring:
- removed parameter from `StreamMessage#getPayloadToSign` which mutated the message object
- remove `StreamMessageUnsigned` and `StreamMessageSigned` interface
- `MessageFactory` has more abstract dependencies:
  - `authentication`, `streamRegistry` and `groupKeyQueue` instead of multiple functions
  - easier to see the big picture of `MessageFactory` dependencies
  - can pass test stub instances easily (added `createStreamRegistryCached` and `createGroupKeyQueue` utility methods)
  - uses a bit less test-only code as we use a real `GroupKeyQueue`
- encapsulate caching functions of `StreamRegistryCached`
  - users of that component aren't supposed to call `getStream#clearMatching` etc.
  
### Future improvements
  
- `MessageChain` is actually just a wrapper to `prevMsgRef`. We could store a `prevMsgRef` for each `partition`+`msgChain` pair instead of storing a `MessageChain`. If we do that createSignedMessage could also get the publisherId directly from authentication (instead of `MessageID` created by `MessageChain`)

### DIFF
--- a/packages/client/test/integration/PublisherKeyExchange.test.ts
+++ b/packages/client/test/integration/PublisherKeyExchange.test.ts
@@ -21,7 +21,8 @@ import { FakeNetworkNode } from '../test-utils/fake/FakeNetworkNode'
 import { fastWallet } from 'streamr-test-utils'
 import { StreamrClient } from '../../src/StreamrClient'
 import { createRandomMsgChainId } from '../../src/publish/MessageChain'
-import { sign } from '../../src/utils/signingUtils'
+import { createSignedMessage } from '../../src/publish/MessageFactory'
+import { createAuthentication } from '../../src/Authentication'
 
 describe('PublisherKeyExchange', () => {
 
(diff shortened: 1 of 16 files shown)


packages/client/test/integration/PublisherKeyExchange.test.ts
```diff
@@ -42,15 +43,15 @@ describe('PublisherKeyExchange', () => {
         return stream
     }
 
-    const createGroupKeyRequest = (
+    const createGroupKeyRequest = async (
```
harbu, line 46: Future refactoring idea: extract a function `createGroupKeyRequest` in source code so test can use that instead?

harbu (approved):

packages/client/test/integration/PublisherKeyExchange.test.ts
teogeb: Alternatively we could use a real `SubscriberKeyExchange` in that test

PR #876, opened 2022-10-10
feat: Require signed messages

teogeb:
`StreamMessage` requires a signature. Non-signed messages are no longer supported. Removed `SignatureType.NONE`, which was previously the default signature type. Now the default is `SignatureType.ETH`.

StreamMessage validates that we pass `signatureType` when we create `StreamMessage`. Therefore if a client would receive a message which is not signed with `ETH` signature type, it would ignore it (it doesn't trigger `sub.on('error')` as the validation fails already on network node level, see `NodeToNode.ts:113`).

As a signature must be passed to `StreamMessage` as a constructor parameter, the signature payload utility function was extracted from `StreamMessage.getPayloadToSign` to a separate utility function outside of `StreamMessage` class.

### Open questions

Maybe we should remove `SignatureType` field altogether? All messages should be signed with that type and we don't currently plan to support other signature types. 

We don't need to modify the field from protocol (so that we can keep backwards-compatibility). But we could remove the field from `StreamMessage`. We could also check that the `signatureType` has a valid value of 2, when we deserialize a `StreamMessage` JSON array.

### DIFF
--- a/packages/client/CHANGELOG.md
+++ b/packages/client/CHANGELOG.md
@@ -22,6 +22,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Replace method `subscription.onResent(listener)` with `subscription.once('resendComplete', listener)`
 - Resend supports multiple storage nodes: the data is fetched from a random storage node
 - Enforce concurrency limit for smart contract calls (per contract, configurable with `maxConcurrentContractCalls` config option)
+- Enforce messages to be signed
+  - a client ignores all non-signed messages it receives
 - Method `stream.update()` parameter `props` is no longer optional
 - Rename method `getStorageNodesOf()` to `getStorageNodes()`
 - Rename method `getStoredStreamsOf()` to `getStoredStreams()`
--- a/packages/client/CHANGELOG.md
+++ b/packages/client/CHANGELOG.md
@@ -22,8 +22,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Replace method `subscription.onResent(listener)` with `subscription.once('resendComplete', listener)`
 - Resend supports multiple storage nodes: the data is fetched from a random storage node
 - Enforce concurrency limit for smart contract calls (per contract, configurable with `maxConcurrentContractCalls` config option)
-- Enforce messages to be signed
-  - a client ignores all non-signed messages it receives
+- Enforce presence of message signatures
+  - all non-signed messages received by client are simply ignored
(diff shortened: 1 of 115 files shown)


packages/client/CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Replace method `subscription.onResent(listener)` with `subscription.once('resendComplete', listener)`
 - Resend supports multiple storage nodes: the data is fetched from a random storage node
 - Enforce concurrency limit for smart contract calls (per contract, configurable with `maxConcurrentContractCalls` config option)
+- Enforce messages to be signed
+  - a client ignores all non-signed messages it receives
```
```suggestion
- Enforce presence of message signatures
  - all non-signed messages received by client are simply ignored
```

packages/client/test/unit/OrderingUtil.test.ts
```diff
@@ -31,53 +42,60 @@ describe('OrderingUtil', () => {
...
         }
-        for (let i = 2; i <= 100000; i++) {
-            expected2.push(createMsg(i, 0, i - 1, 0, {}, 'publisherId2'))
+        for (let i = 2; i <= 100; i++) {
+            expected2.push(await createMsg(i, 0, i - 1, 0, {}, 'publisherId2'))
         }
-        for (let i = 2; i <= 100000; i++) {
-            expected3.push(createMsg(i, 0, i - 1, 0, {}, 'publisherId3'))
+        for (let i = 2; i <= 100; i++) {
+            expected3.push(await createMsg(i, 0, i - 1, 0, {}, 'publisherId3'))
```
harbu: Maybe total 30k?

Actually maybe we can just use fake signatures and keep 300k?

packages/client/src/publish/MessageFactory.ts
```diff
@@ -29,13 +30,19 @@ export const createSignedMessage = async <T>(
     opts: Omit<StreamMessageOptions<T>, 'signature' | 'signatureType' | 'content'> 
     & { serializedContent: string, authentication: Authentication }
 ): Promise<StreamMessage<T>> => {
-    const msg = new StreamMessage<T>({
+    const signature = await opts.authentication.createMessagePayloadSignature(createSignaturePayload({
```
harbu:
```suggestion
    const signature = await opts.authentication.createMessagePayloadSignature(createSignaturePayload({
```

Maybe rethink name `createMessagePayloadSignature`?

harbu (approved):

PR #877, opened 2022-10-10
refactor(client): Message chain handling `MessageFactory`

teogeb:
Simplify message chaining in `MessageFactory`. 

Previously there was `MessageChain` class which was basically just encapsulating a `prevMsgRef` instance. Now `MessageChain` stores `prevMsgRef` objects directly, and `MessageChain` class is no longer needed.

Also error handling of backdated message chaining is now more explicit: `createMessageRef` now throws an error if it is called with a backdated argument. Previously it generated a `prefMsgRef` object, but an error was threw in `StreamMessage` (`StreamMessage:370`).

### Utility function

Added also a utility function `formLookupKey` which makes the code a bit more readable compared to a situation where we'd implement the string manipulation logic in a method which operates on a higher abstraction level  (e.g. in `MessageFactory#createMessage` it is not relevant how the lookup keys for message chains are formed).

### Future improvements

`MessageRef` object could be immutable (e.g. by adding readonly modifiers to timestamp and sequenceNumber fields)? Now `MessageFactory` stores a MessageRef instance internally and also passes the same value to StreamMessage constructor. In theory some component could mutate `MessageFactory's` internal state that way.

Alternatively we could use cloning of objects more aggressively. There is a `clone()` method in `MessageRef`.

If we change MessageRef to be immutable, we could modify also the `MessageID` class. Maybe some other classes, too?


packages/client/src/publish/messageChain.ts
```diff
@@ -0,0 +1,30 @@
+import {
+    MessageRef
+} from 'streamr-client-protocol'
```
harbu:
```suggestion
import { MessageRef } from 'streamr-client-protocol'
```

packages/client/src/utils/utils.ts
```diff
@@ -118,3 +118,10 @@ export class MaxSizedSet<T> {
         this.delegate.delete(value)
     }
 }
+
+// A unique internal identifier to some list of primitive values. Useful
+// e.g. as a map key or a cache key.
+// - we could also support other primitives or optional arguments if needed
```
harbu: This part a bit redundant imo. 
```suggestion
// - we could also support other primitives or optional arguments if needed
```

harbu (approved):

PR #881, opened 2022-10-11
client: use Logger from @streamr/utils 

harbu:
Replace old `debug`-based client logging implementation with `Logger` from package `@streamr/utils` to harmonize logging throughout the monorepo. 

### Key Changes
- Remove method `StreamrClient#enableDebugLogging`. (Is part of public API?)
- Add config `logLevel`  to client configuration for setting desired log level e.g. in browser setting.
- New injectable class `LoggerFactory` for creating "context-aware" instances of `Logger`.
- New error class `StreamrClientError` to replace `ContextError`.
- Remove `debug` from dependencies of client package.
- Remove interface `Context` to simplify class hierarchies. 
- Remove `rootContext`.
- Removed some logging that I felt was redundant. Feel free to refute.
- Changed many `debug` level logs to `trace` as they were indeed very low-level in nature.
- A few refactors throughout the files I touched. E.g. adding `readonly` modifiers to many places that were missing them.

### Further Considerations
- Consistent format and wording of log messages.
- We should really go through our log messages and figure out whether they serve us or not as well as adjust logging levels appropriately. May also want to add logging to some places.
- Make all error types in client package inherit from `StreamrClientError` so that it is apparent when an error was thrown from the client.
- Should we pass `clientId` to network level loggers so that we can cross-reference network level logs to the client that indirectly caused them?

## Checklist
- [x] Updated Changelog
- [x] Updated Documentation


packages/client/src/StreamrClientError.ts
```diff
@@ -0,0 +1,7 @@
+export class StreamrClientError extends Error {
+    constructor(message: string, readonly errorCode?: 'NO_STORAGE_NODES') {
```
harbu: Are codes a good way to identify errors? Or is `instanceof` preferrable? Research.

packages/client/CHANGELOG.md
```diff
@@ -32,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Change storage node assignment event handlers
   - replace method `registerStorageEventListeners(listener)` with `on('addToStorageNode', listener)` and `on('removeFromStorageNode', listener)`
   - replace method `unRegisterStorageEventListeners()` with `off('addToStorageNode', listener)` and `off('removeFromStorageNode', listener)`
+- Remove method `enableDebugLogging`.
```
teogeb: That was an internal method, no need include the change 

(If we include, should mention `disableDebugLogging`, too)

packages/client/README.md
```diff
@@ -742,8 +742,24 @@ Example broker config
...
+  // ... more options
+})
+```
+
+Alternatively, when running your application in Node.js, you can provide the
+logging level via the environment variable `LOG_LEVEL`, for example, by running
+your application as follows:
+
+```bash
+LOG_LEVEL=trace node your-app.js
```
teogeb, line 760: This environment variable is very generic and therefore may collide some other application/component. E.g. currently we control Broker and Tracker log level with this same variable (e.g. we set Broker's log level to `INFO`). Now we get also `StreamrClient` messages to that log. Intentional?

packages/client/src/Config.ts
```diff
@@ -76,8 +76,9 @@ export type MetricsConfig = {
  * @category Important
  */
 export type StrictStreamrClientConfig = {
-    /** Custom human-readable debug id for client. Used in logging. Unique id will be generated regardless. */
+    /** Custom human-readable debug id for client. Used in logging. */
     id?: string
```
teogeb: One idea about StrictStreamrClientConfig was that it'd contain values for all configuration properties. I.e. it merges user-defined config and defaults. And each component can just read values from this structure without applying any fallbacks for optional fields.

If we follow that idea, we'd put `id` and `logLevel` as required fields in this type. The optionality is applied via `StreamrClientConfig` interface below. We'd define the default value (`silent`?) in `STREAM_CLIENT_DEFAULTS`.

packages/client/src/StreamrClientError.ts
teogeb: I'd set `errorCode` as required argument and use `NO_STORAGE_NODES` value where we call this (as this value doesn't make sense when some other component creates a `StreamrClientError` instance)

packages/client/src/StreamrClientError.ts
```diff
@@ -0,0 +1,7 @@
...
+        super(message)
+        Error.captureStackTrace(this, StreamrClientError)
+        this.name = this.constructor.name
```
teogeb, line 6: Is `this.name` a convention of needed for some specific purpose?

packages/client/src/subscribe/MessageStream.ts
```diff
@@ -18,10 +16,8 @@ export class MessageStream<
     OutType extends StreamMessage<T> | unknown = InType
 > extends PushPipeline<InType, OutType> {
     /** @internal */
-    constructor(context: Context) {
+    constructor() {
```
teogeb: Could remove this constructor as it doesn't do anything `PushPipeline` constructor should be marked as internal)

packages/client/src/utils/Pipeline.ts
```diff
@@ -1,15 +1,13 @@
...
-import { ContextError, Context } from './Context'
 import * as G from './GeneratorUtils'
 import { ErrorSignal, Signal } from './Signal'
+import { StreamrClientError } from '../StreamrClientError'
 
 export type PipelineTransform<InType = any, OutType = any> = (src: AsyncGenerator<InType>) => AsyncGenerator<OutType>
-export type FinallyFn = ((err?: Error) => void | Promise<void>)
 
-class PipelineError extends ContextError {}
+class PipelineError extends StreamrClientError {}
```
teogeb: Maybe we don't need this `PipelineError` class (we also removed similar `PushBufferError`)

packages/utils/src/Logger.ts
```diff
@@ -32,13 +32,18 @@ export class Logger {
...
+        module: NodeJS.Module,
+        context?: string,
+        logLevel = process.env.LOG_LEVEL || 'info',
+        destinationStream?: { write(msg: string): void }
+    ) {
         const options: pino.LoggerOptions = {
             name: Logger.createName(module, context),
             enabled: !process.env.NOLOG,
-            level: process.env.LOG_LEVEL || 'info',
+            level: logLevel,
```
teogeb: Is info the default log level for `StreamrClient` in node and browser environments? Previously the client didn't log almost anything by default, do we want to change that?

packages/client/src/Container.ts
```diff
@@ -1,12 +1,14 @@
...
-import { counterId } from './utils/utils'
-import { Debug } from './utils/log'
-import { Context } from './utils/Context'
 import { ConfigInjectionToken, StrictStreamrClientConfig } from './Config'
 import { AuthenticationInjectionToken, createAuthentication } from './Authentication'
+import { counterId } from './utils/utils'
+import { uuid } from './utils/uuid'
+
+function generateClientId(): string {
+    return counterId(process.pid ? `${process.pid}` : `${uuid().slice(-4)}${uuid().slice(0, 4)}`)
```
teogeb: If uuid is just a random string we could just get first n chars from that? Or if it is not a random (but e.g. a time-based) uuid, should we change that to be a random id?

packages/client/README.md
harbu: Intentional, yes.

Controlling what gets logged at a component level would be good but even better would be to be able to do it on a module basis.

packages/client/src/StreamrClientError.ts
harbu: It affects the error message that gets printed. If you don't set the name it will be `Error: something went wrong` vs. `StreamrClientError: something went wrong`.

packages/client/src/StreamrClientError.ts
harbu: Fixed in fafafd2a5e2d44ddf16879b02f512d460a4be828

packages/client/CHANGELOG.md
harbu: Removed in 5ee2a7a4b37a8675cb6342de840dfbfacbb101f5

packages/client/src/Config.ts
harbu: Fixed in 0884b94cf181e2e4a272206e0abb781dd3b4a5c0

packages/client/src/subscribe/MessageStream.ts
harbu: Removed in fde9519d3ff3fdca563342963883e5c18ea74674

packages/client/src/utils/Pipeline.ts
harbu: Fixed in fafafd2a5e2d44ddf16879b02f512d460a4be828

packages/utils/src/Logger.ts
harbu: Previously it didn't log anything since `DEBUG` wasn't enabled. That being said, network package _would_ log at `info` level.

packages/client/src/Container.ts
harbu: Just changed to use `randomString` in 648e1fd356c24ec565b9d6e69f35418cb2f22bbd

PR #896, opened 2022-10-17
refactor(client): Simplify signature validation

teogeb:
Simplified signature validation in `StreamMessageValidator`:
- Removed obsolete checks about the existence of message.signature (the field is always available and checked in `StreamMessage:156`)
- Moved call to `assertSignatureIsValid` to validate that validation is done for all messages
- Reduce method visibility of `assertSignatureIsValid`

Also removed obsolete `Validator` constructor parameter.


packages/client/src/StreamMessageValidator.ts
```diff
@@ -92,6 +92,8 @@ export default class StreamMessageValidator {
             throw new ValidationError('Falsey argument passed to validate()!')
         }
 
+        await StreamMessageValidator.assertSignatureIsValid(streamMessage, this.verify)
```
harbu: Make into normal method since `this.verify` as argument.

harbu (approved):

PR #900, opened 2022-10-18
refactor(client): remove extraneous toLowerCase

harbu:
Remove extraneous uses of `toLowerCase` especially when coupled with `toEthereumAddress` and `EthereumAddress` types


teogeb (approved):
Looks good! Some additional usages which we could modify:
- `NetworkNodeFacade#getEthereumAddressFromNodeId` could return EthereumAddress? 
  - and then remove toLowerCase in `MetricsPublisher`
- `toLowerCase` is not needed in `createSignaturePayload` (`signature.ts`)
- `toLowerCase` is not needed in `NameDirectory.ts`
- could remove "ignores address case" test from `generateMnemonicFromAddress` test

PR #902, opened 2022-10-18
Move `Defer` to @streamr/utils

harbu:
- Move `Defer` to utils package
- Refactor `Defer` as a class, remove unused methods.
- Add unit tests for `Defer`.


packages/utils/src/Defer.ts
```diff
@@ -0,0 +1,68 @@
...
+            throw new Error('invariant violation: resolveFn was undefined')
+        }
+        if (localReject === undefined) {
+            throw new Error('invariant violation: rejectFn was undefined')
+        }
+        this.resolveFn = localResolve
+        this.rejectFn = localReject
+        this.ensureNoopCatchAttached = once(() => {
+            super.catch(() => {})
+        })
```
harbu, line 36: I couldn't figure out a way to get `this.catch(() => {})` or `super.catch(() => {})` to work directly in constructor without leading to `Maximum call stack size exceeded` error. Therefore the choice of "late initialization" pattern.

packages/utils/src/Defer.ts
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Deferred promise allowing external control of resolve / reject.
+ */
+export class Defer<T> extends Promise<T> {
+    private readonly resolveFn: ResolveFn<T>
+    private readonly rejectFn: RejectFn
+    private readonly ensureNoopCatchAttached: () => void
+    private settled = false
+
+    constructor(executor: (resolve: ResolveFn<T>, reject: RejectFn) => void = noopExecutor) {
```
harbu, line 18: This cannot be left as an zero parameter constructor else invoking `resolve` or `reject` will result in `TypeError: Promise resolve or reject function is not callable`. See more discussion here https://github.com/nodejs/node/issues/13678

teogeb (approved):

PR #903, opened 2022-10-18
feat(client): Remove 5 second metrics

teogeb:
Also removed some config duplication.


packages/client/src/MetricsPublisher.ts
```diff
@@ -4,31 +4,12 @@ import { DestroySignal } from './DestroySignal'
...
-        "streamId": "streamr.eth/metrics/nodes/firehose/day"
-    }
-]
+import { ConfigInjectionToken, MetricsPeriodConfig, STREAM_CLIENT_DEFAULTS, StrictStreamrClientConfig } from './Config'
 
 const getPeriodConfig = (rootConfig: StrictStreamrClientConfig): MetricsPeriodConfig[] => {
     switch (rootConfig.metrics) {
         case true:
-            return DEFAULT_PERIODS
+            return (STREAM_CLIENT_DEFAULTS.metrics as any).periods
```
harbu: Why `as any`?

harbu (changes requested):

packages/client/src/MetricsPublisher.ts
teogeb: The type of `STREAMR_CLIENT_DEFAULTS.metrics` is `MetricsConfig`, which could be:
- `{ periods: MetricsPeriodConfig[] }`
- or `boolean`

teogeb: Now using an explicit cast

harbu (approved):

PR #907, opened 2022-10-18
refactor(client): Simplify `MetricsPublisher` start

teogeb:
Guard multiple startups by using `pOnce`. Previously that was guarded by checking a field which was initialized in startup.


packages/client/test/unit/events.test.ts
```diff
@@ -138,4 +139,18 @@ describe('events', () => {
             expect(stop).toBeCalledTimes(2)
         })
     })
+
+    describe('emit', () => {
```
harbu: Was part of other PR?

harbu (approved):

packages/client/test/unit/events.test.ts
teogeb: True

PR #908, opened 2022-10-18
feat(client): Delay before publishing metrics sample

teogeb:
Random 0-30 second delay before each metrics sample publish. If each client would publish samples exactly same time, there would be unnecessary peaks in network traffic on report period end moments (end of each UTC minute, hour and day).

Added a config option for the delay (used currently in `Metrics.test.ts` end-to-end test). Refactored type annotations and config normalization so that `MetricsConfig` support multiple fields .


packages/client/src/MetricsPublisher.ts
```diff
@@ -79,8 +50,10 @@ export class MetricsPublisher {
     }
 
     private async publish(report: MetricsReport, streamId: string, partitionKey: string): Promise<void> {
+        await wait(Math.random() * this.config.maxPublishDelay)
```
harbu, line 51: Since currently there is no way to abort a `wait`, it is footnote worthy to mention that we may attempt to publish after `publisher` has already been destroyed. Not sure if there is any practical way right now to avoid that from happening. (The idea of having a "singleton" factory from which we create all `wait` and `setTimeout` objects in the client would allow us to destroy all such pending tasks on "destroy").

packages/client/src/MetricsPublisher.ts
```diff
@@ -79,8 +50,10 @@ export class MetricsPublisher {
...
         try {
             await this.publisher.publish(streamId, report, {
+                timestamp: report.period.end,
```
harbu, line 54: Hmm.. I'd keep the `timestamp = Date.now()` to more accurately represent when the message was actually published. The "event logical timestamp" is already included in the payload?

packages/client/test/unit/Config.test.ts
```diff
@@ -118,5 +118,44 @@ describe('Config', () => {
             expect((clientOverrides.network.trackers as SmartContractRecord[])[0]).not.toBe(trackers[0])
         })
         
+        describe('metrics', () => {
```
harbu, line 121: Good test! :+1:

harbu (approved):

packages/client/src/MetricsPublisher.ts
teogeb: Yes, Publisher will throw if we call NetworkNodeFacade#publishToNode if a client is already destroyed (`NetworkNodeFacade:212`). It would be good to have a `abortController` support in `wait` so that client can exit immediatelly when `StreamrClient#destroy` is called...

packages/client/src/MetricsPublisher.ts
teogeb: On the other hand, what is the point of having the feature which allows custom timestamp of a message if messages should always have `Date.now()` as the publish timestamp. I find it logical that timestamp represent the data (when it was collected), not when it was delivered.

PR #911, opened 2022-10-19
fix(dht, proto-rpc): clean script should not remove generated code

harbu:
Fix `npm run clean` in both dht and proto-rpc packages to not remove generated protobuf code. Tested by hand and `npm run clean && npm run build` without any changes to git files.


teogeb (approved):

harbu:
Accidentally close...

PR #917, opened 2022-10-20
feat(utils): abortable wait and `asAbortable`

harbu:
## Summary

Add support for `AbortController` in function `wait`. This allows a user to prematurely cancel `wait` leading to promise rejection. Add new exported function `asAbortable` with which one can turn any Promise into one that can be aborted.

## Changes
- Add `AbortController` support to `wait` in similar style as was done with `withTimeout` before.
- Add new function `asAbortable` which can be used to wrap any Promise into one that can be aborted. Resource cleanup, however, is not handled automatically and needs to be implemented by user of said function.
- Implement both `wait` and `withTimeout` using `asAbortable` to avoid repetition.

- The `setTimeout` implementation could be drastically simpler (see my 1st commit e9fc9b57da240e8d434012684ae6334d9584500d) by using [`timers/promises`](https://nodejs.org/api/timers.html#timerspromisessettimeoutdelay-value-options) introduced in Node.js v16. However, this is not supported by browsers.
- Add support for `customErrorContext` similar to `withTimeout`?

### DIFF
--- a/packages/utils/src/asAbortable.ts
+++ b/packages/utils/src/asAbortable.ts
@@ -1,0 +1,44 @@
+export class AbortError extends Error {
+    readonly code = 'AbortError'
+    constructor(customErrorContext?: string) {
+        super(customErrorContext === undefined
+            ? `aborted`
+            : `${customErrorContext} aborted`)
+        Error.captureStackTrace(this, AbortError)
+    }
+}
+
+/**
+ * Wraps a Promise into one that can be aborted with `AbortController`.
+ * Aborting causes the returned Promise to reject with `AbortError` unless
+ * the underlying promise itself has already resolved or rejected.
+ *
+ * Notice that it is the user's responsibility to implement any custom cleanup
+ * logic in a `finally` or `catch` block in case of resources that need to be
+ * freed up.
+ */
+export function asAbortable<T>(
+    promise: Promise<T>,
+    abortController?: AbortController,
+    customErrorContext?: string
+): Promise<T> {
+    if (abortController?.signal.aborted === true) {
+        return Promise.reject(new AbortError(customErrorContext))
+    }
+    let abortListener: () => void
+    return new Promise<T>((resolve, reject) => {
+        if (abortController?.signal !== undefined) {
+            abortListener = () => {
+                reject(new AbortError(customErrorContext))
+            }
+            // TODO remove the type casting when type definition for abortController has been updated to include addEventListener
+            (abortController.signal as any).addEventListener('abort', abortListener)
+        }
+        promise.then(resolve, reject)
+    }).finally(() => {
+        if (abortListener !== undefined) {
+            // TODO remove the type casting when type definition for abortController has been updated to include removeEventListener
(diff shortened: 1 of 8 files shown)


packages/utils/src/withTimeout.ts
```diff
@@ -8,16 +10,6 @@ export class TimeoutError extends Error {
...
-    readonly code = 'AbortError'
-    constructor(customErrorContext?: string) {
-        super(customErrorContext === undefined
-            ? `aborted`
-            : `${customErrorContext} aborted`)
-        Error.captureStackTrace(this, AbortError)
-    }
-}
-
 export const withTimeout = <T>(
```
teogeb, line 13: Would it be possible to use abortable wait in `withTimeout`? Now those method share very similar code.

teogeb (changes requested):

packages/utils/src/withTimeout.ts
harbu: "Fixed" in https://github.com/streamr-dev/network-monorepo/pull/917/commits/e0491b7e04048c8efef229ee7fff0c8dcb374efe, well at least less code repetition

packages/utils/src/asAbortable.ts
```diff
@@ -0,0 +1,44 @@
+export class AbortError extends Error {
+    readonly code = 'AbortError'
```
teogeb, line 2: It would maybe make sense to have this code as value `ABORT`
- as it is an error code, we already know it is an error
- use uppercase notation similarly e.g. in `StreamrClientError`

Or was it so that this follows some other convention, which we prefer?

packages/utils/src/withTimeout.ts
```diff
@@ -1,3 +1,5 @@
+import { asAbortable } from './asAbortable'
+
 export class TimeoutError extends Error {
     readonly code = 'TimeoutError'
```
teogeb, line 4: Similarly, this could be `TIMEOUT` if we want to change the codes

packages/utils/test/asAbortable.test.ts
```diff
@@ -0,0 +1,58 @@
...
+        const abortController = new AbortController()
+        setTimeout(() => {
+            abortController.abort()
+        }, TIME_UNIT)
+        const actual = asAbortable(
+            sleep(2 * TIME_UNIT),
+            abortController,
+            'customError'
+        )
+        return expect(actual).rejects.toEqual(new AbortError('customError'))
```
teogeb, line 33: Note that these kind of assertions don't actually test that we throw `AbortError` (see https://github.com/streamr-dev/network-monorepo/pull/925). These  just check that we throw a correct error message.

teogeb (approved):

packages/utils/src/asAbortable.ts
harbu: Will do in a separate PR to follow.

packages/utils/test/asAbortable.test.ts
harbu: Separate PR

PR #919, opened 2022-10-21
feat(utils): setAbortableTimeout and setAbortableInterval

harbu:
## Summary

Based on #917.

Add exported functions `setAbortableTimeout` and `setAbortableInterval` to `@streamr/utils`. They act similar to their JavaScript counterparts, `setTimeout` and `setInterval`, but with added support for `AbortSignal` to silently abort a timer.

- It seems to be standard practice in JavaScript libraries to accept an `AbortSignal` instead of an `AbortController` for operations that can be aborted. E.g. https://nodejs.org/api/stream.html#streamaddabortsignalsignal-stream, https://nodejs.org/api/stream.html#streamaddabortsignalsignal-stream, and https://developer.mozilla.org/en-US/docs/Web/API/fetch. We should change the rest of the exported functions that take in an `AbortController` to take in `AbortSignal` instead. 
- Lack of `removeEventListener` and `addEventListener` on `AbortSignal` type is cumbersome (lots of casting). 

### DIFF
--- a/packages/utils/src/abortableTimers.ts
+++ b/packages/utils/src/abortableTimers.ts
@@ -1,0 +1,39 @@
+/**
+ * setTimeout with AbortController support. Aborting will simply clear
+ * the timeout silently.
+ */
+export const setAbortableTimeout = createAbortableTimerFn(setTimeout, clearTimeout, true)
+
+/**
+ * setInterval with AbortController support. Aborting will simply clear
+ * the interval silently.
+ */
+export const setAbortableInterval = createAbortableTimerFn(setInterval, clearInterval, false)
+
+function createAbortableTimerFn(
+    setupTimerFn: (cb: () => void, ms?: number) => NodeJS.Timer,
+    clearFn: (ref: NodeJS.Timer) => void,
+    removeListenerAfterCb: boolean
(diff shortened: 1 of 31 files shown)


packages/utils/src/abortableTimers.ts
```diff
@@ -0,0 +1,41 @@
...
+ * setInterval with AbortSignal support. Aborting will simply clear
+ * the interval silently.
+ */
+export const setAbortableInterval = createAbortableTimerFn(setInterval, clearInterval, false)
+
+function createAbortableTimerFn(
+    setupTimerFn: (cb: () => void, ms?: number) => NodeJS.Timer,
+    clearFn: (ref: NodeJS.Timer) => void,
+    removeListenerOnCb: boolean
+): (cb: () => void, ms?: number, abortSignal?: AbortSignal) => NodeJS.Timer {
```
harbu: Let's make return type `void`. If user wants to "clear" timeout / interval they need to pass a composite abort controller (e.g. https://github.com/tjmehta/composite-abort-controller). 

Also make `ms` mandatory. Also make `abortSignal` mandatory. If you just want normal timeout / interval without aborting use the normal node.js built-in variants.

harbu: Fixed in 1299520 and 75908a3

teogeb (approved):

PR #921, opened 2022-10-21
test(client): Fix `no storage assigned` resend test

teogeb:
## Summary

Fixed `"no historical messages available / resendSubscribe / no storage assigned"` test cases which was earlier refactored in this PR: https://github.com/streamr-dev/network-monorepo/pull/795. In that refactoring the test was incorrectly modified to use a stream which has a storage node assigned. Now that test cases creates a separate stream (and doesn't assign it to a storage node).

Added also assertion that `onError` is not called.

## Open questions

Why we have a special handling for `resendSubscribe` (`client.subscribe({..., resend: {}})`) to not to throw if there is no storage node assigned? (`ResendSubscription:60`). Maybe it would make more sense that it would throw, similarly as `client.resend` does?
- if we want to unify the behavior we could remove the `"no historical messages available / resendSubscribe / no storage assigned"` test case altogether

## Other changes

- Improved test naming
- Removed a test case from `"historical messages available"` section. The test case simulated a situation where storage node was not assigned. That didn't didn't make any sense: if there is no storage node, there can't be "historical messages available".
- Added local `publishTestMessages` utility method so that it can publish to a non-default stream.

It is not clear whether we should assert subscription count at the end of each test case. The subscription count is 0 because we iterate the received messages (e.g. by `collect` or by `for await` loop). That triggers the unsubscribe. Therefore those subscription counts asserts our test logic instead of production functionality. Related PR: https://github.com/streamr-dev/network-monorepo/pull/924.

### DIFF
--- a/packages/client/test/integration/Resends2.test.ts
+++ b/packages/client/test/integration/Resends2.test.ts
@@ -1,7 +1,7 @@
 import 'reflect-metadata'
 import fs from 'fs'
 import path from 'path'
-import { StreamMessage } from 'streamr-client-protocol'
+import { StreamID, StreamMessage } from 'streamr-client-protocol'
 import { fastWallet } from 'streamr-test-utils'
 import { createTestStream } from '../test-utils/utils'
 import { getPublishTestStreamMessages, getWaitForStorage, Msg } from '../test-utils/publish'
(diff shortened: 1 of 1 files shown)


harbu:
> Why we have a special handling for resendSubscribe (client.subscribe({..., resend: {}})) to not to throw if there is no storage node assigned? (ResendSubscription:60). Maybe it would make more sense that it would throw, similarly as client.resend does?

Discussed this a bit with @teogeb . On one hand, the user has requested a _resend_ and a subscription, so it would make sense that an error is thrown if the user's request cannot be met due to stream not being assigned to a storage node. On the other hand, there is presumably some amounts of existing code and usage that relies on resend + subscription even though the underlying stream has not been assigned to a storage node. Therefore I think a `logger.warn` would be appropriate here, at least for v7.

harbu (approved):

PR #925, opened 2022-10-24
test(client): Test client error code

teogeb:
## Summary

Added assertions about `StreamrClientError`. Now we always check the error code, and optionally also the error message.

Implemented custom Jest matcher `toThrowStreamError` so that we can easily check these multiple aspects of thrown error.

## Other changes

- Fixed error message for `NO_STORAGE_NODE` error: it uses `streamId` instead of `streamPartId` as assignments are stream specific.
- Renamed `CLIENT_IS_DESTROYED` code to `CLIENT_DESTROYED`. This is not a breaking changes as v6 didn't have any error codes (and we haven't documented the error codes in changelog or in readme).
- Removed invalid test in `Resends2.test.ts`, which seemed to check streamId validity but actually failed for `NO_STORAGE_NODE` (and therefore was a duplicate test).
- Removed duplicate test in `GapFill.test.ts` which tested that client.resend throws `NO_STORAGE_NODE` error (was already tested in `Resends2.test.ts`)

### DIFF
--- a/packages/client/test/test-utils/customMatchers.ts
+++ b/packages/client/test/test-utils/customMatchers.ts
@@ -2,7 +2,7 @@ import { expect } from '@jest/globals'
 import type { MatcherState } from 'expect'
 import { printExpected, printReceived } from 'jest-matcher-utils'
 import { isFunction } from 'lodash'
-import { StreamrClientError } from './../../src/StreamrClientError'
+import { StreamrClientError, StreamrClientErrorCode } from './../../src/StreamrClientError'
 
 // we could ES2015 module syntax (https://jestjs.io/docs/expect#expectextendmatchers),
 // but the IDE doesn't find custom matchers if we do that
@@ -10,7 +10,7 @@ declare global {
     // eslint-disable-next-line @typescript-eslint/no-namespace
     namespace jest {
         interface Matchers<R> {
-            toThrowStreamError(expectedError: Partial<StreamrClientError>): R
+            toThrowStreamError(expectedError: { code: StreamrClientErrorCode, message?: string }): R
         }
     }
 }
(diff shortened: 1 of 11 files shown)


packages/client/test/test-utils/customMatchers.ts
```diff
@@ -0,0 +1,61 @@
...
+        pass: false,
+        message: () => `${description}\nExpected: ${printExpected(expected)}\nReceived: ${printReceived(actual)}`
+    }
+}
+
+const toThrowStreamError = function(
+    this: MatcherState,
+    actual: unknown, // should be (() => StreamrClientError) | StreamrClientError
+    expectedError: { code: StreamrClientErrorCode, message?: string }
+) {
```
harbu: Return type

packages/client/test/test-utils/customMatchers.ts
```diff
@@ -0,0 +1,61 @@
...
+    let actualError
+    if (isFunction(actual)) {
+        try {
+            actual()
+            return {
+                pass: false,
+                message: () => `Function didn't throw`
```
harbu:
```suggestion
                message: () => "Function didn't throw"
```

packages/client/test/test-utils/customMatchers.ts
```diff
@@ -0,0 +1,61 @@
...
+            }
+        } catch (e) {
+            actualError = e
+        }
+    } else {
+        actualError = actual
+    }
+    if (!(actualError instanceof StreamrClientError)) {
```
harbu, line 55: Line break before to separate logically.
```suggestion
    if (!(actualError instanceof StreamrClientError)) {
```

harbu (approved):

PR #926, opened 2022-10-24
refactor(client): Remove `SubscriptionSession` dependency from `Subscription`

teogeb:
## Summary

There was circular dependency between `Subscription` and `SubscriptionSession`. Removed `Subscriptions`'s dependency from `SubscriptionSession`. Removed also internal `Subscription#waitForNeighbours` method which was only used in one test.

## Other changes

- provided direct dependencies for `ResendSubscription` (earlier passed a `DependencyContainer` as an argument)
- removed `StreamrClient#container` private field, which is no longer needed
- removed `@internal` annotations from some private fields

### DIFF
--- a/packages/client/src/subscribe/Subscriber.ts
+++ b/packages/client/src/subscribe/Subscriber.ts
@@ -118,10 +118,7 @@ export class Subscriber {
     }
 
     private async add<T>(streamPartId: StreamPartID): Promise<Subscription<T>> {
-        const subSession = this.getOrCreateSubscriptionSession<T>(streamPartId)
-
-        // create subscription
-        const sub = new Subscription<T>(subSession.streamPartId, this.loggerFactory)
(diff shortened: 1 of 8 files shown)


packages/client/src/subscribe/Subscriber.ts
```diff
@@ -121,7 +121,7 @@ export class Subscriber {
         const subSession = this.getOrCreateSubscriptionSession<T>(streamPartId)
 
         // create subscription
-        const sub = new Subscription<T>(subSession, this.loggerFactory)
+        const sub = new Subscription<T>(subSession.streamPartId, this.loggerFactory)
```
harbu: Could be simplified.
```suggestion
        const sub = new Subscription<T>(streamPartId, this.loggerFactory)
```

harbu (approved):

PR #930, opened 2022-10-25
feat(client): Log warning in resend subscribe if no storage assigned

teogeb:
If storage node is not assigned to a stream and we call `client.resendSubscribe`, it logs a warning. Discussed here: https://github.com/streamr-dev/network-monorepo/pull/921#issuecomment-1288784070

## Testing

Added `FakeLogger` test environment class. It stores log entries to an array, and in test we can check that an expected entry exists in that array.

Created a type annotation for `LogLevel` and used that in `FakeLogger` and `StrictStreamrClientConfig`.


packages/client/test/unit/ResendSubscription.test.ts
```diff
@@ -0,0 +1,46 @@
...
+import { ResendSubscription } from "../../src/subscribe/ResendSubscription"
+import { fromArray } from '../../src/utils/GeneratorUtils'
+import { collect } from '../../src/utils/iterators'
+import { STREAM_CLIENT_DEFAULTS } from './../../src/Config'
+import { StreamrClientError } from './../../src/StreamrClientError'
+import { createMockMessage } from './../test-utils/utils'
+
+const STREAM_PART_ID = StreamPartIDUtils.parse('streamId#0')
+
+describe('ResendSubscription', () => {
```
harbu: Check logging in Resends2.test.ts with integration test as discussed.

teogeb: Fixed

packages/client/test/test-utils/fake/FakeLogger.ts
```diff
@@ -0,0 +1,64 @@
...
+    message: string
+    level: LogLevel
+}
+
+const LOG_LINE_PREFIX_LENGTH = 56 // Logger prefixes each line with level, timestamp and context
+
+export class FakeLogger implements Omit<Logger, 'getFinalLogger'> {
+
+    private entries: LogEntry[] = []
+    private format: (pattern: string, args: any[], cb: (message: string) => void) => void
```
harbu:
```suggestion
    private readonly entries: LogEntry[] = []
    private readonly format: (pattern: string, args: any[], cb: (message: string) => void) => void
```

packages/client/test/test-utils/fake/FakeLogger.ts
```diff
@@ -0,0 +1,64 @@
...
+        this.format = (pattern: string, args: any[], cb: (message: string) => void) => {
+            const logger = new Logger(module, undefined, 'trace', {
+                write: (line: string) => {
+                    // eslint-disable-next-line
+                    const withoutColors = line.replace(/\x1b\[[0-9]+m/gi, '')
+                    const withoutPrefix = withoutColors.substring(LOG_LINE_PREFIX_LENGTH)
+                    const withoutLinefeed = withoutPrefix.substring(0, withoutPrefix.length - 1)
+                    cb(withoutLinefeed)
+                }
+            })
```
harbu, line 26: I think we should create a simple unit test for `FakeLogger` to ensure that it strips away this info correctly. To guard against situations where we would change the format of the real `Logger` but forget to account for the change here.

packages/client/test/test-utils/fake/FakeEnvironment.ts
```diff
@@ -79,6 +87,10 @@ export class FakeEnvironment {
         return this.network
     }
 
+    getLogger(): FakeLogger {
+        return this.logger
+    }
+
```
harbu, line 93: Should we provide a direct access to log entries here to make use easier?
```suggestion
    getLogEntries(): LogEntry[] {
        return this.logger.getEntries()
    }

```

packages/client/test/test-utils/fake/FakeLogger.ts
teogeb: Fixed

packages/client/test/test-utils/fake/FakeLogger.ts
teogeb: Added a unit test

packages/client/test/test-utils/fake/FakeEnvironment.ts
teogeb: I'd keep the current style where each `FakeEnvironment` component is exposed as a getter, and the actual query/command methods are in that component. It is quite straightforward to assert `environment.getLogger().getEntries()` anyway.

This way we follow the "do only one thing, and do it well" principle in `FakeEnvironment`: the environment just manages the component, doesn't operate on them (and doesn't need to know what operations are available).

harbu (approved):

PR #934, opened 2022-10-26
feat(utils): add function `composeAbortSignals`

harbu:
## Summary 
Add new utility function `composeAbortSignals`  to `@streamr/utils`. It can be used to combine multiple instances of `AbortSignal` into one composite `AbortSignal` which acts as a "OR gate", i.e., if any of the supplied signals aborts the composite aborts. 

An imaginary example use case:
```ts
class Database {
    private applicationQuitSignal: AbortSignal
    // rest of fields
    
    doQuery(userRequestedCancelSignal: AbortSignal): ResultSet {
        return dbDriver.query(
            'SELECT * FROM users',
            composeAbortSignals(applicationQuitSignal, userRequestedCancelSignal) // query canceled either when app quits or user cancels request
        )
    }
}
```


packages/utils/src/composeAbortSignals.ts
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Compose a single AbortSignal from multiple AbortSignals with "OR" logic.
+ */
+export function composeAbortSignals(...signals: AbortSignal[]): AbortSignal {
+    if (signals.length === 0) {
+        throw new Error('must provide at least one AbortSignal')
+    }
+
+    const preAbortedSignal = findFirstPreAbortedSignal(signals)
```
teogeb:
```suggestion
    const preAbortedSignal = signals.find((s) => (s.aborted === true))
```

We can then remove the `findFirstPreAbortedSignal` method

packages/utils/test/composeAbortSignals.test.ts
```diff
@@ -0,0 +1,109 @@
...
+            const listener = jest.fn()
+            composedSignal.addEventListener('abort', listener)
+            controllers[3].abort()
+            controllers[6].abort()
+            expect(listener).toHaveBeenCalledTimes(1)
+        })
+
+        it('onabort if set is invoked when a signal aborts', () => {
+            const listener = jest.fn()
+            ;(composedSignal as any).onabort = listener
```
teogeb: Comment about why casting is needed? (also line 54)

Maybe some rewording for the test name?

packages/utils/test/composeAbortSignals.test.ts
```diff
@@ -0,0 +1,109 @@
...
+            expect(listener).toHaveBeenCalledTimes(0)
+            controllers[0].abort()
+            expect(listener).toHaveBeenCalledTimes(0)
+        })
+    })
+
+    it('throws error if composing empty list', () => {
+        expect(() => {
+            composeAbortSignals()
+        }).toThrowError(new Error('must provide at least one AbortSignal'))
```
teogeb:
```suggestion
        }).toThrow('must provide at least one AbortSignal')
```

As the `jest` assertion doesn't check the error class anyway.

(same change for line `107`)

packages/utils/src/composeAbortSignals.ts
harbu: Fixed 👍

packages/utils/test/composeAbortSignals.test.ts
harbu: First part fixed, I think name is fine?

packages/utils/test/composeAbortSignals.test.ts
harbu: Fixed 👍

packages/utils/test/composeAbortSignals.test.ts
teogeb: Does it mean something like "if `onbort` listener is set, it is invoked when a signal aborts?"

teogeb (approved):

PR #936, opened 2022-10-26
chore(network): drastically downsize propagation buffer

harbu:
## Summary

Reduce the default maximum active propagation tasks parameter in network node, i.e. `Propagation#DEFAULT_MAX_MESSAGES`, from 10000 to 30. Having the "propagation buffer" be too large can cause extremely busty behavior in large topologies leading to message delivery failure and high resource consumption.

## Limitations and Future improvements

- Temporarily had to adjust broker test `multiple-publisher-plugins.test.ts` to wait a bit after publishing 1st message (so that connections get opened). Otherwise it would publish 120 messages at once, leading to most of them dropping off the buffer. This should be addressed in NET-919.

- Consider adding a rate limit to how fast a buffer gets emptied. This could allow for a larger buffer whilst keeping the bursty-ness in check.


juslesan (approved):
LGTM,

In the future the size of the messages in the buffer should also be considered!

PR #938, opened 2022-10-27
refactor(utils): remove "as any" casts and undefined checks

harbu:
## Changes

- With the new version of Node.js type definitions `@types/nodes` introduced in 074d31611c94bd69dfe4b5fecd072b80374cc69c,  interface `AbortSignal` now properly inherits from `EventTarget` removing the need for `as any` casts to add and remove event listeners.
- Remove undefined checks for parameter `abortSignal` in `setAbortableTimeout` and `setAbortableInterval`. No need as the parameter is not optional.

### DIFF
--- a/packages/utils/src/abortableTimers.ts
+++ b/packages/utils/src/abortableTimers.ts
@@ -16,23 +16,16 @@ function createAbortableTimerFn(
     removeListenerOnCb: boolean
 ): (cb: () => void, ms: number, abortSignal: AbortSignal) => void {
     return (callback, ms, abortSignal): void => {
-        if (abortSignal?.aborted) {
+        if (abortSignal.aborted) {
             return
         }
-        let abortListener: () => void
-        if (abortSignal !== undefined) {
-            abortListener = () => {
-                clearFn(timeoutRef)
-                // TODO remove the type casting when type definition for abortController has been updated to include addEventListener
-                ;(abortSignal as any).removeEventListener('abort', abortListener)
-            }
-            // TODO remove the type casting when type definition for abortController has been updated to include addEventListener
-            (abortSignal as any).addEventListener('abort', abortListener)
+        const abortListener = () => {
+            clearFn(timeoutRef)
         }
+        abortSignal.addEventListener('abort', abortListener, { once: true })
         const timeoutRef = setupTimerFn(() => {
-            if (abortListener !== undefined && removeListenerOnCb) {
-                // TODO remove the type casting when type definition for abortController has been updated to include removeEventListener
-                (abortSignal as any).removeEventListener('abort', abortListener)
+            if (removeListenerOnCb) {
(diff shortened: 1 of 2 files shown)


packages/utils/src/abortableTimers.ts
```diff
@@ -16,23 +16,16 @@ function createAbortableTimerFn(
...
-                clearFn(timeoutRef)
-                // TODO remove the type casting when type definition for abortController has been updated to include addEventListener
-                ;(abortSignal as any).removeEventListener('abort', abortListener)
-            }
-            // TODO remove the type casting when type definition for abortController has been updated to include addEventListener
-            (abortSignal as any).addEventListener('abort', abortListener)
+        const abortListener = () => {
+            clearFn(timeoutRef)
         }
+        abortSignal.addEventListener('abort', abortListener, { once: true })
```
teogeb, line 25: Maybe worth mentioning in PR description that we now use `{ once: true }` (potentially less memory leaks or otherwise affects functionality).

packages/utils/src/asAbortable.ts
```diff
@@ -22,23 +22,20 @@ export function asAbortable<T>(
...
-        if (abortSignal !== undefined) {
-            abortListener = () => {
-                reject(new AbortError(customErrorContext))
-            }
-            // TODO remove the type casting when type definition for abortController has been updated to include addEventListener
-            (abortSignal as any).addEventListener('abort', abortListener)
+        abortListener = () => {
+            reject(new AbortError(customErrorContext))
         }
+        abortSignal.addEventListener('abort', abortListener)
```
teogeb, line 36: Should we use `{ once: true }` here, too?

packages/utils/src/abortableTimers.ts
harbu: The functionality should be the same. Before this change we did the cleanup inside `abortListener` by-hand, i.e.
```ts
abortListener = () => {
    clearFn(timeoutRef)
    // TODO remove the type casting when type definition for abortController has been updated to include addEventListener
    ;(abortSignal as any).removeEventListener('abort', abortListener)
}
```

Now we just added `{ once: true }` to let the library do the cleaning on our behalf therefore simplifying our code.

packages/utils/src/asAbortable.ts
harbu: In the `finally` block, the event listener gets removed in any case. We could add `{ once: true }` but that would mean that in the case of an abort signal occurring (before `promise` resolves), the `abortListener` would already been removed when we arrive at the `finally` block. Removing a non-existing listener is no-op so it probably wouldn't really matter. So a matter of taste I guess. It would be a bit more "playing it safe" to add the `{ once: true }` but OTOH, it's not really needed in this case.

teogeb (approved):

PR #942, opened 2022-10-27
test(utils): Browser tests

teogeb:
## Summary

Run all tests in `utils` package also in a browser environment.

## Related changes

- Fixed import of wait in `Metric.test.ts` (it imported it from the `@streamr/utils` package)
- Changed test names to be strings instead of class references. When class references were used, the test name was not shown in karma output (it printed the test name as `null`)

## Browser test warnings

There are this kind of warning messages in karma output log:
```
WARNING in ../../node_modules/jest-util/build/requireOrImportModule.js 53:27-44
Critical dependency: the request of a dependency is an expression
 @ ../../node_modules/jest-util/build/index.js 162:2-36
 @ ../../node_modules/expect/build/asymmetricMatchers.js 27:16-36
 @ ../../node_modules/expect/build/index.js 18:26-57
 @ ./karma-setup.js 9:15-41
```

It is not clear what is the root cause of those warnings. All tests seem to run ok anyway. Similar warnings are also shown when browser tests are run for the  `network` package.

## Future improvements

- One of the test cases in `Logger.test.ts` is disabled when it is run in browser environment. If possible, we should enable that test case? 
  - The test case has a comment which tells why it has been disabled. 
  - It was most likely disabled already when `Logger` class was in the network package. 
- NET-920: The `waitForCondition` test utility method is temporarily inlined. If we import it from `streamr-test-utils` there seem to be some dependency error related to the fact that `streamr-test-utils` depends on this utils package. The error message is: `"Can't resolve '@streamr/utils' in '/Users/teogeb/workspace/streamr/network-monorepo/packages/test-utils/dist/src'"`
- Maybe part of NET-921: If we don't need to use features from Node environment, we could maybe use headless Chrome browser instead of Electron.

### DIFF
--- a/packages/utils/test/Metric.test.ts
+++ b/packages/utils/test/Metric.test.ts
@@ -1,6 +1,49 @@
-import { waitForCondition } from 'streamr-test-utils'
 import { wait } from '../src/wait'
 import { AverageMetric, CountMetric, LevelMetric, MetricsContext, MetricsReport, RateMetric } from '../src/Metric'
+
+// TODO import this from streamr-test-utils if possible
+// in browser tests we get error: "Can't resolve '@streamr/utils' in '/Users/teogeb/workspace/streamr/network-monorepo/packages/test-utils/dist/src'"
+const waitForCondition = async (
+    conditionFn: () => (boolean | Promise<boolean>),
+    timeout = 5000,
+    retryInterval = 100,
+    onTimeoutContext?: () => string
+): Promise<void> => {
+    // create error beforehand to capture more usable stack
+    const err = new Error(`waitForCondition: timed out before "${conditionFn.toString()}" became true`)
+    return new Promise((resolve, reject) => {
+        let poller: NodeJS.Timeout | undefined = undefined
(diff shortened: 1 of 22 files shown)


packages/utils/test/Metric.test.ts
```diff
@@ -1,7 +1,50 @@
-import { waitForCondition } from 'streamr-test-utils'
-import { wait } from '@streamr/utils'
+import { wait } from '../src/wait'
 import { AverageMetric, CountMetric, LevelMetric, MetricsContext, MetricsReport, RateMetric } from '../src/Metric'
 
+// TODO import this from streamr-test-utils if possible
+// in browser tests we get error: "Can't resolve '@streamr/utils' in '/Users/teogeb/workspace/streamr/network-monorepo/packages/test-utils/dist/src'"
```
harbu: Make reference to NET-920.

harbu (commented):
Figure out if `karma-jest` can be removed?

harbu (approved):

PR #948, opened 2022-10-31
refactor(client): `HttpUtil` returns `AsyncIterable`

teogeb:
Refactored `HttpUtil#fetchHttpStream` to return `AsyncIterable` instead of `Readable`. The method operates internally with `Readable` stream, but it is easier for users of that component to receive the data as `AsyncIterable` (similarly as other components do).

Added unit test for `HttpUtil#fetchHttpStream`

Simplified `Resend#fetchStream` to use new helper utility to calculate the message count (for logging).

Removed obsolete `opts` parameter and `startTime` debug data from `HttpUtil#fetchHttpStream`.


packages/client/src/HttpUtil.ts
```diff
@@ -80,20 +80,18 @@ export class HttpUtil {
         this.logger = loggerFactory.createLogger(module)
     }
 
-    async fetchHttpStream(
+    async* fetchHttpStream<T>(
```
harbu, line 83: I would leave out the generic since we can't make guarantees about what type gets responded from the HTTP server.

packages/client/src/HttpUtil.ts
```diff
@@ -80,20 +80,18 @@ export class HttpUtil {
...
         const response = await fetchResponse(url, this.logger, {
-            signal: abortController.signal,
-            ...opts,
+            signal: abortController.signal
         })
         if (!response.body) {
             throw new Error('No Response Body')
         }
 
+        let stream: Readable | undefined
```
harbu, line 94: This variable is shadowed by `const stream` on line 99. Should remove the `const` from that line.

harbu (approved):

packages/client/src/HttpUtil.ts
teogeb: 👍 Fixed

packages/client/src/HttpUtil.ts
teogeb: Let's remove the `StreamMessage` typing annotations in a separate PR

PR #952, opened 2022-11-01
feat(client): Allow non authenticated publish

teogeb:
Allow client to publish messages even if it is not authenticated. If there is no authentication in given config, random private key is generated and messages are published using that private key.

- Removed `isUnauthenticated` method and calls to it as it is no longer relevant whether client has been authenticated explicitly (user-defined private key or ethereum), or implicitly (randomly generated private key)
- Refactored auth config to be optional: if auth is given, it must be either a private key authentication or ethereum authentication
  - not an empty object anymore
  - not `{ unauthenticated: true }` or `{ sessionToken: string }` anymore
- Removed two obsolete interfaces: `AuthenticatedAuthConfig` and `UnauthenticatedAuthConfig`


packages/client/CHANGELOG.md
```diff
@@ -58,6 +58,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Remove client configuration option `client.network.name`
 - Remove client configuration option `client.debug`
 - Remove (non-functional) client configuration options `autoConnect`, `autoDisconnect`, and `maxRetries`
+- Remove `AuthenticatedAuthConfig` and `UnauthenticatedAuthConfig` interfaces
```
harbu, line 63: Add entry somewhere that `auth: {}` is no longer allowed.

harbu (approved):

packages/client/CHANGELOG.md
teogeb: Fixed

PR #955, opened 2022-11-01
test(network): `IceServer` interface and fix browser turn configuration 

harbu:
## Summary

Take in STUN and TURN server configuration as objects instead of strings, and also fix issue in which TURN details were not passed correctly to `RTCPeerConnection` in browser environment.

Example:
```ts
// OLD FORMAT
stunUrls = ['turn://username@passwordturn.server.com:31313']
```

```ts
// NEW FORMAT
iceServers = [{
    url: 'turn://passwordturn.server.com',
    port: 31313,
    username: 'username',
    password: 'password'
}]
```

### Future improvements
- Include `port` in `url`. Find a good parser utility for parsing the different parts of the URI(?)?

## Changes
- (network) Introduce `IceServer` interface 
- (network) Pass down `iceServers: ReadonlyArray<IceServer>` instead of `stunUrls: string[]` down the code stack
- (network) Add test that verifies that setting a TURN server in browser does not lead to error
- (network) Export constant `STREAMR_ICE_SERVERS` that gets used by client when creating a network node
- (network) Export constant `GOOGLE_STUN_SERVER` used by some tests


packages/network/src/connection/webrtc/WebRtcConnection.ts
```diff
@@ -7,11 +7,18 @@ import { MessageQueue, QueueItem } from '../MessageQueue'
 import { NameDirectory } from '../../NameDirectory'
 import crypto from 'crypto'
 
+export interface IceServer {
+    url: string
+    port: number
```
teogeb, line 12: Would it make sense that we'd remove port field? It is a component of a URL string, like e.g. domain and path are.

(Yes, username and credentials are also officially part of URL definition... But they don't specify the _location_ of the server)

packages/network/src/constants.ts
```diff
@@ -1,2 +1,22 @@
...
+
 export const COUNTER_UNSUBSCRIBE = -1
 export const DEFAULT_MAX_NEIGHBOR_COUNT = 4
+
+export const GOOGLE_STUN_SERVER: IceServer = {
+    url: 'stun:stun.l.google.com',
+    port: 19302
+}
+
+export const PRODUCTION_ICE_SERVERS: ReadonlyArray<IceServer> = Object.freeze([
```
teogeb: Could maybe rename this to `STREAMR_ICE_SERVERS`? Would be in sync with the `GOOGLE_... ` name.

packages/network/src/connection/webrtc/WebRtcConnection.ts
```diff
@@ -7,11 +7,18 @@ import { MessageQueue, QueueItem } from '../MessageQueue'
...
+    username?: string
+    credential?: string
```
teogeb: Do we want to support multiple credential types (by adding an optional `credentialType` field, https://developer.mozilla.org/en-US/docs/Web/API/RTCIceServer)?

Maybe it is best to use username+password authentication as we currently do. But then we could consider naming `credential` to `password`.

harbu: Yeah make sense

packages/network/src/connection/webrtc/WebRtcConnection.ts
harbu: Yeah that would work. We just need a reliable way of extracting port from url as node-datachannel library requires this.

teogeb (approved):

PR #962, opened 2022-11-03
feat(client)!: Stream metadata public API

teogeb:
## Summary

This is a breaking change.

Stream metadata is no longer stored as fields of `Stream` object. Therefore e.g. partition could and other metadata is now be queried via new  `stream.getMetadata()` method. Previously it was possible to access metadata directly.

```
const description = stream.getMetadata().description
// was: const description = stream.description
```

## Related changes

Removed `Stream#toObject` method. The new `stream.getMetadata()` implements almost the same logig:
- the `Stream#toObject` contained also the stream
- it used reflection to collect fields from `Stream` object

## Partitions

As partition count is also part of stream metadata, it is queried via the same method. We could have a separate method for this if needed. But it is quite straightforward to access it via `stream.getMetadata().partitions`.

The partitions property of `StreamMetadata` is a a required property. In StreamProperties object it was optional, but in `Stream` an it was initialized to `1` if no value was provided.

## Refactoring

- Modified `StreamRegistry#createStream` and `StreamRegistry#updateStream` to operate on `StreamID` and `StreamMetadata` instead of a raw object. Now The parsing is done in `StreamrClient`. Also moved `getOrCreateStream` from `StreamRegistry` to `StreamrClient` as it operates on raw objects.
- `StreamMessageValidator` used an interface called `StreamMetadata`, but now it uses just a simple method which returns the stream partition count.
- Renamed `Stream#parsePropertiesFromMetadata` to `Stream#parseMetadata` and inlined `StreamRegistry#formMetadata` (internal methods)

## Future improvements

Added several TODOs, which we should implement either in this PR or later:

- Currently we collect all properties from contract's metadata JSON to Stream metadata. There is no picking, e.g. currently the metadata object can have also other values than those listed in `StreamMetadata` interface.

- Currently we don't validate the metadata properties in any way. We should maybe check that all required parameters are available and have valid values: in practice we'd validate that the metadata contains `partitions` field, and it has value between `0` and `99`.

- In `StreamrClient#updateStream` we don't currently require any metadata fields. It is too easy to write invalid metadata (e.g. a JSON which doesn't contain the partition count) by calling that method. The method doesn't do any merging with the existing metadata field and most likely the contract doesn't do either. The logic of that method is different from e.g. `Stream#update` which does merge the current metadata with the existing metadata. It is also different from e.g. `createStream` which defaults partition count to `1` if none is set.
- Maybe it would make sense that we'd reimplement `StreamrClient#updateStream` just as a delegate to `Stream#update`? (or that `Stream#update` would delegate `StreamrClient#updateStream`)
- If we do that we could also change the signature so that the methods gets two parameters: `streamId` and `metadata`

- Config field of metadata is currently defaulted to empty array in `Stream` constructor. Therefore if is kind of "required property". It is a property used in frontend, but not part of e.g. the Streamr protocol. Maybe we should not set that to default value and frontend could just fallback to empty array if there is no config definition available in the metadata. (Alternatively we could annotate the config property as required parameter, but that is maybe not an optimal solution.)

### DIFF
--- a/packages/client/CHANGELOG.md
+++ b/packages/client/CHANGELOG.md
@@ -25,6 +25,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Enforce concurrency limit for smart contract calls (per contract, configurable with `maxConcurrentContractCalls` config option)
 - Enforce presence of message signatures
   - all non-signed messages received by client are simply ignored
+- Stream metadata (e.g. `partitions`) is no longer stored as fields of `Stream` object
+  - query the data by calling `stream.getMetadata()`
 - Method `stream.update()` parameter `props` is no longer optional
 - Rename method `getStorageNodesOf()` to `getStorageNodes()`
 - Rename method `getStoredStreamsOf()` to `getStoredStreams()`
@@ -52,6 +54,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - use `destroy()` instead
 - Remove method `unsubscribeAll()`
   - use `unsubscribe()` without arguments to same effect
+- Remove method `stream.toObject()`
+  - use `stream.getMetadata()` to get metadata (doesn't contain stream id)
 - Remove properties `subscription.onMessage`, `onStart`, and `onError`
   - use `subscription.on('error', cb)` to add an error listener
 - Remove configuration option `groupKeys`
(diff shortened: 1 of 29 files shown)


packages/client/CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Enforce concurrency limit for smart contract calls (per contract, configurable with `maxConcurrentContractCalls` config option)
 - Enforce presence of message signatures
   - all non-signed messages received by client are simply ignored
+- Stream metadata (e.g. `partitions`) is no longer stored as fields of `Stream` object
+  - query the data by calling `stream.getMetadata()`
```
```suggestion
- Stream metadata now accessed through `stream.getMetadata()`
  - e.g. usages of `stream.partitions` has changed to `stream.getMetadata().partitions`
```

harbu (approved):

PR #965, opened 2022-11-04
feat(client): Paginated stored streams queries

teogeb:
`StreamrClient#getStoredStreams` uses paginated GraphQL query. Previously it queried only one page of results, and therefore the result set was limited to 1000 streams.

## Other changes

- Using orderBy in all paginated GraphQL queries (it is the default but maybe good to have anyway for clarity)
- Created new  `GraphQLQuery` which contains a query string and optionally some variables for the query. Previously we needed to convert the query to JSON before passing it to `sendQuery`/`fetchPaginatedQuery`.

## Future improvements

- We could `blockNumber` collecting functionality from `getStoredStreams` to `GraphQLClient` as the logic is not specific to that query.
- It would be good to have pagination also for some other GraphQL queries e.g. for `getPermissions`.


packages/client/src/registry/searchStreams.ts
```diff
@@ -128,7 +128,7 @@ const buildQuery = (
             $canGrant: Boolean
             $id_gt: String
         ) {
-            permissions (first: ${pageSize} ${GraphQLClient.createWhereClause(variables)}) {
+            permissions (first: ${pageSize} ordeBy: "id" ${GraphQLClient.createWhereClause(variables)}) {
```
harbu:
```suggestion
            permissions (first: ${pageSize} orderBy: "id" ${GraphQLClient.createWhereClause(variables)}) {
```

harbu (approved):

packages/client/src/registry/searchStreams.ts
harbu: Resolved already.

PR #967, opened 2022-11-04
fix(client): Query stored streams for non-existent storage node

teogeb:
If `StreamrClient#getStoredStreams` was called for a non-existent storage node it threw an exception. Fixed the bug by adding a `null` check.

- bug was introduced here: https://github.com/streamr-dev/network/pull/965
- error: `TypeError: Cannot read properties of null (reading 'storedStreams')`
- GraphQL response: `{ node: null, _meta: { block: { number: ... } } }`

Improved also test coverage:
- added two test cases
- merged one test to another

### DIFF
--- a/packages/client/test/end-to-end/StorageNodeRegistry.test.ts
+++ b/packages/client/test/end-to-end/StorageNodeRegistry.test.ts
@@ -60,6 +60,34 @@ describe('StorageNodeRegistry', () => {
         stored.streams.forEach((s) => expect(s).toBeInstanceOf(Stream))
     }, TEST_TIMEOUT)
 
+    it('no storage node', async () => {
+        stream = await createTestStream(creatorClient, module)
+        const id = randomEthereumAddress()
+        const stored = await creatorClient.getStoredStreams(id)
+        expect(stored.streams).toEqual([])
+        expect(stored.blockNumber).toBeNumber()
+    }, TEST_TIMEOUT)
+
+    it('no assignments', async () => {
+        const storageNodeWallet = new Wallet(await fetchPrivateKeyWithGas())
+        const storageNodeManager = new StreamrClient({
+            ...ConfigTest,
+            auth: {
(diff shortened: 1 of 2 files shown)


packages/client/test/end-to-end/StorageNodeRegistry.test.ts
```diff
@@ -56,6 +56,36 @@ describe('StorageNodeRegistry', () => {
         expect(storageNodes).toHaveLength(0)
         stored = await creatorClient.getStoredStreams(DOCKER_DEV_STORAGE_NODE)
         expect(stored.streams.some((s) => s.id === stream.id)).toBe(false)
+        expect(stored.streams.length).toBeGreaterThanOrEqual(0)
+        stored.streams.forEach((s) => expect(s).toBeInstanceOf(Stream))
+    }, TEST_TIMEOUT)
+
+    it('no storage node', async () => {
+        stream = await createTestStream(creatorClient, module)
```
harbu: Why do we need to create stream here?

packages/client/test/end-to-end/StorageNodeRegistry.test.ts
```diff
@@ -56,6 +56,36 @@ describe('StorageNodeRegistry', () => {
...
+            auth: {
+                privateKey: storageNodeWallet.privateKey
+            },
+            network: {
+                ...ConfigTest.network,
+                id: storageNodeWallet.address
+            }
+        })
+        await storageNodeManager.setStorageNodeMetadata({ http: 'mock-url' })
+        stream = await createTestStream(creatorClient, module)
```
harbu: Why do we need to create stream here?

harbu (approved):

packages/client/test/end-to-end/StorageNodeRegistry.test.ts
teogeb: No need 👍 Fixed

PR #975, opened 2022-11-08
refactor(client): Remove subconfig interfaces

teogeb:
Removed config injection tokens. The configuration is injected to components all components which need to access it, and the components read the config options from that object. There is no need to divide the config to multiple subconfigs and inject those to the components.  The new injection model is straightforward, similar to what `Authentication` already used.

In components we annotate the config type e.g. with `Pick<StrictStreamrClientConfig, 'decryption'>` if a components consumes only one aspect of the client config.

Removed all subconfig types:
- `SubscriberConfig`
- `NetworkConfig`
- `DecryptionConfig`
- `CacheConfig`
- `MetricsConfig` and `MetricsPeriodConfig`
- `TimeoutsConfig`
- `EthereumConfig`
- `AuthConfig`

When we don't use the subconfig types we can quite easily see the structure of `StrictStreamrClientConfig`. Previously it was easy to see what config blocks are there, but config details were maybe somewhat hidden. 
- The subconfig types were exported primarily for API docs, it is not very likely that the types were extensively used by the end-users

## Other changes

- Improved `StreamRegistryCached` 
  - Now wrapped functions are initialized in constructor. It ensures that the config property can access the config property which is initialized in the constructor. Previously browser tests failed as we created the methods for the object before constructor has been invoked. This way it is also easier to see the what kind of state handling is done in the class.
  - strict return types for cache
  - removed unnecessary `await`
- Improved null checks in `Ethereum.ts`
  - `contracts.ethereumNetworks` is required in strict config as there's a default value for it
  - RPC array length check is checked in schema
  - other small null checks
- Removed obsolete constructor parameter from `Resends`
- Removed unnecessary check from `Authentication`
- Some tests are now using `createPrivateKeyAuthentication` method  instead of generic `createAuthentication`
- Removed `ProvideConfig` alias name

## Future improvements

- Remove xor-operation from auth config? It could be `PrivateKeyAuthConfig | ProviderAuthConfig` instead of `XOR<PrivateKeyAuthConfig, ProviderAuthConfig>` if we want to improve readability? On the other hand, the xor is more strictly what should be configured. But we don't use xors in any other places where we can configure either X or Y.
- It could make sense that the main config interface would be `StreamrClientConfig` instead of `StrictStreamrClientConfig`. That way users could easily see what are the configuration options. The fact that we merge the user-defined config to the defaults to make the strict version of the config is just an implementation detail and should be not part or API docs.


harbu (approved):
I think the `Pick<>` pattern is good for tests. We can pass actual types without having to cast `as any`.

harbu:
Ah is an update to CHANGELOG needed?

PR #979, opened 2022-11-09
fix(network): fix MessageQueue size

harbu:
## Summary

- Fix bug(?) in which MessageQueue default buffer size was huge (potentially `1048576` passed from `WebRtcEndpoint.ts`).
- Reduce default MessageQueue buffer size from 5000 to 500. 


juslesan (approved):
Good catch

PR #990, opened 2022-11-16
refactor(utils): convert stop functions to abortSignal

harbu:
## Summary

Convert util functions returning stop function (return type of `{ stop(): void }` ) to take in an `AbortSignal` instead.

## Changes
- Functions `createReportProducer`, `scheduleAtFixedRate`, `scheduleAtInterval` accept an `AbortSignal` as parameter.
- Fix client and broker usages of said functions.


packages/broker/src/plugins/brubeckMiner/BrubeckMinerPlugin.ts
```diff
@@ -65,7 +64,11 @@ export class BrubeckMinerPlugin extends Plugin<BrubeckMinerPluginConfig> {
 
         await this.subscribe()
 
-        this.rewardSubscriptionRetryRef = setTimeout(() => this.subscriptionIntervalFn(), this.subscriptionRetryInterval)
+        setAbortableTimeout(
```
teogeb: It would maybe make sense to convert this to use `scheduleAtInterval`. We set a timeout here to execute a function for the first time, and then on the function we set another timeout so that the operation waits `subscriptionRetryInterval` between each invocation.

packages/utils/test/Metric.test.ts
```diff
@@ -11,27 +11,27 @@ describe('metrics', () => {
 
         let context: MetricsContext
         let reports: (MetricsReport & { generationTime: number })[]
-        let producer: { stop: () => void }
-
+        let abortController: AbortController
```
teogeb, line 14:
```suggestion
        let abortController: AbortController
        
```

teogeb (approved):

PR #998, opened 2022-11-18
test(client): Refactor unsubscribe tests

teogeb:
Refactored unsubscribe tests. Moved the test cases to a separate file.

Some test logic has changed:

- Removed `"unsubscribe before subscribed"` test case which did unsubscribe and unsubscribe calls parallel. 
  - It is maybe ok that the state of the subscription is not well-defined after the parallel calls as we can't assume which call was completed first. We could have 0 subscriptions (i.e. subscribe added, then unsubscribe removed) or 1 subscription (unsubscribe was no-op as there was no subscription yet, then subscribe added).
  - Removed the similar assertion also from `"unsubscribe after subscribed"` test case. 

- Added a test case for `Subscription#unsubscribe` to assert that the list of active subscriptions is modified.

- Added an assertion which ensures that `StreamrClient#unsubscribe` does the same cleanup as `Subscription#unsubscribe`.
  - That is done by checking that `StreamrClient#unsubscribe` calls `Subscription#unsubscribe`. In the future we may want to refactor the unsubscribe logic to be more straightforward: e.g. so that `StreamrClient#unsubscribe` just calls `Subscription#unsubscribe` and all cleanup is done there.

- Merged test cases `"can unsubscribe inside resend"` and `"ends resend if unsubscribed"` which were duplicates.


packages/client/test/integration/unsubscribe.test.ts
```diff
@@ -0,0 +1,139 @@
...
+        jest.spyOn(sub, 'unsubscribe')
+        expect(await client.getSubscriptions()).toHaveLength(1)
+
+        await client.unsubscribe(sub)
+
+        expect(await client.getSubscriptions()).toHaveLength(0)
+        expect(sub.unsubscribe).toBeCalled()
+    })
+
+    it('Subscription#unsubscribe', async () => {
```
harbu: I'd put this test before the previous one due to the previous test consisting of this behavior + additional.

packages/client/test/integration/unsubscribe.test.ts
```diff
@@ -0,0 +1,139 @@
...
+        expect(await client.getSubscriptions()).toHaveLength(1)
+
+        await sub.unsubscribe()
+
+        expect(await client.getSubscriptions()).toHaveLength(0)
+    })
+
+    it('twice', async () => {
+        const sub = await client.subscribe(stream.id, () => {})
+        jest.spyOn(sub, 'unsubscribe')
```
harbu: unused

packages/client/test/integration/unsubscribe.test.ts
```diff
@@ -0,0 +1,139 @@
...
+        const received: Message[] = []
+        for await (const m of sub) {
+            received.push(m)
+            setImmediate(() => {
+                sub.unsubscribe()
+            })
+        }
+
+        expect(received).toHaveLength(1)
+        expect(received[0].signature).toEqual(published!.signature)
```
harbu: Can be removed.

packages/client/test/integration/unsubscribe.test.ts
```diff
@@ -0,0 +1,139 @@
...
+        jest.spyOn(sub, 'unsubscribe')
+        expect(await client.getSubscriptions()).toHaveLength(1)
+
+        await sub.unsubscribe()
+        await sub.unsubscribe()
+
+        expect(await client.getSubscriptions()).toHaveLength(0)
+    })
+
+    it('asynchronously', async () => {
```
harbu:
```suggestion
    it('can unsubscribe inside async iteration', async () => {
```

harbu (approved):

PR #999, opened 2022-11-18
refactor(client): Simplify `RSAKeyPair`

teogeb:
Simplified internal object creation logic in `RSAKeyPair`.

All components which used `RSAKeyPair` created it by calling `RSAKeyPair.create()`. In this PR we annotated the constructor as private to enforce that logic. After that enforcement we can be sure that any `RsaKeyPair` instance is always "ready"  when it is given to the caller component. Therefore:
- we don't need a separate `InitializedRSAKeyPair` type 
- we don't need `isReady` method, and calls to that method
- `publicKey` and `privateKey` fields can't be undefined as we set the during the initialization

Refactored the factory methods `keyPairServer` and `keyPairBrowser` to return `RSAKeyPair` instances instead of modifying an existing instance. Also renamed the methods to make it clear that those are environment-specific variants of the same functionality. 

Removed caching of the factory promise `(_generateKeyPairPromise`). No need to have caching as the factory is called only once for each instance.

Annotated `publicKey` and `privateKey` fields private as those are always accessed via the accessors.

Annotated return type of `getSubtle()` and removed an unnecessary null check related to that.

## Future improvements

- After this PR the only place where we use `ts-toolbelt` library is used only for MaybeAsync helper type. The dependency could be dropped if we inline/remove `MaybeAsync`.

### DIFF
--- a/packages/client/src/encryption/RSAKeyPair.ts
+++ b/packages/client/src/encryption/RSAKeyPair.ts
@@ -1,5 +1,4 @@
 import crypto from 'crypto'
-import { O } from 'ts-toolbelt'
 import { promisify } from 'util'
 
 const { webcrypto } = crypto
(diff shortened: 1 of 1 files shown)


packages/client/src/encryption/RSAKeyPair.ts
```diff
@@ -41,59 +40,31 @@ async function exportCryptoKey(key: CryptoKey, { isPrivate = false } = {}): Prom
...
-    public publicKey: string | undefined
-    private _generateKeyPairPromise: Promise<void> | undefined
-
-    async onReady(): Promise<void> {
-        if (this.isReady()) { return undefined }
-        return this.generateKeyPair()
-    }
+    // the keys are in PEM format
+    public privateKey: string
+    public publicKey: string
```
harbu:
```suggestion
    public readonly privateKey: string
    public readonly publicKey: string
```
Can use `readonly` here?

packages/client/src/encryption/RSAKeyPair.ts
```diff
@@ -41,59 +40,31 @@ async function exportCryptoKey(key: CryptoKey, { isPrivate = false } = {}): Prom
...
-        if (!this._generateKeyPairPromise) {
-            this._generateKeyPairPromise = (typeof window !== 'undefined')
-                ? this.keyPairBrowser()
-                : this.keyPairServer()
-        }
-        return this._generateKeyPairPromise
+    static async create(): Promise<RSAKeyPair> {
+        return (typeof window !== 'undefined')
+            ? RSAKeyPair.create_browserEnvironment()
+            : RSAKeyPair.create_serverEnvironment()
```
harbu, line 64: Why not typical naming convention?
```
RSAKeyPair.createBrowserEnvironment()
RSAKeyPair.createServerEnvironment()
```

harbu (approved):

packages/client/src/encryption/RSAKeyPair.ts
teogeb: True, and actually can also be private

packages/client/src/encryption/RSAKeyPair.ts
teogeb: Not sure if we have a convention for methods? Usually we have a separate class, e.g. `BrowserPersistence` and `ServerPersistence`.

Without underscores the methods would would create _environments_, not _key pairs_. E.g. the earlier name `keyPairBrowser` was very confusing: it sounded like it'd return some kind of searching iterator (a method which _browses_ a key from a set of keys).

PR #1003, opened 2022-11-22
refactor(client): Config normalization

teogeb:
Refactored config handling. Metrics and private key configs are now normalized locally (in `MetricPublisher` and in `Authentication`, i.e. in the components which need the normalized structure).
  - added also a unit test for `MetricsPublisher`

## Other config-related changes

- fixed schema description
- removed `encryptionKeys` config option from schema
   - should have removed it already in https://github.com/streamr-dev/network/pull/839
- reorder configs so that the options are in same order in all definitions: `StrictStreamrClientConfig`, `STREAM_CLIENT_DEFAULTS` and in schema
- remove unnecessary check for non-nullable `streamRegistryChainRPCs` field
- remove gap fill defaults from `OrderingUtils` as there are always values for those config options (either given by user or client defaults)
  - Simplified also `boolean` evaluation

### DIFF
--- a/packages/client/src/subscribe/OrderMessages.ts
+++ b/packages/client/src/subscribe/OrderMessages.ts
@@ -43,8 +43,8 @@ export class OrderMessages<T> {
         this.onOrdered = this.onOrdered.bind(this)
         this.onGap = this.onGap.bind(this)
         this.maybeClose = this.maybeClose.bind(this)
-        const { gapFillTimeout, retryResendAfter, maxGapRequests, orderMessages = true, gapFill = true } = this.config
-        this.enabled = !!(gapFill && maxGapRequests)
+        const { gapFillTimeout, retryResendAfter, maxGapRequests, orderMessages, gapFill } = this.config
+        this.enabled = gapFill && (maxGapRequests > 0)
         this.orderMessages = orderMessages
         this.orderingUtil = new OrderingUtil(
             this.onOrdered,
--- a/packages/client/test/test-utils/utils.ts
+++ b/packages/client/test/test-utils/utils.ts
@@ -8,7 +8,7 @@ import { StreamrClient } from '../../src/StreamrClient'
 import { counterId } from '../../src/utils/utils'
 import { Stream, StreamMetadata } from '../../src/Stream'
 import { ConfigTest } from '../../src/ConfigTest'
-import { STREAM_CLIENT_DEFAULTS, StreamrClientConfig } from '../../src/Config'
+import { StreamrClientConfig } from '../../src/Config'
 import { GroupKey } from '../../src/encryption/GroupKey'
 import { addAfterFn } from './jest-utils'
 import { GroupKeyStore } from '../../src/encryption/GroupKeyStore'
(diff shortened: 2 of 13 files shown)


packages/client/src/subscribe/OrderMessages.ts
```diff
@@ -43,8 +43,8 @@ export class OrderMessages<T> {
         this.onOrdered = this.onOrdered.bind(this)
         this.onGap = this.onGap.bind(this)
         this.maybeClose = this.maybeClose.bind(this)
-        const { gapFillTimeout, retryResendAfter, maxGapRequests, orderMessages = true, gapFill = true } = this.config
-        this.enabled = !!(gapFill && maxGapRequests)
+        const { gapFillTimeout, retryResendAfter, maxGapRequests, orderMessages, gapFill } = this.config
```
harbu, line 46: What about the default values before (i.e. `orderMessages = true` and `gapFill = true`)? Is this functionally equivalent?

packages/client/test/test-utils/utils.ts
```diff
@@ -18,13 +18,14 @@ import { Authentication, createPrivateKeyAuthentication } from '../../src/Authen
...
 import { LoggerFactory } from '../../src/utils/LoggerFactory'
+import { waitForCondition } from '@streamr/utils'
 
 const logger = new Logger(module)
 
 export function mockLoggerFactory(clientId?: string): LoggerFactory {
     return new LoggerFactory({
         id: clientId ?? counterId('TestCtx'),
-        logLevel: STREAM_CLIENT_DEFAULTS.logLevel
+        logLevel: 'info'
```
harbu: Useful to mirror default log level?

packages/client/src/subscribe/OrderMessages.ts
teogeb: Not a functionality change, the fallbacks were never used. Same default values in https://github.com/streamr-dev/network/blob/6753475f770de8dda94cdea7252603aad8c187b5/packages/client/src/Config.ts#L137

packages/client/test/test-utils/utils.ts
teogeb: Related to this PR: https://github.com/streamr-dev/network/pull/1005

packages/client/test/unit/MetricsPublisher.test.ts
```diff
@@ -0,0 +1,151 @@
...
+        const eventEmitter = new StreamrClientEventEmitter()
+        new MetricsPublisher(
+            publisher as any,
+            node as any,
+            eventEmitter,
+            destroySignal,
+            config
+        )
+
+        eventEmitter.emit('subscribe', undefined)
```
harbu, line 42: What is the purpose of this emit?

packages/client/test/unit/MetricsPublisher.test.ts
```diff
@@ -0,0 +1,151 @@
...
+                    duration: PERIOD_DURATION
+                }],
+                maxPublishDelay: 1
+            }
+        }
+        startMetricsPublisher(config)
+        
+        await waitForCalls(publishReportMessage, 1)
+
+        const [ streamId, reportContent, publishMetadata ] = publishReportMessage.mock.calls[0]
```
harbu:
```suggestion
        const [streamId, reportContent, publishMetadata] = publishReportMessage.mock.calls[0]
```

PR #1009, opened 2022-11-23
client(docs): add typedoc documentation to public APIs

harbu:
## Summary

Add some documentation to public APIs using tsdoc.

Not everything exported has documentation, so further work is needed. But main parts are now documented.


packages/client/src/Message.ts
```diff
@@ -2,17 +2,51 @@ import { EthereumAddress } from '@streamr/utils'
...
+     */
     streamId: StreamID
+
+    /**
+     * The partition number the message was published to.
+     */
     streamPartition: number
+
+    /**
+     * The timestamp of when the message ws published.
```
harbu:
```suggestion
     * The timestamp of when the message was published.
```

packages/client/src/Message.ts
```diff
@@ -2,17 +2,51 @@ import { EthereumAddress } from '@streamr/utils'
...
+     */
     timestamp: number
+
+    /**
+     * Tiebreaker used to determine order in the case of multiple messages having the same exact timestamp.
```
harbu:
```suggestion
     * Tiebreaker used to determine order in the case of multiple messages within a message chain having the same exact timestamp.
```

packages/client/src/Stream.ts
```diff
@@ -28,12 +28,34 @@ import { Message } from './Message'
 import { convertStreamMessageToMessage } from './Message'
 
 export interface StreamMetadata {
+    /**
+     * Determines how many (stream) partitions does this stream consist of.
```
harbu:
```suggestion
     * Determines how many partitions this stream consist of.
```

packages/client/src/Stream.ts
```diff
@@ -115,7 +139,10 @@ export class Stream {
     }
 
     /**
-     * Persist stream metadata updates.
+     * Updates the metadata of the stream.
```
harbu:
```suggestion
     * Updates the metadata of the stream by merging.
```

packages/client/src/StreamrClient.ts
```diff
@@ -184,10 +247,33 @@ export class StreamrClient {
...
+    /**
+     * Waits for a message to be stored by a storage node.
+     *
+     * @param message - the message to be awaited for
+     * @param options - additional options for controlling waiting and message matching
+     * @returns resolved promise if message was found in storage before timeout, otherwise rejects
+     */
     waitForStorage(message: Message, options?: {
+        /**
+         * Determines how often should storage node should be polled.
```
harbu:
```suggestion
         * Determines how often should storage node be polled.
```

packages/client/src/StreamrClient.ts
```diff
@@ -184,10 +247,33 @@ export class StreamrClient {
...
+         */
         interval?: number
+        /**
+         * Timeout after which to give up if message was not seen
```
harbu:
```suggestion
         * Timeout after which to give up if message was not seen.
```

packages/client/src/StreamrClient.ts
```diff
@@ -230,15 +332,34 @@ export class StreamrClient {
...
+     */
     deleteStream(streamIdOrPath: string): Promise<void> {
         return this.streamRegistry.deleteStream(streamIdOrPath)
     }
 
+    /**
+     * Searches for streams based on given criteria.
+     *
+     * @param term - a search term that should be contained in either the stream id or metadata of a result
+     * @param permissionFilter - check that given permissions should be in effect
```
harbu:
```suggestion
     * @param permissionFilter - permissions that should be in effect for a result
```

packages/client/src/Stream.ts
```diff
@@ -115,7 +139,10 @@ export class Stream {
     }
 
     /**
-     * Persist stream metadata updates.
+     * Updates the metadata of the stream by merging.
+     *
+     * @param metadata - the new metadata (merged with existing metadata)
+     * @returns if successful, a resolved promise
```
teogeb: The "if successful, a resolved promise" phrase is maybe redundant, as it's the default functionality of how async methods work. 

If we have some methods which are somewhat likely to reject (i.e. more likely than other methods), I'd mention a list of possible error causes and their error codes in `@returns` description. But this method is maybe not very likely to reject?

packages/client/src/StreamrClient.ts
```diff
@@ -58,6 +60,12 @@ export class StreamrClient {
     private readonly streamIdBuilder: StreamIDBuilder
     private readonly eventEmitter: StreamrClientEventEmitter
 
+    /**
+     * Instantiates a new {@link StreamrClient} instance.
```
teogeb: Maybe redundant as that's what all constructors do. Less documentation makes it easier to focus to the important parts.

We could rename `options` parameter to config (to have consistent naming). Then we could also remove that parameter description.

(The `parentContainer` parameter was removed in a separate PR.)

packages/client/src/StreamrClient.ts
```diff
@@ -87,7 +95,14 @@ export class StreamrClient {
     // --------------------------------------------------------------------------------------------
 
     /**
+     * Publishes a message to a stream partition in the network.
+     *
      * @category Important
+     *
+     * @param streamDefinition - the stream or stream partition to publish the message to
+     * @param content - the content / payload of the message
```
teogeb: Always content, not payload. If by payload we mean the raw payload string (`serializedContent` field).

packages/client/src/StreamrClient.ts
```diff
@@ -87,7 +95,14 @@ export class StreamrClient {
...
+     * @param metadata - provide additional metadata to be included in the message or to control the publishing process
+     * @returns if successful, the unencrypted published message including metadata
```
teogeb: Is that true? If I remember correctly, it is encrypted if private stream is used

packages/client/src/StreamrClient.ts
```diff
@@ -99,6 +114,15 @@ export class StreamrClient {
         return convertStreamMessageToMessage(result)
     }
 
+    /**
+     * Manually update the current encryption key of a given stream.
+     *
+     * @remarks Only affects the encryption key used when acting as publisher.
```
teogeb: Could maybe combine the description and remarks here? 

Actually it doesn't update the _current_ encryption key if rotate method is used, but instead it queues the key to be used after one message has been published (i.e. the message which uses the previous key and delivered this key inside the `nextGroupKey` field).

packages/client/src/StreamrClient.ts
```diff
@@ -124,7 +158,14 @@ export class StreamrClient {
     // --------------------------------------------------------------------------------------------
 
     /**
+     * Subscribes to a stream partition in the network.
+     *
      * @category Important
+     *
+     * @param options - the stream or stream partition to subscribe to,
```
teogeb, line 157: Always a partition, not the whole stream. (Also line `193`)

Maybe we should just say that by default we use the first partion. That fact is now somehow hidden, maybe because we planned (or are still planning?) to change the semantics of how subscribe works if no partition is given.

packages/client/src/StreamrClient.ts
```diff
@@ -168,8 +221,18 @@ export class StreamrClient {
     // --------------------------------------------------------------------------------------------
 
     /**
-     * Call last/from/range as appropriate based on arguments
+     * Performs a resend of stored historical data.
```
teogeb, line 213: Just "fetches"? At least to me the term "resend" is not very intuitive as long as resends are not subscriptions...

packages/client/src/StreamrClient.ts
```diff
@@ -168,8 +221,18 @@ export class StreamrClient {
...
+     *
      * @category Important
+     *
+     * @remarks The given stream needs to be assigned to at least one storage node, otherwise this method will outright
```
teogeb: This could be in `@returns` section?

packages/client/src/StreamrClient.ts
```diff
@@ -184,10 +247,33 @@ export class StreamrClient {
...
         timeout?: number
+
+        /**
+         * Controls size of internal resend used in polling.
+         */
         count?: number
+
+        /**
+         * Used to set a custom message equality operator.
+         * @param msgTarget - message being waited for (i.e. `message`)
```
teogeb, line 260: Maybe we should annotate `messageMatchFn` as `@internal`?

packages/client/src/StreamrClient.ts
```diff
@@ -198,14 +284,25 @@ export class StreamrClient {
     // --------------------------------------------------------------------------------------------
 
     /**
+     * Gets a stream.
```
teogeb, line 273: The method name tells that already (also e.g. line `299`).

packages/client/src/StreamrClient.ts
```diff
@@ -247,41 +368,102 @@ export class StreamrClient {
     // Permissions
     // --------------------------------------------------------------------------------------------
 
+    /**
+     * Gets all valid publishers of a stream.
+     *
+     * @param streamIdOrPath - the stream id
```
teogeb: Remove this description (also e.g. line 384)

packages/client/src/StreamrClient.ts
```diff
@@ -247,41 +368,102 @@ export class StreamrClient {
...
+     * @returns async iterable collection of {@link EthereumAddress} (automatic paging)
```
teogeb: Paging is an implementation detail (also e.g. line 385)

packages/client/src/StreamrClient.ts
```diff
@@ -247,41 +368,102 @@ export class StreamrClient {
...
+     */
     setPermissions(...items: {
         streamId: string
         assignments: PermissionAssignment[]
     }[]): Promise<void> {
         return this.streamRegistry.setPermissions(...items)
     }
 
+    /**
+     * Checks whether a given (user) address is a valid publisher of a stream.
```
teogeb: Either word `user` or `address` is redundant? Maybe could just say `"... given user is a valid..."`

packages/client/src/StreamrClient.ts
```diff
@@ -247,41 +368,102 @@ export class StreamrClient {
...
+     *
+     * @param streamIdOrPath - the stream id
+     * @param userAddress - the Ethereum address of the user
+     * @returns resolves with true/false, rejects if check could not be performed
```
teogeb: That is the normal `Promise` functionality, and `true`/`false` return type is obvious.

packages/client/src/StreamrClient.ts
```diff
@@ -290,30 +472,78 @@ export class StreamrClient {
...
+     */
     async removeStreamFromStorageNode(streamIdOrPath: string, nodeAddress: string): Promise<void> {
         return this.streamStorageRegistry.removeStreamFromStorageNode(streamIdOrPath, toEthereumAddress(nodeAddress))
     }
 
+    /**
+     * Checks whether a stream is being stored by a storage node.
+     *
+     * @param streamIdOrPath - the stream id
+     * @param nodeAddress - Ethereum address of the storage node
```
teogeb: Could rename the parameter to `storageNode`

teogeb (approved):

PR #1016, opened 2022-11-24
chore(client): remove extraneous generics

harbu:
## Summary

Remove extraneous and occasionally misleading generics. Many were related to`StreamMessage<T>` where the type `T` could not actually  be guaranteed to be what is claimed.

- Potentially some missed spots here and there. However, public API should be pretty clean from these now.

### DIFF
--- a/packages/client/src/StreamrClient.ts
+++ b/packages/client/src/StreamrClient.ts
@@ -102,7 +102,7 @@ export class StreamrClient {
      * @category Important
      *
      * @param streamDefinition - the stream or stream partition to publish the message to
-     * @param content - the content (the payload) of the message (usually an JSON serializable object)
+     * @param content - the content (the payload) of the message (must be JSON serializable)
      * @param metadata - provide additional metadata to be included in the message or to control the publishing process
      * @returns the published message (note: the field {@link Message.content} is encrypted if the stream is private)
      */
--- a/packages/client/src/StreamrClient.ts
+++ b/packages/client/src/StreamrClient.ts
@@ -102,13 +102,13 @@ export class StreamrClient {
      * @category Important
      *
      * @param streamDefinition - the stream or stream partition to publish the message to
-     * @param content - the content (the payload) of the message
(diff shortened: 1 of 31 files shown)


packages/client/src/StreamrClient.ts
```diff
@@ -102,13 +102,13 @@ export class StreamrClient {
      * @category Important
      *
      * @param streamDefinition - the stream or stream partition to publish the message to
-     * @param content - the content (the payload) of the message
+     * @param content - the content (the payload) of the message (usually an JSON serializable object)
```
teogeb: Needs to be JSON serializable object as we create a message which has content type of JSON (the default content type, `StreamMessage:107`)

packages/client/src/subscribe/OrderMessages.ts
```diff
@@ -146,8 +146,8 @@ export class OrderMessages<T> {
         }
     }
 
-    transform(): (src: AsyncGenerator<StreamMessage<T>, any, unknown>) => AsyncGenerator<StreamMessage<T>, void, unknown> {
-        return async function* Transform(this: OrderMessages<T>, src: AsyncGenerator<StreamMessage<T>>) {
+    transform(): (src: AsyncGenerator<StreamMessage, any, unknown>) => AsyncGenerator<StreamMessage, void, unknown> {
```
teogeb: Could simplify AsyncGenerator generics to be just `AsyncGenerator<StreamMessage>` (in this PR on in a separate PR)

packages/client/src/StreamrClient.ts
harbu: Addressed in a5e9c88e854ce597f323272888a637a5150aa834 with wording change "must be"

packages/client/src/subscribe/OrderMessages.ts
harbu: Fixed in a5e9c88e854ce597f323272888a637a5150aa834

PR #1052, opened 2022-12-27
feat: lit protocol key exchange

harbu:
## Summary

Add support for retrieving and storing encryption keys via lit-protocol. When enabled we will first try to retrieve / store key using lit protocol, if this doesn't work, we use our own key-exchange system.

## Changes

- All lit-protocol library interactions are encapsulated into class `LitProtocolFacade`. 
- Add new class `GroupKeyManager` that is used to retrieve and store encryption keys. It contains the logic for iterating over different key exchange strategies until a key is found (or successfully generated and stored).
    - Retrieve key: (1) local, (2) lit-protcol, and (3) Streamr key-exchange
    - Store key: (1) generate with lit-protocol, and if that fails, (2) generate locally. We _always_ store encryption key into local store.

- lit-protocol is hard coded to work with main net Polygon and pre-defined streamRegistryChainAddress
- no integration tests for lit-protocol
    - implement in future when it becomes possible to run own nodes and override streamRegistryChainAddress and RPC settings
- Since we can't write integration tests for lit-protocol, do we need some automated tests to run against production?
- Combine `encryption` and `decryption` client config blocks


packages/client/package.json
```diff
@@ -118,6 +119,7 @@
     "reflect-metadata": "^0.1.13",
     "secp256k1": "^4.0.2",
     "sha3": "^2.1.4",
+    "siwe": "^1.1.6",
```
harbu: Might be better to use `lit-siwe`?

packages/client/package.json
```diff
@@ -97,6 +97,7 @@
     "@ethersproject/transactions": "^5.5.0",
     "@ethersproject/wallet": "^5.5.0",
     "@ethersproject/web": "^5.5.0",
+    "@lit-protocol/lit-node-client": "^2.0.40",
```
harbu, line 100: Can this be narrowed?

packages/client/src/encryption/LitProtocolKeyStore.ts
```diff
@@ -0,0 +1,144 @@
...
+            comparator: '=',
+            value: "true",
+        },
+    }
+])
+
+const signAuthMessage = async (authentication: Authentication) => {
+    const domain = "localhost"
+    const uri = "https://localhost/login"
+    const statement = "This is a test statement. You can put anything you want here."
```
harbu: Can these take on any value?

packages/client/src/encryption/LitProtocolKeyStore.ts
```diff
@@ -0,0 +1,144 @@
...
+    const address = ethers.utils.getAddress(await authentication.getAddress())
```
harbu:
```suggestion
    const address = ethers.utils.getAddress(await authentication.getAddress()) // requires checksum casing
```

packages/client/src/encryption/LitProtocolKeyStore.ts
```diff
@@ -0,0 +1,144 @@
...
+        sig: signature,
+        derivedVia: "web3.eth.personal.sign",
+        signedMessage: messageToSign,
+        address
+    }
+}
+
+@scoped(Lifecycle.ContainerScoped)
+export class LitProtocolKeyStore {
+    private readonly litNodeClient: LitJsSdk.LitNodeClient | undefined
```
harbu:
```suggestion
    private readonly litNodeClient?: LitJsSdk.LitNodeClient
```

packages/client/src/encryption/LitProtocolKeyStore.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    constructor(
+        @inject(AuthenticationInjectionToken) private readonly authentication: Authentication,
+        @inject(ConfigInjectionToken) private readonly config: Pick<StrictStreamrClientConfig, 'contracts' | 'litProtocolEnabled'>
+    ) {
+        if (config.litProtocolEnabled) {
+            logger.debug('lit-protocol enabled')
+            this.litNodeClient = new LitJsSdk.LitNodeClient({
+                alertWhenUnauthorized: false,
+                debug: true
```
harbu: We need to make `debug: true` configurable

packages/client/src/encryption/LitProtocolKeyStore.ts
```diff
@@ -0,0 +1,144 @@
...
+            })
+        } else {
+            logger.debug('lit-protocol disabled')
+            this.litNodeClient = undefined
+        }
+    }
+
+    async store(streamId: StreamID, symmetricKey: Uint8Array): Promise<GroupKey | undefined> {
+        if (this.litNodeClient === undefined) {
```
harbu: Instead of this, let's check whether litProtocol is enabled outside this class (i.e. by the user of this class)

packages/client/src/encryption/LitProtocolKeyStore.ts
```diff
@@ -0,0 +1,144 @@
...
+            return undefined
+        }
+        try {
+            await this.litNodeClient.connect()
```
harbu: Is there some sort of auto-connect? What about lifecycle of session? Can we loose connectivity, e.g? Should we actually disconnect after using lit and then re-connect when we re-need something from lit?

packages/client/src/encryption/LitProtocolKeyStore.ts
```diff
@@ -0,0 +1,144 @@
...
+            const authSig = await signAuthMessage(this.authentication)
+            const encryptedSymmetricKey = await this.litNodeClient.saveEncryptionKey({
+                evmContractConditions: formEvmContractConditions(this.config.contracts.streamRegistryChainAddress, streamId),
```
harbu: Worth currying `formEvmContractConditions`? (First argument is immutable in this case)

packages/client/src/encryption/LitProtocolKeyStore.ts
```diff
@@ -0,0 +1,144 @@
...
+    const signature = await authentication.createMessageSignature(messageToSign)
+    return {
+        sig: signature,
+        derivedVia: "web3.eth.personal.sign",
+        signedMessage: messageToSign,
+        address
+    }
+}
+
+@scoped(Lifecycle.ContainerScoped)
```
harbu:
```suggestion
/**
*  This class only operates with Polygon production network and therefore ignores contracts config.
**/
@scoped(Lifecycle.ContainerScoped)
```

packages/client/src/encryption/LitProtocolKeyStore.ts
```diff
@@ -0,0 +1,144 @@
...
+            }
+            const groupKeyId = LitJsSdk.uint8arrayToString(encryptedSymmetricKey, 'base16')
+            return new GroupKey(groupKeyId, Buffer.from(symmetricKey))
+        } catch (e) {
+            logger.warn('encountered error when trying to store key on lit-protocol: %s', e)
+            return undefined
+        }
+    }
+
+    async get(streamId: StreamID, encryptedSymmetricKey: string): Promise<GroupKey | undefined> {
```
harbu:
```suggestion
    async get(streamId: StreamID, groupKeyId: string): Promise<GroupKey | undefined> {
```

packages/client/src/encryption/LitProtocolKeyStore.ts
```diff
@@ -0,0 +1,144 @@
...
+        if (this.litNodeClient === undefined) {
+            return undefined
+        }
```
harbu: Optimization to consider: immediately return undefined if group key id is clearly not in the right format. (It wasn't stored in lit protocol but was generated by client locally.)

packages/client/src/publish/GroupKeyQueue.ts
```diff
@@ -38,16 +42,28 @@ export class GroupKeyQueue {
...
         await this.store.add(newKey, this.streamId)
         this.currentGroupKey = newKey
         this.queuedGroupKey = undefined
         return newKey
     }
+
+    private async generateNewKey(): Promise<GroupKey> {
+        // 1st try lit-protocol, if a key cannot be generated and stored, then generate group key locally
+        const litProtocolGroupKey = await this.litProtocolKeyStore.store(this.streamId, crypto.randomBytes(32))
+        return litProtocolGroupKey !== undefined ? litProtocolGroupKey : GroupKey.generate()
```
harbu: Something like this?
```suggestion
        const keyData =  crypto.randomBytes(32)
        // 1st try lit-protocol, if a key cannot be generated and stored, then generate group key locally
        const litProtocolGroupKey = await this.litProtocolKeyStore.store(this.streamId, keyData)
        return litProtocolGroupKey !== undefined ? litProtocolGroupKey : GroupKey.generate(keyData)
```

packages/client/src/subscribe/Decrypt.ts
```diff
@@ -47,7 +49,13 @@ export class Decrypt {
...
 
             let groupKey = await this.groupKeyStore.get(groupKeyId, streamMessage.getStreamId())
-            if (groupKey === undefined) {
+            if (groupKey === undefined) { // 1st try lit protocol
+                groupKey = await this.litProtocolKeyStore.get(streamMessage.getStreamId(), groupKeyId)
+                if (groupKey !== undefined) {
+                    await this.groupKeyStore.add(groupKey, streamMessage.getStreamId())
+                }
+            }
+            if (groupKey === undefined) { // 2nd try group-key exchange
```
harbu: Violation of SRP: we should have some sort of `GroupKeyManager#get(keyId)`. The specifics of how to fetch a key (and in which order we try different key store services) is a separate question from decrypting a message.

packages/client/package.json
harbu: Using `lit-siwe` instead now. Doesn't make the webpack bundle smaller but seems like the appropriate choice nonetheless.

packages/client/src/encryption/LitProtocolKeyStore.ts
harbu: Resolved

packages/client/test/test-utils/utils.ts
```diff
@@ -163,8 +168,31 @@ export const createStreamRegistryCached = (opts: {
...
+        new StreamrClientEventEmitter(),
+        new DestroySignal(),
+        {
+            encryption: {
+                litProtocolEnabled: false,
+                litProtocolLogging: false
+            },
+            decryption: {
+                maxKeyRequestsPerSecond: 10,
+                keyRequestTimeout: 0
```
teogeb: To keep the existing functionality of `Decrypt.test.ts`:

```suggestion
                keyRequestTimeout: 50
```

The value is not relevant for other test cases which currently use this fake `GroupKeyManager`.

packages/client/src/encryption/GroupKeyManager.ts
```diff
@@ -0,0 +1,74 @@
...
+import { ConfigInjectionToken, StrictStreamrClientConfig } from '../Config'
+import { StreamrClientEventEmitter } from '../events'
+import { DestroySignal } from '../DestroySignal'
+import crypto from 'crypto'
+import { uuid } from '../utils/uuid'
+
+@scoped(Lifecycle.ContainerScoped)
+export class GroupKeyManager {
+    constructor(
+        @inject(GroupKeyStore) private readonly groupKeyStore: GroupKeyStore,
```
teogeb: The `@inject` annotations are not needed when we inject `scoped` classes. Could remove the annotations from lines `17`-`21` (but not `22`) to highlight that the injection of `config` field is different compared to the other fields.

packages/client/src/encryption/LitProtocolFacade.ts
```diff
@@ -0,0 +1,155 @@
...
+@scoped(Lifecycle.ContainerScoped)
+export class LitProtocolFacade {
+    private readonly logger: Logger
+    private litNodeClient?: LitJsSdk.LitNodeClient
+    private connectLitNodeClient?: () => Promise<void>
+
+    constructor(
+        @inject(AuthenticationInjectionToken) private readonly authentication: Authentication,
+        @inject(ConfigInjectionToken) private readonly config: Pick<StrictStreamrClientConfig, 'contracts' | 'encryption'>,
+        private readonly loggerFactory: LoggerFactory
```
teogeb: The `loggerFactory` is not needed a field of `LitProtocolFacade`:

```suggestion
        loggerFactory: LoggerFactory
```

Could also consider defining all fields their types explicitly, i.e. not using the `private` modifier shortcut?  That's the "convention" we are currently using, and it makes it easier to see the complete list of class fields in one place.

The list of fields is usually important first place to check when someone is learning the functionality of a new class. Therefore having the list increases the code readability significantly, which is maybe more important compared to the benefits of the `private` modifier.

packages/client/src/Config.ts
```diff
@@ -154,6 +160,11 @@ export const STREAM_CLIENT_DEFAULTS:
     maxGapRequests: 5,
     retryResendAfter: 5000,
     gapFillTimeout: 5000,
+
+    encryption: {
```
teogeb, line 178: Maybe it would make sense to combine encryption and decryption blocks at some point. In a separate PR? If we'd keep these as separate blocks, it could kind of indicate that LIT is used only for encryption and not for decryption.

PR #1059, opened 2023-01-13
refactor(client): Publish error

teogeb:
Refactored the error message which is shown if a user tries to publish to a stream, but doesn't have permission to do so.

New error message:

```
StreamrClientError: Failed to publish to stream 0x7e5f4552091a69125d5dfcb7b8c2659029395bdf/foo. Cause: You don't have permission to publish to this stream. Using address: 0xe1ab8145f7e55dc933d51a18c793f901a3a0b276
    at .../network-monorepo/packages/client/src/publish/Publisher.ts:100:23
    at processTicksAndRejections (node:internal/process/task_queues:95:5) {
  code: MISSING_PERMISSION
}
```

Before this PR:

```
PublishError: Failed to publish to stream 0x7e5f4552091a69125d5dfcb7b8c2659029395bdf/foo (timestamp=1234567890123), cause: 0xe1ab8145f7e55dc933d51a18c793f901a3a0b276 is not a publisher on stream 0x7e5f4552091a69125d5dfcb7b8c2659029395bdf/foo
    at .../network-monorepo/packages/client/src/publish/Publisher.ts:121:23
    at processTicksAndRejections (node:internal/process/task_queues:95:5) {
  streamId: '0x7e5f4552091a69125d5dfcb7b8c2659029395bdf/foo',
  timestamp: 1673562505056,
  [cause]: Error: 0xe1ab8145f7e55dc933d51a18c793f901a3a0b276 is not a publisher on stream 0x7e5f4552091a69125d5dfcb7b8c2659029395bdf/foo
      at MessageFactory.createMessage (.../network-monorepo/packages/client/src/publish/MessageFactory.ts:75:19)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async .../network-monorepo/packages/client/src/publish/Publisher.ts:110:33
}
```

Now we use the standard `StreamrClientError` instead of a custom error class (`PublishError`). The main benefit of using the standard class is the error code field. In this case, the error code is `MISSING_PERMISSION`, which clearly communicates the root cause of the error.

### Removed metadata

The custom class contained some metadata about the originating message:
- `streamId`
- `timestamp`

These fields are not included in the new error. This way the error message is shorter and more readable.

In practice the metadata was quite redundant:
- the stream id is included in the error message
- if a log file is used, timestamp information can be read from it (as any log file would most likely have timestamps for the log lines)

### Possible future enhancements

`StreamrClientError` could support error chaining. Some situations are easier to debug if we have a full chain of causes. For end users, a chain is maybe harder to decipher as the root cause is "hidden" at the end of the chain. Error chaining is currently not supported in Chrome, see the comment in `PublishError`.

Some error constructors use `Error.captureStackTrace`. Would we get better stack traces in some environments if we added that functionality to `StreamrClientError`?

### DIFF
--- a/packages/client/src/StreamrClientError.ts
+++ b/packages/client/src/StreamrClientError.ts
@@ -1,4 +1,10 @@
-export type StreamrClientErrorCode = 'NO_STORAGE_NODES' | 'INVALID_ARGUMENT' | 'CLIENT_DESTROYED' | 'PIPELINE_ERROR'
+export type StreamrClientErrorCode = 
+    'MISSING_PERMISSION' | 
+    'NO_STORAGE_NODES' | 
+    'INVALID_ARGUMENT' | 
+    'CLIENT_DESTROYED' | 
+    'PIPELINE_ERROR' |
+    'UNKNOWN_ERROR'
 
 export class StreamrClientError extends Error {
     constructor(message: string, public readonly code: StreamrClientErrorCode) {
--- a/packages/client/src/publish/MessageFactory.ts
+++ b/packages/client/src/publish/MessageFactory.ts
@@ -18,6 +18,7 @@ import { Mapping } from '../utils/Mapping'
 import { Authentication } from '../Authentication'
 import { StreamRegistryCached } from '../registry/StreamRegistryCached'
 import { formLookupKey } from '../utils/utils'
+import { StreamrClientError } from '../StreamrClientError'
 
 export interface MessageFactoryOptions {
     streamId: StreamID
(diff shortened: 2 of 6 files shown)


harbu:
> Some error constructors use Error.captureStackTrace. Would we get better stack traces in some environments if we added that functionality to StreamrClientError?

Would be worth having I think. Just need to check that the function exists before invoking otherwise browser goes kaput.

harbu (approved):

fonty1 (approved):
LGTM. When the new docs land and we have a stable link to `grantPermissions` examples,  we could perhaps link to that advice as a future improvement.

PR #1063, opened 2023-01-17
feat: enable TCP turn

juslesan:
## Summary

Makes adding TCP TURN servers possible.

## Changes

- Add configurability for TCP TURN
- Add TCP TURN to the client config by default.

As the list of IceServers increases the time to establish connections could increase. It is possible that the default timeouts for opening connections need to be adjusted after this change, especially if overlay performance is detected to be reduced.

### DIFF
--- a/packages/client/src/config.schema.json
+++ b/packages/client/src/config.schema.json
@@ -172,6 +172,9 @@
                             },
                             "password": {
                                 "type": "string"
+                            },
+                            "tcp": {
+                                "type": "boolean"
                             }
                         }
                     },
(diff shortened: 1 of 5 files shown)


packages/client/src/config.schema.json
```diff
@@ -172,6 +172,9 @@
                             },
                             "password": {
                                 "type": "string"
+                            },
+                            "tcp": {
+                                "type": "boolean"
                             }
```
harbu, line 178:
```suggestion
                            "tcp": {
                                "type": "boolean",
                                "default": false
                            }
```

harbu (approved):
lgtm!

packages/client/src/config.schema.json
juslesan: This would add `tpc: false` to all configs (STUN only included) which is not strictly required. ie:
` {
              "port": 5349,
              "url": "stun:stun.streamr.network",
               "tcp": false
 }
`

harbu: Ah I see, keep it as is then.

PR #1064, opened 2023-01-19
style: Class parameter property

teogeb:
New `eslint` rule which disallows disallow parameter properties in class constructors: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/parameter-properties.md

With this rule we make sure that all fields of the classes are defined in one place. Therefore it is easy to see the complete list of fields for any class. That information is usually very important in order to understand what a class does.

E.g. 
```
class Foo {

    private x: number
    
    constructor(private y: number) {
        this.x = 123
    }
}
```

is no longer allowed. Instead we should use:

```
class Foo {

    private x: number
    private y: number

    constructor(y: number) {
        this.x = 123
        this.y = y
    }
}
```

This style makes the code a bit longer and it duplicates some information (the name and type of the fields). Therefore it is a trade-off between readability and writability. 

With this rule we can also avoid some bugs, which we have had in our codebase:
- "accidental" fields: a parameter is defined in constructor using e.g. `private` keyword, but it is not needed outside of constructor (and therefore it should not be a field)
- duplicate field initialization: e.g. having extra `this.foo = foo` statement in a constructor for a field which uses `private` keyword


teogeb:
@juslesan, @ptesavol If you prefer the previous style, we could revert the changes related to  `dht` and `proto-rpc` packages and add exceptions for the `eslint` rule instead?

packages/client/src/subscribe/Resends.ts
```diff
@@ -84,17 +89,21 @@ export class Resends {
...
-        streamrClientEventEmitter: StreamrClientEventEmitter,
-        destroySignal: DestroySignal,
+        @inject(StreamStorageRegistry) streamStorageRegistry: StreamStorageRegistry,
+        @inject(delay(() => StorageNodeRegistry)) storageNodeRegistry: StorageNodeRegistry,
+        @inject(delay(() => StreamRegistryCached)) streamRegistryCached: StreamRegistryCached,
+        @inject(HttpUtil) httpUtil: HttpUtil,
+            groupKeyStore: GroupKeyStore,
+            subscriberKeyExchange: SubscriberKeyExchange,
+            streamrClientEventEmitter: StreamrClientEventEmitter,
+            destroySignal: DestroySignal,
```
harbu: Weird indentation or just github ui issue?

harbu (approved):

harbu (approved):

PR #1072, opened 2023-01-31
feat(broker): Config overrides from environment variables

teogeb:
**BREAKING CHANGE!**

Changed the syntax of the environment variables. Now any configuration option can be defined by using an environment variable.

The new syntax is like this:
```
STREAMR__BROKER__CLIENT__AUTH__PRIVATE_KEY = '0x111'
STREAMR__BROKER__CLIENT__GAP_FILL = 'false'
STREAMR__BROKER__CLIENT__NETWORK__TRACKERS_1__ID = 'tracker1-id'
STREAMR__BROKER__CLIENT__NETWORK__TRACKER_PING_INTERVAL = '123'
STREAMR__BROKER__AUTHENTICATION__KEYS_1 = 'key-1'
STREAMR__BROKER__PLUGINS__BRUBECK_MINER__BENEFICIARY_ADDRESS = '0x222'
```

We support strings, numbers, booleans and `null` as the values of the environment variables.

### Removed functionality

Before this PR we supported only two variables: private key and beneficiary address. This syntax is no longer supported:
```
OVERRIDE_BROKER_PRIVATE_KEY = '0x111'
OVERRIDE_BROKER_BENEFICIARY_ADDRESS = '0x222'
```

### DIFF
--- a/packages/broker/configuration.md
+++ b/packages/broker/configuration.md
@@ -15,8 +15,10 @@ The names start with `STREAMR__BROKER__` and each configuration block is separat
 
 It is possible to defined arrays by adding a numeration suffix to a block/property:
 ```
-STREAMR__BROKER__CLIENT__NETWORK__TRACKERS_1__ID = 'foo'
-STREAMR__BROKER__AUTHENTICATION__KEYS_1 = 'bar'
+STREAMR__BROKER__CLIENT__NETWORK__TRACKERS_1__ID = '0x1234'
+STREAMR__BROKER__CLIENT__NETWORK__TRACKERS_2__ID = '0x5678'
+STREAMR__BROKER__AUTHENTICATION__KEYS_1 = 'foo'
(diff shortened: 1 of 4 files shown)


packages/broker/configuration.md
```diff
@@ -4,7 +4,35 @@ See [config.schema.json](src/config/config.schema.json).
...
+    },
+    ...
+}
+```
+
+All environment variable names start with `STREAMR__BROKER__` and each configuration block is separated by double underscore. Blocks and properties are defined in *CONSTANT_CASE* instead of *camelCase*.
+
+If the value is defined both in an environment variable and the configuration file, the environment variable value is used.
+
+It is possible to defined arrays by adding a numeration suffix to a block/property:
```
fonty1:
```suggestion
It is possible to define arrays by adding a numeration suffix to a block/property:
```

fonty1 (approved):
LGTM

juslesan (approved):

PR #1073, opened 2023-01-31
NET-943:  Fallback proxies

juslesan:
## Summary

This is a breaking change as it changes the client's public API! 

The proxy pattern is changed to function based on selecting `k` nodes to connect to out of `n` candidates. The client API is changed as a result of the changes. However, new clients should be able to open proxy connections to old clients.

## Changes

- The ProxyConnectionManager is split to two separate files for the client and server sides.
- If a disconnection on a Proxy Connection occurs, the client will select a new random known candidate node and attempt to open a Proxy Connection to it.
- The `k` value an be increased and reduced after candidates have been set. The node will attempt to connect up to or disconnect down to the new number of target connections.

Directionality (publish / subscribe ) is now only set when candidates are set. When candidates are removed the candidates are not differentiated based on direction. (It was never the intention that nodes should be able to open proxy connections to both directions on a single stream. In this case the full node pattern should be used)


packages/client/README.md
```diff
@@ -694,16 +694,18 @@ Proxy publishing and subscribing are handled on the network overlay level. This
 ```js
 
 // Open publish proxy to multiple nodes on stream
-await publishingClient.openProxyConnections(stream, ['0x11111...', '0x22222...'], ProxyDirection.PUBLISH)
+await publishingClient.setProxies(stream, ['0x11111...', '0x22222...'], ProxyDirection.PUBLISH)
 
 // Remove publish proxy to multiple nodes on stream
-await publishingClient.closeProxyConnections(stream, ['0x11111...', '0x22222...'], ProxyDirection.PUBLISH)
+await publishingClient.setProxies(stream, [], ProxyDirection.PUBLISH)
```
teogeb: Could the comment or description be something like: "To use nodes "0x11111 and 0x22222 as proxies when subscribing to `stream`"?

And the comment or description about remove something like "To remove some/all proxies, call the same method with a different set of nodes. If the node list is empty, proxies are no longer used for the given stream."

packages/client/README.md
```diff
@@ -694,16 +694,18 @@ Proxy publishing and subscribing are handled on the network overlay level. This
...
 
-// Open publish proxy to multiple nodes on stream
-await publishingClient.openProxyConnections(stream, ['0x11111...', '0x22222...'], ProxyDirection.SUBSCRIBE)
+```
 
-// Remove publish proxy to multiple nodes on stream
-await publishingClient.closeProxyConnections(stream, ['0x11111...', '0x22222...'], ProxyDirection.SUBSCRIBE)
+By default the client will attempt to open proxy connections to all of the nodes set in  `setProxies`. If you want to limit the number of connections by setting a fourth parameter. In this approach, if the client is disconnected from one of the nodes it will attempt to connect to another node by random.
```
teogeb:
```suggestion
By default the client will attempt to open proxy connections to all of the nodes set in  `setProxies`. You can limit the number of connections by setting the `connectionCount` parameter. In this approach, if the client is disconnected from one of the nodes it will attempt to connect to another node by random.
```

packages/network/src/logic/Node.ts
```diff
@@ -435,6 +423,10 @@ export class Node extends EventEmitter {
     }
 
     isProxiedStreamPart(streamPartId: StreamPartID, direction: ProxyDirection): boolean {
-        return this.streamPartManager.isBehindProxy(streamPartId) && this.proxyStreamConnectionManager.isProxiedStreamPart(streamPartId, direction)
+        return this.streamPartManager.isBehindProxy(streamPartId) && this.proxyStreamConnectionClient.isProxiedStreamPart(streamPartId, direction)
+    }
+
+    getProxiedStreamPartConnectionNodeIds(streamPartId: StreamPartID): NodeId[] {
```
teogeb: This method is not needed?

packages/network/src/logic/proxy/ProxyStreamConnectionClient.ts
```diff
@@ -0,0 +1,358 @@
...
+    ACCEPTED,
+}
+
+interface ProxyCandidate {
+    state?: State
+    direction: ProxyDirection
+    userId: string
+}
+
+interface ProxyTargets {
```
teogeb: What the `target` word means here? Or something like ProxyDefinition, if it defined which nodes we use as proxies for a stream? If it is just a definition, then `numOfTargets` can be `connectionCount` as in the public API?

What's the correlation between `connections` map, or in `ProxyCandidate.state`: if a new connection is made, do we need to update both?

packages/network/src/logic/proxy/ProxyStreamConnectionClient.ts
```diff
@@ -0,0 +1,289 @@
...
+            nodeIds: new Set(nodeIds),
+            userId: await getUserId(),
+            direction,
+            connectionCount: connectionCount ?? nodeIds.length
+        })
+        await this.updateConnections(streamPartId)
+    }
+
+    private async updateConnections(streamPartId: StreamPartID): Promise<void> {
+        if (!this.definitions.has(streamPartId)) {
```
teogeb: We can maybe remove this check?

- If `this.definitions.has(streamPartId)` is false, we have not ever called `setProxies` for this streamPart. 
- Therefore this call must be from Node:352
- And the Node must have been behind a proxy (`this.isBehindProxy(streamPartId)`, so that `removeNodeFromAllStreamParts` returns the value to `onNodeDisconnected`)
- A node can't be behind a proxy if user hasn't called `setProxies()`?

packages/network/src/logic/proxy/ProxyStreamConnectionClient.ts
```diff
@@ -0,0 +1,289 @@
...
+            return
+        }
+        await Promise.all(this.getInvalidConnections(streamPartId).map(async (id) => {
+            logger.debug(`Closing invalid connection to ${id} on ${streamPartId}`)
```
teogeb: Actually as we now have logging in the root methods, we may not need these log messages? (This line and lines 127` and `216`). If we'd keep this, the log would indicate that we open/close a connection multiple times when we do it just once.
```
logger.info(`Open proxy connection to ${targetNodeId} on ${streamPartId}`)
logger.info(`Close proxy connection to ${targetNodeId} on ${streamPartId}`)
```

packages/network/src/logic/proxy/ProxyStreamConnectionClient.ts
```diff
@@ -0,0 +1,289 @@
...
+        const proxiesToAttempt = shuffle(Array.from(definition.nodeIds.keys()).filter((id) =>
+            !this.getConnections(streamPartId).has(id)
+        )).map((id) => id).splice(0, connectionCount)
+        await Promise.all(proxiesToAttempt.map((id) =>
+            this.attemptConnection(streamPartId, id, definition.direction, definition.userId)
+        ))
+    }
+
+    private async attemptConnection(streamPartId: StreamPartID, nodeId: NodeId, direction: ProxyDirection, userId: string): Promise<void> {
+        await Promise.all([
```
teogeb, line 143: Would it make sense to use `runAndWaitForEvents` utility from the `utils` package? With that we could simplify the code and guarantee that we set the event listener before we do the actual operation which triggers the event.

packages/network/src/logic/proxy/ProxyStreamConnectionClient.ts
```diff
@@ -0,0 +1,289 @@
...
+    private async initiateConnection(
+        streamPartId: StreamPartID,
+        targetNodeId: string,
+        direction: ProxyDirection,
+        userId: string
+    ): Promise<void> {
+        if (!this.streamPartManager.isSetUp(streamPartId)) {
+            this.streamPartManager.setUpStreamPart(streamPartId, true)
+        } else if (!this.streamPartManager.isBehindProxy(streamPartId)) {
+            const reason = `Could not open a proxy ${direction} stream connection ${streamPartId}, non-proxy stream already exists`
```
teogeb: Would it make sense if `setProxies` would already check this and fail if we try to set a proxy for a stream which is already used non-proxied way? Could also move the check at line `182` to `setProxies` as it is a similar violation.

packages/network/src/logic/proxy/ProxyStreamConnectionClient.ts
```diff
@@ -0,0 +1,289 @@
...
+        this.connections.get(streamPartId)!.set(targetNodeId, direction)
+        try {
+            await this.connectAndHandshake(streamPartId, targetNodeId, direction, userId)
+        } catch (err) {
+            logger.warn(`Failed to create a proxy ${direction} stream connection to ${targetNodeId} for stream ${streamPartId}:\n${err}`)
+            this.removeConnection(streamPartId, targetNodeId)
+            this.emit(Event.CONNECTION_REJECTED, targetNodeId, streamPartId, direction, err)
+        } finally {
+            const trackerId = this.trackerManager.getTrackerId(streamPartId)
+            this.trackerManager.disconnectFromSignallingOnlyTracker(trackerId)
```
teogeb: If this connection fails, but there were other connections, should we disconnect from the tracker or not?

packages/network/src/logic/proxy/ProxyStreamConnectionClient.ts
```diff
@@ -0,0 +1,289 @@
...
+        }
+    }
+
+    private async connectAndHandshake(streamPartId: StreamPartID, targetNodeId: NodeId, direction: ProxyDirection, userId: string): Promise<void> {
+        const trackerId = this.trackerManager.getTrackerId(streamPartId)
+        const trackerAddress = this.trackerManager.getTrackerAddress(streamPartId)
+        await this.trackerManager.connectToSignallingOnlyTracker(trackerId, trackerAddress)
```
teogeb: Could refactor `connectToSignallingOnlyTracker` to take `streamId` as a parameter, as it can get `trackerId` and `trackerAddress` from that?

packages/network/src/logic/proxy/ProxyStreamConnectionClient.ts
```diff
@@ -0,0 +1,289 @@
...
+            || this.definitions.get(streamPartId)!.direction !== this.getConnections(streamPartId).get(id)
+        )
+    }
+
+    private async openRandomConnections(streamPartId: StreamPartID, connectionCount: number): Promise<void> {
+        logger.debug(`Open ${connectionCount} random connections on ${streamPartId}`)
+        const definition = this.definitions.get(streamPartId)!
+        const proxiesToAttempt = shuffle(Array.from(definition.nodeIds.keys()).filter((id) =>
+            !this.getConnections(streamPartId).has(id)
+        )).map((id) => id).splice(0, connectionCount)
```
teogeb:
```suggestion
        const candidates = Array.from(definition.nodeIds.keys()).filter((id) => !this.getConnections(streamPartId).has(id)
        const proxiesToAttempt = sampleSize(candidates, connectionCount)
```

Similar also to line 217.  Uses `sampleSize` from Lodash: https://lodash.com/docs/4.17.15#sampleSize

packages/network/src/logic/proxy/ProxyStreamConnectionClient.ts
```diff
@@ -0,0 +1,289 @@
...
+            this.removeConnection(streamPartId, targetNodeId)
+        }
+    }
+
+    private getConnections(streamPartId: StreamPartID): Map<NodeId, ProxyDirection> {
+        return this.connections.get(streamPartId) ?? new Map()
+    }
+
+    private hasConnection(nodeId: NodeId, streamPartId: StreamPartID): boolean {
+        if (!this.definitions.has(streamPartId)) {
```
teogeb: Would it make sense to remove this check? Then hasConnection and getConnections would have the same logic, i.e. just analyzing `this.connections` field.

packages/network/src/logic/proxy/ProxyStreamConnectionClient.ts
```diff
@@ -0,0 +1,289 @@
...
+            return false
+        }
+        return this.getConnections(streamPartId).has(nodeId)
+    }
+
+    private removeConnection(streamPartId: StreamPartID, nodeId: NodeId): void {
+        if (this.connections.has(streamPartId)) {
```
teogeb: We could use `hasConnection` here?

packages/network/src/logic/proxy/ProxyStreamConnectionClient.ts
```diff
@@ -0,0 +1,289 @@
...
+        }
+    }
+
+    async onNodeDisconnected(streamPartId: StreamPartID, nodeId: NodeId): Promise<void> {
+        this.removeConnection(streamPartId, nodeId)
+        await retry(() => this.updateConnections(streamPartId), 'Updating proxy connections', this.abortController.signal)
+    }
+
+    isProxiedStreamPart(streamPartId: StreamPartID, direction: ProxyDirection): boolean {
+        if (this.definitions.get(streamPartId) && this.getConnections(streamPartId).size > 0) {
```
teogeb: This method answers the question whether some `streamPart` is proxied or not? If we called setProxies for the streamPart but have not yet formed a connections (or the connections are lost and reconnect again), this should return `true`? I.e. should we remove the `this.getConnections(streamPartId).size > 0` check?

packages/network/src/logic/proxy/ProxyStreamConnectionServer.ts
```diff
@@ -0,0 +1,109 @@
...
+export interface ProxyStreamConnectionServerOptions {
+    streamPartManager: StreamPartManager
+    nodeToNode: NodeToNode
+    propagation: Propagation
+    node: Node
+    acceptProxyConnections: boolean
+}
+
+interface ProxyConnection {
+    direction: ProxyDirection // Direction is from the clients point of view
```
teogeb:
```suggestion
    direction: ProxyDirection // Direction is from the client's point of view
```

packages/network/src/logic/proxy/ProxyStreamConnectionServer.ts
```diff
@@ -0,0 +1,109 @@
...
+        })
+
+        this.nodeToNode.on(NodeToNodeEvent.LEAVE_REQUEST_RECEIVED, (message, nodeId) => {
+            this.processLeaveRequest(message, nodeId)
+        })
+    }
+
+    private async processHandshakeRequest(message: ProxyConnectionRequest, nodeId: NodeId): Promise<void> {
+        const streamPartId = message.getStreamPartID()
+        const isAccepted = this.streamPartManager.isSetUp(streamPartId) && this.acceptProxyConnections
```
teogeb: If `this.streamPartManager.isSetUp(streamPartId)` true because of some side effect of the fact that we accept proxy connections to that stream part? Or should we just check this.acceptProxyConnections here and setup the `streamPart` to `streamPartManager` if necessary?

packages/network/src/logic/proxy/ProxyStreamConnectionServer.ts
```diff
@@ -0,0 +1,109 @@
...
+        }
+        this.connections.get(streamPartId)!.set(nodeId, {
+            direction,
+            userId
+        })
+    }
+
+    private processLeaveRequest(message: UnsubscribeRequest, nodeId: NodeId): void {
+        const streamPartId = message.getStreamPartID()
+        if (this.streamPartManager.isSetUp(streamPartId)
```
teogeb: Would it make sense to check whether we have a connection `in this.connections`? Instead of checking the facts from `streamPartManager`.

packages/network/test/integration/proxy-connections.test.ts
```diff
@@ -258,42 +253,112 @@ describe('Proxy connection tests', () => {
...
         await Promise.all([
             waitForEvent(contactNode, NodeEvent.ONE_WAY_CONNECTION_CLOSED),
-            onewayNode.closeProxyConnection(defaultStreamPartId, 'contact-node', ProxyDirection.PUBLISH),
+            onewayNode.setProxies(defaultStreamPartId, [], ProxyDirection.PUBLISH, async () => 'publisher'),
         ])
         // @ts-expect-error private
         expect(contactNode.streamPartManager.isSetUp(defaultStreamPartId)).toBeTrue()
     })
 
+    it('will not connect above set target limit', async () => {
```
teogeb, line 293: Maybe should also have a test for situation where `setProxies` is called with a `connectionCount` that is greater than the `nodeIds` array length? Should fail with a validation error?

packages/network/test/integration/proxy-connections.test.ts
```diff
@@ -68,39 +73,36 @@ describe('Proxy connection tests', () => {
             tracker?.stop(),
             onewayNode?.stop(),
             contactNode?.stop(),
-            contactNode2?.stop(),
+            contactNode2?.stop()
         ])
     })
 
     it('publisher node can form proxy connections', async () => {
-        await onewayNode.openProxyConnection(defaultStreamPartId, 'contact-node', ProxyDirection.PUBLISH, 'publisher')
```
teogeb, line 76: All operations in this test are using this stream part. Maybe it would make sense to drop the `default` prefix because it is a bit misleading.

packages/network/test/integration/proxy-connections.test.ts
```diff
@@ -307,6 +372,5 @@ describe('Proxy connection tests', () => {
                 signature: 'signature'
             }))
         ])
-    }, 20100)
-
+    }, 60000)
```
teogeb: Does the connection process take significantly longer than previously?

packages/client/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Proxy enhancements: It is possible to set a limit for the number of proxy connections while having a larger set of nodes to choose from
```
teogeb: Could mention that there is an change in the API: 
- use `.setProxies` instead of `.openProxyConnections` and `.closeProxyConnections`

How difficult it would be to support the old methods temporarily so that we don't need to do a breaking change? We'd deprecate the methods, but they would work ok (just delegating the call to `setProxies`?)

packages/client/README.md
```diff
@@ -694,16 +694,17 @@ Proxy publishing and subscribing are handled on the network overlay level. This
 ```js
 
 // Open publish proxy to multiple nodes on stream
-await publishingClient.openProxyConnections(stream, ['0x11111...', '0x22222...'], ProxyDirection.PUBLISH)
+await publishingClient.setProxies(stream, ['0x11111...', '0x22222...'], ProxyDirection.PUBLISH)
```
teogeb: The method doesn't handle a stream, but a stream part. We could highlight that fact in the readme, and  also rename the variable.

(One alternative would be to have the first parameter as `StreamDefinition[]` so that we could set proxy for the whole stream easily.)

packages/network/src/logic/proxy/ProxyStreamConnectionServer.ts
juslesan: The server should not accept any proxy connections to streams that it has not already joined. This is one of the core requirements of the feature.

packages/network/src/logic/proxy/ProxyStreamConnectionClient.ts
juslesan: The main problem with `runAndRaceEvents` here is that it does not account for the values that are emitted in the events. If multiple handshakes are performed at once `runAndRace` is unable point the `CONNECTION_ACCEPTED` and `CONNECTION_REJECTED` events to their connection attempts.

packages/network/test/integration/proxy-connections.test.ts
juslesan: The test was intentionally made to last longer to simulate a longer downtime from the server. The test could be made faster by adding configurability to the reconnection interval

packages/client/src/ProxyPublishSubscribe.ts
```diff
@@ -18,17 +18,14 @@ export class ProxyPublishSubscribe {
         this.streamIdBuilder = streamIdBuilder
     }
 
-    async openProxyConnections(streamDefinition: StreamDefinition, nodeIds: string[], direction: ProxyDirection): Promise<void> {
+    async setProxies(
```
teogeb: We could inline this method to `StreamrClient#setProxies`? Then we wouldn't need this class.

packages/client/src/NetworkNodeFacade.ts
```diff
@@ -39,9 +39,13 @@ export interface NetworkNodeStub {
     /** @internal */
     stop: () => Promise<unknown>
     /** @internal */
-    openProxyConnection: (streamPartId: StreamPartID, nodeId: string, direction: ProxyDirection, userId: string) => Promise<void>
-    /** @internal */
-    closeProxyConnection: (streamPartId: StreamPartID, nodeId: string, direction: ProxyDirection) => Promise<void>
+    setProxies: (
```
teogeb, line 41: The `hasProxyConnection` is maybe also an internal method, similarly to `setProxies`?

packages/network/src/logic/Node.ts
```diff
@@ -31,8 +32,6 @@ export enum Event {
     DUPLICATE_MESSAGE_RECEIVED = 'streamr:node:duplicate-message-received',
     NODE_SUBSCRIBED = 'streamr:node:subscribed-successfully',
     NODE_UNSUBSCRIBED = 'streamr:node:node-unsubscribed',
-    PROXY_CONNECTION_ACCEPTED = 'streamr:node:proxy-connection-accepted',
-    PROXY_CONNECTION_REJECTED = 'streamr:node:proxy-connection-rejected',
     ONE_WAY_CONNECTION_CLOSED = 'stream:node-one-way-connection-closed',
```
teogeb, line 35: The ` ONE_WAY_CONNECTION_CLOSED` event is used only in tests? Would it make sense to remove the event and use e.g. waitForCondition in test to wait for connection to close? Something like this:

```
onewayNode.setProxies(streamPartId, [], ProxyDirection.PUBLISH, async () => 'publisher')
await waitForCondition(() => {
    // @ts-expect-error private
    return !onewayNode.streamPartManager.hasOutOnlyConnection(streamPartId, 'contact-node-2')
})
```

If we remove that event, we can also remove `private readonly node: Node` field from `ProxyStreamConnectionClient` and `ProxyStreamConnectionServer`.

packages/network/src/logic/proxy/ProxyStreamConnectionClient.ts
```diff
@@ -0,0 +1,266 @@
...
+            this.emit(Event.CONNECTION_REJECTED, targetNodeId, streamPartId, direction, err)
+        } finally {
+            this.trackerManager.removeSignallingOnlySession(streamPartId, targetNodeId)
+        }
+    }
+
+    private async connectAndHandshake(streamPartId: StreamPartID, targetNodeId: NodeId, direction: ProxyDirection, userId: string): Promise<void> {
+        await this.trackerManager.addSignallingOnlySession(streamPartId, targetNodeId)
+        const trackerId = this.trackerManager.getTrackerId(streamPartId)
+        await withTimeout(this.nodeToNode.connectToNode(targetNodeId, trackerId, false), this.nodeConnectTimeout)
```
teogeb, line 185: The 2nd parameter of `connectToNode` is `trackerAddress`. Is that method parameter maybe wrongly named, or should we here pass Trackers's address (i.e. the websocket endpoint address), or the `trackerId` as the 2nd argument?

packages/network/src/logic/proxy/ProxyStreamConnectionClient.ts
```diff
@@ -0,0 +1,266 @@
...
+            return result
+        } catch (e: any) {
+            logger.warn(`${description} failed, retrying in ${delay} ms`)
+        }
+        await wait(delay, abortSignal)
+    }
+}
+
+export class ProxyStreamConnectionClient extends EventEmitter {
+    private readonly trackerManager: TrackerManager
```
teogeb, line 63: As we have quite many fields in this class, it would maybe improve readability if we'd reordered the fields somehow. E.g. 
1. state-related fields (`definitions`, `connections`), 
2. configuration-related fields (`nodeConnectionTimeout`)
3. dependencies

Could apply the same also for `ProxyStreamConnectionServer`.

packages/network/src/logic/proxy/ProxyStreamConnectionClient.ts
```diff
@@ -0,0 +1,266 @@
...
+            connectionCount: connectionCount ?? nodeIds.length
+        })
+        await this.updateConnections(streamPartId)
+    }
+
+    private async updateConnections(streamPartId: StreamPartID): Promise<void> {
+        await Promise.all(this.getInvalidConnections(streamPartId).map(async (id) => {
+            await this.closeConnection(streamPartId, id)
+        }))
+        const connectionCountDiff =  this.definitions.get(streamPartId)!.connectionCount - this.getConnections(streamPartId).size
```
teogeb:
```suggestion
        const connectionCountDiff = this.definitions.get(streamPartId)!.connectionCount - this.getConnections(streamPartId).size
```

packages/network/src/logic/proxy/ProxyStreamConnectionClient.ts
```diff
@@ -0,0 +1,266 @@
...
+        if (connectionCountDiff > 0) {
+            await this.openRandomConnections(streamPartId, connectionCountDiff)
+        } else if (connectionCountDiff < 0) {
+            await this.closeRandomConnections(streamPartId, -connectionCountDiff)
+        }
+    }
+
+    private getInvalidConnections(streamPartId: StreamPartID): string[] {
+        return Array.from(this.getConnections(streamPartId).keys()).filter((id) =>
+            !this.definitions.get(streamPartId)!.nodeIds.has(id)
```
teogeb: Use helper variable to make it bit easier to read? `const definition = this.definitions.get(streamPartId)!`

packages/network/src/logic/proxy/ProxyStreamConnectionClient.ts
```diff
@@ -0,0 +1,266 @@
...
+            || this.definitions.get(streamPartId)!.direction !== this.getConnections(streamPartId).get(id)
+        )
+    }
+
+    private async openRandomConnections(streamPartId: StreamPartID, connectionCount: number): Promise<void> {
+        const definition = this.definitions.get(streamPartId)!
+        const proxiesToAttempt = sampleSize(Array.from(definition.nodeIds.keys()).filter((id) =>
+            !this.getConnections(streamPartId).has(id)
+        ), connectionCount).map((id) => id)
```
teogeb:
```suggestion
        ), connectionCount)
```

Not needed?

packages/network/src/logic/proxy/ProxyStreamConnectionClient.ts
```diff
@@ -0,0 +1,266 @@
...
+
+    private hasConnection(nodeId: NodeId, streamPartId: StreamPartID): boolean {
+        return this.getConnections(streamPartId).has(nodeId)
+    }
+
+    private removeConnection(streamPartId: StreamPartID, nodeId: NodeId): void {
+        if (this.hasConnection(nodeId, streamPartId)) {
+            this.connections.get(streamPartId)!.delete(nodeId)
+        }
+        this.streamPartManager.removeNodeFromStreamPart(streamPartId, nodeId)
```
teogeb, line 217: Would it make sense for `removeNodeFromStreamPart` to check whether `neighbors`, `inOnly`, `outOnly` are all empty? And in that situation it'd call `removeStreamPart(streamPartId)` automatically.

Also noticed that `unsubscribeFromStreamPartOnNode` doesn't call `removeStreamPart` after it calls `removeNodeFromStreamPart`. Is that use case different from this? Or a bug?

packages/network/src/logic/proxy/ProxyStreamConnectionClient.ts
```diff
@@ -0,0 +1,266 @@
...
+
+        if (this.definitions.get(streamPartId)!.nodeIds.size === 0 && this.getConnections(streamPartId).size === 0) {
+            this.streamPartManager.removeStreamPart(streamPartId)
+        }
+    }
+
+    private processHandshakeResponse(message: ProxyConnectionResponse, nodeId: NodeId): void {
+        const streamPartId = message.getStreamPartID()
+        if (message.accepted) {
+            if (message.direction === ProxyDirection.PUBLISH) {
```
teogeb, line 231: Currently we add an item to this.connections in `initiateConnection` and remove it if the connection attempt fails. Would we have some race conditions if we'd add the connection later, e.g. here when know the connection is now really opened? If that works ok, the state keeping would be bit simpler.

packages/network/src/logic/proxy/ProxyStreamConnectionServer.ts
```diff
@@ -0,0 +1,110 @@
...
+    private readonly propagation: Propagation
+
+    constructor(opts: ProxyStreamConnectionServerOptions) {
+        this.streamPartManager = opts.streamPartManager
+        this.nodeToNode = opts.nodeToNode
+        this.node = opts.node
+        this.acceptProxyConnections = opts.acceptProxyConnections
+        this.propagation = opts.propagation
+        this.connections = new Map()
+        this.nodeToNode.on(NodeToNodeEvent.PROXY_CONNECTION_REQUEST_RECEIVED, (message,  nodeId) => {
```
teogeb:
```suggestion
        this.nodeToNode.on(NodeToNodeEvent.PROXY_CONNECTION_REQUEST_RECEIVED, (message, nodeId) => {
```

packages/network/src/logic/proxy/ProxyStreamConnectionServer.ts
```diff
@@ -0,0 +1,110 @@
...
+        })
+
+        this.nodeToNode.on(NodeToNodeEvent.LEAVE_REQUEST_RECEIVED, (message, nodeId) => {
+            this.processLeaveRequest(message, nodeId)
+        })
+    }
+
+    private async processHandshakeRequest(message: ProxyConnectionRequest, nodeId: NodeId): Promise<void> {
+        const streamPartId = message.getStreamPartID()
+        const isAccepted =  this.acceptProxyConnections && this.streamPartManager.isSetUp(streamPartId)
```
teogeb:
```suggestion
        const isAccepted = this.acceptProxyConnections && this.streamPartManager.isSetUp(streamPartId)
```

packages/network/src/logic/proxy/ProxyStreamConnectionServer.ts
```diff
@@ -0,0 +1,110 @@
...
+        if (isAccepted) {
+            if (message.direction === ProxyDirection.PUBLISH) {
```
teogeb, line 58: Could move `this.addConnection(streamPartId, nodeId, message.direction message.userId)` out from the `if`-blocks as that is done in both branches.

packages/network/src/logic/proxy/ProxyStreamConnectionServer.ts
```diff
@@ -0,0 +1,110 @@
...
+        if (!this.connections.has(streamPartId)) {
+            this.connections.set(streamPartId, new Map())
+        }
+        this.connections.get(streamPartId)!.set(nodeId, {
+            direction,
+            userId
+        })
+    }
+
+    private processLeaveRequest(message: UnsubscribeRequest, nodeId: NodeId): void {
```
teogeb, line 78: If we remove the emitting of `ONE_WAY_CONNECTION_CLOSED` as suggested in a separate comment, this method is basically just a delegator. Could  we always call  `removeConnection`, without checking `hasConnection`? 

If that is the case, we could maybe inline this method?

packages/network/src/logic/TrackerConnector.ts
```diff
@@ -23,7 +23,7 @@ export class TrackerConnector {
     private maintenanceTimer?: NodeJS.Timeout | null
     private readonly maintenanceInterval: number
     private connectionStates: Map<TrackerId, ConnectionState>
-    private readonly signallingOnlyTrackers: Set<TrackerId>
+    private readonly signallingOnlySessions: Map<StreamPartID, Set<string>>
```
teogeb:
```suggestion
    private readonly signallingOnlySessions: Map<StreamPartID, Set<NodeId>>
```

packages/network/test/integration/proxy-connections.test.ts
```diff
@@ -68,93 +73,82 @@ describe('Proxy connection tests', () => {
...
         ])
     })
 
     it('publisher node can form proxy connections', async () => {
-        await onewayNode.openProxyConnection(defaultStreamPartId, 'contact-node', ProxyDirection.PUBLISH, 'publisher')
-        await onewayNode.openProxyConnection(defaultStreamPartId, 'contact-node-2', ProxyDirection.PUBLISH, 'publisher')
+        await onewayNode.setProxies(streamPartId, ['contact-node', 'contact-node-2'], ProxyDirection.PUBLISH, async () => 'publisher')
         // @ts-expect-error private
-        expect(onewayNode.streamPartManager.getOutboundNodesForStreamPart(defaultStreamPartId)).toContainValues(['contact-node', 'contact-node-2'])
+        expect(onewayNode.streamPartManager.getOutboundNodesForStreamPart(streamPartId)).toContainValues(['contact-node', 'contact-node-2'])
```
teogeb: The assertions are quite long. We could maybe use helper function to make these more readable?

E.g.
```
const state = getState(onewayNode)
expect(state.outboundNodes).toContainValues(['contact-node', 'contact-node-2'])
expect(state.neighbours).toBeArrayOfSize(0)
```

The `getState` could be something like this:
```
const getState = (node: NetworkNode) => {
    // @ts-expect-error private
    const streamPartManager = node.streamPartManager
    if (streamPartManager.isSetUp(streamPartId)) {
        return {
            isSetUp: true,
            inboundNodes: streamPartManager.getInboundNodesForStreamPart(streamPartId),
            outboundNodes: streamPartManager.getOutboundNodesForStreamPart(streamPartId),
            neighbours: streamPartManager.getNeighborsForStreamPart(streamPartId),
            hasInOnlyConnection: (otherNodeId: NodeId) => streamPartManager.hasInOnlyConnection(streamPartId, otherNodeId),
            hasOutOnlyConnection: (otherNodeId: NodeId) => streamPartManager.hasOutOnlyConnection(streamPartId, otherNodeId),
        }
    } else {
        return {
            isSetUp: false
        }
    }
}
```

packages/network/test/integration/proxy-connections.test.ts
```diff
@@ -24,18 +25,22 @@ describe('Proxy connection tests', () => {
...
         })
+    }
+
+    beforeEach(async () => {
+        tracker = await startTestTracker({
+            port: 30353
+        })
+        trackerInfo = tracker.getConfigRecord()
+        contactNode = createContactNode()
         await contactNode.start()
```
teogeb:
```suggestion
        contactNode.start()
```

Also at line 53

packages/network/test/integration/proxy-connections.test.ts
```diff
@@ -68,93 +73,82 @@ describe('Proxy connection tests', () => {
...
+        expect(onewayNode.streamPartManager.getInboundNodesForStreamPart(streamPartId)).toContainValues(['contact-node', 'contact-node-2'])
         // @ts-expect-error private
-        expect(onewayNode.streamPartManager.getNeighborsForStreamPart(defaultStreamPartId)).toBeArrayOfSize(0)
+        expect(onewayNode.streamPartManager.getNeighborsForStreamPart(streamPartId)).toBeArrayOfSize(0)
     })
 
     it('publisher node can close proxy connections', async () => {
-        await onewayNode.openProxyConnection(defaultStreamPartId, 'contact-node-2', ProxyDirection.PUBLISH, 'publisher')
-        await onewayNode.openProxyConnection(defaultStreamPartId, 'contact-node', ProxyDirection.PUBLISH, 'publisher')
+        await onewayNode.setProxies(streamPartId, ['contact-node', 'contact-node-2'], ProxyDirection.PUBLISH, async () => 'publisher')
```
teogeb: User ids are not relevant in these tests. Maybe we could hide that aspect into a helper function:
```
const getUserId = async () => 'mock-user-id'
```

```suggestion
        await onewayNode.setProxies(streamPartId, ['contact-node', 'contact-node-2'], ProxyDirection.PUBLISH, getUserId)
```

packages/network/src/logic/proxy/ProxyStreamConnectionClient.ts
juslesan: The second parameter is indeed the tracker's id (Ethereum address in production)

packages/network/src/logic/Node.ts
juslesan: I would prefer to keep the code as is. There is really no way in the current network to know if an interaction between network participants has been fulfilled successfully outside of emitting events upon completion. `waitForCondition` can be used to check that a node has reached a desired state but it will not ensure that an operation has been successfully completed.

packages/network/src/logic/proxy/ProxyStreamConnectionClient.ts
juslesan: Yes the use case is different and `unsubscribeFromStreamPartOnNode` is poorly named.

packages/network/src/logic/proxy/ProxyStreamConnectionServer.ts
juslesan: Removed checking hasConnection, see comments related to `ONE_WAY_CONNECTION_CLOSED` above.

teogeb (approved):

PR #1081, opened 2023-02-13
feat(client): Key request timeout for v5

teogeb:
If client can't fetch a group key from a publisher, it timeouts with an error. Before this PR the request didn't have timeout and therefore it could freeze the whole subscribe/resend process.

The default timeout is 30s, and can be configured with `keyRequestTimeout` configuration option.


juslesan (approved):
LGTM, it might be useful to have a way to clean up the timeouts but perhaps it not strictly needed in this case

PR #1085, opened 2023-02-15
feat(client): NET-932 Use all RPC providers

teogeb:
All contract providers are used to query the tracker registry, storage node registry and stream storage registry. Previously only first contract provider was used.

Stream registry contract queries are done in sequence. Previously contracts were queried in parallel using `Promise.any()`.

### Other small refactoring

Inlined `Ethereum#getStreamRegistryChainProvider` and renamed `Ethereum#getAllStreamRegistryChainProviders`


packages/client/src/utils/promises.ts
```diff
@@ -270,3 +270,21 @@ export const withThrottling = (fn: (...args: any[]) => Promise<any>, maxInvocati
...
+        throw new Error('no tasks')
+    }
+    let firstError: any
+    for (const fn of fns) {
+        try {
+            const promise = fn()
+            return await promise
+        } catch (e: any) {
+            if (firstError === undefined) {
+                firstError = e
```
juslesan, line 285: Wondering if it could be useful to report all of the errors for users. Although, most of the time if all of the requests fail it's likely due to the same reason

juslesan (approved):

PR #1099, opened 2023-02-25
refactor(broker): Remove `ApiAuthenticator` rule object

teogeb:
Removed `ApiAuthenticator` rule object. Instead we use `ApiAuthentication` data structure + `isValidAuthentication` helper function.


juslesan (approved):
LGTM

PR #1109, opened 2023-03-02
feat(network, client): webrtc port range and max message size configurability

juslesan:
## Summary

Configurability for the UDP port range configurability and max message size in webrtc in the network and the client.

## Changes

-  Added configurability to the NetworkNode for setting port ranges for webrtc connections.
-  Added an option for setting max sizes for webrtc messages in the NetworkNode

### DIFF
--- a/packages/client/src/Config.ts
+++ b/packages/client/src/Config.ts
@@ -10,7 +10,7 @@ import CONFIG_SCHEMA from './config.schema.json'
 import { TrackerRegistryRecord } from '@streamr/protocol'
 import { LogLevel } from '@streamr/utils'
 
-import { IceServer, Location } from '@streamr/network-node'
+import { IceServer, Location, WebRtcPortRange } from '@streamr/network-node'
 import type { ConnectionInfo } from '@ethersproject/web'
 import { generateClientId } from './utils/utils'
 
@@ -95,6 +95,8 @@ export interface StreamrClientConfig {
         newWebrtcConnectionTimeout?: number
         webrtcDatachannelBufferThresholdLow?: number
         webrtcDatachannelBufferThresholdHigh?: number
+        webrtcDatachannelMaxMessageSize?: number
+        webrtcPortRange?: WebRtcPortRange
         /**
          * The maximum amount of outgoing messages to be buffered on a single WebRTC connection.
          */
@@ -104,6 +106,7 @@ export interface StreamrClientConfig {
         rttUpdateTimeout?: number
         iceServers?: ReadonlyArray<IceServer>
         location?: Location
+
     }
 
     contracts?: {
--- a/packages/client/src/Config.ts
+++ b/packages/client/src/Config.ts
@@ -106,7 +106,6 @@ export interface StreamrClientConfig {
         rttUpdateTimeout?: number
         iceServers?: ReadonlyArray<IceServer>
         location?: Location
-
     }
 
     contracts?: {
--- a/packages/client/src/Config.ts
+++ b/packages/client/src/Config.ts
@@ -95,7 +95,7 @@ export interface StreamrClientConfig {
         newWebrtcConnectionTimeout?: number
         webrtcDatachannelBufferThresholdLow?: number
         webrtcDatachannelBufferThresholdHigh?: number
-        webrtcDatachannelMaxMessageSize?: number
+        webrtcMaxMessageSize?: number
         webrtcPortRange?: WebRtcPortRange
         /**
          * The maximum amount of outgoing messages to be buffered on a single WebRTC connection.
--- a/packages/network/src/connection/webrtc/WebRtcConnection.ts
+++ b/packages/network/src/connection/webrtc/WebRtcConnection.ts
@@ -21,15 +21,15 @@ export interface ConstructorOptions {
     routerId: string
     iceServers: ReadonlyArray<IceServer>
     pingInterval: number
-    bufferThresholdLow?: number
-    bufferThresholdHigh?: number
-    maxMessageSize?: number
-    newConnectionTimeout?: number
-    maxPingPongAttempts?: number
-    flushRetryTimeout?: number
     messageQueue: MessageQueue<string>
     deferredConnectionAttempt: DeferredConnectionAttempt
     portRange: WebRtcPortRange
+    maxMessageSize: number
+    bufferThresholdLow?: number
+    bufferThresholdHigh?: number
+    newConnectionTimeout?: number
+    maxPingPongAttempts?: number
+    flushRetryTimeout?: number
 }
 
 export interface WebRtcPortRange {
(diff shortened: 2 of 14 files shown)


packages/network/src/connection/webrtc/WebRtcEndpoint.ts
```diff
@@ -90,6 +94,8 @@ export class WebRtcEndpoint extends EventEmitter implements IWebRtcEndpoint {
         this.sendBufferMaxMessageCount = webrtcSendBufferMaxMessageCount
         this.disallowPrivateAddresses = webrtcDisallowPrivateAddresses
         this.maxMessageSize = maxMessageSize
+        this.portRangeBegin = portRangeBegin
+        this.portRangeEnd = portRangeEnd
```
harbu: I believe you still need to pass these values to the constructor of `WebRtcConnection` down in method `createConnection` in this class?

harbu (approved):

packages/network/src/createNetworkNode.ts
```diff
@@ -24,6 +24,9 @@ export interface NetworkNodeOptions extends AbstractNodeOptions {
     trackerConnectionMaintenanceInterval: number
     webrtcDisallowPrivateAddresses: boolean
     acceptProxyConnections: boolean
+    webrtcDatachannelMaxMessageSize?: number
+    webrtcDatachannelPortRangeBegin?: number
```
teogeb: To make the configuration easy to use, I'd maybe use only one parameter `webrtcPortRange` which would enclose the lower and upper bound into an object:
```
{
  min: number
  max: number
}
```

Most likely end-users would almost always define both the upper and lower bound? 

There are already quite many options for `createNetworkNode`, and therefore I'd like to keep the number of options minimal. When we have less parameters, it is easier for end-users to find relevant config options that they need to adjust. The same applies when this configurability is applied to `StreamrClient`.

Also dropped `datachannel` from the name as datachannel is an implementation specific detail? The user is just configuring WebRTC port ranges and we must deliver the configuration wherever is applicable (e.g. to the `datachannel` library. Therefore also `webrtcDatachannelMaxMessageSize` could be shortened to `webrtcMaxMessageSize`.

(Noticed that we need to pass this to `node-datachannel` as two parameters. But that internal implementation is not relevant for end-users, and therefore we can just use one parameter here if we want.)

teogeb (changes requested):

teogeb:
In https://github.com/streamr-dev/network/pull/1027 we all `NetworkNodeOptions` required. 

If we want to continue that style, these new options would also be required, and the defaults would be in client's config schema and in the `TEST_CONFIG` constant.

packages/network/src/connection/webrtc/WebRtcConnection.ts
```diff
@@ -29,6 +29,12 @@ export interface ConstructorOptions {
     flushRetryTimeout?: number
     messageQueue: MessageQueue<string>
     deferredConnectionAttempt: DeferredConnectionAttempt
+    portRange: WebRtcPortRange
+}
+
+export interface WebRtcPortRange {
+    begin: number
+    end: number
 }
```
teogeb: I'd rename the fields to min and max so that we are consistent with our naming. 

Typically when we have some kind of range with upper and/or lower bounds, we name fields using min and max prefixes. E.g. `minTimestamp`+`maxTimestamp`, `minPropagationTargets`, `maxMessages`, `maxNeighborsPerNode`...

```suggestion
export interface WebRtcPortRange {
    min: number
    max: number
}
```

packages/client/src/Config.ts
```diff
@@ -104,6 +106,7 @@ export interface StreamrClientConfig {
         rttUpdateTimeout?: number
         iceServers?: ReadonlyArray<IceServer>
         location?: Location
+
```
```suggestion
```

packages/network/test/utils.ts
```diff
@@ -50,6 +50,7 @@ export const createTestWebRtcEndpoint = (
         webrtcDatachannelBufferThresholdHigh ?? TEST_CONFIG.webrtcDatachannelBufferThresholdHigh,
         TEST_CONFIG.webrtcSendBufferMaxMessageCount,
         webrtcDisallowPrivateAddresses ?? false,
+        { begin: 6000, end: 65535 }
```
teogeb: This could refer to `TEST_CONFIG.webrtcPortRange`, similarly to other options?

packages/network/src/connection/webrtc/WebRtcConnection.ts
```diff
@@ -125,12 +132,13 @@ export abstract class WebRtcConnection extends ConnectionEmitter {
         messageQueue,
         deferredConnectionAttempt,
         pingInterval,
+        portRange = { begin: 6000, end: 65535 },
```
teogeb: Does the default have any effect as `portRange` is a required field in `ConstructorOptions`?

packages/network/src/connection/webrtc/WebRtcEndpoint.ts
```diff
@@ -72,6 +73,7 @@ export class WebRtcEndpoint extends EventEmitter implements IWebRtcEndpoint {
         webrtcDatachannelBufferThresholdHigh: number,
         webrtcSendBufferMaxMessageCount: number,
         webrtcDisallowPrivateAddresses: boolean,
+        portRange: WebRtcPortRange,
```
teogeb, line 76: Maybe `maxMessageSize` could be a required parameter (similarly to all other fields)? It is a required parameter of `NetworkNodeOptions` and therefore we always have a value there.

packages/network/src/connection/webrtc/WebRtcConnection.ts
```diff
@@ -125,12 +132,13 @@ export abstract class WebRtcConnection extends ConnectionEmitter {
...
         bufferThresholdHigh = 2 ** 17,
         bufferThresholdLow = 2 ** 15,
         newConnectionTimeout = 15000,
         maxPingPongAttempts = 5,
         flushRetryTimeout = 500,
-        maxMessageSize = 1048576,
+        maxMessageSize = 1048576
```
teogeb: It would make sense to have this as a required field `ConstructorOptions`, as it is also required in `NetworkNodeOptions`? If we do so, we don't need a default value here (and can be closer to the goal of keeping all defaults in client's config schema and `TEST_CONFIG`).

packages/network/src/createNetworkNode.ts
```diff
@@ -77,7 +86,9 @@ export const createNetworkNode = ({
         webrtcDatachannelBufferThresholdLow,
         webrtcDatachannelBufferThresholdHigh,
         webrtcSendBufferMaxMessageCount,
-        webrtcDisallowPrivateAddresses
+        webrtcDisallowPrivateAddresses,
+        webrtcPortRange,
+        webrtcDatachannelMaxMessageSize
```
teogeb: Does the `webrtcDatachannelMaxMessageSize` limit the size of all WebRTC messages? Or are there non-datachannel messages, to which this limit does not apply? 

If the limit is generic, and `datachannel` is just an implementation detail, I'd shorten the variable to `webrtcMaxMessageSize` To me it sounds that it would be more understandable that way. If I've understood correctly, the `datachannel` term doesn't mean anything for our end-users, and therefore we should not use that term in the field name.

teogeb (changes requested):

teogeb (approved):
LGTM 👍 Could update client's changelog

PR #1158, opened 2023-03-31
feat: add tsdoc documentation to Config

harbu:
## Summary

Add tsdoc documentation to main `Config` object.

### DIFF
--- a/packages/client/src/Config.ts
+++ b/packages/client/src/Config.ts
@@ -50,38 +50,91 @@ export interface EthereumNetworkConfig {
 export interface StreamrClientConfig {
     /** Custom human-readable debug id for client. Used in logging. */
     id?: string
-    logLevel?: LogLevel
+
     /**
-    * Authentication: identity used by this StreamrClient instance.
-    * Can contain member privateKey or (window.)ethereum
+     * Override the default logging level.
+     */
+    logLevel?: LogLevel
+
+    /**
+    * The Ethereum identity to be used by the client. Either a private key
+     * or a window.ethereum object.
     */
     auth?: PrivateKeyAuthConfig | ProviderAuthConfig
 
-    /** Attempt to order messages */
+    /**
+     * Set to true to enable message ordering.
+     *
+     * Due to the distributed nature of the network, messages may occasionally
+     * arrive out-of-order to the client. By enabling message ordering, the
+     * client will reorder the received messages in the intended order for you.
+     *
+     * */
     orderMessages?: boolean
-    gapFill?: boolean
-    maxGapRequests?: number
-    retryResendAfter?: number
-    gapFillTimeout?: number
 
     /**
-     * Message encryption/decryption
+     * Set to true to enable gap filling.
+     *
+     * Some messages may occasionally not reach the client due to networking
+     * issues. Missing messages form gaps that are often detectable and
+     * retrievable on demand. By enabling gap filling, you enable the client to
+     * detect and fix gaps automatically for you.
+     */
+    gapFill?: boolean
+
+    /**
+     * When gap filling is enabled, this option controls the maximum amount of
+     * times a gap will try to be filled before giving up and proceeding
+     * forwards.
+     */
+    maxGapRequests?: number
+
+    /**
+     * When gap filling is enabled and a gap is encountered, the amount of time
+     * to wait before attempting to _actively_ fill in the gap.
+     *
+     * Rationale: the data may just be arriving out-of-order and the missing
+     * messages may be on their way. For efficiency, it makes sense to wait a
+     * little before actively attempting to fill in the gap, as this involves
+     * a resend request to (and response from) a storage node.
+     */
+    gapFillTimeout?: number
+
+    retryResendAfter?: number
+
+    /**
+     * Controls how messages encryption and decryption should be handled.
      */
     encryption?: {
         /**
          * Enable experimental Lit Protocol key exchange.
          *
-         * When enabled encryption key storing and fetching will be primarily done through the Lit Protocol and
+         * When enabled encryption key storing and fetching will primarily be done through the Lit Protocol and
          * secondarily through the standard Streamr key-exchange system.
          */
         litProtocolEnabled?: boolean
+
         /**
          * Enable log messages of the Lit Protocol library to be printed to stdout.
          */
         litProtocolLogging?: boolean
+
         // TODO keyRequestTimeout and maxKeyRequestsPerSecond config options could be applied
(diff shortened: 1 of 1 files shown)


packages/client/src/Config.ts
```diff
@@ -50,64 +50,241 @@ export interface EthereumNetworkConfig {
...
+         *
+         * In streams with 1000+ publishers, it is important to limit the amount
+         * of control message traffic that gets generated to avoid network buffers
+         * from overflowing.
+         */
         maxKeyRequestsPerSecond?: number
     }
 
+    /**
+     * These settings apply to the network node of the client.
```
fonty1:
```suggestion
     * These settings determine how the client performs and interacts with the Streamr Network.
```

fonty1: Not exactly sure how to best describe this section, I was thrown off a bit with "network node of the client" though

packages/client/src/Config.ts
```diff
@@ -50,64 +50,241 @@ export interface EthereumNetworkConfig {
...
+        /**
+         * The maximum outgoing message size accepted by WebRTC connections.
+         * Messages exceeding the maximum size are simply discarded.
+         */
         webrtcMaxMessageSize?: number
+
+        /**
+         * Defines a custom UDP port range to be used for WebRTC connections.
+         * This port range should not be restricted by enclosing firewalls
+         * or VPC configurations.
```
fonty1: Virtual Port-Channel or Virtual private cloud? Probs best to add the full name here for searchability also.

packages/client/src/Config.ts
```diff
@@ -50,64 +50,241 @@ export interface EthereumNetworkConfig {
...
 
-    /** Attempt to order messages */
+    /**
+     * Set to true to enable message ordering.
+     *
+     * Due to the distributed nature of the network, messages may occasionally
+     * arrive to the client out-of-order. By enabling message ordering, the
+     * client will reorder received messages to the intended order for you.
+     *
+     * */
```
teogeb: The first line is kind of trivial (guessable from the field name). We could consider if it is more straightforward to write this without introducing thee extra term. 

I.e. skip the middle step in this chain: "I want to client to reorder messages" -> "I need to enable message ordering" -> "I set this to `true`".

```suggestion
    /**
  
     * Due to the distributed nature of the network, messages may occasionally
     * arrive to the client out-of-order. Set this option to `true` if you want that the 
     * client reorders the received messages to the intended order.
     *
     * */
```

packages/client/src/Config.ts
```diff
@@ -50,64 +50,241 @@ export interface EthereumNetworkConfig {
...
-    retryResendAfter?: number
+
+    /**
+     * When gap filling is enabled and a gap is encountered, this option
+     * defines the amount of time to wait before attempting to _actively_ fill
+     * in the gap.
+     *
+     * Rationale: data may just be arriving out-of-order and the missing
+     * message(s) may be on their way. For efficiency, it makes sense to wait a
+     * little before actively attempting to fill in a gap, as this involves
```
teogeb, line 97: Could move this description of active filling to `maxGapRequest` section as the concept of active filling is coined there for the first time.

packages/client/src/Config.ts
```diff
@@ -50,64 +50,241 @@ export interface EthereumNetworkConfig {
...
+     * a resend request / response interaction with a storage node.
+     */
     gapFillTimeout?: number
 
     /**
-     * Message encryption/decryption
+     * When gap filling is enabled and a gap is encountered, a resend request
```
teogeb, line 103: This is not always done, only after `gapFillTimeout`. 

We could say something like "If we make a storage node request in order to fill a gap, and the request fails, the client can retry the fetch from (another?) storage. This option controls how long to wait for a storage node response from the storage node before proceeding to the next attempt.

The number is in milliseconds."

(The unit of time, i.e. milliseconds, could be included in each of the sections where applicable: `retryResendAfter`, `keyRequestTimeout`, `newWebrtcConnectionTimeout`).

packages/client/src/Config.ts
```diff
@@ -50,64 +50,241 @@ export interface EthereumNetworkConfig {
...
+    /**
+     * Controls how messages encryption and decryption should be handled and
+     * how encryption keys should be exchanged.
      */
     encryption?: {
         /**
          * Enable experimental Lit Protocol key exchange.
          *
-         * When enabled encryption key storing and fetching will be primarily done through the Lit Protocol and
+         * When enabled encryption key storing and fetching will primarily be done through the Lit Protocol and
```
teogeb: Link to LIT protocol web site?

packages/client/src/Config.ts
```diff
@@ -50,64 +50,241 @@ export interface EthereumNetworkConfig {
...
     network?: {
+        /**
+         * The network-wide identifier of this node. Should be unique
+         * within the Streamr Network.
+         */
         id?: string
+
+        /**
+         * Whether to accept proxy connections. Enabling this option allows
+         * the network node to act as proxy on behalf of other nodes / clients.
```
```suggestion
         * this network node to act as proxy on behalf of other nodes / clients.
```

packages/client/src/Config.ts
```diff
@@ -50,64 +50,241 @@ export interface EthereumNetworkConfig {
...
         disconnectionWaitTime?: number
+
+        /**
+         * Defines how often to ping connected nodes to determine connection
+         * aliveness.
+         */
         peerPingInterval?: number
+
+        /**
+         * Determines how often, at most, to include RTT statistics in status
```
```suggestion
         * Determines how often, at most, to include round-trip time statistics in status
```

packages/client/src/Config.ts
harbu: Well... there is a network node within the client and these settings are passed to it when constructing it. However, maybe the user doesn't want to know this level of detail so I think your suggestion makes sense here.

packages/client/src/Config.ts
harbu: Changed wording to emphasize conditionality of resend request:
> When gap filling is enabled and a gap is encountered, a resend request
_may eventually_ be sent ...

harbu: I would leave out the details "of requesting from another storage node".

packages/client/src/Config.ts
harbu: There is no guarantee of a reading order here so each section should sort of stand on its own (to the extent that this is possible without excessive repeating). If anything I would imagine the `gapFill` option to be referenced first.

packages/client/src/Config.ts
harbu: Added units to descriptions

PR #1168, opened 2023-04-07
Fix IP addresses in the ConfigTest.ts file

victorshevtsov:
## Summary

The `ConfigTest.ts` file has a mixture of IP addresses of `127.0.0.1` and `10.200.10.1`
That works fine if you run the tests on the same machine you run `streamr-docker-dev` (both IPs are accessible).
But that does not work if `streamr-docker-dev` is running separately and the ports are tunneled to the dev machine.

## Changes
There are just 3 replacements of `10.200.10.1` with `127.0.0.1`


harbu (approved):
LGTM! Thanks for the contribution! :+1:

PR #1174, opened 2023-04-12
feat(broker): [NET-945] ask for beneficiary address in broker config wizard

harbu:
## Summary

Add beneficiary address prompt to broker config wizard.

An example prompt:
```
> Do you want to participate in mining and staking? Yes
> By default the generated / imported Ethereum identity will be used for staking, would you like to change this to a different address (i.e. set a beneficiary address)? Yes
> Please provide a beneficiary (Ethereum) address 0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
``` 

### DIFF
--- a/packages/broker/src/config/ConfigWizard.ts
+++ b/packages/broker/src/config/ConfigWizard.ts
@@ -160,7 +160,7 @@ const createPluginPrompts = (): Array<inquirer.Question | inquirer.ListQuestion 
         type: 'confirm',
         name: 'wantToSetBeneficiaryAddress',
         // eslint-disable-next-line max-len
-        message: `By default the generated / imported Ethereum identity will be used for staking, would you like to change this to a different address (i.e. set a beneficiary address)?`,
+        message: 'It is recommended to set a separate beneficiary address for security reasons. Would you like to set one?',
         default: false,
         when: (answers: inquirer.Answers) => {
             return answers.enableMinerPlugin
@@ -170,7 +170,7 @@ const createPluginPrompts = (): Array<inquirer.Question | inquirer.ListQuestion 
     const beneficiaryAddressPrompt: inquirer.DistinctQuestion = {
         type: 'input',
         name: 'beneficiaryAddress',
-        message: `Please provide a beneficiary (Ethereum) address`,
+        message: 'Please provide a beneficiary address (public key)',
         when: (answers: inquirer.Answers) => {
             return answers.wantToSetBeneficiaryAddress
         },
--- a/packages/broker/src/config/ConfigWizard.ts
+++ b/packages/broker/src/config/ConfigWizard.ts
@@ -238,11 +238,8 @@ export const getConfig = (privateKey: string, pluginsAnswers: PluginAnswers): an
     })
 
     if (pluginsAnswers.enableMinerPlugin) {
-        config.plugins.brubeckMiner = {}
-        if (pluginsAnswers.beneficiaryAddress !== undefined) {
-            config.plugins.brubeckMiner = {
-                beneficiaryAddress: pluginsAnswers.beneficiaryAddress
-            }
+        config.plugins.brubeckMiner = {
+            beneficiaryAddress: pluginsAnswers.beneficiaryAddress
         }
     }
 
--- a/packages/broker/src/config/ConfigWizard.ts
+++ b/packages/broker/src/config/ConfigWizard.ts
@@ -24,6 +24,9 @@ export interface PluginAnswers extends Answers {
     mqttPort?: string
     httpPort?: string
     enableMinerPlugin: boolean
+    wantToSetBeneficiaryAddress?: boolean
+    beneficiaryAddress?: string
+
 }
 
 export interface StorageAnswers extends Answers {
@@ -153,7 +156,41 @@ const createPluginPrompts = (): Array<inquirer.Question | inquirer.ListQuestion 
         default: true
     }
 
-    return [selectApiPluginsPrompt, ...portPrompts, minerPluginPrompt]
+    const wantToSetBeneficiaryAddressPrompt: inquirer.DistinctQuestion = {
+        type: 'confirm',
+        name: 'wantToSetBeneficiaryAddress',
+        // eslint-disable-next-line max-len
+        message: `By default the generated / imported Ethereum identity will be used for staking, would you like to change this to a different address (i.e. set a beneficiary address)?`,
+        default: false,
+        when: (answers: inquirer.Answers) => {
+            return answers.enableMinerPlugin
+        }
+    }
+
+    const beneficiaryAddressPrompt: inquirer.DistinctQuestion = {
+        type: 'input',
+        name: 'beneficiaryAddress',
+        message: `Please provide a beneficiary (Ethereum) address`,
+        when: (answers: inquirer.Answers) => {
+            return answers.wantToSetBeneficiaryAddress
+        },
+        validate: (input: string): string | boolean => {
+            try {
+                toEthereumAddress(input)
+                return true
+            } catch (e: any) {
+                return 'Invalid Ethereum address provided.'
+            }
+        }
+    }
+
+    return [
+        selectApiPluginsPrompt,
+        ...portPrompts,
+        minerPluginPrompt,
+        wantToSetBeneficiaryAddressPrompt,
+        beneficiaryAddressPrompt
+    ]
 }
 
 export const PROMPTS = {
@@ -202,6 +239,11 @@ export const getConfig = (privateKey: string, pluginsAnswers: PluginAnswers): an
 
     if (pluginsAnswers.enableMinerPlugin) {
         config.plugins.brubeckMiner = {}
+        if (pluginsAnswers.beneficiaryAddress !== undefined) {
+            config.plugins.brubeckMiner = {
+                beneficiaryAddress: pluginsAnswers.beneficiaryAddress
+            }
+        }
     }
 
     return config
--- a/packages/broker/CHANGELOG.md
+++ b/packages/broker/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add beneficiary address prompt to config wizard
+
 ### Changed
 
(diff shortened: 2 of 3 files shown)


packages/broker/src/config/ConfigWizard.ts
```diff
@@ -202,6 +239,11 @@ export const getConfig = (privateKey: string, pluginsAnswers: PluginAnswers): an
 
     if (pluginsAnswers.enableMinerPlugin) {
         config.plugins.brubeckMiner = {}
+        if (pluginsAnswers.beneficiaryAddress !== undefined) {
+            config.plugins.brubeckMiner = {
+                beneficiaryAddress: pluginsAnswers.beneficiaryAddress
+            }
+        }
     }
```
teogeb:
```suggestion
    if (pluginsAnswers.enableMinerPlugin) {
        config.plugins.brubeckMiner = {
            beneficiaryAddress: pluginsAnswers.beneficiaryAddress
        }
    }
```

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -153,7 +156,41 @@ const createPluginPrompts = (): Array<inquirer.Question | inquirer.ListQuestion
         default: true
     }
 
-    return [selectApiPluginsPrompt, ...portPrompts, minerPluginPrompt]
+    const wantToSetBeneficiaryAddressPrompt: inquirer.DistinctQuestion = {
+        type: 'confirm',
+        name: 'wantToSetBeneficiaryAddress',
+        // eslint-disable-next-line max-len
+        message: `By default the generated / imported Ethereum identity will be used for staking, would you like to change this to a different address (i.e. set a beneficiary address)?`,
```
teogeb: Could make the message shorter so that it is more understandable. E.g. we know from privateKeyAnswers whether we generated or imported the key, and choose message:
`By default the generated Ethereum identity...` or `By default the imported Ethereum identity`.

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -153,7 +156,41 @@ const createPluginPrompts = (): Array<inquirer.Question | inquirer.ListQuestion
...
+        default: false,
```
teogeb, line 164: Not sure if it is possible to make this prompt less complex... Is the core question something like this:
```
You can mine the tokens to your own wallet (the imported Ethereum identity provided at step X) or some other wallet. That wallet should contain the stake.`
```
Could also consider using enum instead of boolean?
```
Choose wallet: own / other
```

teogeb:
One option which we talked with @harbu: Maybe we could re-structure some options under "Advanced customization". 

The wizard would ask only basic questions: the Ethereum address, mining participation and which API plugins to enable. Then it would ask whether whether the user would want do advanced customization or not.

If the user answers yes, we'd ask plugin ports, beneficiary address etc. Most of the users would just say that they don't need advanced config and therefore we don't even need to bother them about the advanced topics.

packages/broker/src/config/ConfigWizard.ts
lelas: "Ethereum identity" could cause some confusion ("Ethereum address" ditto), considering far from everyone understands that Ethereum and Polygon addresses are one and the same. Change to "Polygon wallet address"?

harbu:
> One option which we talked with @harbu: Maybe we could re-structure some options under "Advanced customization".
> 
> The wizard would ask only basic questions: the Ethereum address, mining participation and which API plugins to enable. Then it would ask whether whether the user would want do advanced customization or not.
> 
> If the user answers yes, we'd ask plugin ports, beneficiary address etc. Most of the users would just say that they don't need advanced config and therefore we don't even need to bother them about the advanced topics.

As you said "Most of the users would just say that they don't need advanced config and therefore we don't even need to bother them about the advanced topics."

So maybe we should leave beneficiary address and ports at the very end? 

Example prompt of skipping advanced stuff:
```
> Do you want to generate a new Ethereum private key or import an existing one? Generate
> We strongly recommend backing up your private key. It will be written into the config file, but would you also like to see this sensitive 
information on screen now? No
> Select the plugins to enable websocket
> Do you want to participate in mining and staking? Yes
> Do you want to customize the chosen settings (for advanced users)? No
> Select a path to store the generated config in /home/harbu/.streamr/config/default.json
```

Example prompt of when user wants to set advanced stuff:
```
> Do you want to generate a new Ethereum private key or import an existing one? Generate
> We strongly recommend backing up your private key. It will be written into the config file, but would you also like to see this sensitive 
information on screen now? No
> Select the plugins to enable websocket
> Do you want to participate in mining and staking? Yes
> Do you want to customize the chosen settings (for advanced users)? Yes
> Provide a port for the websocket Plugin [Enter for default: 7170] 7170
> Provide the  beneficiary address [Enter for default: <GENERATED_KEY_EARLIER>] 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
> Select a path to store the generated config in /home/harbu/.streamr/config/default.json
```

Thoughts @lelas @fonty1 ?

packages/broker/src/config/ConfigWizard.ts
fonty1: A few more wordy suggestions,

`The Ethereum identity that has been generated or imported will be utilized for staking by default. Do you want to modify this and assign a different address as the beneficiary?`

Or if we know the generated/import decision,
`The Ethereum identity that has been generated will be utilized for staking by default. Do you want to modify this and assign a different address as the beneficiary?`

`The Ethereum identity that has been imported will be utilized for staking by default. Do you want to modify this and assign a different address as the beneficiary?`

packages/broker/src/config/ConfigWizard.ts
fonty1: > Change to "Polygon wallet address"?

Hmm, yeah good point. This is probably easier to understand.

fonty1:
> So maybe we should leave beneficiary address and ports at the very end?

I like shuffling the plugin/port decisions to the advanced section, but I feel strongly that the beneficiary address prompt should be apart of the basic config because we do recommend it in the docs and it's arguably the safer/preffered way to stake for most users IMHO. Maybe @lelas has more in the field experience towards that assertion.

lelas:
> > So maybe we should leave beneficiary address and ports at the very end?
> 
> I like shuffling the plugin/port decisions to the advanced section, but I feel strongly that the beneficiary address prompt should be apart of the basic config because we do recommend it in the docs and it's arguably the safer/preffered way to stake for most users IMHO. Maybe @lelas has more in the field experience towards that assertion.

Fully agree! And since plugin/ports are unnecessary for node operators, it makes perfect sense to keep that in the advanced section.

harbu:
Based on this discussion I would agree (and perhaps you all would too?) that we should keep the beneficiary address prompt out of the advanced section (if such section were created).

What is left then is to finalize the the wording for the prompt for the beneficiary address. The question here is whether it is better for the description to be long and informative or shorter but quicker to glance on repetitive use of the wizard?

@teogeb @lelas @fonty1

fonty1:
> What is left then is to finalize the the wording for the prompt for the beneficiary address. The question here is whether it is better for the description to be long and informative or shorter but quicker to glance on repetitive use of the wizard?

I'm fine with either, slight bias towards long and informative.

harbu:
```
> It is recommended to set a separate beneficiary address for security reasons. Would you like to set one? Y
> Please provide a beneficiary address 0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
```
How about the following @fonty1 ?

lelas:
> What is left then is to finalize the the wording for the prompt for the beneficiary address. The question here is whether it is better for the description to be long and informative or shorter but quicker to glance on repetitive use of the wizard?

I'd prefer longer and more informative, void of ambiguity and using well-known terms. Ideally very clearly stating it's the public key they need to use/enter/paste, not the private key, taking the Dumbest Assumable User into consideration.

Would it make sense to show the beneficiary address question before "Do you want to generate a new Ethereum private key or import an existing one?". If they use a beneficiary address, simply generate a private (signing) key without asking? If not, ask if they want to generate or import.

"Please provide a beneficiary address" -> "Please provide a beneficiary address (public key)"

harbu:
> > What is left then is to finalize the the wording for the prompt for the beneficiary address. The question here is whether it is better for the description to be long and informative or shorter but quicker to glance on repetitive use of the wizard?
> 
> I'd prefer longer and more informative, void of ambiguity and using well-known terms. Ideally very clearly stating it's the public key they need to use/enter/paste, not the private key, taking the Dumbest Assumable User into consideration.
> 
> Would it make sense to show the beneficiary address question before "Do you want to generate a new Ethereum private key or import an existing one?". If they use a beneficiary address, simply generate a private (signing) key without asking? If not, ask if they want to generate or import.
> 
> "Please provide a beneficiary address" -> "Please provide a beneficiary address (public key)"

How about the following then:
```
> It is recommended to set a separate beneficiary address for security reasons. Would you like to set one? Y
> Please provide a beneficiary address (public key) 0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
```

ping (at least) @lelas ?

fonty1:
> > > What is left then is to finalize the the wording for the prompt for the beneficiary address. The question here is whether it is better for the description to be long and informative or shorter but quicker to glance on repetitive use of the wizard?
> > 
> > 
> > I'd prefer longer and more informative, void of ambiguity and using well-known terms. Ideally very clearly stating it's the public key they need to use/enter/paste, not the private key, taking the Dumbest Assumable User into consideration.
> > Would it make sense to show the beneficiary address question before "Do you want to generate a new Ethereum private key or import an existing one?". If they use a beneficiary address, simply generate a private (signing) key without asking? If not, ask if they want to generate or import.
> > "Please provide a beneficiary address" -> "Please provide a beneficiary address (public key)"
> 
> How about the following then:
> 
> ```
> > It is recommended to set a separate beneficiary address for security reasons. Would you like to set one? Y
> > Please provide a beneficiary address (public key) 0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
> ```
> 
> ping (at least) @lelas ?

LGTM

PR #1196, opened 2023-04-19
refactor: move event utility functions to `utils` package

teogeb:
Move two `ObservableEventEmitter` class and `initEventGateway` utility function from `client` package to `utils` package.


packages/utils/src/events.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+}
+
+/*
+ * Initializes a gateway which can produce events to the given emitter. The gateway is running
+ * when there are any listeners for the given eventName: the start() callback is called
+ * when a first event listener for the event name is added, and the stop() callback is called
+ * when the last event listener is removed.
+ */
+export const initEventGateway = <E extends Events<E>, T extends keyof E, P>(
```
harbu: I'd almost think about moving this to its own file.

harbu (approved):

PR #1197, opened 2023-04-19
feat(client): [NET-964] Listener for `createStream` event

teogeb:
## Summary

Observe `StreamCreated`  contract event and emit it as `createStream` client event:

```ts
client.on('createStream', (event: StreamCreationEvent) => {
    console.log(event.streamId)
    console.log(event.metadata)
    console.log(event.blockNumber)
})
```

We do we emit event if metadata is not valid for a stream.

## Refactoring

The new `initContractEventGateway` utility function is used also in `StreamStorageRegistry`.


harbu:
> Do we emit any event if metadata is not valid?

It would make sense to stay consistent with how e.g. `getStream` works.

packages/client/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add support for observing stream creation events
```
```suggestion
- Add support for observing stream creation events with `.on('createStream')`
```

harbu (approved):

PR #1205, opened 2023-04-20
chore(broker): improvements to `WebsocketPlugin`

harbu:
## Summary

Add a few improvements to `WebsocketPlugin`.

Partly fixes NET-966.

## Changes

- Add more logging statements to help users see what happened & is happening.
- Attach a `socketId` to log messages to differentiate between WebSocket connections in the presence of multiple.
- Add error handling around `client#subscribe` to avoid unhandled promise rejections.
- Add cleanup logic so that subscriptions are unsubscribed from when the (a) WebSocket client disconnects or (b) Subscribing to multiple partitions fails.

### DIFF
--- a/packages/broker/src/plugins/websocket/WebsocketServer.ts
+++ b/packages/broker/src/plugins/websocket/WebsocketServer.ts
@@ -6,7 +6,7 @@ import { once } from 'events'
 import { Socket } from 'net'
 import qs, { ParsedQs } from 'qs'
 import StreamrClient from 'streamr-client'
-import { Logger } from '@streamr/utils'
+import { Logger, randomString } from '@streamr/utils'
 import { addPingSender, addPingListener, Connection } from './Connection'
 import { ApiAuthentication, isValidAuthentication } from '../../apiAuthentication'
 import { PublishConnection } from './PublishConnection'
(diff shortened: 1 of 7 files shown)


packages/broker/src/plugins/websocket/WebsocketServer.ts
```diff
@@ -85,11 +90,18 @@ export class WebsocketServer {
...
+            connection.init(ws, socketId, this.streamrClient, payloadFormat).then(() => {
+                addPingListener(ws)
+                if (this.pingSendInterval !== 0) {
+                    addPingSender(ws, socketId, this.pingSendInterval, this.disconnectTimeout)
+                }
+                return
+            }, (err) => {
+                logger.warn('Close connection', { socketId, reason: err?.message })
+                ws.close()
+            })
```
teogeb: The return statement at line `100` is not needed?

Maybe `async await` and `setImmediate` would make the flow more straightforward:
```suggestion
            setImmediate(async () => {
                try {
                    await connection.init(ws, socketId, this.streamrClient, payloadFormat)
                } catch (err) {
                    logger.warn('Close connection', { socketId, reason: err?.message })
                    ws.close()
                    return
                }
                addPingListener(ws)
                if (this.pingSendInterval !== 0) {
                    addPingSender(ws, socketId, this.pingSendInterval, this.disconnectTimeout)
                }
            })
```

harbu: Fixed in e579e9d0403b351ad93afb490afde67d55ba40b4 but without setImmediate.

PR #1210, opened 2023-04-21
rm husky and commitling

harbu:
## Summary

Remove husky and commit linting from project. No need to enforce on a per-commit basis as we use squash merge. 

(Feel free to install a local commit linter if you want.)


teogeb (approved):
👍

PR #1215, opened 2023-04-21
refactor(broker): use raw subscribe in subscriber plugin

harbu:
## Summary

Use new client feature raw subscription 482e471fb1823fc0558fb9cdd271d64e06605f8d to simplify `SubscriberPlugin`. Notice deprecation of `subscriptionRetryInterval`.

Partly fixes NET-954.

`StreamrClient` subscriptions should stay in effect until unsubscribe or the client is destroyed. This is why I removed the subscribe interval logic. There is a slight risk that a subscription could be lost earlier if we have a bug in client. But I think "dogfooding" here makes sense since such a (theoretical) bug would _should_ be catched and fixed!


packages/broker/src/plugins/subscriber/SubscriberPlugin.ts
```diff
@@ -9,60 +10,39 @@ interface ConfigStream {
...
-        this.subscriptionIntervalRef = setTimeout(() => this.subscriptionIntervalFn(), this.subscriptionRetryInterval)
+        this.subscriptions = await Promise.all(this.streamParts.map(({ id, partition }) => (
+            this.streamrClient.subscribe({ id, partition, raw: true })
+        )))
     }
 
     async start(): Promise<void> {
         await this.subscribeToStreamParts()
-        this.subscriptionIntervalRef = setTimeout(() => this.subscriptionIntervalFn(), this.subscriptionRetryInterval)
         logger.info('Started subscriber plugin')
```
teogeb, line 27: Could move this logging to the start of the method, and use the same wording as in the `stop` method (i.e. `Start subscriber plugin`).

packages/broker/src/plugins/subscriber/SubscriberPlugin.ts
```diff
@@ -9,60 +10,39 @@ interface ConfigStream {
...
-    private readonly subscriptionRetryInterval: number
-    private subscriptionIntervalRef: NodeJS.Timeout | null
+    private readonly streamParts: { id: string, partition: number }[]
+    private subscriptions: Subscription[] = []
 
     constructor(options: PluginOptions) {
         super(options)
         this.streamParts = this.pluginConfig.streams.map((stream) => {
-            return toStreamPartID(toStreamID(stream.streamId), stream.streamPartition)
+            const streamPart = toStreamPartID(toStreamID(stream.streamId), stream.streamPartition)
```
teogeb: Do we actually need this "transformation" to type safe streamId + partition? We could use the config array directly in the `this.streamrClient.subscribe` call. That way the the values are validated by the client immediately when plugin's `start()` method is called.

teogeb (changes requested):
Could add a changelog entry and/or automatic migration which removed `subscriptionRetryInterval` config option?

> StreamrClient subscriptions should stay in effect until unsubscribe or the client is destroyed. This is why I removed the subscribe interval logic.

This is a great improvement! 👍

packages/utils/src/allOrCleanup.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Given an iterable of promises, settles them one at a time.
+ * If all of them resolve, returns the list of values.
+ * If one of them rejects, provides a callback for cleaning up the
+ * preceding (already resolved) values.
+ */
+export async function allOrCleanup<T>(
```
teogeb: Nice utility 👍 Could consider if we want to name this to e.g. to `promiseTransaction`? It either succeeds or fails, just like a DB transaction, and can be rolled back.

packages/broker/src/plugins/subscriber/SubscriberPlugin.ts
harbu: Removed.

packages/utils/src/allOrCleanup.ts
harbu: Renamed.

packages/broker/src/plugins/subscriber/SubscriberPlugin.ts
harbu: Not going to do in this PR. Reason: naming convention with rest of plugins.

PR #1230, opened 2023-04-24
Docs

nicoburkart:
## Summary

Merged the old documentation repository (Docusaurus docs) into the monorepo

## Changes

Removed the old ./git folder inside of the old documentation folder. Git flow now works with this repository.

Added the docs to root/docs after consultation with Eric. 

Renamed old "docs" folder in root repository to "network-docs"

We need to add automation for deployment to prod.

## Note for reviewer:
You can ignore all the markdown and yml files
Please focus mostly on the merge of the repository. Is the folder structure right, in your opinion?


docs/.nvmrc
```diff
@@ -0,0 +1 @@
+v16.16.0
```
harbu: Do you think this would work with node.js v18 as well? We are using v18 in the rest of the monorepo so would be handy to be able to use same node installation here too.
```suggestion
v18.16.0
```

package.json
```diff
@@ -49,5 +49,8 @@
         "typescript": "^4.9.3",
         "yarn": "^1.22.17",
         "zx": "^1.14.1"
+    },
+    "devDependencies": {
+        "typedoc": "^0.24.1"
     }
 }
```
harbu: A bit unclear to me, what's the intention of this change?

harbu (commented):
> Please focus mostly on the merge of the repository. Is the folder structure right, in your opinion?

I think it makes sense the way it is setup. 

An argument could be made for or against including the generated API docs markdown in the change sets (i.e. `docs/docs/api/`). I'm fine with including them as is done here.

What is the process for updating those API markdown files btw? Does it happen automatically when running `npm run build`  in the docs directory?

teogeb:
> An argument could be made for or against including the generated API docs markdown in the change sets (i.e. `docs/docs/api/`)

I'd leave out the generated API docs. It is just duplicate information as the files can be generated with `npm run docs` whenever we need them.

As the public documentation always describes the latest published client/broker, we need some way to postpone version-specific document updates in the automated deployment process.  The API docs are always in this category of postponed publish process, and therefore we need to generate them only very rarely.

The current process is that we write examples and TSDocs about new features in PRs and merge those changes to `main` when the feature is ready. If possible, we'd like to keep that same process. 

> We need to add automation for deployment to prod.

That automated process should maybe have at least 3 different "topics":
- general documentation, which can be published immediately
- client-specific docs (examples, API docs etc.), which can be published when new client version is released
- broker-specific docs (examples), which can be published when new broker version is released

nicoburkart:
> What is the process for updating those API markdown files btw? Does it happen automatically when running npm run build in the docs directory?

Yes, right now the process is that after the build the API docs get generated based on whatever is defined in the client package. It is possible to prevent this.

In addition, after the build is finished, there is a custom script we run to refine the output from the docusaurus plugin to fit our needs. This process is going to be implemented in the deployment process/pipeline and will be automated.

harbu:
Hmm.. looks like tsdoc doesn't support the following attributes:
```

> streamr-client@8.3.0 docs
> typedoc --options typedoc.js

[error] Tried to set an option (entryDocument) that was not declared. You may have meant:
	entryPoints
	out
[error] Tried to set an option (hideInPageTOC) that was not declared. You may have meant:
	name
	theme
	cname
	hideGenerator
	help
	media
	emit
	readme
```

PR #1231, opened 2023-04-24
fix: use merge instead of object spread

teogeb:
### Issue

It is very easy to merge two JavaScript objects incorrectly using the spread operator. Typically when we want merge some defaults and overrides, we use `const result = { ...defaults, ...overrides }`. While it works, it may not produce the expected value if any property of `overrides` is `undefined`.

For all _missing_ properties in the `overrides` object, the functionality is correct. But if there is an object property and the value of that property is `undefined`, the merged object doesn't fallback to the default value. 

E.g.
```ts
interface SomeOptions {
    a: string
}
const defaultValues = {
    a: 'foobar'
}
const overrideValues: Partial<SomeOptions> = {
    a: undefined,
}
const result = {
    ...defaultValues,
    ...overrideValues
}
console.log(result) // { a: undefined }
```

Any property which is annotated with question mark (directly, or via `Partial`) can have explicit value of `undefined`.

Note also that e.g. in VS Code and IntelliJ IDEA the inferred type of `result` is shown as:
```ts
{
    a: string;
}
```

(There is a TypeScript configuration option to control this: https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes)

### Fix

We need to use something else than the spread operator.

In this PR we added a new `merge` utility function to `@streamr/utils`. It does a simple non-recursive merge and doesn't mutate any of the argument objects. But it skips all `undefined` properties.

Alternatively we could use `merge` from `lodash` if recursive merge is needed (and if mutating the target argument is ok).

With both methods we get valid merged object:
```ts
console.log(merge(defaultValues, overrideValues))  // { a: 'foobar' }
```

### DIFF
--- a/packages/protocol/test/unit/protocol/message_layer/StreamMessage.test.ts
+++ b/packages/protocol/test/unit/protocol/message_layer/StreamMessage.test.ts
@@ -10,7 +10,7 @@ import '../../../../src/protocol/message_layer/StreamMessageSerializerV32'
 import { Serializer } from '../../../../src/Serializer'
 import { toStreamID } from '../../../../src/utils/StreamID'
 import { StreamPartIDUtils } from '../../../../src/utils/StreamPartID'
-import merge from 'lodash/merge'
+import { merge } from '@streamr/utils'
 
 const content = {
     hello: 'world',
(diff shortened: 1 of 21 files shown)


packages/protocol/test/unit/protocol/message_layer/StreamMessage.test.ts
```diff
@@ -10,6 +10,7 @@ import '../../../../src/protocol/message_layer/StreamMessageSerializerV32'
 import { Serializer } from '../../../../src/Serializer'
 import { toStreamID } from '../../../../src/utils/StreamID'
 import { StreamPartIDUtils } from '../../../../src/utils/StreamPartID'
+import merge from 'lodash/merge'
```
harbu: lodash on purpose?

harbu (approved):

PR #1240, opened 2023-04-25
feat(client): [NET-968] Missing partition count in metadata

teogeb:
When we parse stream data, we fallback to partition count 1 if there is no information about the partition count in the contract metadata:
- if the metadata JSON doesn't have `partitions` field
- if the metadata is not a JSON

In all other cases the functionality is the same as previously: i.e. invalid partition count throws an error. That error is now `StreamrClientError` instead of generic `Error` so that we can provide an error code to end users.

### Open questions

- What abstraction level we should use in our error codes? Is `INVALID_STREAM_METADATA` too specific, or should we just have some generic `VALIDATION_FAILED` error code?

### DIFF
--- a/packages/client/src/Stream.ts
+++ b/packages/client/src/Stream.ts
@@ -24,6 +24,7 @@ import { Subscription } from './subscribe/Subscription'
 import { LoggerFactory } from './utils/LoggerFactory'
 import { Message } from './Message'
 import { convertStreamMessageToMessage } from './Message'
+import { StreamrClientError } from './StreamrClientError' 
 
 export interface StreamMetadata {
     /**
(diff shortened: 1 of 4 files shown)


packages/client/src/Stream.ts
```diff
@@ -283,14 +284,26 @@ export class Stream {
...
+            try {
+                ensureValidStreamPartitionCount(json.partitions)
+                return json
+            } catch (_ignored) {
+                throw err
+            }
+        } else {
+            return {
+                partitions: 1
+            }
```
harbu: Rest of data should be retained.

packages/client/src/Stream.ts
```diff
@@ -283,14 +284,26 @@ export class Stream {
...
+        let json
         try {
-            // TODO we could pick the fields of StreamMetadata explicitly, so that this
-            // object can't contain extra fields
-            const json = JSON.parse(metadata)
-            ensureValidStreamPartitionCount(json.partitions)
-            return json
-        } catch (error) {
-            throw new Error(`Could not parse properties from onchain metadata: ${metadata}`)
+            json = JSON.parse(metadata)
```
harbu, line 297: Empty string handling (and test for it)

harbu (approved):

harbu:
> What abstraction level we should use in our error codes? Is INVALID_STREAM_METADATA too specific, or should we just have some generic VALIDATION_FAILED error code?

Specific feels better here.

PR #1241, opened 2023-04-25
chore(client): Remove obsolete `CacheFn` utility

teogeb:
Remove obsolete `CacheFn` utility. We always use `CacheAsyncFn` instead. 

Also removed obsolete `opts` rest parameters from `CacheAsyncFn`.


harbu (approved):
👍 💯

PR #1264, opened 2023-04-26
refactor: browser-test-runner package

harbu:
## Summary

Created a new sub-package `browser-test-runner` and moved all karma and electron related test configuration files and dependencies there. This avoids having duplicated configuration files and dependencies.

Each sub-package retains a `karma.config.js` that calls code in `browser-test-runner` and overrides default settings as needed.

*Somewhat breaking change*: network-node is no longer includes webpack bundle when published.

- The `test-browser` scripts of sub-packages still have a reference to `karma` executable even though they don't directly import it. However things work in practice due to the executable being present in the monorepo shared node_modules.
  - I tried to implement running karma programmatically (http://karma-runner.github.io/6.4/dev/public-api.html) from `browser-test-runner` package to get rid of the direct `karma` executable references, but I couldn't get things to behave similar to `karma start` within reasonable effort. (E.g. random errors, Webpack not starting)
- It is possible to write karma and webpack configs in TypeScript. 


packages/browser-test-runner/webpack.config.js
```diff
@@ -0,0 +1,76 @@
...
+                }),
+                new webpack.ProvidePlugin({
+                    process: 'process/browser'
+                }),
+                new webpack.ProvidePlugin({
+                    Buffer: ['buffer', 'Buffer']
+                }),
+            ],
+            resolve: {
+                extensions: ['.tsx', '.ts', '.js'],
```
teogeb: We don't maybe need `.tsx` handling as we don't have any?

packages/browser-test-runner/karma.config.js
```diff
@@ -0,0 +1,50 @@
...
+                CustomElectron: {
+                    base: 'Electron',
+                    browserWindowOptions: {
+                        webPreferences: {
+                            contextIsolation: false,
+                            preload: __dirname + '/preload.js',
+                            webSecurity: false,
+                            sandbox: false
+                        },
+                        show: false // set to true to show the electron window
```
teogeb: Would it useful to have a variable, e.g.`DEBUG_MODE` , which we'd use here and at line 43? It would be `false` by default, but we'd be able to toggle debug mode on and off by modifying the variable.

teogeb (approved):

packages/browser-test-runner/karma.config.js
harbu: Added

packages/browser-test-runner/webpack.config.js
harbu: Added

PR #1265, opened 2023-04-27
chore(client): run tests using karma / electron

harbu:
## Summary

Run the unit, integration, and end-to-end tests of package `client` under Karma (Electron).

## Changes
- Remove `babel-plugin-lodash` from webpack (we use features of `lodash` it strips away)
- Remove `concurrently` package
- GitHub action jobs for `client-browser` and `client-browser-e2e`
- Disable random test order in browser test runner (some of our tests, regrettably, assume sequential execution)
- Add `toThrowStreamrError` matcher to browser test runner
- Add `it.skip` and `describe.skip` support to browser test runner
- Three npm scripts: `test-browser-unit`, `test-browser-integration`, `test-browser-end-to-end`
- Remove `jest.setTimeout` and replace with `it(..., ..., TIMEOUT)` instead (jest.setTimeout doesn't work in browser env)
- Create smoke tests for webpack minified output
- Add `test-utils` utilities `testOnlyInNodeJs` and `describeOnlyInNodeJs` for declaring tests that should be skipped in Electron
- Change the way contract members are enumared thru (new helper function `getAllPropertyNames`) for browser compatibility

- Rename utility functions `testOnlyInNodeJs` => `it.onlyInNodeJs` and `describeOnlyInNodeJs` => `describe.onlyInNodeJs`
- Remove `git-revision-webpack-plugin` from webpack
- There is a couple of `TODO` tests that should work but don't. Figure out why


packages/client/src/utils/uuid.ts
```diff
@@ -3,7 +3,7 @@ import uniqueId from 'lodash/uniqueId'
 
 export const SEPARATOR = '-'
 
-let UUID: string
+let UUID: string | undefined
```
harbu: Needed?

packages/client/test/test-utils/utils.ts
```diff
@@ -45,8 +45,9 @@ export const uid = (prefix?: string): string => counterId(`p${process.pid}${pref
 
 const getTestName = (module: NodeModule): string => {
     const fileNamePattern = new RegExp('.*/(.*).test\\...')
-    const groups = module.filename.match(fileNamePattern)
-    return (groups !== null) ? groups[1] : module.filename
+    const moduleFilename = (module.filename ?? module.id)
```
```suggestion
    const moduleFilename = (module.filename ?? module.id) // browser has no filename
```

packages/client/test/integration/Resends2.test.ts
```diff
@@ -428,6 +428,9 @@ describe('Resends2', () => {
     })
 
     it('decodes resent messages correctly', async () => {
```
harbu, line 430: For these cases it would be nice to have a special `it` that skips the test properly when running in Electron.

packages/client/test/integration/gap-fill.test.ts
```diff
@@ -65,6 +65,9 @@ describe('gap fill', () => {
     })
 
     it('failing storage node', async () => {
+        if (isRunningInElectron()) { // TODO: why doesn't this work in electron?
```
harbu: Possible bug?

packages/browser-test-runner/custom-matcher.js
```diff
@@ -0,0 +1,47 @@
...
+ * TODO: Get rid of this file before merging, no copy-paste code please.
+ */
+const formError = (description, expected, actual) => {
+    return {
+        pass: false,
+        message: () => `${description}\nExpected: ${printExpected(expected)}\nReceived: ${printReceived(actual)}`
+    }
+}
+
+const toThrowStreamrError = (actual, expectedError) => {
```
harbu: Figure this out

packages/browser-test-runner/package.json
```diff
@@ -21,6 +21,7 @@
     "buffer": "^6.0.3",
     "electron": "^28.1.3",
     "expect": "^29.6.2",
+    "jest-matcher-utils": "^29.5.0",
```
harbu, line 24: Needed?

packages/client/src/utils/contract.ts
```diff
@@ -139,8 +139,16 @@ export const createDecoratedContract = <T extends Contract>(
...
+    function getAllPropertyNames(obj: object): string[] {
+        const proto = Object.getPrototypeOf(obj)
+        const inherited = (proto) ? getAllPropertyNames(proto) : []
+        return [...new Set(Object.getOwnPropertyNames(obj).concat(inherited))]
+    }
+
     // copy own properties and inherited properties (e.g. contract.removeAllListeners)
-    for (const key in contract) {
+    // eslint-disable-next-line no-prototype-builtins
+    for (const key of getAllPropertyNames(contract)) {
```
harbu, line 151: Is this needed?

packages/client/webpack.config.js
```diff
@@ -63,7 +57,7 @@ module.exports = (env, argv) => {
         },
         plugins: [
             gitRevisionPlugin,
-            new webpack.EnvironmentPlugin({
+            new webpack.EnvironmentPlugin({ // TODO: remove?
```
harbu:
```suggestion
            new webpack.EnvironmentPlugin({
```

packages/client/src/utils/uuid.ts
harbu: Was not needed, resolving.

packages/client/src/utils/contract.ts
harbu: Is needed, otherwise in browser [end-to-end tests](https://github.com/streamr-dev/network/actions/runs/7712499768/job/21020246975):

```
1) event listener: picks up add and remove events
     StorageNodeRegistry
     TypeError: opts.sourceEmitter.on is not a function
    at /tmp/_karma_webpack_610790/commons.js:66166:24
    at EventEmitter.<anonymous> (/tmp/_karma_webpack_610790/commons.js:100933:18)
    at EventEmitter.emit (/tmp/_karma_webpack_610790/commons.js:133517:33)
    at StreamrClientEventEmitter.on (/tmp/_karma_webpack_610790/commons.js:99667:21)
    at StreamrClient.on (/tmp/_karma_webpack_610790/commons.js:53192:25)
    at UserContext.<anonymous> (/tmp/_karma_webpack_610790/commons.js:68473:20)
    at Generator.next (<anonymous>)
    at asyncGeneratorStep (/tmp/_karma_webpack_610790/commons.js:68399:103)
    at _next (/tmp/_karma_webpack_610790/commons.js:68400:194)

2) listener
     StreamRegistry createStream
     TypeError: opts.sourceEmitter.on is not a function
    at /tmp/_karma_webpack_610790/commons.js:66166:24
    at EventEmitter.<anonymous> (/tmp/_karma_webpack_610790/commons.js:100933:18)
    at EventEmitter.emit (/tmp/_karma_webpack_610790/commons.js:133517:33)
    at StreamrClientEventEmitter.on (/tmp/_karma_webpack_610790/commons.js:99667:21)
    at StreamrClient.on (/tmp/_karma_webpack_610790/commons.js:53192:25)
    at UserContext.<anonymous> (/tmp/_karma_webpack_610790/commons.js:68726:14)
    at Generator.next (<anonymous>)
    at asyncGeneratorStep (/tmp/_karma_webpack_610790/commons.js:68652:103)
    at _next (/tmp/_karma_webpack_610790/commons.js:68653:194)
    at /tmp/_karma_webpack_610790/commons.js:68653:364
```

packages/browser-test-runner/package.json
harbu: it is for custom matchers, checked with grep.

packages/browser-test-runner/custom-matcher.js
harbu: Figured out.

packages/client/test/integration/StreamrClient.test.ts
```diff
@@ -150,7 +150,7 @@ describe('StreamrClient', () => {
         })
     })
 
-    describe('utf-8 encoding', () => {
```
juslesan, line 153: I suppose UTF-8 encoding should work in the browser as well. We don't validate it in the browser tests now at all since the input in this test and `integration/Resends2.test.ts` is read from a file. I suppose  the file is useful since the content covers a wide array of characters.

packages/client/test/end-to-end/publish-subscribe-via-proxy.test.ts
```diff
@@ -1,17 +1,17 @@
...
 import { StreamPermission } from '../../src/permission'
 import { createTestClient, createTestStream } from '../test-utils/utils'
 
-const TIMEOUT = 15 * 1000
+const TIMEOUT = 30 * 1000
 const SUBSCRIBE_WAIT_TIME = 2000
 const WEBSOCKET_PORT = 14231
 
-describe('publish/subscribe via proxy', () => {
+describeOnlyInNodeJs('publish/subscribe via proxy', () => { // Cannot run proxy server in browser
```
juslesan, line 14: Why is it not possible to run the proxy server in the electron? The `preload.js` in `browser-test-runner` it is possible to run WS servers in electron. (ConfigTest.ts enables WS servers for all other end-to-end test anyhow)

harbu: Is there a trick to get it working? When I enable the test I get the following errors:
```
1) publish
     publish/subscribe via proxy
     Error: Timeout - Async function did not complete within 30000ms (custom timeout)
    at <Jasmine>

2) subscribe
     publish/subscribe via proxy
     Error: Timeout - Async function did not complete within 30000ms (custom timeout)
    at <Jasmine>
     TypeError: Cannot read properties of undefined (reading 'content')
    at UserContext.<anonymous> (/tmp/_karma_webpack_25212/commons.js:69299:32)
    at Generator.next (<anonymous>)
    at asyncGeneratorStep (/tmp/_karma_webpack_25212/commons.js:69252:103)
    at _next (/tmp/_karma_webpack_25212/commons.js:69253:194)
```

packages/client/test/integration/StreamrClient.test.ts
harbu: Both tests read the file. I made it so that in the browser we use a hardcoded string instead of reading the file: 930233bf20d95463234bbbea84277e024f8cecb0. I'm guessing it more or less covers the UTF-8 case.

packages/client/test/end-to-end/publish-subscribe-via-proxy.test.ts
juslesan: Yup, I had a closer look and realised that the client webpack always starts the node in browser mode. This means that starting the WS server is disabled even in Electron. Probably best to keep the test disabled in the browser then.

juslesan: The test could pass if we allow proxy connections to be formed with webrtc. (Requires the proxied client to join the DHT temporarily or at least to open a connection to an entry point to forward signaling messages)

juslesan (approved):

packages/browser-test-runner/karma-setup.js
```diff
@@ -36,22 +38,29 @@ jest.runOnlyPendingTimers = timers.runOnlyPendingTimers
...
 jest.useRealTimers = timers.useRealTimers
+
 // eslint-disable-next-line no-undef
 jasmine.DEFAULT_TIMEOUT_INTERVAL = 15000
+
 // eslint-disable-next-line no-underscore-dangle
 jest._checkFakeTimers = timers._checkFakeTimers
 
+// eslint-disable-next-line no-undef
+jasmine.getEnv().configure({ random: false }) // disable random test order
```
teogeb, line 49: Add TODO to re-enable random order in the future

packages/browser-test-runner/karma-setup.js
```diff
@@ -36,22 +38,29 @@ jest.runOnlyPendingTimers = timers.runOnlyPendingTimers
...
+
 Object.assign(jest, timers)
 
 expect.extend(jestExtendedMatchers)
 
 // Add missing Jest functions
 window.test = window.it
 window.test.each = (inputs) => (testName, test) =>
-    inputs.forEach((args) => window.it(testName, () => test(...args)))
+    inputs.forEach((args) => window.it(format(testName, args), () => test(args)))
```
teogeb, line 58: Is the removal of `...` intentional? Does it fix a bug in this test setup script?

packages/client/karma-setup.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * This setup script is executed indirectly by Karma via the `browser-test-runner`
+ * package. See karma-config.js in that package for more details.
+ */
+
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const { toThrowStreamrError } = require('./test/test-utils/customMatchers')
+
+if (toThrowStreamrError !== undefined) {
```
teogeb, line 9: In which situation `toThrowStreamrError` can be `undefined`?

packages/client/src/utils/contract.ts
```diff
@@ -139,8 +139,16 @@ export const createDecoratedContract = <T extends Contract>(
     const result: any = {
         eventEmitter
     }
+
+    function getAllPropertyNames(obj: object): string[] {
```
teogeb, line 143: This change is to improve browser compatibility? Could mention this change in the PR description (just in case this introduces some regression which we need to debug later)

packages/client/test/browser-smoke-test/browser.js
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable no-undef, @typescript-eslint/no-require-imports */
+
+const TARGET_SUBSCRIBE_MSG_COUNT = 50
+const RESEND_SUBSCRIBE_MSG_COUNT = TARGET_SUBSCRIBE_MSG_COUNT - 10
```
teogeb, line 4: Add comment why we don't expect to be able to receive all published messages?

packages/client/test/end-to-end/publish-subscribe-via-proxy.test.ts
teogeb: Maybe add a TODO to enable this test in a separate PR

packages/client/test/integration/Resends2.test.ts
```diff
@@ -428,9 +428,10 @@ describe('Resends2', () => {
...
-        const publishedMessage = Msg({
-            content: fs.readFileSync(path.join(__dirname, '../data/utf8Example.txt'), 'utf8')
-        })
+        const content = isRunningInElectron()
+            ? Buffer.from('Hęłłö Wørłd! 1234 @#$$%^&*() 你好，世界 Привет, мир こんにちは世界', 'utf-8').toString('utf-8')
+            : fs.readFileSync(path.join(__dirname, '../data/utf8Example.txt'), 'utf8')
```
teogeb, line 433: We could use the hardcoded data also in NodeJS and remove the `utf8Example.txt`?

packages/client/karma-setup.js
harbu: There's something weird about the Karma runner here that requires this undefined check.

Tests that use `toThrowStreamrError` definitely have it available as they pass without problem. If I don't do this setup I will see that those tests would error. So we are definitely extending jest matches with something that is defined.

_However_ at the very end of the test suite, Karma will report that we are trying to extend an `undefined` matcher. My guess is that Karma is for some reason running this setup script many times and perhaps at the very end the  `./test/test-utils/customMatchers` is somehow unloaded and thus the variable becomes `undefined`. 

But it's a bit of a mystery and imo not worth delving more time into.

packages/browser-test-runner/karma-setup.js
harbu: I believe the comment here is enough? I would prefer to add a TODO comment on those individual tests.

packages/client/src/utils/contract.ts
harbu: Added mention to PR description

packages/client/test/integration/Resends2.test.ts
harbu: Actually let's create express server and serve this file up that way in both instances.

PR #1290, opened 2023-05-02
fix(network): do not send duplicate messages when buffers full

harbu:
## Summary

Fixes an issue in the `network` package in which duplicate messages would be sent out when the send buffer of a WebRTC connection (on Node.js) had items in it.

## Detailed explanation

The root of this bug is in how our code interpreted the returned boolean value of method `dataChannel#send` of  `node-datachannel`. Our code assumed that this boolean value indicates the success of sending a message. However, it actually indicates whether the message could be sent immediately or whether it had to be buffered by the library (see https://github.com/murat-dogan/node-datachannel/issues/159). So when the return value was `false`, our code would re-send the same message later (up to 10 times) even though the message was already accepted by the library to be sent when its internal buffer cleared a bit.

Another side-effect of this bug: we may have warn logged messages as not having been sent even though they probably were (including duplicates). 

Fixed this issue by ignoring the return value. We already implement low / high buffer handling on our end (with the assistance of `dataChannel#setBufferedAmountLowThreshold`.

### DIFF
--- a/packages/network/src/connection/webrtc/WebRtcConnection.ts
+++ b/packages/network/src/connection/webrtc/WebRtcConnection.ts
@@ -417,10 +417,13 @@ export abstract class WebRtcConnection extends ConnectionEmitter {
                 let sent = false
                 let isOpen
                 try {
+                    if (this.isOpen()) {
+                        this.doSendMessage(queueItem.getMessage())
+                        sent = true
+                    }
                     // this.isOpen() is checked immediately after the call to node-datachannel.sendMessage() as if
                     // this.isOpen() returns false after a "successful" send, the message is lost with a near 100% chance.
                     // This does not work as expected if this.isOpen() is checked before sending a message
-                    sent = this.isOpen() && this.doSendMessage(queueItem.getMessage())
                     isOpen = this.isOpen()
                     sent = sent && isOpen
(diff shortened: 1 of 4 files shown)


packages/network/src/connection/webrtc/WebRtcConnection.ts
```diff
@@ -417,10 +417,13 @@ export abstract class WebRtcConnection extends ConnectionEmitter {
...
+                        this.doSendMessage(queueItem.getMessage())
+                        sent = true
+                    }
                     // this.isOpen() is checked immediately after the call to node-datachannel.sendMessage() as if
                     // this.isOpen() returns false after a "successful" send, the message is lost with a near 100% chance.
                     // This does not work as expected if this.isOpen() is checked before sending a message
-                    sent = this.isOpen() && this.doSendMessage(queueItem.getMessage())
                     isOpen = this.isOpen()
                     sent = sent && isOpen
                     this.messagesSent += 1
```
teogeb, line 429: If `sent` is false, we should not increase `messageSent` and `bytesSent`. Also we could move the increment of sendFailures into the else branch and remove the `return`.

Maybe in a separate PR.

teogeb (approved):

PR #1301, opened 2023-05-02
test(cli-tools): Add end-to-end tests

teogeb:
## Summary

Add end-to-end test for `cli-tools` package. Most functionality is now under tests.

## Implementation

We build the `cli-tools` package before running the tests. It allows us to run the scripts with `node` instead of e.g. `ts-node`. Alternatively we could install `ts-node` as a dev dependency and run scripts like this: `../../node_modules/.bin/ts-node bin/streamr-wallet-whoami.ts`. But with that approach the startup of the interpreter is significantly slower (~30s).

- Add tests for the `governance vote` command
- Improve test coverage so that all command line arguments have some test case
- Move `collect` to `@streamr/utils` package and also refactor client to use this shared function

### DIFF
--- a/packages/cli-tools/jest.config.js
+++ b/packages/cli-tools/jest.config.js
@@ -1,0 +1,2 @@
+const rootConfig = require('../../jest.config')
+module.exports = rootConfig
--- a/packages/cli-tools/jest.config.js
+++ b/packages/cli-tools/jest.config.js
@@ -1,2 +1,5 @@
 const rootConfig = require('../../jest.config')
-module.exports = rootConfig
+module.exports = {
+    ...rootConfig,
+    globalTeardown: './jest.teardown.js'
+}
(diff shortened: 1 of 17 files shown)


packages/cli-tools/jest.config.js
```diff
@@ -0,0 +1,2 @@
+const rootConfig = require('../../jest.config')
+module.exports = rootConfig
```
harbu:
```suggestion
module.exports = {
    ...rootConfig,
    globalTeardown: './jest.teardown.js'
}
```

packages/cli-tools/test/utils.ts
```diff
@@ -0,0 +1,84 @@
...
+            yield line
+            buffer = buffer.substring(delimeterPos + 1)
+        }
+    }
+    if (buffer !== '') {
+        yield buffer
+    }
+}
+
+export const collect = async <T>(source: AsyncIterable<T>, maxCount?: number): Promise<T[]> => {
```
harbu, line 68: Move to `utils` and use also in client? Maybe separate PR?

harbu (approved):

PR #1310, opened 2023-05-03
chore(cli-tools): Type annotations for `options`

teogeb:
Add `Options` interface for each command which annotates the configuration options of that command.

Also simplified boolean handling in some classes.


harbu (commented):
LGTM; don't really understand the `assertBothOrNoneDefined` change though.

harbu (approved):

PR #1357, opened 2023-05-09
NET-974: external ip

juslesan:
## Summary

Make it possible to configure external IPs for the WebRTC connections. This is useful in cases where the peer has a public IP address but the host does not know of it. This happens in cases where the node is behind a Full Cone NAT. For example, in AWS it is common that the host's network interface does not know it's public IP.

The external IP address is injected into WebRTC IceCandidate messages when the type is known to be `host`, making connecting to such peers possible without the use of STUN.

## Changes

-  Added configurability for NodeJS WebRTC connections for an external IP address
- Added configurability all the way to the client

- The solution should not work in cases where the the Node is behind a Full Cone NAT that randomizes the ports that are used. It works only if the port local port mappings are preserved on the public side.

### DIFF
--- a/packages/client/src/Config.ts
+++ b/packages/client/src/Config.ts
@@ -285,8 +285,11 @@ export interface StreamrClientConfig {
         /**
          * Used to assign a custom external IP address for the node.
          * Useful in cases where the node has a public IP address but
-         * the hosts network interface does not know of it
-         */
+         * the hosts network interface does not know of it.
+         * 
+         * Works only if the Full Cone NAT that the node is behind preserves local   
+         * port mappings on the public side. 
+        */
         externalIp?: ExternalIP
     }
 
--- a/packages/client/src/Config.ts
+++ b/packages/client/src/Config.ts
@@ -283,7 +283,9 @@ export interface StreamrClientConfig {
         location?: Location
 
         /**
-         * Used to assign a custom external IP address for this nodes 
+         * Used to assign a custom external IP address for the node.
+         * Useful in cases where the node has a public IP address but
+         * the hosts network interface does not know of it
          */
         externalIp?: ExternalIP
     }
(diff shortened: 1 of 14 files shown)


packages/client/src/config.schema.json
```diff
@@ -179,6 +179,18 @@
...
+                        {
+                            "type": "null"
+                        },
+                        {
+                            "type": "string",
+                            "format": "ipv4"
+                        }
+                    ],
+                    "default": null
+                },
```
harbu: You can simplify this to the following and then the value will be `undefined` by default. Allowing you to have type `string` instead of `ExternalIP = string | null`.

```suggestion
                "externalIp": {
                     "type": "string",
                     "format": "ipv4"
                },
```

packages/client/src/Config.ts
```diff
@@ -281,6 +281,13 @@ export interface StreamrClientConfig {
          * IP lookup).
          */
         location?: Location
+
+        /**
+         * Used to assign a custom external IP address for the node.
+         * Useful in cases where the node has a public IP address but
+         * the hosts network interface does not know of it
+         */
```
harbu: Maybe we should add the limitation here for future reference.

```suggestion
         * Used to assign a custom external IP address for the node.
         * Useful in cases where the node has a public IP address but
         * the hosts network interface does not know of it.
         * 
         * Works only if the Full Cone NAT that the node is behind preserves local   
         * port mappings on the public side. 
         */
```

packages/network/src/connection/webrtc/ParsedLocalCandidate.ts
```diff
@@ -0,0 +1,39 @@
+export enum CandidateType {
+    HOST = 'host',
+    SRFLX = 'srflx',
+    PRFLX = 'prflx',
+    RELAY = 'relay'
+}
+
+export class ParsedLocalCandidate {
```
harbu, line 8: nitpick: would be much simpler as a helper function than a class imo.

harbu (approved):

PR #1376, opened 2023-05-16
feat(broker): MQTT plugin partition support

teogeb:
MQTT plugin can publish and subscribe to specific partition. The information can be encoded into the topic string, e.g. 
```
await client.publish('/foobar?partition=5', ...)
```

### DIFF
--- a/packages/broker/src/plugins/mqtt/Bridge.ts
+++ b/packages/broker/src/plugins/mqtt/Bridge.ts
@@ -60,10 +60,10 @@ export class Bridge implements MqttServerListener {
 
     async onSubscribed(topic: string, clientId: string): Promise<void> {
         logger.info('Handle client subscribe', { clientId, topic })
-        const streamId = this.getStreamPartition(topic)
-        const existingSubscription = this.getSubscription(streamId)
+        const streamPart = this.getStreamPartition(topic)
+        const existingSubscription = this.getSubscription(streamPart)
         if (existingSubscription === undefined) {
-            const streamrClientSubscription = await this.streamrClient.subscribe(streamId, (content: any, metadata: MessageMetadata) => {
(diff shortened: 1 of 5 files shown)


packages/broker/src/plugins/mqtt/Bridge.ts
```diff
@@ -57,7 +60,7 @@ export class Bridge implements MqttServerListener {
 
     async onSubscribed(topic: string, clientId: string): Promise<void> {
         logger.info('Handle client subscribe', { clientId, topic })
-        const streamId = this.getStreamId(topic)
+        const streamId = this.getStreamPartition(topic)
```
harbu:
```suggestion
        const streamPart = this.getStreamPartition(topic)
```
Rename var?

packages/broker/test/unit/helpers/parser.test.ts
```diff
@@ -0,0 +1,35 @@
...
+        })
+    })
+
+    describe('parseQueryAndBase', () => {
+        it('with parameters', () => {
+            const { base, query } = parseQueryAndBase('foobar?lorem=ipsum')
+            expect(base).toBe('foobar')
+            expect(query['lorem']).toBe('ipsum')
+        })
+        it('without parameters', () => {
```
harbu, line 29: Will it handle `parseQueryAndBase('foobar?')` properly?

harbu (approved):

packages/broker/test/unit/helpers/parser.test.ts
teogeb: Depends which is expected behavior. If we define that the topic format is `streamId?query`, and query can be empty (like e.g. in http urls), then this return base as `foobar`. We wouldn't support stream ids which contain question marks here. 

I'll add a test case.

PR #1391, opened 2023-05-18
refactor(client): Refactor message validator

teogeb:
Refactored `Validator` to a utility function and inline `StreamMessageValidator` class into it.

Validator no longer uses `pOrderedResolve`. It is not needed as `subscribePipeline` and `MsgChainUtils` handle the ordering of validation requests internally (by iterating each message chain pipeline sequentially).

Also merged two separate group key messages validation functions.

### Future improvements
- merge `validateStreamMessage.test.ts` and `validateStreamMessage2.test.ts`


packages/client/src/utils/validateStreamMessage.ts
```diff
@@ -0,0 +1,114 @@
...
+        // it is handling?
+        if (!err.streamMessage) {
+            err.streamMessage = msg
+        }
+        throw err
+    })
+}
+
+/**
+ * Checks that the given StreamMessage is satisfies the requirements of the protocol.
```
harbu:
```suggestion
 * Checks that the given StreamMessage satisfies the requirements of the protocol.
```

harbu (approved):

PR #1403, opened 2023-05-22
add internal declaration to methods & variables that have internally declared vars, types or methods

yaruno:
This should fix the issue I have on including current client version and compiling with typescript compiler. Let me know what you think.


harbu (approved):
LGTM; we should probably make an automated smoke test to check that the generated type declarations are internally consistent. But we shall do that in a separate PR.

PR #1422, opened 2023-05-28
refactor(client): Simplify message cloning

teogeb:
Use `StreamMessage#clone` instead of `StreamMessage.deserialize(streamMessage.serialize())`.

Also removed cloning from `FakeNetwork` and `FakeStorage` node as after PR #574 `decrypt()` no longer mutates `StreamMessage` instances.

### DIFF
--- a/packages/client/test/test-utils/fake/FakeNetwork.ts
+++ b/packages/client/test/test-utils/fake/FakeNetwork.ts
@@ -47,10 +47,7 @@ export class FakeNetwork {
         * TODO: should we change the serialization or the test? Or keep this hack?
         */
         recipients.forEach((n) => {
-            n.messageListeners.forEach((listener) => {
-                // return a clone as client mutates message when it decrypts messages
-                listener(msg.clone())
-            })
+            n.messageListeners.forEach((listener) => listener(msg))
         })
         this.sends.push({
             message: msg,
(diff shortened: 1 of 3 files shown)


packages/client/test/test-utils/fake/FakeNetwork.ts
```diff
@@ -46,13 +46,8 @@ export class FakeNetwork {
...
         * TODO: should we change the serialization or the test? Or keep this hack?
         */
-        const serialized = msg.serialize()
         recipients.forEach((n) => {
-            n.messageListeners.forEach((listener) => {
-                // return a clone as client mutates message when it decrypts messages
-                const deserialized = StreamMessage.deserialize(serialized)
-                listener(deserialized)
-            })
+            n.messageListeners.forEach((listener) => listener(msg))
```
harbu, line 50: No need to clone here after all?

packages/client/test/test-utils/fake/FakeStorageNode.ts
```diff
@@ -92,11 +92,7 @@ export class FakeStorageNode extends FakeNetworkNode {
         if (messages !== undefined) {
             const firstIndex = Math.max(messages.length - count, 0)
             const lastIndex = Math.min(firstIndex + count, messages.length - 1)
-            return messages.slice(firstIndex, lastIndex + 1).map((msg: StreamMessage) => {
-                // return a clone as client mutates message when it decrypts messages
-                const serialized = msg.serialize()
-                return StreamMessage.deserialize(serialized)
-            })
+            return messages.slice(firstIndex, lastIndex + 1)
```
harbu, line 95: No need to clone here after all?

harbu (approved):

PR #1432, opened 2023-05-30
client: fix ts-essentials skipLibCheck issue

harbu:
Fix issue in `streamr-client` type declarations. If a TypeScript project with setting `skipLibCheck: false` uses `streamr-client` it will not build due to error caused by streamr-client type declarations.

### Further improvements
Any other solution that avoids having to include `ts-essentials` as a dependency?


yaruno:
LGTM

yaruno (approved):
LGTM, Should fix the issue

PR #1458, opened 2023-06-04
fix(client): Subscribe pipeline clears cache on decrypt error

teogeb:
Small fix to subscribe pipeline error handling: if there was a decrypt error, the cache was not cleared. The `try-catch` never caught anything as we didn't await the promise which was returned from `decrypt()`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ found [here](packages/broker/CHANGELOG.md).
 
 #### Fixed
 
-- Clear cache when message decryption fails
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ found [here](packages/broker/CHANGELOG.md).
 
 #### Fixed
 
+- Clear cache when message decryption fails
```
```suggestion
- Clear permissions cache when message decryption fails (https://github.com/streamr-dev/network/pull/1458)
```

harbu (approved):

PR #1460, opened 2023-06-05
feat(client): [NET-585]: Gap fill in resends

teogeb:
Message ordering and gap filling is functionality is determined by storage node count.

### Changes

If there is only one storage node, we don't do message ordering or gap filling for resends.
- we do't need ordering as the the data fetched via HTTP returns the messages in ascending order
- as there is no other storage node in that case, we can't do gap fills from any other source

If there are multiple storage nodes, we can fetch the gaps from another storage node. But we explicitly disable the use of  the same storage node, which was used to fetch the resend data.

### Implementation

If there is only one storage node, we override the value of `config.orderMessages `to `false` when we create the pipeline which transforms the resend messages. That disables message ordering and the gap filling.

If there are multiple storage nodes, the `getStorageNodes` callback which is created for the pipeline returns only all other storage nodes, not the storage node which we randomly chose for this resend. We need to pass the new `getStorageNodes` callback to quite many methods. We can simplify this later when we refactor some `Resends` methods (see "Future improvements", and possibly refactor the ordering utility).

### Future improvements

The method signature of some `Resends` methods are quite long as we introduced yet another parameter for those methods. The signature could be changed to options and simplified in another PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ found [here](packages/broker/CHANGELOG.md).
 
 #### Fixed
 
+- Clear permissions cache when message decryption fails (https://github.com/streamr-dev/network/pull/1458)
+
 #### Security
 
 ### cli-tools
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ found [here](packages/broker/CHANGELOG.md).
 
 #### Changed
 
+- Optimize message ordering and gap handling of resend requests (https://github.com/streamr-dev/network/pull/1460)
+
 #### Deprecated
 
 #### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ found [here](packages/broker/CHANGELOG.md).
 
 #### Changed
 
-- Optimize message ordering and gap handling of resend requests (https://github.com/streamr-dev/network/pull/1460)
+- Optimize message ordering and gap handling of resends (https://github.com/streamr-dev/network/pull/1460)
 
 #### Deprecated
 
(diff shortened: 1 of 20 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ found [here](packages/broker/CHANGELOG.md).
 
 #### Changed
 
+- Optimize message ordering and gap handling of resend requests (https://github.com/streamr-dev/network/pull/1460)
```
```suggestion
- Optimize message ordering and gap handling of resends (https://github.com/streamr-dev/network/pull/1460)
```

packages/client/test/unit/Resends.test.ts
```diff
@@ -51,25 +52,51 @@ describe('Resends', () => {
...
+        const msg3 = await messageFactory.createMessage({ foo: 3 }, { timestamp: 3000 })
+        const msg4 = await messageFactory.createMessage({ foo: 4 }, { timestamp: 4000 })
+        const allMessages = [msg1, msg2, msg3, msg4]
+        const storageNodeAddress1 = randomEthereumAddress()
+        const storageNodeAddress2 = randomEthereumAddress()
+        const resends = createResends({
+            [storageNodeAddress1]: without(allMessages, msg2),
+            [storageNodeAddress2]: without(allMessages, msg3)
+        })
+        const messageStream = await resends.last(STREAM_PART_ID, { count: 2 }, false, async () => [storageNodeAddress1, storageNodeAddress2])
```
harbu:
```suggestion
        const messageStream = await resends.last(STREAM_PART_ID, { count: 4 }, false, async () => [storageNodeAddress1, storageNodeAddress2])
```

harbu (approved):

PR #1464, opened 2023-06-05
refactor(client): `MessagePipelineFactory`

teogeb:
New factory class which provides an easy way to create message pipelines. There are multiple DI dependencies needed to create a pipeline, and therefore the factory can abstract out the dependency handling.

### DIFF
--- a/packages/client/src/subscribe/Resends.ts
+++ b/packages/client/src/subscribe/Resends.ts
@@ -4,18 +4,15 @@ import random from 'lodash/random'
 import without from 'lodash/without'
 import { Lifecycle, delay, inject, scoped } from 'tsyringe'
 import { ConfigInjectionToken, StrictStreamrClientConfig } from '../Config'
-import { DestroySignal } from '../DestroySignal'
 import { HttpUtil, createQueryString } from '../HttpUtil'
 import { Message } from '../Message'
 import { StreamrClientError } from '../StreamrClientError'
-import { GroupKeyManager } from '../encryption/GroupKeyManager'
 import { StorageNodeRegistry } from '../registry/StorageNodeRegistry'
-import { StreamRegistryCached } from '../registry/StreamRegistryCached'
 import { StreamStorageRegistry } from '../registry/StreamStorageRegistry'
 import { counting } from '../utils/GeneratorUtils'
 import { LoggerFactory } from '../utils/LoggerFactory'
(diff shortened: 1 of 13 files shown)


packages/client/src/subscribe/Resends.ts
```diff
@@ -175,16 +164,17 @@ export class Resends {
...
-        const config = (nodeAddresses.length > 1) ? this.config : { ...this.config, orderMessages: false }
-        const messageStream = (raw === false) ? createSubscribePipeline({
+        const messageStream = (raw === false) ? this.messagePipelineFactory.createMessagePipeline({
             streamPartId,
+            /*
+             * Disable ordering if the source of this resend is the only storage node. In that case there is no
+             * another storage node from which we could fetch the gaps. When we set "disableMessageOrdering"
+             * to true, we disable both gap filling and message ordering. As resend messages always arrive 
+             * in ascending order, we don't need the ordering functionality. But as that flag controls also 
+             * gap filling, we can use it only if gap filling is not needed.
```
harbu:
```suggestion
             * Disable ordering if the source of this resend is the only storage node. In that case there is no
             * other storage node from which we could fetch the gaps. When we set "disableMessageOrdering"
             * to true, we disable both gap filling and message ordering. As resend messages always arrive 
             * in ascending order, we don't need the ordering functionality.
```

packages/client/src/subscribe/SubscriptionSession.ts
```diff
@@ -31,28 +25,17 @@ export class SubscriptionSession {
...
+        this.pipeline = messagePipelineFactory.createMessagePipeline({
             streamPartId,
-            getStorageNodes: (streamId: StreamID) => streamStorageRegistry.getStorageNodes(streamId),
-            resends,
-            groupKeyManager,
-            streamRegistryCached,
-            loggerFactory,
-            destroySignal,
-            config: config
+            getStorageNodes: (streamId: StreamID) => streamStorageRegistry.getStorageNodes(streamId)
```
harbu: Could this be the default value (i.e. we can leave this out?)

packages/client/src/subscribe/Resends.ts
harbu: Maybe the previous "value injection" pattern was better?

harbu (approved):

PR #1466, opened 2023-06-05
refactor(client): Extract `waitForStorage` function

teogeb:
Extracted `waitForStorage` function from `Resends` class. 

The client defaults for the function are applied in `StreamrClient#waitForStorage` and that method uses `merge` from `@streamr/utils`. Changed also `createStream` method to use the same merge function (instead of using a function from `lodash`).


packages/client/src/StreamrClient.ts
```diff
@@ -285,7 +285,17 @@ export class StreamrClient {
...
-        return this.resends.waitForStorage(message, options)
+        const defaultOptions = {
+            // eslint-disable-next-line no-underscore-dangle
+            interval: this.config._timeouts.storageNode.retryInterval,
+            // eslint-disable-next-line no-underscore-dangle
+            timeout: this.config._timeouts.storageNode.timeout,
+            count: 100,
+            messageMatchFn: (msgTarget: Message, msgGot: Message) => {
+                return msgTarget.signature === msgGot.signature
+            }
```
harbu: Given that this is the only implementation in use in our code base, should it just be hard coded into `waitForStorage`?

harbu (approved):

PR #1491, opened 2023-06-09
refactor(client): Throw `StreamrClientError` if stream not found from registry 

teogeb:
Throw `StreamrClientError(..., 'STREAM_NOT_FOUND')` if we query a stream from registry, which does not exists. 

Before this PR an error was throw, but it was not an instance of `StreamrClientError`, but a `HttpUtils.NotFound` error. It is preferred that all errors generated by the client are instances of `StreamrClientError`.

This changes also the the content of `error.code` field from `NOT_FOUND` to `STREAM_NOT_FOUND` (and removes the fallback field `error.errorCode`). Therefore this could be a breaking change. But error codes are not documented in the public API, and therefore this can also be considered as internal refactoring. There can be applications which assume this undocumented behavior.

### Future improvements

- `HttpUtils.ERROR_CODES` are specific to the storage node implementation, not generic HTTP error codes. We should move the logic to `Resends` and use standard `StreamrClientError` instead of custom error class.
- Specify to TSDocs of `getStream` (and maybe other places) which error is thrown. Could also document other `StreamrClientError` throws for other methods similarly.


harbu (approved):
Let's think thru implications to semver before merging

PR #1505, opened 2023-06-12
feat(client): Configurable RSA key length

teogeb:
Added RSA key length as a new configuration option: `encryption.rsaKeyLength`

In tests we use shorter RSA key (640 bits instead of 4096 bits) as tests don't need very safe RSA keys. This change reduced the execution time of Integration tests significantly: 
- before ~50s
- after ~37s


CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ found [here](packages/broker/CHANGELOG.md).
 
 #### Added
 
+- Add optinal config option `encryption.rsaKeyLength` to control the strength of RSA encryption in key-exchange (https://github.com/streamr-dev/network/pull/1505)
```
```suggestion
- Add optional config option `encryption.rsaKeyLength` to control the strength of RSA encryption in key-exchange (https://github.com/streamr-dev/network/pull/1505)
```

harbu (approved):

PR #1513, opened 2023-06-12
NET-527 maintain operator value service

samigabor:
Added MaintainOperatorValueService, helper and tests


packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
```diff
@@ -0,0 +1,32 @@
+import { Logger } from '@streamr/utils'
+import { Contract } from 'ethers'
+import { Operator, StreamrConfig, operatorABI, streamrConfigABI } from '@streamr/network-contracts'
+import { OperatorServiceConfig } from './OperatorPlugin'
+
+const logger = new Logger(module)
+
+export class MaintainOperatorValueHelper {
+    config: OperatorServiceConfig
+    operator: Operator
```
harbu:
```suggestion
    private readonly config: OperatorServiceConfig
    private readonly operator: Operator
```
We can tighten the access modifiers here.

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    constructor(config: OperatorServiceConfig) {
+        logger.trace('MaintainOperatorValueHelper created')
```
harbu: I'm guessing this log line isn't very relevant anymore?
```suggestion
```

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
```diff
@@ -0,0 +1,32 @@
...
+        const streamrConfig = new Contract(streamrConfigAddress, streamrConfigABI, this.config.provider) as unknown as StreamrConfig
+        return (await streamrConfig.poolValueDriftLimitFraction()).toBigInt()
+    }
+
+    async getApproximatePoolValuesPerSponsorship(): Promise<any> {
+        return await this.operator.getApproximatePoolValuesPerSponsorship()
+    }
+
+    async updateApproximatePoolvalueOfSponsorships(sponsorshipAddresses: string[]): Promise<void> {
+        await (await this.operator.updateApproximatePoolvalueOfSponsorships(sponsorshipAddresses)).wait()
```
harbu, line 31: More releated to the network-contracts package: should the smart contract function be named `updateApproximatePoolValuesOfSponsorships` so the naming convention would be consistent, i.e. `Poolvalue` vs `PoolValue`?

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    /// returns a wei value (1 ETH means 100%)
+    async getPenaltyLimitFraction(): Promise<bigint> {
+        const streamrConfigAddress = await this.operator.streamrConfig()
+        const streamrConfig = new Contract(streamrConfigAddress, streamrConfigABI, this.config.provider) as unknown as StreamrConfig
+        return (await streamrConfig.poolValueDriftLimitFraction()).toBigInt()
+    }
+
+    async getApproximatePoolValuesPerSponsorship(): Promise<any> {
```
harbu: Add type information

```suggestion
    async getApproximatePoolValuesPerSponsorship(): Promise<{
        sponsorshipAddresses: string[]
        approxValues: BigNumber[]
        realValues: BigNumber[]
    }> {
```

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
```diff
@@ -0,0 +1,32 @@
...
+    config: OperatorServiceConfig
+    operator: Operator
+
+    constructor(config: OperatorServiceConfig) {
+        logger.trace('MaintainOperatorValueHelper created')
+        this.config = config
+        this.operator = new Contract(config.operatorContractAddress, operatorABI, this.config.signer) as unknown as Operator
+    }
+
+    /// returns a wei value (1 ETH means 100%)
```
harbu:
```suggestion
    // returns a wei value (1 ETH means 100%)
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+import { OperatorServiceConfig } from './OperatorPlugin'
+
+const logger = new Logger(module)
+
+const CHECK_VALUE_INTERVAL = 1000 * 60 * 60 * 24 // 1 day
+
+export class MaintainOperatorValueService {
+    config: OperatorServiceConfig
+    private checkValueInterval: NodeJS.Timeout | null = null
+    private helper: MaintainOperatorValueHelper
```
harbu:
```suggestion
    private readonly config: OperatorServiceConfig
    private checkValueInterval: NodeJS.Timeout | null = null
    private readonly helper: MaintainOperatorValueHelper
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    constructor(config: OperatorServiceConfig) {
+        logger.trace('MaintainOperatorValueService created')
```
harbu: Logger statement probably unnecessary, so could be removed?

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+        this.config = config
+        this.helper = new MaintainOperatorValueHelper(config)
+    }
+
+    // eslint-disable-next-line class-methods-use-this
+    start(penaltyLimitFraction?: bigint): void {
+        logger.info('MaintainOperatorValueService started')
```
harbu: The name of the module `MaintainOperatorValueService` will already be included in the log statement so you can simplify this to:
```suggestion
        logger.info('Started')
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+    private checkValueInterval: NodeJS.Timeout | null = null
+    private helper: MaintainOperatorValueHelper
+
+    constructor(config: OperatorServiceConfig) {
+        logger.trace('MaintainOperatorValueService created')
+        this.config = config
+        this.helper = new MaintainOperatorValueHelper(config)
+    }
+
+    // eslint-disable-next-line class-methods-use-this
```
harbu:
```suggestion
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+    start(penaltyLimitFraction?: bigint): void {
+        logger.info('MaintainOperatorValueService started')
+        this.checkValue(this.config.operatorContractAddress, penaltyLimitFraction)
+        this.checkValueInterval = setInterval(() => {
+            this.checkValue(this.config.operatorContractAddress, penaltyLimitFraction)
+        }, CHECK_VALUE_INTERVAL)
+    }
+
+    async checkValue(operatorContractAddress: string, penaltyLimitFraction?: bigint): Promise<void> {
+        logger.info(`Check approximate value for operator ${operatorContractAddress}`)
```
harbu:
```suggestion
        logger.info('Check approximate value for operator', { operatorContractAddress })
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+    // eslint-disable-next-line class-methods-use-this
+    start(penaltyLimitFraction?: bigint): void {
+        logger.info('MaintainOperatorValueService started')
+        this.checkValue(this.config.operatorContractAddress, penaltyLimitFraction)
+        this.checkValueInterval = setInterval(() => {
+            this.checkValue(this.config.operatorContractAddress, penaltyLimitFraction)
+        }, CHECK_VALUE_INTERVAL)
+    }
+
+    async checkValue(operatorContractAddress: string, penaltyLimitFraction?: bigint): Promise<void> {
```
harbu: Can be made `private` and first parameter is unnecessary as de facto field `this.config.operatorContractAddress` is used.

```suggestion
    private async checkValue(penaltyLimitFraction?: bigint): Promise<void> {
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+    private helper: MaintainOperatorValueHelper
+
+    constructor(config: OperatorServiceConfig) {
+        logger.trace('MaintainOperatorValueService created')
+        this.config = config
+        this.helper = new MaintainOperatorValueHelper(config)
+    }
+
+    // eslint-disable-next-line class-methods-use-this
+    start(penaltyLimitFraction?: bigint): void {
```
harbu: Should `penaltyLimitFraction` be passed in constructor instead? As `start` is invoked only once, this value should never change anyways?

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+                approxValue: approxValues[i],
+                realValue: realValues[i],
+                diff: realValues[i].sub(approxValues[i])
+            }
+            sponsorships.push(sponsorship)
+            totalDiff = totalDiff + sponsorship.diff.toBigInt()
+            totalApprox = totalApprox + sponsorship.approxValue.toBigInt()
+        }
+
+        if (!penaltyLimitFraction) {
```
harbu: We prefer to use explicit undefined checks since the equality won't then match zero / empty situations (e.g. `!BigInt(0) === true`)
```suggestion
        if (penaltyLimitFraction === undefined) {
```

However, if the intention is to match both undefined _and zero_ then we just spell those cases out as two conditions: `penaltyLimitFraction === undefined || penaltyLimitFraction === BigInt(0)`

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+        }, CHECK_VALUE_INTERVAL)
+    }
+
+    async checkValue(operatorContractAddress: string, penaltyLimitFraction?: bigint): Promise<void> {
+        logger.info(`Check approximate value for operator ${operatorContractAddress}`)
+
+        const { sponsorshipAddresses, approxValues, realValues } = await this.helper.getApproximatePoolValuesPerSponsorship()
+        let totalDiff = BigInt(0)
+        let totalApprox = BigInt(0)
+        const sponsorships = []
```
harbu: Add typing information
```suggestion
        const sponsorships: { address: string, approxValue: BigNumber, realValue: BigNumber, diff: BigNumber }[] = []
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        if (!penaltyLimitFraction) {
+            penaltyLimitFraction = await this.helper.getPenaltyLimitFraction()
+        }
+
+        const threshold = totalApprox * penaltyLimitFraction / BigInt(1e18)
+        if (totalDiff > threshold) {
+            // sort sponsorships by diff in descending order
+            const sortedSponsorships = sponsorships.sort((a: any, b: any) => b.diff - a.diff)
```
harbu, line 59: After adding type info as suggested above, we can get rid of the any casts here for improved type checking:
```suggestion
            const sortedSponsorships = sponsorships.sort((a, b) => b.diff - a.diff)
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+            for (const sponsorship of sortedSponsorships) {
+                diff = diff + sponsorship.diff.toBigInt()
+                neededSponsorshipsCount += 1
+                if (diff > threshold) {
+                    break
+                }
+            }
+            
+            // pick the first entries needed to get the total diff under the threshold
+            const neededSponsorshipAddresses = sortedSponsorships.slice(0, neededSponsorshipsCount).map((sponsorship: any) => sponsorship.address)
```
harbu: You can get rid of this any cast after adding typing info to the `sortedSponsorships` array as suggested above.
```suggestion
            const neededSponsorshipAddresses = sortedSponsorships.slice(0, neededSponsorshipsCount).map((sponsorship) => sponsorship.address)
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+            logger.info(`Updating ${neededSponsorshipsCount} sponsorships. Threshold (${threshold}), diff ${diff}/${totalDiff}`)
```
harbu:
```suggestion
            logger.info('Updating sponsorships', { neededSponsorshipsCount, threshold, diffPercentage: diff / totalDiff })
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+            await this.helper.updateApproximatePoolvalueOfSponsorships(neededSponsorshipAddresses)
+            logger.info(`Updated sponsorships!`)
```
harbu:
```suggestion
            logger.info('Updated sponsorships!')
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+
+    // eslint-disable-next-line class-methods-use-this
```
harbu:
```suggestion
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+    async stop(): Promise<void> {
+        logger.info('MaintainOperatorValueService stopped')
```
harbu:
```suggestion
        logger.info('Stop')
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+    constructor(config: OperatorServiceConfig) {
+        logger.trace('MaintainOperatorValueService created')
+        this.config = config
+        this.helper = new MaintainOperatorValueHelper(config)
+    }
+
+    // eslint-disable-next-line class-methods-use-this
+    start(penaltyLimitFraction?: bigint): void {
+        logger.info('MaintainOperatorValueService started')
+        this.checkValue(this.config.operatorContractAddress, penaltyLimitFraction)
```
harbu: Both these `this.checkValue` invocations need a catch statement to avoid unhandled promise rejections. You can, for example, either add them here directly
```ts
this.checkValue(this.config.operatorContractAddress, penaltyLimitFraction).catch((err) => {
    logger.warn('Encountered error while checking value', { err })
})
``` 
Or you can wrap the the check value method inside a try catch, e.g.
```ts
async checkValue(penaltyLimitFraction?: bigint): Promise<void> {
    try {
        // all existing logic here
    } catch (err) {
        logger.warn('Encountered error while checking value', { err })
    }
}
```

packages/broker/src/plugins/operator/OperatorPlugin.ts
```diff
@@ -31,6 +32,7 @@ export class OperatorPlugin extends Plugin<OperatorPluginConfig> {
     private readonly maintainOperatorContractService = new MaintainOperatorContractService()
     private readonly voteOnSuspectNodeService = new VoteOnSuspectNodeService()
     private readonly maintainTopologyService: MaintainTopologyService
+    private maintainOperatorValueService: MaintainOperatorValueService
```
harbu:
```suggestion
    private readonly maintainOperatorValueService: MaintainOperatorValueService
```

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,159 @@
...
+import { Contract } from "@ethersproject/contracts"
+
+import { deployOperatorContract } from "./deployOperatorContract"
+import { deploySponsorship } from "./deploySponsorshipContract"
+import { MaintainOperatorValueService } from "../../../../src/plugins/operator/MaintainOperatorValueService"
+import { OperatorServiceConfig } from "../../../../src/plugins/operator/OperatorPlugin"
+
+const config = Chains.load()["dev1"]
+const adminPrivKey = "0x4059de411f15511a85ce332e7a428f36492ab4e87c7830099dadbf130f1896ae"
+const theGraphUrl = `http://${process.env.STREAMR_DOCKER_DEV_HOST}:8000/subgraphs/name/streamr-dev/network-subgraphs`
```
harbu:
```suggestion
const theGraphUrl = `http://${process.env.STREAMR_DOCKER_DEV_HOST ?? '127.0.0.1'}:8000/subgraphs/name/streamr-dev/network-subgraphs`
```

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,159 @@
...
+
+const logger = new Logger(module)
+
+jest.setTimeout(60 * 1000)
```
harbu: Better to set this per-test than globally, e.g.

```ts

beforeEach({
    // ...
}, 60 * 1000)

it('something something', {
    // ...
}, 60 * 1000)
```

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    let opertatorConfig: OperatorServiceConfig
+
+    const deployNewOperator = async () => {
+        const operatorWallet = Wallet.createRandom().connect(provider)
+        logger.debug("Funding", { address: operatorWallet.address })
+        await (await token.transfer(operatorWallet.address, parseEther("1000"))).wait()
+        await (await adminWallet.sendTransaction({
+            to: operatorWallet.address,
+            value: parseEther("1")
+        })).wait()
```
harbu: We should probably use the test helper function `generateWalletWithGasAndTokens` here to avoid repetition (and to have retry handling for nonce issues)

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    let provider: Provider
+    let operatorWallet: Wallet
+    let operatorContract: Operator
+    let token: IERC677
+    let adminWallet: Wallet
+    let streamId1: string
+    let streamId2: string
+
+    let opertatorConfig: OperatorServiceConfig
```
harbu:
```suggestion
    let operatorConfig: OperatorServiceConfig
```

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,159 @@
...
+        const operatorWallet = Wallet.createRandom().connect(provider)
+        logger.debug("Funding", { address: operatorWallet.address })
+        await (await token.transfer(operatorWallet.address, parseEther("1000"))).wait()
+        await (await adminWallet.sendTransaction({
+            to: operatorWallet.address,
+            value: parseEther("1")
+        })).wait()
+
+        logger.debug("Deploying operator contract")
+        const operatorContract = await deployOperatorContract(config, operatorWallet)
```
harbu: There's an even more simple version of deploying operator contract in `smartContractUtils.ts`. No need to pass first argument `config`.

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,159 @@
...
+            provider,
+            theGraphUrl,
+            fetch,
+            signer: operatorWallet
+        }
+        return { operatorWallet, operatorContract }
+    }
+
+    beforeEach(async () => {
+        provider = new JsonRpcProvider(chainURL)
```
harbu: You can use `getProvider` from `smartContractUtils.ts` for re-use (not having to dig chainURL yourself).

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,159 @@
...
+        adminWallet = new Wallet(adminPrivKey, provider)
+
+        token = new Contract(config.contracts.LINK, tokenABI, adminWallet) as unknown as IERC677
+        const timeString = (new Date()).getTime().toString()
+        const streamPath1 = "/operatorvalueservicetest-1-" + timeString
+        const streamPath2 = "/operatorvalueservicetest-2-" + timeString
+        streamId1 = adminWallet.address.toLowerCase() + streamPath1
+        streamId2 = adminWallet.address.toLowerCase() + streamPath2
+        const streamRegistry = new Contract(config.contracts.StreamRegistry, streamRegistryABI, adminWallet) as unknown as StreamRegistry
+        logger.debug(`creating stream with streamId1 ${streamId1}`)
```
harbu:
```suggestion
        logger.debug(`Creating stream with streamId1 ${streamId1}`)
```

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,159 @@
...
+        await (await streamRegistry.createStream(streamPath1, "metadata")).wait()
+        logger.debug(`creating stream with streamId2 ${streamId2}`)
```
harbu:
```suggestion
        logger.debug(`Creating stream with streamId2 ${streamId2}`)
```

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,159 @@
...
+        expect(await token.balanceOf(sponsorship2.address)).toEqual(parseEther("300"))
+        
+        const maintainOperatorValueService = new MaintainOperatorValueService(opertatorConfig)
+
+        const totalValueInSponsorshipsBefore = await operatorContract.totalValueInSponsorshipsWei()
+        const penaltyFraction = 0.0005 // * 1e18
+        const threshold = 200 * penaltyFraction // 0.1
+
+        // wait for sponsorships to accumulate earnings so approximate values differ enough form the real values
+        await new Promise((resolve) => setTimeout(resolve, 3000)) // sleep 3 sec
```
harbu: We have a utility for this in `@streamr/utils` package
```suggestion
        await wait(3000)
```

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,159 @@
...
+
+        await waitForCondition(async () => await operatorContract.totalValueInSponsorshipsWei() > totalValueInSponsorshipsBefore, 10000, 1000)
+        
+        const { sponsorshipAddresses, approxValues, realValues } = await operatorContract.getApproximatePoolValuesPerSponsorship()
+        let diff = BigInt(0)
+        for (let i = 0; i < sponsorshipAddresses.length; i++) {
+            diff = realValues[i].toBigInt() - approxValues[i].toBigInt()
+        }
+        
+        expect(await operatorContract.totalValueInSponsorshipsWei() > totalValueInSponsorshipsBefore)
```
harbu: You get improved jest error messages when using more specific assertions:
```suggestion
        expect(await operatorContract.totalValueInSponsorshipsWei()).toBeGreaterThan(totalValueInSponsorshipsBefore)
```

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,159 @@
...
+        expect(diff < threshold)
```
harbu:
```suggestion
        expect(diff).toBeLessThan(threshold)
```

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,159 @@
...
+        expect(await token.balanceOf(sponsorship2.address)).toEqual(parseEther("300"))
+        
+        const maintainOperatorValueService = new MaintainOperatorValueService(opertatorConfig)
+
+        const totalValueInSponsorshipsBefore = await operatorContract.totalValueInSponsorshipsWei()
+        const penaltyFraction = 0.0005 // * 1e18
+        const threshold = 200 * penaltyFraction // 0.02
+
+        // wait for sponsorships to accumulate earnings so approximate values differ enough form the real values
+        await new Promise((resolve) => setTimeout(resolve, 3000)) // sleep 3 sec
```
harbu:
```suggestion
        await wait(3000)
```

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,159 @@
...
+        await waitForCondition(async () => await operatorContract.totalValueInSponsorshipsWei() > totalValueInSponsorshipsBefore, 10000, 1000)
+        
+        const { sponsorshipAddresses, approxValues, realValues } = await operatorContract.getApproximatePoolValuesPerSponsorship()
+        let diff = BigInt(0)
+        for (let i = 0; i < sponsorshipAddresses.length; i++) {
+            diff = realValues[i].toBigInt() - approxValues[i].toBigInt()
+        }
+        
+        expect(await operatorContract.totalValueInSponsorshipsWei() > totalValueInSponsorshipsBefore)
+        expect(diff < threshold)
```
harbu: Same comment as above regarding more explicit assertions.

harbu (changes requested):
Good stuff; see my comments on changes, I can clarify them over a call if you want

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+        this.checkValueInterval = setInterval(() => {
```
teogeb: Maybe we want to ensure that there can't be two `checkValue` calls to be running in paraller?

Could use `scheduleAtInterval` helper from @streams/utils to achieve this. By using it there would be `CHECK_VALUE_INTERVAL` milliseconds _between_ two requests, i.e. the waiting starts only after the previous task has been completed.

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+import { MaintainOperatorValueHelper } from "./MaintainOperatorValueHelper"
+import { OperatorServiceConfig } from './OperatorPlugin'
+
+const logger = new Logger(module)
+
+const CHECK_VALUE_INTERVAL = 1000 * 60 * 60 * 24 // 1 day
+
+export class MaintainOperatorValueService {
+    private readonly config: OperatorServiceConfig
+    private checkValueInterval: NodeJS.Timeout | null = null
```
teogeb: It would maybe improve readability if most relevant fields are listed first? E.g. state fields before dependencies and those before configurations?

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+    start(penaltyLimitFraction?: bigint): void {
+        logger.info('MaintainOperatorValueService started')
+        this.checkValue(this.config.operatorContractAddress, penaltyLimitFraction)
+        this.checkValueInterval = setInterval(() => {
+            this.checkValue(this.config.operatorContractAddress, penaltyLimitFraction)
+        }, CHECK_VALUE_INTERVAL)
+    }
+
+    async checkValue(operatorContractAddress: string, penaltyLimitFraction?: bigint): Promise<void> {
+        logger.info(`Check approximate value for operator ${operatorContractAddress}`)
```
teogeb: This is maybe `debug` level logging?

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+            sponsorships.push(sponsorship)
+            totalDiff = totalDiff + sponsorship.diff.toBigInt()
+            totalApprox = totalApprox + sponsorship.approxValue.toBigInt()
+        }
+
+        if (!penaltyLimitFraction) {
+            penaltyLimitFraction = await this.helper.getPenaltyLimitFraction()
+        }
+
+        const threshold = totalApprox * penaltyLimitFraction / BigInt(1e18)
```
teogeb: Could introduce a constant for `1e18` magic number so that it has a name.

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+        this.checkValueInterval = setInterval(() => {
+            this.checkValue(this.config.operatorContractAddress, penaltyLimitFraction)
+        }, CHECK_VALUE_INTERVAL)
+    }
+
+    async checkValue(operatorContractAddress: string, penaltyLimitFraction?: bigint): Promise<void> {
+        logger.info(`Check approximate value for operator ${operatorContractAddress}`)
+
+        const { sponsorshipAddresses, approxValues, realValues } = await this.helper.getApproximatePoolValuesPerSponsorship()
+        let totalDiff = BigInt(0)
```
teogeb, line 35: Would it make sense to unify the data types so that all numbers are either integers (`BigInt`) or floating point numbers (`BigNumber`)? 

Do we support that the pool values could be floating point values, or are those always integers?

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,81 @@
...
+                if (diff > threshold) {
+                    break
+                }
+            }
+            
+            // pick the first entries needed to get the total diff under the threshold
+            const neededSponsorshipAddresses = sortedSponsorships.slice(0, neededSponsorshipsCount).map((sponsorship: any) => sponsorship.address)
+            logger.info(`Updating ${neededSponsorshipsCount} sponsorships. Threshold (${threshold}), diff ${diff}/${totalDiff}`)
+            await this.helper.updateApproximatePoolvalueOfSponsorships(neededSponsorshipAddresses)
+            logger.info(`Updated sponsorships!`)
```
teogeb:
```suggestion
            logger.info('Updated sponsorships')
```

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
```diff
@@ -0,0 +1,32 @@
+import { BigNumber, Contract } from 'ethers'
+import { Operator, StreamrConfig, operatorABI, streamrConfigABI } from '@streamr/network-contracts'
+import { OperatorServiceConfig } from './OperatorPlugin'
+
+export class MaintainOperatorValueHelper {
```
teogeb, line 5: Maybe we could remove this helper class as there is not much business logic implemented here?

We could have `operator` and `streamrConfig` fields in MaintainOperatorValueService and just call the relevant methods of those classes from the service? That way both contracts would be initialized just once and the logic is very straightforward.

packages/client/test/integration/multiple-clients.test.ts
```diff
@@ -174,7 +174,7 @@ describe('PubSub with multiple clients', () => {
 
             checkMessages(published, receivedMessagesMain)
             checkMessages(published, receivedMessagesOther)
-        }, 10 * 1000)
+        })
```
teogeb, line 177: Why this has changed?

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,159 @@
...
+import { streamRegistryABI } from "@streamr/network-contracts"
+import { Contract } from "@ethersproject/contracts"
+
+import { deployOperatorContract } from "./deployOperatorContract"
+import { deploySponsorship } from "./deploySponsorshipContract"
+import { MaintainOperatorValueService } from "../../../../src/plugins/operator/MaintainOperatorValueService"
+import { OperatorServiceConfig } from "../../../../src/plugins/operator/OperatorPlugin"
+
+const config = Chains.load()["dev1"]
+const adminPrivKey = "0x4059de411f15511a85ce332e7a428f36492ab4e87c7830099dadbf130f1896ae"
```
teogeb: This could refer to `ADMIN_WALLET_PK` constant if we export that from `smartContractUtils.ts`.

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,159 @@
...
+        const timeString = (new Date()).getTime().toString()
+        const streamPath1 = "/operatorvalueservicetest-1-" + timeString
+        const streamPath2 = "/operatorvalueservicetest-2-" + timeString
+        streamId1 = adminWallet.address.toLowerCase() + streamPath1
+        streamId2 = adminWallet.address.toLowerCase() + streamPath2
+        const streamRegistry = new Contract(config.contracts.StreamRegistry, streamRegistryABI, adminWallet) as unknown as StreamRegistry
+        logger.debug(`creating stream with streamId1 ${streamId1}`)
+        await (await streamRegistry.createStream(streamPath1, "metadata")).wait()
+        logger.debug(`creating stream with streamId2 ${streamId2}`)
+        await (await streamRegistry.createStream(streamPath2, "metadata")).wait();
```
teogeb: We could use StreamrClient to create the streams:

```suggestion
        const client = new StreamrClient({
            auth: {
                privateKey: adminWallet.address
            }
        })
        stream1 = await client.createStream(`/operatorvalueservicetest-1-${Date.now()}`)
        stream2 = await client.createStream(`/operatorvalueservicetest-2-${Date.now()}`)
        await client.destroy()
```

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,159 @@
...
+        // deploy 2 sponsorships, sponsor 200 & stake 100 into both of them
+        const sponsorship1 = await deploySponsorship(config, operatorWallet, { streamId: streamId1 })
+        const sponsorship2 = await deploySponsorship(config, operatorWallet, { streamId: streamId2 })
+        await (await token.connect(operatorWallet).transferAndCall(sponsorship1.address, parseEther("200"), "0x")).wait()
+        await (await token.connect(operatorWallet).transferAndCall(sponsorship2.address, parseEther("200"), "0x")).wait()
+        await (await operatorContract.stake(sponsorship1.address, parseEther("100"))).wait()
+        await (await operatorContract.stake(sponsorship2.address, parseEther("100"))).wait()
+
+        expect(await token.balanceOf(sponsorship1.address)).toEqual(parseEther("300")) // 200 sponsored + 100 staked
+        expect(await token.balanceOf(sponsorship2.address)).toEqual(parseEther("300"))
```
teogeb: Maybe could use a `for` loop as exactly the same action is done for both streams?

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,159 @@
...
+        
+        const maintainOperatorValueService = new MaintainOperatorValueService(opertatorConfig)
+
+        const totalValueInSponsorshipsBefore = await operatorContract.totalValueInSponsorshipsWei()
+        const penaltyFraction = 0.0005 // * 1e18
```
teogeb: What the `1e18` multiplication comment means? (If internal representation of the value in contract level then maybe not relevant in this service level test?)

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,159 @@
...
+        await (await token.connect(operatorWallet).transferAndCall(operatorContract.address, parseEther("200"), operatorWallet.address)).wait()
+        // deploy 2 sponsorships, sponsor 200 & stake 100 into both of them
+        const sponsorship1 = await deploySponsorship(config, operatorWallet, { streamId: streamId1 })
+        const sponsorship2 = await deploySponsorship(config, operatorWallet, { streamId: streamId2 })
+        await (await token.connect(operatorWallet).transferAndCall(sponsorship1.address, parseEther("200"), "0x")).wait()
+        await (await token.connect(operatorWallet).transferAndCall(sponsorship2.address, parseEther("200"), "0x")).wait()
+        await (await operatorContract.stake(sponsorship1.address, parseEther("100"))).wait()
+        await (await operatorContract.stake(sponsorship2.address, parseEther("100"))).wait()
+
+        expect(await token.balanceOf(sponsorship1.address)).toEqual(parseEther("300")) // 200 sponsored + 100 staked
```
teogeb: Could introduce constants e.g. SPONSOR_AMOUNT and STAKE_AMOUT which we'd use here. Then no need for the comments about the exact values at lines 94, 100 and 105.

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,159 @@
...
+            diff = realValues[i].toBigInt() - approxValues[i].toBigInt()
+        }
+        
+        expect(await operatorContract.totalValueInSponsorshipsWei() > totalValueInSponsorshipsBefore)
+        expect(diff < threshold)
+
+        await maintainOperatorValueService.stop()
+    })
+
+    it("needs only one sponsorship to stay over the threshold", async () => {
```
teogeb: Is this test case seems to be completely identical to the previous test case? Only comments are different. Does this test something different even if there are no different code lines?

teogeb (changes requested):

packages/client/test/integration/multiple-clients.test.ts
samigabor: It was set globally and made it per-test, as suggested

samt1803: it timed out on that test on some of the CI runs. Defaulting on the global val (60sec?) fixed it

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
samigabor: It's a typo. Fixed it for the service method, but the contract one can stay the same for now. Once the ["pull earnings into operator contract"](https://linear.app/streamr/issue/ETH-535/poolvalue-update-mechanism-should-pull-earnings-into-operatorcontract) feature is implemented, it will change name anyways (e.g. does more than just update)

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
samigabor: Moved it to constructor

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
samigabor: Updated

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: Changed to a "for of" loop

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: Ups, seems like the limit fraction was overwritten somehow and was updating only one sponsorship in both cases. Changed it back and it now updated differently for each test case.

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
samigabor: Keept to info, but updated some messages

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
samigabor: BigNumber will be handled in a future ticket when upgrading to ethers6

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
samigabor: We'll keep the "helper" as-is for now

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: Moved the common setup to beforeEach and shared method

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: Changed

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: Updated

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
samigabor: Changed to scheduleAtInterval

harbu (approved):

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,86 @@
...
+const logger = new Logger(module)
+
+const CHECK_VALUE_INTERVAL = 1000 * 60 * 60 * 24 // 1 day
+const ONE_ETHER = BigInt(1e18)
+
+export class MaintainOperatorValueService {
+    private penaltyLimitFraction: bigint
+    private readonly helper: MaintainOperatorValueHelper
+    private readonly abortController: AbortController
+    private readonly config: OperatorServiceConfig
```
teogeb: This field is redundant (the value if never read)

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,86 @@
...
+                approxValue: approxValues[i],
+                realValue: realValues[i],
+                diff: realValues[i].sub(approxValues[i])
+            }
+            sponsorships.push(sponsorship)
+            totalDiff = totalDiff + sponsorship.diff.toBigInt()
+            totalApprox = totalApprox + sponsorship.approxValue.toBigInt()
+        }
+
+        if (this.penaltyLimitFraction === undefined || this.penaltyLimitFraction === BigInt(0)) {
```
teogeb: Can `penaltyLimitFraction` be undefined? If so, should the type be be `bigint | undefined`?

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,86 @@
...
+            for (const sponsorship of sortedSponsorships) {
+                diff = diff + sponsorship.diff.toBigInt()
+                neededSponsorshipsCount += 1
+                if (diff > threshold) {
+                    break
+                }
+            }
+            
+            // pick the first entries needed to get the total diff under the threshold
+            const neededSponsorshipAddresses = sortedSponsorships.slice(0, neededSponsorshipsCount).map((sponsorship) => sponsorship.address)
```
teogeb, line 74: As this is a helper value for line 78, it would make it a bit easier to read if this value is introduced right before that line. (I.e. swap lines 77 and 78, maybe even a black line between is quite separate action to the return value calculation?

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,133 @@
...
+            signer: operatorWallet
+        }
+        return { operatorWallet, operatorContract }
+    }
+
+    const getDiffBetweenApproxAndRealValues = async (): Promise<bigint> => {
+        const { sponsorshipAddresses, approxValues, realValues } = await operatorContract.getApproximatePoolValuesPerSponsorship()
+        let diff = BigInt(0)
+        for (let i = 0; i < sponsorshipAddresses.length; i++) {
+            diff = realValues[i].toBigInt() - approxValues[i].toBigInt()
```
teogeb, line 57: My IDE shows an error:`number` can't be assigned to `bigint`.

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,133 @@
...
+        const operatorContract = await deployOperatorContract(operatorWallet)
+        logger.debug(`Operator deployed at ${operatorContract.address}`)
+        operatorConfig = {
+            operatorContractAddress: operatorContract.address,
+            provider,
+            theGraphUrl,
+            fetch,
+            signer: operatorWallet
+        }
+        return { operatorWallet, operatorContract }
```
teogeb, line 50: As this returns "an operator", it would make sense to simplify the names just to `wallet` and `contract`. Then we could it easily:
```
const operator = await deployNewOperator()
await token.connect(operator.wallet)
```

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,133 @@
...
+        for (const streamId of [streamId1, streamId2]) {
+            const sponsorship = await deploySponsorship(config, operatorWallet, { streamId })
+            await (await token.connect(operatorWallet).transferAndCall(sponsorship.address, parseEther(`${SPONSOR_AMOUNT}`), "0x")).wait()
+            await (await operatorContract.stake(sponsorship.address, parseEther(`${STAKE_AMOUNT}`))).wait()
+            expect(await token.balanceOf(sponsorship.address)).toEqual(parseEther(`${SPONSOR_AMOUNT + STAKE_AMOUNT}`))
+        }
+    }, 60 * 1000)
+
+    it("updates both sponsorships to stay over the threshold", async () => {
+        const penaltyFraction = parseEther("0.001")
```
teogeb: Is this `200` same as `SPONSOR_AMOUNT`? Could use the constant here. Also at line 115.

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,133 @@
...
+        await waitForCondition(async () => await operatorContract.totalValueInSponsorshipsWei() > totalValueInSponsorshipsBefore, 10000, 1000)
+        
+        const diff = await getDiffBetweenApproxAndRealValues()
+        expect((await operatorContract.totalValueInSponsorshipsWei()).toBigInt()).toBeGreaterThan(totalValueInSponsorshipsBefore.toBigInt())
+        expect(diff).toBeLessThan(threshold)
+
+        await maintainOperatorValueService.stop()
+    }, 60 * 1000)
+
+    it("needs only one sponsorship to stay over the threshold", async () => {
```
teogeb: The only difference between the test cases is the `penaltyFraction` value. Could use Jest's `each` helper to avoid repetition: https://dev.to/bgord/simplify-repetitive-jest-test-cases-with-test-each-310m

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,133 @@
...
+        const penaltyFraction = parseEther("0.001")
+        const threshold = penaltyFraction.mul(200).toBigInt()
+        const maintainOperatorValueService = new MaintainOperatorValueService(operatorConfig, penaltyFraction.toBigInt())
+
+        const totalValueInSponsorshipsBefore = await operatorContract.totalValueInSponsorshipsWei()
+
+        // wait for sponsorships to accumulate earnings so approximate values differ enough form the real values
+        await wait(3000)
+
+        maintainOperatorValueService.start()
```
teogeb: Not too relevant with the current implementation, but would be good to `await` start as it returns a promise.

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -0,0 +1,86 @@
...
+        await scheduleAtInterval(
+            () => this.checkValue().catch((err) => {
+                logger.warn('Encountered error while checking value', { err })
+            }),
+            CHECK_VALUE_INTERVAL,
+            true,
+            this.abortController.signal
+        )
+
+    }
```
teogeb:
```suggestion
        )
    }
```

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,133 @@
...
+
+        await waitForCondition(async () => await operatorContract.totalValueInSponsorshipsWei() > totalValueInSponsorshipsBefore, 10000, 1000)
```
teogeb: What is the data type of `operatorContract.totalValueInSponsorshipsWei()`? We can do comparison operators to that with `>`, but it is not a `bigint` as we convert it `toBigInt` in 128?

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,133 @@
...
+        ({ operatorWallet, operatorContract } = await deployNewOperator())
+
+        await (
+            await token.connect(operatorWallet).transferAndCall(operatorContract.address, parseEther(`${STAKE_AMOUNT * 2}`), operatorWallet.address)
+        ).wait()
+        for (const streamId of [streamId1, streamId2]) {
+            const sponsorship = await deploySponsorship(config, operatorWallet, { streamId })
+            await (await token.connect(operatorWallet).transferAndCall(sponsorship.address, parseEther(`${SPONSOR_AMOUNT}`), "0x")).wait()
+            await (await operatorContract.stake(sponsorship.address, parseEther(`${STAKE_AMOUNT}`))).wait()
+            expect(await token.balanceOf(sponsorship.address)).toEqual(parseEther(`${SPONSOR_AMOUNT + STAKE_AMOUNT}`))
```
teogeb: Why `expect`? Maybe should remove the expect assertion as this is just a test setup, not checking any logic about the service?

teogeb (changes requested):

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
samigabor: Removed.

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
samigabor: It should not be undefined since if this optional param is not provided it will be set to BigInt(0). Removed the undefined check.

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: Updated

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: Updated

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
samigabor: Removed blank line

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: Env issue. Leave as-is

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: Will leave as-is, as discussed.

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: Hmm, that's cool. 
Changed to test.each

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: Uses ".gt" instead of "<" according to ethers docs

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: Removed since they're not testing this service, but rather the contract implementation

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
samigabor: Swapped those lines.

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: Seems like isolated issue. No changes needed.

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -88,30 +89,7 @@ describe("MaintainOperatorValueService", () => {
         }
     }, 60 * 1000)
 
-    it("updates both sponsorships to stay over the threshold", async () => {
-        const penaltyFraction = parseEther("0.001")
-        const poolValue = STAKE_AMOUNT * 2
```
teogeb: What is the relation of const `poolValue = STAKE_AMOUNT * 2` and `SPONSOR_AMOUNT`? Seems that both have the same value of `200`.

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -23,6 +23,7 @@ const logger = new Logger(module)
 
 const SPONSOR_AMOUNT = 200
 const STAKE_AMOUNT = 100
+const cases = [parseEther("0.001"), parseEther("0.0005")]
```
teogeb: This could be inlined to line 92. (Or if we need to use constant, could rename to CASES as it is a constant.)

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    const getDiffBetweenApproxAndRealValues = async (): Promise<bigint> => {
+        const { sponsorshipAddresses, approxValues, realValues } = await operatorContract.getApproximatePoolValuesPerSponsorship()
+        let diff = BigInt(0)
+        for (let i = 0; i < sponsorshipAddresses.length; i++) {
+            diff = realValues[i].toBigInt() - approxValues[i].toBigInt()
+        }
+        return diff
+    }
+
+    beforeEach(async () => {
```
teogeb, line 62: Can `penaltyFraction` be 0? If it can, should we test that functionality?

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: Pool value should actually be: `poolValue = STAKE_AMOUNT * 2 + 0` and zero would be the freeFunds (e.g. DATA balance in operator contract, not yet put to work). STAKE_AMOUNT is not dependent of SPONSOR_AMOUNT and vice-versa. They are different values and came form different sources (e.g. stake amount comes from user delegations and sponsor amount comes from user sponsorings).

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: Inlined it.

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: Nope. 
Zero would mean no difference between approximate pool values and the real pool values is allowed => operator is always in breach of the rules => always slashed.

It will be something like 0.1 (10%), but I used very small values to not wait a long time for enough earnings to accumulate in order to reach the threshold.

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
teogeb: Would it make sense to have different values if not dependent?

samigabor: Doesn't hurt, for sure, but at the same time I'd say it doesn't really matter in this case. Changed it anyways to make it obvious that the sponsor amount does not need to be equal with the sum of the stakings.

teogeb (approved):
👍

PR #1542, opened 2023-06-15
Maintain topology service

samt1803:
Added maintainTopology service and test to Operatorplugin


samt1803:
wrong destination branch

PR #1550, opened 2023-06-19
test(client): Test server utility

teogeb:
Create `startTestServer` utility for creating simple `expess` server. There were two tests which uses such test server, and now those tests cases create the server by calling the function. 

The shutdown logic has been improved: it now waits for the `close` event.


harbu (approved):

packages/client/test/test-utils/utils.ts
```diff
@@ -223,3 +226,23 @@ export const waitForCalls = async (mockFunction: jest.Mock<any>, n: number): Pro
         return `Timeout while waiting for calls: got ${mockFunction.mock.calls.length} out of ${n}`
     })
 }
+
+export const startTestServer = async (
```
harbu: It could be nice to have this function as part of `test-utils` package. But up to you to decide whether or not it is worth the move.

harbu (approved):

PR #1554, opened 2023-06-19
fix(client): Error handling of `fetchHttpStream`

teogeb:
If the HTTP `fetch` stream fails, it now propagates the error to the piped stream (the stream which transforms chunks of texts to lines). The underlying stream throws an error e.g. if we call `AbortController#abort()`. 

Before this PR the error in the underlying stream would just stop the output of that stream, but the transform stream didn't stop as it was not notified by the error. The `AsyncGenerator` never completed in that case.

Also replaced `parseError` functionality: the `fetchHttpStream` now throws an explicit `FetchHttpStreamResponseError` if the response is not ok (i.e. HTTP 200) instead of calling `parseError` to form the error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ found [here](packages/broker/CHANGELOG.md).
 #### Fixed
 
 - Fix error handling in message pipeline (https://github.com/streamr-dev/network/pull/1479)
+- Fix resend freeze issue (https://github.com/streamr-dev/network/pull/1554)
 
 #### Security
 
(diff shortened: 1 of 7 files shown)


harbu:
> changelog?
- Fix resend freeze / stuck issue (https://github.com/streamr-dev/network/pull/1554)

harbu (approved):
Related to your open question, I think it is unconventional for the `fetchHttpStream` to take in a function for transforming errors in the first place. Instead I would just catch the error in the place where it is invoked and then do the transformation there if any error is thrown.

teogeb:
> Related to your open question, I think it is unconventional for the `fetchHttpStream` to take in a function for transforming errors in the first place. Instead I would just catch the error in the place where it is invoked and then do the transformation there if any error is thrown.

Changed this. Added additional `transformError` utility function to transform the error. Alternative way would be decrease significantly the use of our async helpers (like `map` and `forEach`) and implement this in-place.

PR #1573, opened 2023-06-21
feat: operator fleet state

harbu:
## Summary

Add utility class `OperatorFleetState` that subscribes to a coordination stream and maintains a set of active online nodes based on incoming heartbeat messages.

## Future improvements
- Instead of running `pruneOfflineNodes` at a predefined interval, we could pick the lowest timestamp in the `Map` and subtract Date.now() from it to determine when the next pruning should occur. (Remember to account for jitter!)


packages/broker/src/plugins/operator/OperatorFleetState.ts
```diff
@@ -0,0 +1,82 @@
...
+    removed: (nodeId: string) => void
+}
+
+export class OperatorFleetState extends EventEmitter<OperatorFleetStateEvents> {
+    private readonly streamrClient: StreamrClient
+    private readonly coordinationStreamId: StreamID
+    private readonly timeProvider: () => number
+    private readonly pruneAgeInMs: number
+    private readonly pruneIntervalInMs: number
+    private readonly nodes = new Map<string, number>()
```
teogeb: Would it make sense to rename this to e.g. to `nodeActivityTimestamps = new Map<NodeId, number>`  or `latestNodeTimestamps = new Map<NodeId, number>`

packages/broker/src/plugins/operator/OperatorFleetState.ts
```diff
@@ -0,0 +1,82 @@
...
+        this.pruneNodesIntervalRef = setInterval(() => this.pruneOfflineNodes(), this.pruneIntervalInMs)
+    }
+
+    async destroy(): Promise<void> {
+        clearInterval(this.pruneNodesIntervalRef)
+        await this.subscription?.unsubscribe()
+    }
+
+    getNodeIds(): string[] {
+        this.pruneOfflineNodes()
```
teogeb: Would it be ok if there is no side-effect in this node? I.e. it would just return whichever state it currently has? 

Any component which uses this data needs anyway listen to subsequent changes (the `added` and `removed` events).

packages/broker/test/unit/plugins/operator/OperatorFleetState.test.ts
```diff
@@ -0,0 +1,122 @@
+import { OperatorFleetState } from '../../../../src/plugins/operator/OperatorFleetState'
+import { mock, MockProxy } from 'jest-mock-extended'
+import StreamrClient, { MessageListener, Subscription } from 'streamr-client'
+import { toEthereumAddress, wait, waitForCondition, waitForEvent } from '@streamr/utils'
+import { toStreamID } from '@streamr/protocol'
+import { eventsWithArgsToArray } from '@streamr/test-utils'
+
+const ADDRESS = toEthereumAddress('0x61BBf708Fb7bB1D4dA10D1958C88A170988d3d1F')
```
teogeb: Could we use `randomEthereumAddress()` here, or do we need this address specifically?

packages/broker/test/unit/plugins/operator/OperatorFleetState.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    })
+
+    it('can handle invalid messages in coordination stream', async () => {
+        await state.start()
+        await setTimeAndPublishMessage(10, { foo: 'bar', 'lorem': 666 })
+        await setTimeAndPublishMessage(10, {})
+        expect(state.getNodeIds()).toEqual([])
+    })
+
+    it('handles nodes coming online', async () => {
```
teogeb, line 68: Could handle case where node publishes some other valid `msgType` to the coordination stream?

packages/broker/src/plugins/operator/OperatorFleetState.ts
harbu: Good idea, changed!

packages/broker/src/plugins/operator/OperatorFleetState.ts
harbu: Yeah maybe better without side-effect. I'll write up in the PR description about future improvements better interval logic.

packages/broker/test/unit/plugins/operator/OperatorFleetState.test.ts
harbu: Changed.

packages/broker/test/unit/plugins/operator/OperatorFleetState.test.ts
harbu: Good catch. This was actually a bug, we should only update the timer when the msgType is `heartbeat` (I suppose)

packages/broker/src/plugins/operator/OperatorFleetState.ts
```diff
@@ -0,0 +1,83 @@
...
+    removed: (nodeId: string) => void
+}
+
+export class OperatorFleetState extends EventEmitter<OperatorFleetStateEvents> {
+    private readonly streamrClient: StreamrClient
+    private readonly coordinationStreamId: StreamID
+    private readonly timeProvider: () => number
+    private readonly pruneAgeInMs: number
+    private readonly pruneIntervalInMs: number
+    private readonly nodeActivityTimestamps = new Map<string, number>()
```
harbu: Maybe like this?
```suggestion
    private readonly heartbeatTimestamps = new Map<NodeId, number>()
```

PR #1575, opened 2023-06-21
Refactor operator plugin

harbu:
## Summary

Do some refactoring of code.

Changes include:
- Adjust log levels and messages
- Remove unneeded eslint exceptions
- Use `EthereumAddress` type to avoid having to lowercase in many places
- Use `StreamID` type more
- Remove unused code
- Adjust field access modifiers
- Add some missing error handling


packages/broker/src/plugins/operator/MaintainTopologyHelper.ts
```diff
@@ -125,29 +124,28 @@ export class MaintainTopologyHelper extends EventEmitter<MaintainTopologyHelperE
         }
         const parseItems = (response: any) => {
             if (!response.operator) {
-                this.logger.error(`Operator ${this.address.toLowerCase()} not found in TheGraph`)
+                logger.debug('Unable to find operator in The Graph', { operatorContractAddress: this.operatorContractAddress })
```
samt1803: is this not an error?

packages/broker/src/plugins/operator/MaintainTopologyHelper.ts
```diff
@@ -125,29 +124,28 @@ export class MaintainTopologyHelper extends EventEmitter<MaintainTopologyHelperE
...
                 this.sponsorshipCountOfStream.set(streamId, sponsorshipCount)
-                this.logger.info(`added ${stake.sponsorship.id} to stream ${streamId} with sponsorshipCount ${sponsorshipCount}`)
             }
         }
         return Array.from(this.sponsorshipCountOfStream.keys())
     }
 
     stop(): void {
-        this.provider.removeAllListeners()
+        // TODO: remove our listeners from operatorContract
```
samt1803, line 149: good find this was a bug!

samt1803 (approved):
looks good!

packages/broker/src/plugins/operator/MaintainTopologyHelper.ts
harbu: Good point. I guess this is somewhat of a situation that would warrant user intervention as described here: https://github.com/streamr-dev/network/blob/main/internal-docs/logging-practices.md#use-log-levels-appropriately. I'll change it back to error.

packages/broker/src/plugins/operator/MaintainTopologyHelper.ts
```diff
@@ -3,108 +3,107 @@ import { Provider } from "@ethersproject/providers"
...
 
     /**
-     * Emitted when a unstaked from ALL Sponsorships for the given stream
+     * Emitted when un-staked from all Sponsorships for the given stream
      */
-    removeStakedStream: (streamId: string) => void
+    removeStakedStream: (streamId: StreamID) => void
 }
 
 export class MaintainTopologyHelper extends EventEmitter<MaintainTopologyHelperEvents> {
```
teogeb, line 33: I'd consider putting `eventEmitter` as a field instead of extending `EventEmitter`, as event handling is not a core part of this helper.

packages/broker/src/plugins/operator/MaintainTopologyHelper.ts
```diff
@@ -3,108 +3,107 @@ import { Provider } from "@ethersproject/providers"
...
-            if (!streamId) {
-                this.logger.error("Sponsorship not found!")
+            if (streamId === undefined) {
+                logger.debug('Unable to find streamId for sponsorship', { sponsorshipAddress })
                 return
             }
             this.streamIdOfSponsorship.delete(sponsorshipAddress)
-            const sponsorshipCount = (this.sponsorshipCountOfStream.get(streamId) || 1) - 1
+            const sponsorshipCount = (this.sponsorshipCountOfStream.get(streamId) ?? 1) - 1
             this.sponsorshipCountOfStream.set(streamId, sponsorshipCount)
```
teogeb, line 83: Is this intentional to have `1` as a default if there are no sponsorships? Or should it be `0` as in line 67?

packages/broker/src/plugins/operator/MaintainTopologyHelper.ts
```diff
@@ -3,108 +3,107 @@ import { Provider } from "@ethersproject/providers"
...
+        if (initialStreams.length > 0) {
+            this.emit('addStakedStream', initialStreams)
         }
     }
 
-    async getStreamId(sponsorshipAddress: string): Promise<string> {
-        const bounty = new Contract(sponsorshipAddress, sponsorshipABI, this.contract.provider as Provider) as unknown as Sponsorship
-        return bounty.streamId()
+    async getStreamId(sponsorshipAddress: string): Promise<StreamID> {
+        const bounty = new Contract(sponsorshipAddress, sponsorshipABI, this.operatorContract.provider as Provider) as unknown as Sponsorship
```
teogeb: Rename variable to `sponsorship`

packages/broker/src/plugins/operator/MaintainTopologyHelper.ts
```diff
@@ -3,108 +3,107 @@ import { Provider } from "@ethersproject/providers"
...
 
-    private async pullStakedStreams(requiredBlockNumber: number): Promise<string[]> {
-        this.logger.info(`getStakedStreams for ${this.address.toLowerCase()}`)
+    private async pullStakedStreams(requiredBlockNumber: number): Promise<StreamID[]> {
         const createQuery = (lastId: string, pageSize: number) => {
             return {
                 query: `
                     {
-                        operator(id: "${this.address.toLowerCase()}") {
+                        operator(id: "${this.operatorContractAddress}") {
```
teogeb, line 106: Could fix the indentation at lines `109`-`113`

packages/broker/src/plugins/operator/MaintainTopologyHelper.ts
```diff
@@ -125,29 +124,28 @@ export class MaintainTopologyHelper extends EventEmitter<MaintainTopologyHelperE
...
+        const queryResult = this.theGraphClient.queryEntities<any>(createQuery, parseItems) // TODO: add type
 
         for await (const stake of queryResult) {
-            if (stake.sponsorship.stream && stake.sponsorship.stream.id
-                && this.streamIdOfSponsorship.get(stake.sponsorship.id) !== stake.sponsorship.stream.id) {
-                const streamId = stake.sponsorship.stream.id
-                this.streamIdOfSponsorship.set(stake.sponsorship.id, stake.sponsorship.stream.id)
+            const sponsorshipId = stake.sponsorship?.id
+            const streamId = toStreamIDSafe(stake.sponsorship?.stream?.id)
+            if (streamId !== undefined && sponsorshipId !== undefined && this.streamIdOfSponsorship.get(sponsorshipId) !== streamId) {
```
teogeb, line 139: Are all null checks needed? Maybe the GraphQL schema guarantees that we always have e.g. `stream` field inside a `Sponsorship` entity?

packages/broker/src/plugins/operator/MaintainTopologyService.ts
```diff
@@ -1,49 +1,38 @@
...
-    } catch {
-        return undefined
-    }
-}
+const logger = new Logger(module)
 
 export class MaintainTopologyService {
     private readonly streamrClient: StreamrClient
     private readonly maintainTopologyHelper: MaintainTopologyHelper
     private readonly subscriptions = new Multimap<StreamID, Subscription>()
```
teogeb, line 12: Could put the `subscription` s the first field as this is the main state of the class. Others are just dependencies.

packages/broker/src/plugins/operator/MaintainTopologyService.ts
```diff
@@ -69,16 +58,13 @@ export class MaintainTopologyService {
         }
     }
 
-    private parseStreamIdWrapper(
+    private oneByOne(
```
teogeb: Why sequential processing is needed?

packages/broker/src/plugins/operator/MaintainTopologyHelper.ts
harbu: That would make the `sponsorshipCount` negative which would be invalid.

packages/broker/src/plugins/operator/MaintainTopologyHelper.ts
harbu: good catch, fixed

packages/broker/src/plugins/operator/MaintainTopologyService.ts
harbu: I prefer to have fields ordered from tightest to least tight (w.r.t visibility and readonlyness), E.g.
```
private readonly
private readonly
protected readonly
protected readonly
public readonly
public readonly

private
private
protected
protected
public
public
```

packages/broker/src/plugins/operator/MaintainTopologyService.ts
harbu: It is needed because events `addStakedStream` and `removeStakedStream` can lead to subscriptions (and unsubscribes) on multiple stream partitions. Because `subscribe` and `unsubscribe` are asynchronous operations, there is a risk of interleaving of subscribe / unsubscribe operations leading to potential race conditions.  The test titled `"handles concurrency properly"` in MaintainTopologyService.test.ts verifies that this is handled properly

packages/broker/src/plugins/operator/MaintainTopologyHelper.ts
harbu: Future improvement imo

packages/broker/src/plugins/operator/MaintainTopologyHelper.ts
harbu: I'll add a `TODO` on this as the point of this PR is to just improve the existing code somewhat and not attempt to solve everything.

packages/broker/src/plugins/operator/MaintainTopologyHelper.ts
harbu: Removed defensive programming

teogeb (approved):

PR #1579, opened 2023-06-21
refactor(client): Remove `GapFillFailedError`

teogeb:
When an error occurs in message ordering, the `OrderedMsgChain` calls an `onError` callback which is registered by the `OrderMessages` class. There is a parameter in that call, but `OrderMessages` doesn't use the value.

See https://github.com/streamr-dev/network/blob/73b83c0cc2af716a3b09c71f13cb05665cee25fe/packages/client/src/subscribe/OrderMessages.ts#L51

Refactored error handler to be just a callback without arguments and removed the custom `GapFillFailedError` class.

### Open questions

In PR https://github.com/streamr-dev/network/pull/1570 we changed the error handling functionality so that the pipeline can continue to process messages after error. There we consider the error case as an empty result. But if we want we could notify subscription somehow if there is an error.

If we want to notify, should we emit error only when there is a storage node error, or maybe also when we can't fill the gap (i.e. don't find the missing messages from the storage node). See: https://github.com/streamr-dev/network/blob/73b83c0cc2af716a3b09c71f13cb05665cee25fe/packages/client/test/unit/OrderedMsgChain.test.ts#L429

- If we want to notify the subscription, we should use `StreamrClientError` instead of custom `GapFillFailedError`
- It currently not possible to add error listeners for resend streams (i.e. when user calls `StreamrClient#resend()`)


harbu (approved):

teogeb:
Let's implement the error handling in NET-987.

PR #1583, opened 2023-06-22
refactor(client): Refactor gap filling

teogeb:
## Summary

Refactored gap filling functionality:
- `OrderedMessageChain.ts` has similar functionality to the previous `OrderedMsgChain.ts`, but has rewritten completely. It represents a chain of ordered messages: i.e. a new message can be appended to the chain if it references to the current latest message of the chain (if message references are used). The chain emits event when a gap is found or resolved. 
- `GapFiller` contains the gap filling logic. It listens to `gapFound` event and starts a resend for the gap. It also listens to `gapResolved` event so that it can abort an ongoing resend if all missing messages were received from another source (the real-time pipeline), or if the gap fill task timeouts.
- `OrderMessage` has similar functionality as previously, i.e. it provides us an iterator to iterate the merge outputs of multiple ordered chains. It also uses the client configs to configure `GapFiller` and `OrderedMessageChain` classes.

## Other changes

`OrderedMessageChain.ts` now uses a custom `Heap` implementation as the `heap` package didn't handle uniqueness very well. The heap package didn't allow us to use the provided comparator when checking whether an item is included in the heap (it used simple `array.indexOf` instead). 
- The custom `Heap` implements efficient binary search for finding and inserting items. But it is still maybe slower than the `heap` package? See the benchmark in https://github.com/streamr-dev/network/pull/1628

The `getStorageNodes` parameter in `Resends#resend` is no longer optional. That change simplifies the storage node handling in `GapFiller`.

The `OrderedMessageChain` emits also `unfillableGap` events. We can now easily listen to those events and implement NET-987 (in a separate PR).

## Future improvements
 
- Maybe we don't need `OrderMessages` class? In `messagePipeline` we could maybe use `GapFilledMessageChain` directly as in `MsgChainUtil` we are always handling just one chain. The multi-chain functionality is needed only in `resendSubscription`, and the relevant logic could be inlined there?
- The caching of storage node is now in OrderMessages. We could use application-level caching instead.
- `getStorageNodes` handling in `Resends#resend` could be simpler: maybe we could pass a storage node as an argument (i.e. choose the target storage node outside of `Resends` class). We could also pass information about the storage node to `messagePipeline`: if a `sourceStorageNode` is not `undefined`, it can initialize a `OrderMessages` instance which uses all other storage nodes but excludes the given storage node. If we use application-level caching of storage nodes, it is ok that `OrderMessages` would query the storage node list (as it was surely just queried by the `resend()` method and is now in cache).
- The gaps that we find from the chain and the unfillable gaps are almost the same concepts, but not exactly (as we emit unfillable gaps for gaps which we haven't "found"). We can consider if we want to use same term for both or not.


packages/client/src/subscribe/ordering/OrderedMessageChain.ts
```diff
@@ -0,0 +1,188 @@
...
+
+const areEqualRefs = (ref1: MessageRef, ref2: MessageRef) => {
+    return ref1.compareTo(ref2) === 0
+}
+
+/*
+ * This class represents a chain of messages, in which message are in ascencing order. 
+
+ * There is always one message reference (timestamp + sequence number) to the head of the chain, 
+ * and a message can be immediately added to the chain if the "prevMsgRef" of the new message reference
```
harbu:
```suggestion
 * and a message can be immediately added to the chain if the "prevMsgRef" of the new message is equal
```

packages/client/src/subscribe/ordering/OrderedMessageChain.ts
```diff
@@ -0,0 +1,188 @@
...
+ * added to the chain, the "orderedMessageAdded" event is emitted.
+ * 
+ * If the new message can't be immediately added to the chain, there must be some missing messages
+ * which we need before the message can be processed. In that case we emit the "gapFound" event.
+ * 
+ * Typically the "gapFound" triggers some component to provide us more messages (e.g. fetching
+ * messages from a storage node). Also the same component which produced the original message continues 
+ * to provide more messages. Both of these sources provide messages by calling the "addMessage" method.
+ * 
+ * The sources may provide us messages which can be added be immediately added to the chain. In that case
```
harbu:
```suggestion
 * The sources may provide us messages which can be immediately added to the chain. In that case
```

packages/client/src/subscribe/ordering/OrderedMessageChain.ts
```diff
@@ -0,0 +1,188 @@
...
+            })
+            this.eventEmitter.emit('gapResolved')
+        }
+    }
+
+    private isNextOrderedMessage(msg: StreamMessage) {
+        const previousRef = msg.getPreviousMessageRef()
+        return (previousRef !== null)
+            ? (this.lastOrderedMsg === undefined) || areEqualRefs(previousRef, this.lastOrderedMsg.getMessageRef())
+            : true
```
harbu:
```suggestion
        return (previousRef === null) || (this.lastOrderedMsg === undefined) || areEqualRefs(previousRef, this.lastOrderedMsg.getMessageRef()
 
```

packages/client/src/subscribe/ordering/OrderedMessageChain.ts
```diff
@@ -0,0 +1,188 @@
...
+            this.consumePendingOrderedMessages(
+                (msg) => this.isNextOrderedMessage(msg)
+            )
+        }
+    }
+
+    resolveMessages(to: MessageRef): void {
+        if ((this.lastOrderedMsg === undefined) || (this.lastOrderedMsg.getMessageRef().compareTo(to) < 0)) {
+            this.consumePendingOrderedMessages(
+                (msg) => (msg.getMessageRef().compareTo(to) <= 0) ? true : this.isNextOrderedMessage(msg)
```
harbu:
```suggestion
                (msg) => (msg.getMessageRef().compareTo(to) <= 0) || this.isNextOrderedMessage(msg)
```

packages/client/src/subscribe/ordering/OrderMessages.ts
```diff
@@ -0,0 +1,116 @@
...
+        retryWaitTime: config.retryResendAfter,
+        maxRequestsPerGap: (config.gapFill) ? config.maxGapRequests : 0,
+        abortSignal
+    })
+    gapFiller.start()
+    return chain
+}
+
+/**
+ * Manages message ordering and gap filling. Provides an iterator to read the sequence
```
harbu:
```suggestion
 * Manages message ordering and gap filling (per stream part). Provides an iterator to read the sequence
```

packages/client/src/subscribe/ordering/GapFiller.ts
```diff
@@ -0,0 +1,125 @@
...
+        this.chain = opts.chain
+        this.resend = opts.resend
+        this.getStorageNodeAddresses = opts.getStorageNodeAddresses
+        this.initialWaitTime = opts.initialWaitTime
+        this.retryWaitTime = opts.retryWaitTime
+        this.maxRequestsPerGap = opts.maxRequestsPerGap
+        opts.abortSignal.addEventListener('abort', () => {
+            if (this.currentTask !== undefined) {
+                this.currentTask.abortController.abort()
+            }
```
harbu:
```suggestion
            this.currentTask?.abortController.abort()
```

packages/client/src/subscribe/ordering/GapFiller.ts
```diff
@@ -0,0 +1,125 @@
...
+        })
+    }
+
+    start(): void {
+        this.chain.on('gapFound', (gap: Gap) => this.onGapFound(gap))
+        this.chain.on('gapResolved', () => this.onGapResolved())
+    }
```
harbu, line 55: Is there a reason not to do this already in constructor?

packages/client/src/subscribe/ordering/GapFiller.ts
```diff
@@ -0,0 +1,125 @@
...
+        }
+        try {
+            await runAbortableTask(async () => {
+                await wait(this.initialWaitTime, abortController.signal)
+                if (this.maxRequestsPerGap > 0) {
+                    await this.fetchFromStorageNode(gap, abortController.signal)
+                }
+                /*
+                 * The "fetchFromStorageNode" typically provides all the missing messages and the chain emits the
+                 * "gapResolved" event. In that case "chain.once('gapResolved')" callback aborts this task
```
harbu:
```suggestion
                 * "gapResolved" event. In that case "chain.on('gapResolved')" callback aborts this task 
```

packages/client/src/subscribe/ordering/OrderMessages.ts
```diff
@@ -0,0 +1,116 @@
...
+                this.abortController.signal
+            )
+            chain.on('orderedMessageAdded', (msg: StreamMessage) => this.onOrdered(msg))
+            return chain
+        })
+    }
+
+    private onOrdered(orderedMessage: StreamMessage): void {
+        if (!this.outBuffer.isDone()) {
+            this.outBuffer.push(orderedMessage)
```
harbu: Should we catch the promise here?

packages/client/src/subscribe/ordering/OrderMessages.ts
```diff
@@ -0,0 +1,116 @@
...
+        try {
+            for await (const msg of src) {
+                if (this.abortController.signal.aborted) {
+                    return
+                }
+                const chain = await this.chains.get(msg.getPublisherId(), msg.getMsgChainId())
+                chain.addMessage(msg)
+            }
+            await Promise.all(this.chains.values().map((chain) => chain.waitUntilIdle()))
+            this.outBuffer.endWrite()
```
harbu, line 107: Is there a risk here we end up calling `this.outBuffer.endWrite()` before we have even seen messages from certain chains (that just haven't arrived to our network node yet)?

packages/client/test/unit/OrderMessages2.test.ts
```diff
@@ -127,53 +132,76 @@ describe.skip(OrderingUtil, () => {
             actual[msg.getPublisherId()].push(msg.getTimestamp())
         }
 
-        const gapHandler = async (from: MessageRef, to: MessageRef, context: MsgChainContext) => {
-            const requestedMessages = groundTruthMessages[context.publisherId].filter(({ delivery, timestamp }) => {
-                return delivery === Delivery.GAP_FILL && (timestamp > from.timestamp && timestamp <= to.timestamp)
-            })
+        const gapHandler = async (from: number, to: number, publisherId: string): Promise<PushPipeline<StreamMessage>> => {
+            const pipeline = new PushPipeline<StreamMessage>
+            const requestedMessages = sortedUniqBy(groundTruthMessages[publisherId].filter(({ delivery, timestamp }) => {
```
harbu: Why is `sortedUniqBy` now needed?

packages/client/test/unit/GapFiller.test.ts
```diff
@@ -0,0 +1,241 @@
...
+    })
+}
+
+const MAX_REQUESTS_PER_GAP = 3
+const INITIAL_WAIT_TIME = 50
+
+describe('GapFiller', () => {
+
+    let chain: OrderedMessageChain
+    const onOrderedMessageAdded = jest.fn()
```
harbu:
```suggestion
    let onOrderedMessageAdded
```
This should be set in `beforeEach` so it is reset for each test?

packages/client/src/subscribe/ordering/OrderMessages.ts
teogeb: Won't throw an error. But we can wrap it with `executeSafePromise` wrapper to communicate that fact (and be super safe to ensure that this won't throw if the `PushBuffer`#push implementation changes and starts to throw).

packages/client/test/unit/OrderMessages2.test.ts
teogeb: Not needed, removed

packages/client/src/subscribe/ordering/GapFiller.ts
teogeb: Maybe. But as constructors don't usually initialize heavy tasks, it could be a kind of surprise that the whole gap fill process starts just by creating an object instance?

packages/utils/src/executeSafePromise.ts
```diff
@@ -0,0 +1,12 @@
+import { Logger } from './Logger'
+
+const logger = new Logger(module)
+
+export const executeSafePromise = async <T>(createPromise: () => Promise<T>): Promise<T> => {
+    try {
+        return await createPromise()
+    } catch (error: any) {
+        logger.error(`Assertion failed! A safe promise threw an error: ${error?.message}`, { stack: error?.stack })
```
harbu:
```suggestion
        logger.error('Assertion failure!', { message: error?.message, stack: error?.stack })
```

harbu (approved):

PR #1586, opened 2023-06-22
Use cleanup branch of streamr-docker-dev

hpihkala:
Testing whether CI runs successfully against `cleanup` branch of `streamr-docker-dev`


harbu:
Tested, can be removed?

PR #1618, opened 2023-06-28
style: add `quotes` eslint rule

teogeb:
Enforces the use of single quotes instead of double quotes. Converted all double quotes by running `eslint --fix` for each package.


harbu (approved):
LGTM; can you still verify that this merges nicely with 1.0?

PR #1628, opened 2023-06-30
feat(client): [NET-865]: gap filling strategy option

teogeb:
## Summary

Add the ability to control the gap fill strategy with `gapFillStrategy` config option.

By default this is `light` to be in line with current behavior. However, if more pedantic gap filling is wanted by the user, it is now possible to enable it by setting the option to `full`.

## Open questions

Would it make sense to unify `gapFillStrategy` and the the `gapFill` boolean flag? E.g. `gapFillStrategy='passive'` would correspond to current `gapFill=false`. Maybe `orderMessages` flag shouldn't be a separate flag?

## Benchmark

Typically the new implementation does fewer resend requests compared to the previous implementation (the requests cover larger message ranges). That is most likely the root cause why the `OrderMessages2.test.ts` unit test is slower than the similar `OrderingUtil2.test.ts` in In PR https://github.com/streamr-dev/network/pull/732. The new tests takes about 23 seconds to run, whereas the old tests took takes about 19 seconds to run (i.e. the the new one takes is 20% more time). 

As `OrderMessages2.test.ts` test is quite slow, the test  is currently disabled (by Jest `skip`). It would be good to run it manually when we make changes to ordering functionality.


packages/client/src/subscribe/ordering/GapFiller.ts
```diff
@@ -5,10 +5,13 @@ import { Gap, OrderedMessageChain } from './OrderedMessageChain'
 
 const logger = new Logger(module)
 
+export type GapFillStrategy = 'light' | 'full'
+
 interface GapFillerOptions {
     chain: OrderedMessageChain
     resend: (gap: Gap, storageNodeAddress: EthereumAddress, abortSignal: AbortSignal) => AsyncGenerator<StreamMessage>
     getStorageNodeAddresses: () => Promise<EthereumAddress[]>
+    strategy: 'full' | 'light'
```
harbu:
```suggestion
    strategy: GapFillStrategy
```

packages/client/src/subscribe/ordering/OrderedMessageChain.ts
```diff
@@ -107,19 +107,23 @@ export class OrderedMessageChain {
         }
     }
 
-    resolveMessages(): void {
+    resolveMessages(to: MessageRef | undefined, gapCheckEnabled: boolean = true): void {
```
harbu: I would not use a default boolean value here for the 2nd parameter.

```suggestion
    resolveMessages(to: MessageRef | undefined, gapCheckEnabled: boolean): void {
```

packages/client/src/subscribe/ordering/GapFiller.ts
```diff
@@ -72,9 +77,13 @@ export class GapFiller {
...
                  * this task has not been aborted by either of these reasons, we resolve the gap manually as we
-                 * don't try to fill it anymore. We also drop all accumulated gaps.
+                 * don't try to fill it anymore.
                  */
-                this.chain.resolveMessages()
+                if (this.strategy === 'full') {
+                    this.chain.resolveMessages(gap.to.getMessageRef())
+                } else if (this.strategy === 'light') {
+                    this.chain.resolveMessages(undefined, false)
+                }
```
harbu:
```suggestion
                } else if (this.strategy === 'light') {
                    this.chain.resolveMessages(undefined, false)
                } else {
                    throw new Error('unexpected gap strategy')
                }
```

harbu (approved):

PR #1629, opened 2023-07-03
Watch if operators are in breach of the threshold

samigabor:
## Summary

[Tries to find Operator contracts that are in breach with the update rules, to cash in rewards. Periodically select an Operator contract and inspect if total diff in accumulated rewards is over the threshold. If so, auto-withdraw them from sponsorship to operator contract and get a share of the operator's cut, for doing his job.](https://linear.app/streamr/issue/ETH-537/operatorvaluebreachwatcher)

## Changes

Watcher implementation and maintain operator update to auto-withdraw.
Other fastchain related fixes.


.github/workflows/validate.yml
```diff
@@ -110,6 +110,14 @@ jobs:
...
       run-entry-point: true
       command: npm run test-integration
+  broker-e2e-fast:
+    needs: build
+    uses: ./.github/workflows/test-setup.yml
+    with:
+      package: broker
+      docker-services: cassandra init-keyspace parity-sidechain-node0 graph-deploy-streamregistry-subgraph deploy-network-subgraphs deploy-network-subgraphs-fastchain
+      run-entry-point: true
+      command: npm run test-operator-plugins
```
harbu: fast-chain is now in use in normal broker-e2e tests, so this can be removed?

packages/broker/package.json
```diff
@@ -22,14 +22,15 @@
     "test": "jest test/unit test/integration && npm run test-sequential",
     "test-unit": "jest test/unit",
     "test-sequential": "jest --forceExit --maxWorkers=1 test/sequential # always run sequential tests with maxWorkers=1",
-    "test-integration": "jest --forceExit test/integration && npm run test-sequential"
+    "test-integration": "jest --testPathIgnorePatterns='test/integration/plugins/operator/' --forceExit test/integration && npm run test-sequential",
+    "test-operator-plugins": "jest --forceExit test/integration/plugins/operator"
```
harbu: fast-chain is now in use in normal broker-e2e tests, so this can be reverted?

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
```diff
@@ -1,33 +1,127 @@
...
     }
 
-    // returns a wei value (1 ETH means 100%)
+    async getRandomOperator(): Promise<EthereumAddress> {
+        const latestBlock = await this.operator.provider.getBlockNumber()
+        const queryFilter = '' // e.g. (first: 10, orderBy: poolValue, orderDirection: desc)
+        const operators = await this.getOperatorAddresses(latestBlock, queryFilter)
+        // filter out my own operator
+        const operatorAddresses = operators.filter((id) => id !== this.config.operatorContractAddress)
+        logger.info(`Found ${operatorAddresses.length} operators`, { operatorAddresses })
```
harbu: Let's keep details at `debug` level as our end-users don't need to know about this details unless they want to diagnose an issue and enable `LOG_LEVEL=debug`.

```suggestion
        logger.debug(`Found ${operatorAddresses.length} operators`, { operatorAddresses })
```

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
```diff
@@ -1,33 +1,127 @@
...
+        const operator = new Contract(operatorContractAddress, operatorABI, this.config.signer) as unknown as Operator
+        const minSponsorshipEarningsWei = BigNumber.from(this.config.minSponsorshipEarnings ?? 0)
+        const { sponsorshipAddresses: allSponsorshipAddresses, earnings } = await operator.getEarningsFromSponsorships()
+
+        const sponsorships = allSponsorshipAddresses
+            .map((address, i) => ({ address, earnings: earnings[i] }))
+            .filter((sponsorship) => sponsorship.earnings.gte(minSponsorshipEarningsWei))
+            .sort((a, b) => Number(b.earnings.sub(a.earnings).toBigInt())) // TODO: after Node 20, use .toSorted() instead
+            .slice(0, this.config.maxSponsorshipsCount) // take all if maxSponsorshipsCount is undefined
+        const sponsorshipAddresses = sponsorships.map((sponsorship) => sponsorship.address as EthereumAddress)
```
harbu: If you use the `EthereumAddress` type it is assumed that it is (a) a valid address and (b) in lower-case. Here the first one most likely always holds since it is read from a smart contract but the second one not.

So we should avoid casting here and instead use the helper function `toEthereumAddress`:

```suggestion
        const sponsorshipAddresses = sponsorships.map((sponsorship) => toEthereumAddress(sponsorship.address))
```

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
```diff
@@ -1,33 +1,127 @@
...
     }
 
-    async updateApproximatePoolValueOfSponsorships(sponsorshipAddresses: string[]): Promise<void> {
-        await (await this.operator.updateApproximatePoolvalueOfSponsorships(sponsorshipAddresses)).wait()
+    private async getOperatorAddresses(requiredBlockNumber: number, queryFilter: string): Promise<EthereumAddress[]> {
+        const createQuery = () => {
+            return {
+                query: `
+                    {
+                        operators${queryFilter} {
```
harbu: Parameter `queryFilter` left blank in actual uses of method. I would remove parameter.
```suggestion
    private async getOperatorAddresses(requiredBlockNumber: number): Promise<EthereumAddress[]> {
        const createQuery = () => {
            return {
                query: `
                    {
                        operators {
```

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
```diff
@@ -1,33 +1,127 @@
...
+                            number
+                            }
+                        }
+                    }
+                    `
+            }
+        }
+        const parseItems = (response: any) => {
+            if (!response.operators) {
+                logger.error('Unable to find operators in The Graph')
```
harbu: In a sense a "valid" situation (given no operators around) and not an error. I'd remove logging.
```suggestion
```

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
```diff
@@ -1,33 +1,127 @@
...
+                return []
+            }
+            return response.operators
+        }
+        this.theGraphClient.updateRequiredBlockNumber(requiredBlockNumber)
+        const queryResult = this.theGraphClient.queryEntities<any>(createQuery, parseItems)
```
harbu:
```suggestion
        const queryResult = this.theGraphClient.queryEntities<{ id: string }>(createQuery, parseItems)
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -1,81 +1,49 @@
...
-const ONE_ETHER = BigInt(1e18)
+const DEFAULT_CHECK_VALUE_INTERVAL_MS = 1000 * 60 * 60 * 24 // 1 day
 
 export class MaintainOperatorValueService {
-    private penaltyLimitFraction: bigint
+    private readonly withdrawLimitSafetyFraction: bigint
+    private penaltyLimitFraction?: bigint
     private readonly helper: MaintainOperatorValueHelper
     private readonly abortController: AbortController
+    private readonly checkIntervalMs: number
```
harbu: (nitpick) To be in line with rest of code base, we've usually used "inMs" suffix. 
```suggestion
    private readonly checkIntervalInMs: number
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -1,81 +1,49 @@
...
-    constructor(config: OperatorServiceConfig, penaltyLimitFraction = BigInt(0)) {
-        this.penaltyLimitFraction = penaltyLimitFraction
+    constructor(config: OperatorServiceConfig, withdrawLimitSafetyFraction = 0.5, checkValueIntervalMs = DEFAULT_CHECK_VALUE_INTERVAL_MS) {
+        this.withdrawLimitSafetyFraction = BigInt(withdrawLimitSafetyFraction * 1e18)
         this.helper = new MaintainOperatorValueHelper(config)
         this.abortController = new AbortController()
+        this.checkIntervalMs = checkValueIntervalMs
+    }
+
+    async checkMyUnwithdrawnEarnings(): Promise<void> {
```
harbu: A private helper method.
```suggestion
    private async checkMyUnwithdrawnEarnings(): Promise<void> {
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -1,81 +1,49 @@
...
+        const safeUnwithdrawnEarningsFraction = this.penaltyLimitFraction! * this.withdrawLimitSafetyFraction / BigInt(1e18)
+        logger.info(` -> is ${fraction / BigInt(1e16)}% > ${safeUnwithdrawnEarningsFraction / BigInt(1e16)}% ?`)
+        if (fraction > safeUnwithdrawnEarningsFraction) {
+            logger.info('Withdrawing earnings from sponsorships', { sponsorshipAddresses })
+            await this.helper.withdrawMyEarningsFromSponsorships(sponsorshipAddresses)
+        }
     }
 
     async start(): Promise<void> {
+        this.penaltyLimitFraction = await this.helper.getPenaltyLimitFraction()
```
harbu, line 30: (nitpick) Let's move `start` method right after constructor for clarity. (Since everything else is a private helper method)

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -1,81 +1,49 @@
...
+
         await scheduleAtInterval(
-            () => this.checkValue().catch((err) => {
-                logger.warn('Encountered error while checking value', { err })
+            () => this.checkMyUnwithdrawnEarnings().catch((err) => {
+                logger.warn('Encountered error during checkUnwithdrawnEarnings', { err })
```
harbu: We could even elevate this to `error` level since this isn't a typical error scenario, right?
```suggestion
                logger.error('Encountered error while checking unwithdrawn earnings', { err })
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -1,81 +1,49 @@
...
-            
-            logger.info('Update approximate pool values of sponsorships', { threshold, diffPercentage: diffSum / totalDiff })
-            // pick the first entries needed to get the total diff under the threshold
-            const neededSponsorshipAddresses = sortedSponsorships.slice(0, neededSponsorshipsCount).map((sponsorship) => sponsorship.address)
-            await this.helper.updateApproximatePoolValueOfSponsorships(neededSponsorshipAddresses)
-        }
-    }
-
     async stop(): Promise<void> {
         logger.info('Stop')
```
harbu: Not really needed imo
```suggestion
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -1,81 +1,49 @@
...
-        this.penaltyLimitFraction = penaltyLimitFraction
+    constructor(config: OperatorServiceConfig, withdrawLimitSafetyFraction = 0.5, checkValueIntervalMs = DEFAULT_CHECK_VALUE_INTERVAL_MS) {
+        this.withdrawLimitSafetyFraction = BigInt(withdrawLimitSafetyFraction * 1e18)
         this.helper = new MaintainOperatorValueHelper(config)
         this.abortController = new AbortController()
+        this.checkIntervalMs = checkValueIntervalMs
+    }
+
+    async checkMyUnwithdrawnEarnings(): Promise<void> {
+        logger.info('Check unwithdrawn earnings % of total value, and check if it\'s above the safe threshold')
```
harbu: This is good to log imo (since it happens only hourly), can we make the log message a bit more succinct so it is clear to the end-user what is happening?

Maybe something like?
```suggestion
        logger.info('Check whether it is time to withdraw my earnings')
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -1,81 +1,49 @@
...
+        const { fraction, sponsorshipAddresses } = await this.helper.getMyUnwithdrawnEarnings()
+        const safeUnwithdrawnEarningsFraction = this.penaltyLimitFraction! * this.withdrawLimitSafetyFraction / BigInt(1e18)
```
harbu: I noticed we multiplied `withdrawLimitSafetyFraction` with `1e18` in constructor and here we divide by `1e18` reversing that multiplication? So would it be equivalent just to remove the multiplication (in the constructor) and division (here)?
 
```suggestion
        const safeUnwithdrawnEarningsFraction = this.penaltyLimitFraction! * this.withdrawLimitSafetyFraction / BigInt(1e18)
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -1,81 +1,49 @@
...
+        logger.info(` -> is ${fraction / BigInt(1e16)}% > ${safeUnwithdrawnEarningsFraction / BigInt(1e16)}% ?`)
```
harbu: Development-time debugging log? If so remove, or if you feel it is useful to have around maybe make into `trace` level, e.g.

```suggestion
        logger.trace(` -> is ${fraction / BigInt(1e16)}% > ${safeUnwithdrawnEarningsFraction / BigInt(1e16)}% ?`)
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -1,81 +1,49 @@
...
+        if (fraction > safeUnwithdrawnEarningsFraction) {
+            logger.info('Withdrawing earnings from sponsorships', { sponsorshipAddresses })
```
harbu:
```suggestion
            logger.info('Withdraw earnings from sponsorships', { sponsorshipAddresses })
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -1,81 +1,49 @@
...
+            await this.helper.withdrawMyEarningsFromSponsorships(sponsorshipAddresses)
+        }
```
harbu: May be useful to let the user know also when we decide not to withdraw.

```suggestion
            await this.helper.withdrawMyEarningsFromSponsorships(sponsorshipAddresses)
        } else {
            logger.info('Skip withdrawing earnings', { fraction, safeUnwithdrawnEarningsFraction })
        }
```

.github/workflows/validate.yml
```diff
@@ -110,6 +110,14 @@ jobs:
       docker-services: cassandra init-keyspace dev-chain-fast deploy-network-subgraphs-fastchain
       run-entry-point: true
       command: npm run test-integration
+  broker-e2e-fast:
```
teogeb: The streamr-1.0 branch now uses fastchain, so we don't need this if you merge latest from that branch

packages/broker/src/plugins/operator/OperatorValueBreachWatcher.ts
```diff
@@ -0,0 +1,59 @@
...
+
+const logger = new Logger(module)
+
+const CHECK_VALUE_INTERVAL = 1000 * 60 * 60 // 1 hour
+
+export class OperatorValueBreachWatcher {
+    private penaltyLimitFractionCached?: bigint
+    private readonly abortController: AbortController
+    
+    readonly helper: MaintainOperatorValueHelper
```
harbu, line 14:
```suggestion
    // public access modifier for tests 
    readonly helper: MaintainOperatorValueHelper
```

packages/broker/src/plugins/operator/OperatorValueBreachWatcher.ts
```diff
@@ -0,0 +1,59 @@
...
+                logger.warn('Encountered error while watching operators', { err })
+            }),
+            CHECK_VALUE_INTERVAL,
+            true,
+            this.abortController.signal
+        )
+    }
+
+    async stop(): Promise<void> {
+        logger.info('Stop')
```
harbu:
```suggestion
```

packages/broker/package.json
```diff
@@ -22,14 +22,15 @@
     "test": "jest test/unit test/integration && npm run test-sequential",
     "test-unit": "jest test/unit",
     "test-sequential": "jest --forceExit --maxWorkers=1 test/sequential # always run sequential tests with maxWorkers=1",
-    "test-integration": "jest --forceExit test/integration && npm run test-sequential"
+    "test-integration": "jest --testPathIgnorePatterns='test/integration/plugins/operator/' --forceExit test/integration && npm run test-sequential",
```
teogeb: It would be good to run all integration tests with `test-integration`.

Is there a need for this change as we can run single test or subset of tests from command line with `npx`:
`npx jest test/integration/plugins/operator/OperatorPlugin.test.ts`

packages/broker/src/plugins/operator/OperatorValueBreachWatcher.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    async getPenaltyLimitFraction(): Promise<bigint> {
+        if (!this.penaltyLimitFractionCached) {
+            this.penaltyLimitFractionCached = await this.helper.getPenaltyLimitFraction()
+        }
+        return this.penaltyLimitFractionCached
+    }
+
+    async start(): Promise<void> {
```
harbu: (nitpick): Move this method right after constructor.

packages/broker/src/plugins/operator/OperatorValueBreachWatcher.ts
```diff
@@ -0,0 +1,59 @@
...
+        await scheduleAtInterval(
+            () => this.checkRandomUnwithdrawnEarnings().catch((err) => {
+                logger.warn('Encountered error while watching operators', { err })
```
harbu:
```suggestion
                logger.warn('Encountered error', { err })
```

packages/broker/src/plugins/operator/OperatorValueBreachWatcher.ts
```diff
@@ -0,0 +1,59 @@
...
+    private readonly abortController: AbortController
+    
+    readonly helper: MaintainOperatorValueHelper
+
+    constructor(config: OperatorServiceConfig) {
+        this.helper = new MaintainOperatorValueHelper(config)
+        this.abortController = new AbortController()
+    }
+
+    async checkRandomUnwithdrawnEarnings(): Promise<void> {
```
harbu:
```suggestion
    private async checkRandomUnwithdrawnEarnings(): Promise<void> {
```

packages/broker/src/plugins/operator/OperatorValueBreachWatcher.ts
```diff
@@ -0,0 +1,59 @@
...
+        const randomOperatorAddress = await this.helper.getRandomOperator()
+        logger.info('Checking unwithdrawn earnings', { randomOperatorAddress })
```
harbu:
```suggestion
        logger.info('Check unwithdrawn earnings', { randomOperatorAddress })
```

packages/broker/src/plugins/operator/OperatorValueBreachWatcher.ts
```diff
@@ -0,0 +1,59 @@
...
+        return this.checkUnwithdrawnEarningsOf(randomOperatorAddress)
+    }
+
+    async checkUnwithdrawnEarningsOf(targetOperatorAddress: EthereumAddress): Promise<void> {
```
harbu:
```suggestion
    private async checkUnwithdrawnEarningsOf(targetOperatorAddress: EthereumAddress): Promise<void> {
```

packages/broker/src/plugins/operator/OperatorValueBreachWatcher.ts
```diff
@@ -0,0 +1,59 @@
...
+        logger.info('Check unwithdrawn earnings and check if they are above the safe threshold')
```
harbu: We already log an info statement the the caller `checkRandomUnwithdrawnEarnings`. A bit redundant here (from end-user perspective).
```suggestion
```

packages/broker/src/plugins/operator/OperatorValueBreachWatcher.ts
```diff
@@ -0,0 +1,59 @@
...
+        const { fraction, sponsorshipAddresses } = await this.helper.getUnwithdrawnEarningsOf(targetOperatorAddress)
+        const limit = await this.getPenaltyLimitFraction()
+        logger.info(` -> is ${fraction} > ${limit}?`)
```
harbu: Looks like a development-time debug statement. Remove, or move to trace level:
```suggestion
        logger.trace(` -> is ${fraction} > ${limit}?`)
```

packages/broker/src/plugins/operator/OperatorValueBreachWatcher.ts
```diff
@@ -0,0 +1,59 @@
...
+        if (fraction > limit) {
+            logger.info('Withdrawing earnings from sponsorships', { sponsorshipAddresses })
```
harbu:
```suggestion
            logger.info('Withdraw earnings from sponsorships (target operator value in breach)', { targetOperatorAddress, sponsorshipAddresses, fraction, limit })
```

packages/broker/src/plugins/operator/OperatorValueBreachWatcher.ts
```diff
@@ -0,0 +1,59 @@
...
+            await this.helper.triggerWithdraw(targetOperatorAddress, sponsorshipAddresses)
+        }
+    }
+
+    async getPenaltyLimitFraction(): Promise<bigint> {
```
harbu:
```suggestion
    private async getPenaltyLimitFraction(): Promise<bigint> {
```

packages/broker/test/integration/plugins/operator/OperatorValueBreachWatcher.test.ts
```diff
@@ -0,0 +1,128 @@
...
+        const operatorConfig = {
+            operatorContractAddress: toEthereumAddress(operatorContract.address),
+            provider,
+            theGraphUrl,
+            signer: operatorWallet,
+            maxSponsorshipsCount: 20,
+            minSponsorshipEarnings: 1 // full tokens
+        }
+        return { operatorWallet, operatorContract, operatorConfig }
+    }
```
harbu: Can we use pre-existing `setupOperatorContract` to avoid repeating code?

packages/broker/test/integration/plugins/operator/OperatorValueBreachWatcher.test.ts
```diff
@@ -0,0 +1,128 @@
...
+            `/operatorvaluewatchertest-${Date.now()}`,
+            '{"partitions":1}')
+        ).wait()
+        streamId = createStreamReceipt.events?.find((e) => e.event === 'StreamCreated')?.args?.id
+        const streamExists = await streamRegistry.exists(streamId)
+        logger.debug('Stream created:', { streamId, streamExists })
+        // TODO: use createClient once configs allow it. For now I'm creating the stream directly using the contract
+        // const client = createClient(STREAM_CREATION_KEY)
+        // streamId = (await createTestStream(client, module)).id
+        // await client.destroy()
```
harbu: You can use the `client` to create the stream now since we are using fast-chain by default in client now.

packages/broker/test/integration/plugins/operator/OperatorValueBreachWatcher.test.ts
```diff
@@ -0,0 +1,128 @@
...
+
+        token = new Contract(fastChainConfig.contracts.DATA, tokenABI) as unknown as TestToken
+    }, 60 * 1000)
+
+    it('can find a random operator, excluding himself', async () => {
```
harbu: This is really testing the method `getRandomOperator` of class `MaintainOperatorValueHelper`.

I would create a new integration test `MaintainOperatorValueHelper.test.ts" and move this test there.

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -1,125 +1,108 @@
...
 const logger = new Logger(module)
 
-const SPONSOR_AMOUNT = 250
-const STAKE_AMOUNT = 100
-
 const STREAM_CREATION_KEY = '0xb1abdb742d3924a45b0a54f780f0f21b9d9283b231a0a0b35ce5e455fa5375e7'
 
-// test is outdated, is completely rewritten and will be merged with PR #1629 
-describe.skip('MaintainOperatorValueService', () => {
+async function getTotalUnwithdrawnEarnings(operatorContract: Operator): Promise<bigint> {
```
harbu: Same function duplicated in OperatorValueBreachWatcher.test.ts? Extract into common helper function and use in both to avoid code repetition

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -1,125 +1,108 @@
...
+            operatorContractAddress: toEthereumAddress(operatorContract.address),
+            provider,
+            theGraphUrl,
+            signer: operatorWallet,
+            maxSponsorshipsCount: 20,
+            minSponsorshipEarnings: 1 // full tokens
         }
-        return totalDiff
+        return { operatorWallet, operatorContract, operatorConfig }
     }
```
harbu: Can we use pre-existing `setupOperatorContract` for this? (To avoid code repetition)

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -1,125 +1,108 @@
...
-            await (await token.connect(operatorWallet).transferAndCall(sponsorship.address, parseEther(`${SPONSOR_AMOUNT}`), '0x')).wait()
-            await (await operatorContract.stake(sponsorship.address, parseEther(`${STAKE_AMOUNT}`))).wait()
-        }
+        streamId = createStreamReceipt.events?.find((e) => e.event === 'StreamCreated')?.args?.id
+        const streamExists = await streamRegistry.exists(streamId)
+        logger.debug('Stream created:', { streamId, streamExists })
+        // TODO: use createClient once configs allow it. For now I'm creating the stream directly using the contract
+        // const client = createClient(STREAM_CREATION_KEY)
+        // streamId = (await createTestStream(client, module)).id
+        // await client.destroy()
```
harbu: fast-chain is enabled in client now in streamr-1.0 branch. So you can use `client` to create streams.

harbu (changes requested):

packages/broker/package.json
jtakalai: the reason for splitting operator tests out was to run them with fast-chain. So if now all `test-integration` already work with fast-chain, then this split isn't needed.

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
jtakalai: there should be pagination coming up (for the random selection), so the queryFilter is probably needed

teogeb: The pagination is automatically handled if we inject the the parameters of `createQuery` (`lastId: string, pageSize: number`) to line `101`. (see e.g. https://github.com/streamr-dev/network/blob/f782bb9155e2d399fe5424d56e9a6b411945bc27/packages/broker/src/plugins/operator/MaintainTopologyHelper.ts#L107)

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
```diff
@@ -1,33 +1,124 @@
...
+        })
     }
 
-    // returns a wei value (1 ETH means 100%)
+    async getRandomOperator(): Promise<EthereumAddress | undefined> {
+        const latestBlock = await this.operator.provider.getBlockNumber()
+        const queryFilter = '' // e.g. (first: 10, orderBy: poolValue, orderDirection: desc)
+        const operators = await this.getOperatorAddresses(latestBlock, queryFilter)
+        // filter out my own operator
+        const operatorAddresses = operators.filter((id) => toEthereumAddress(id) !== toEthereumAddress(this.config.operatorContractAddress))
```
teogeb: both of the comparison operands have already the type of `EthereumAddress`

```suggestion
        const operatorAddresses = operators.filter((id) => id !== this.config.operatorContractAddress)
```

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
```diff
@@ -1,33 +1,124 @@
...
+        logger.info(`Found ${operatorAddresses.length} operators`, { operatorAddresses })
+        const randomIndex = Math.floor(Math.random() * operatorAddresses.length)
+        return operatorAddresses[randomIndex]
```
teogeb: `import sample from 'lodash/sample'` and

```suggestion
        return sample(operatorAddresses)
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -1,81 +1,49 @@
...
+    private readonly withdrawLimitSafetyFraction: bigint
+    private penaltyLimitFraction?: bigint
     private readonly helper: MaintainOperatorValueHelper
     private readonly abortController: AbortController
+    private readonly checkIntervalMs: number
 
-    constructor(config: OperatorServiceConfig, penaltyLimitFraction = BigInt(0)) {
-        this.penaltyLimitFraction = penaltyLimitFraction
+    constructor(config: OperatorServiceConfig, withdrawLimitSafetyFraction = 0.5, checkValueIntervalMs = DEFAULT_CHECK_VALUE_INTERVAL_MS) {
+        this.withdrawLimitSafetyFraction = BigInt(withdrawLimitSafetyFraction * 1e18)
```
teogeb: Use `ONE_ETHER` from the helper (also at line 26)

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -1,81 +1,49 @@
...
-    private penaltyLimitFraction: bigint
+    private readonly withdrawLimitSafetyFraction: bigint
+    private penaltyLimitFraction?: bigint
     private readonly helper: MaintainOperatorValueHelper
     private readonly abortController: AbortController
+    private readonly checkIntervalMs: number
 
-    constructor(config: OperatorServiceConfig, penaltyLimitFraction = BigInt(0)) {
-        this.penaltyLimitFraction = penaltyLimitFraction
+    constructor(config: OperatorServiceConfig, withdrawLimitSafetyFraction = 0.5, checkValueIntervalMs = DEFAULT_CHECK_VALUE_INTERVAL_MS) {
```
teogeb: The default value of `withdrawLimitSafetyFraction` could be a constant, similarly to `checkValueIntervalMs`.

packages/broker/src/plugins/operator/OperatorPlugin.ts
```diff
@@ -50,7 +54,9 @@ export class OperatorPlugin extends Plugin<OperatorPluginConfig> {
             operatorContractAddress: toEthereumAddress(this.pluginConfig.operatorContractAddress),
             // TODO read from client, as we need to use production value in production environment (not ConfigTest)
             theGraphUrl: CONFIG_TEST.contracts!.theGraphUrl!,
-            signer: Wallet.createRandom().connect(provider)
+            signer: Wallet.createRandom().connect(provider),
+            maxSponsorshipsCount: 20, // max number of sponsorships to loop over before tx reverts
+            minSponsorshipEarnings: 1 // token value, not wei
```
teogeb: Introduce constants for these?

packages/broker/src/plugins/operator/OperatorPlugin.ts
```diff
@@ -27,6 +28,8 @@ export interface OperatorServiceConfig {
     signer: Signer
     operatorContractAddress: EthereumAddress
     theGraphUrl: string
+    maxSponsorshipsCount?: number
```
teogeb: Maybe a more specific name if this limits just the operator value maintenance, not anything else?

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -1,125 +1,96 @@
...
+            signer: operatorWallet,
+            maxSponsorshipsCount: 20,
+            minSponsorshipEarnings: 1 // full tokens
         }
-        return totalDiff
+        return { operatorWallet, operatorContract, operatorConfig }
     }
 
     beforeAll(async () => {
+        provider = new JsonRpcProvider(`http://${STREAMR_DOCKER_DEV_HOST}:8547`)
```
teogeb: Url from `chainConfig.rpcEndpoints[0].url`

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -1,125 +1,96 @@
...
-        expect((approxValuesAfter[0].toBigInt() > approxValuesBefore[0].toBigInt()
-            || approxValuesAfter[1].toBigInt() > approxValuesBefore[1].toBigInt())
-            && !((approxValuesAfter[0].toBigInt() > approxValuesBefore[0].toBigInt()
-            && approxValuesAfter[1].toBigInt() > approxValuesBefore[1].toBigInt()))).toBeTruthy()
+        // 1000 = check every second
+        const service = new MaintainOperatorValueService(operatorConfig, 0.5, 1000)
+
+        const poolValueBeforeWithdraw = await operatorContract.getApproximatePoolValue()
+
+        // safe threshold is 5 DATA, that's when withdraw happens
```
teogeb: Which calculation or config defines that value of `5`? (Same for value `3` at lines `87`-`88`)

packages/broker/test/integration/plugins/operator/OperatorValueBreachWatcher.test.ts
```diff
@@ -0,0 +1,121 @@
...
+        const { operatorContract, operatorConfig } = await deployNewOperator()
+        // deploy another operator to make sure there are at least 2 operators
+        await deployNewOperator()
+
+        const operatorValueBreachWatcher = new OperatorValueBreachWatcher(operatorConfig)
+        const randomOperatorAddress = await operatorValueBreachWatcher.helper.getRandomOperator()
+        if (randomOperatorAddress === undefined) {
+            throw new Error('No random operator found')
+        }
```
teogeb:
```suggestion
        expect(randomOperatorAddress).toBeDefined()
```

packages/broker/test/integration/plugins/operator/OperatorValueBreachWatcher.test.ts
```diff
@@ -0,0 +1,121 @@
...
+        // check it's a valid operator, deployed by the OperatorFactory
+        const adminWallet = new Wallet(STREAM_CREATION_KEY, provider)
+        const operatorFactory = new Contract(fastChainConfig.contracts.OperatorFactory, operatorFactoryABI, adminWallet) as unknown as OperatorFactory
+        const isDeployedByFactory = await operatorFactory.deploymentTimestamp(randomOperatorAddress)
```
teogeb: Is the variable name correct?

.github/workflows/validate.yml
samigabor: Removed

packages/broker/package.json
samigabor: Removed the split.

.github/workflows/validate.yml
samigabor: Yes, removed it.

packages/broker/package.json
samigabor: yes, reverted

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
samigabor: Changed

packages/broker/test/integration/plugins/operator/OperatorValueBreachWatcher.test.ts
samigabor: Changed it

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: Updated.

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
samigabor: removed param

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
samigabor: Changed

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
samigabor: Moved

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
samigabor: removed

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
samigabor: Changed

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
samigabor: Changed to trace

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
samigabor: Added

packages/broker/src/plugins/operator/OperatorValueBreachWatcher.ts
samigabor: Removed

packages/broker/src/plugins/operator/OperatorValueBreachWatcher.ts
samigabor: Moved

packages/broker/src/plugins/operator/OperatorValueBreachWatcher.ts
samigabor: Updated

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
samigabor: Multiplication needed in the constructor in order to convert it to BigInt. 
And this way, it also reflects the wei value you have in the contact

packages/broker/src/plugins/operator/OperatorValueBreachWatcher.ts
samigabor: Changed

packages/broker/src/plugins/operator/OperatorValueBreachWatcher.ts
samigabor: Updated

packages/broker/test/integration/plugins/operator/OperatorValueBreachWatcher.test.ts
samigabor: Couldn't use that one because `{ operatorSharePercent: 10 }` param was needed

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: Couldn't use that one because { operatorSharePercent: 10 } param was needed

packages/broker/test/integration/plugins/operator/OperatorValueBreachWatcher.test.ts
jtakalai: yes, if it has a deployment timestamp then it was deployed by that factory. I guess it could be cast to bool by adding `.gt(0)` to the end

packages/broker/test/integration/plugins/operator/OperatorValueBreachWatcher.test.ts
jtakalai: one of the points of having that comparison is making it clear to TypeScript that `randomOperatorAddress` can not be undefined. TS can't infer it from `expect`.

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
jtakalai: TODO: add more comments.

5 DATA = 50% of 10 DATA, 50% is the "safe threshold" (we don't want to wait all the way to the limit, lest we be late)
10 DATA = 10% of 100 DATA, 10% is the penalty limit (sum of unwithdrawn earnings may not exceed it) 
100 DATA = pool value

packages/broker/src/plugins/operator/OperatorPlugin.ts
jtakalai: maybe maxSponsorshipsInWithdraw

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: Extracted

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
samigabor: Updated

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
samigabor: Updated

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
samigabor: Updated

packages/broker/src/plugins/operator/OperatorPlugin.ts
samigabor: Updated

packages/broker/src/plugins/operator/OperatorPlugin.ts
samigabor: Updated

packages/broker/test/integration/plugins/operator/OperatorValueBreachWatcher.test.ts
samigabor: Left it as-is for now

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: Added more comments

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: Updated

packages/broker/test/integration/plugins/operator/OperatorValueBreachWatcher.test.ts
samigabor: Updated

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
samigabor: Left the multiplication, but extracted the one ether value into a const (as suggested below)

packages/broker/test/integration/plugins/operator/OperatorValueBreachWatcher.test.ts
harbu: I would solve this by adding the possibility to pass that param into setupOperatorContract; see https://github.com/streamr-dev/network/pull/1763 for my solution idea.

packages/broker/test/integration/plugins/operator/OperatorValueBreachWatcher.test.ts
harbu: I noticed this wasn't moved. I would move it as done here https://github.com/streamr-dev/network/pull/1763

harbu (commented):
Most things looking very good now 👍 , I had couple of suggestions more, I thought easiest maybe just to write them down as code see here: https://github.com/streamr-dev/network/pull/1763 and let me know what you think.

packages/broker/test/integration/plugins/operator/OperatorValueBreachWatcher.test.ts
samigabor: Sorry about this one. I see now I missed it..

packages/broker/test/integration/plugins/operator/OperatorValueBreachWatcher.test.ts
samigabor: Reviewed it. Thanks

harbu (approved):

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
```diff
@@ -1,33 +1,122 @@
...
+                query: `
+                    {
+                        operators {
+                            id
+                        }
+                    }
+                    `
+            }
+        }
+        const parseItems = (response: any) => {
```
teogeb: Stricter type?

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
```diff
@@ -1,33 +1,122 @@
...
+            return response.operators
+        }
+        this.theGraphClient.updateRequiredBlockNumber(requiredBlockNumber)
+        const queryResult = this.theGraphClient.queryEntities<{ id: string }>(createQuery, parseItems)
+
+        const operatorAddresses: EthereumAddress[] = []
+        for await (const operator of queryResult) {
+            operatorAddresses.push(toEthereumAddress(operator.id))
+        }
+        return operatorAddresses
```
teogeb, line 114:
```suggestion
        return queryResult.map((operator) => toEthereumAddress(operator.id))
```

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
```diff
@@ -1,83 +1,58 @@
...
-            
-            logger.info('Update approximate pool values of sponsorships', { threshold, diffPercentage: diffSum / totalDiff })
-            // pick the first entries needed to get the total diff under the threshold
-            const neededSponsorshipAddresses = sortedSponsorships.slice(0, neededSponsorshipsCount).map((sponsorship) => sponsorship.address)
-            await this.helper.updateApproximatePoolValueOfSponsorships(neededSponsorshipAddresses)
+    private async checkMyUnwithdrawnEarnings(): Promise<void> {
+        logger.info('Check whether it is time to withdraw my earnings')
+        const { fraction, sponsorshipAddresses } = await this.helper.getMyUnwithdrawnEarnings()
+        const safeUnwithdrawnEarningsFraction = this.penaltyLimitFraction! * this.withdrawLimitSafetyFraction / BigInt(ONE_ETHER)
+        logger.trace(` -> is ${fraction / BigInt(1e16)}% > ${safeUnwithdrawnEarningsFraction / BigInt(1e16)}% ?`)
```
teogeb: Use `ONE_ETHER`

packages/broker/src/plugins/operator/OperatorPlugin.ts
```diff
@@ -11,12 +11,16 @@ import { Signer } from '@ethersproject/abstract-signer'
...
 import { VoteOnSuspectNodeService } from './VoteOnSuspectNodeService'
 import { MaintainOperatorValueService } from './MaintainOperatorValueService'
+import { OperatorValueBreachWatcher } from './OperatorValueBreachWatcher'
 import { OperatorFleetState } from './OperatorFleetState'
 import { toStreamID } from '@streamr/protocol'
 import { AnnounceNodeToContractService } from './AnnounceNodeToContractService'
 import { AnnounceNodeToContractHelper } from './AnnounceNodeToContractHelper'
 import { CONFIG_TEST } from 'streamr-client'
 
+export const DEFAULT_MAX_SPONSORSHIP_IN_WITHDRAW = 20 // max number of sponsorships to loop over before tx reverts
```
teogeb: What `tx` this refers to? Could update the comment so that it it is understandable without knowning the implementation details?

packages/broker/src/plugins/operator/OperatorValueBreachWatcher.ts
```diff
@@ -0,0 +1,63 @@
...
+const logger = new Logger(module)
+
+const CHECK_VALUE_INTERVAL = 1000 * 60 * 60 // 1 hour
+
+export class OperatorValueBreachWatcher {
+    private penaltyLimitFractionCached?: bigint
+    private readonly abortController: AbortController
+    
+    // public access modifier for tests 
+    readonly helper: MaintainOperatorValueHelper
```
teogeb, line 14: Is it possible to write the test wihtout exposing this? One possible solution would be to implement same pattern which we use `VoteOnSuspectNodeService.test.ts` which deploys an isolated environment?

packages/broker/src/plugins/operator/OperatorValueBreachWatcher.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    private async checkRandomUnwithdrawnEarnings(): Promise<void> {
+        const randomOperatorAddress = await this.helper.getRandomOperator()
+        if (randomOperatorAddress === undefined) {
+            logger.info('No operators found')
+            return
+        }
+        logger.info('Check unwithdrawn earnings', { randomOperatorAddress })
+        return this.checkUnwithdrawnEarningsOf(randomOperatorAddress)
```
teogeb:
```suggestion
        await this.checkUnwithdrawnEarningsOf(randomOperatorAddress)
```

packages/broker/src/plugins/operator/OperatorValueBreachWatcher.ts
```diff
@@ -0,0 +1,63 @@
...
+        logger.trace(` -> is ${fraction} > ${limit}?`)
+        if (fraction > limit) {
+            logger.info('Withdraw earnings from sponsorships (target operator value in breach)',
+                { targetOperatorAddress, sponsorshipAddresses, fraction, limit })
+            await this.helper.triggerWithdraw(targetOperatorAddress, sponsorshipAddresses)
+        }
+    }
+
+    private async getPenaltyLimitFraction(): Promise<bigint> {
+        if (!this.penaltyLimitFractionCached) {
```
teogeb:
```suggestion
        if (!this.penaltyLimitFractionCached !== undefined) {
```

packages/broker/test/integration/plugins/operator/MaintainOperatorValueHelper.test.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import { STREAMR_DOCKER_DEV_HOST } from '../../../utils'
+import { setupOperatorContract, SetupOperatorOpts } from './setupOperatorContract'
+import { getProvider } from './smartContractUtils'
+import { MaintainOperatorValueHelper } from '../../../../src/plugins/operator/MaintainOperatorValueHelper'
+
+const chainConfig = config.dev2
+const theGraphUrl = `http://${STREAMR_DOCKER_DEV_HOST}:8800/subgraphs/name/streamr-dev/network-subgraphs`
+
+const STREAM_CREATION_KEY = '0xb1abdb742d3924a45b0a54f780f0f21b9d9283b231a0a0b35ce5e455fa5375e7'
```
teogeb: Maybe better name if we don't create streams with this key?

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
```diff
@@ -1,125 +1,78 @@
...
-        await maintainOperatorValueService.stop()
+        // 1000 = check every second
+        const service = new MaintainOperatorValueService(operatorConfig, 0.5, 1000)
+
+        const poolValueBeforeWithdraw = await operatorContract.getApproximatePoolValue()
+
+        // 5 DATA = 50% of 10 DATA, 50% is the "safe threshold" (we don't want to wait all the way to the limit, lest we be late)
+        // 10 DATA = 10% of 100 DATA, 10% is the penalty limit (sum of unwithdrawn earnings may not exceed it)
+        // 100 DATA = pool value
+        await service.start()
```
teogeb: Are these comments about line 70-71? Or how they relate to `service.start()`

packages/broker/test/integration/plugins/operator/OperatorPlugin.test.ts
```diff
@@ -23,7 +23,7 @@ describe('OperatorPlugin', () => {
             privateKey: brokerWallet.privateKey,
             extraPlugins: {
                 operator: {
-                    operatorContractAddress: '0x4A5C0EC07F7ddBd4B6050181638e24b0153991b2'
+                    operatorContractAddress: '0x139dfa493a45364b598f2f98e504192819082c85'
```
teogeb, line 26: Why this is changed?

packages/broker/test/integration/plugins/operator/OperatorValueBreachWatcher.test.ts
```diff
@@ -0,0 +1,84 @@
...
+        await (await operatorContract.stake(sponsorship1.address, parseEther('100'))).wait()
+
+        const sponsorship2 = await deploySponsorship(chainConfig, operatorWallet, { streamId, earningsPerSecond: parseEther('2') })
+        await (await token.connect(operatorWallet).transferAndCall(sponsorship2.address, parseEther('250'), '0x')).wait()
+        await (await operatorContract.stake(sponsorship2.address, parseEther('100'))).wait()
+
+        const operatorValueBreachWatcher = new OperatorValueBreachWatcher(watcherConfig)
+
+        const poolValueBeforeWithdraw = await operatorContract.getApproximatePoolValue()
+        const allowedDifference = poolValueBeforeWithdraw.div('10').toBigInt()
```
teogeb: Why `10`? Can we read/calculate that value from somewhere?

packages/broker/test/integration/plugins/operator/deploySponsorshipContract.ts
```diff
@@ -1,39 +1,50 @@
...
+        earningsPerSecond = parseEther('0.01'),
     } = {},
 ): Promise<Sponsorship> {
+    const { contracts } = chainConfig
+    logger.debug('deploySponsorship', { contracts, deployerBalance: await deployer.getBalance() })
+
+    const sponsorshipFactory = new Contract(contracts.SponsorshipFactory, sponsorshipFactoryABI, deployer) as unknown as SponsorshipFactory
+    const streamRegistry = new Contract(contracts.StreamRegistry, streamRegistryABI, deployer) as unknown as StreamRegistry
+
+    const streamExists = await streamRegistry.exists(streamId)
```
teogeb: Why this check is added? The core functionality of this method is to deploy a sponsorship, and it is maybe ok if this test helper function relies that it is passed valid data. (As we don't need to provide extra dependencies just for checking the dependencies.)

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
samigabor: Updated

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
samigabor: Hmm, that would not work though.. queryResult is an AsyncGenerator..

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
samigabor: The 1e16 is 0.01 ether and that value is only used in those dev logs.

packages/broker/src/plugins/operator/OperatorPlugin.ts
samigabor: Updated

packages/broker/src/plugins/operator/OperatorValueBreachWatcher.ts
samigabor: It's not possible to deploy a new dev env for these tests (using streamrEnvDeployer) because we also need to query the existing subgraph. The alternative would be to build a new subgraph image on the fly for the newly generated addresses

packages/broker/src/plugins/operator/OperatorValueBreachWatcher.ts
samigabor: Updated

packages/broker/test/integration/plugins/operator/MaintainOperatorValueHelper.test.ts
samigabor: Updated

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: The idea was to explain through numbers what will happen when the service starts (what's the penalty limit and the safe threshold for the operator)

packages/broker/test/integration/plugins/operator/OperatorPlugin.test.ts
samigabor: It was an outdated address (not anymore part of the config) after the fast-chain started to be used. Not sure how ci skipped it and got into streamr-1.0 (no impl for the operator plugin service was added in this PR). The new address is the one that's now always deployed into the fast-chain container.

packages/broker/test/integration/plugins/operator/OperatorValueBreachWatcher.test.ts
samigabor: It's the 10% penalty limit. Removed the hard-coded value and retrieved from the config

packages/broker/test/integration/plugins/operator/deploySponsorshipContract.ts
samigabor: Not needed as we can indeed assume the factory will correctly deploy the stream. Removed it.

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
teogeb: I don't still understand the comment... Maybe clarify a bit?

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
```diff
@@ -1,33 +1,122 @@
...
+                    {
+                        operators {
+                            id
+                        }
+                    }
+                    `
+            }
+        }
+        const parseItems = (response: { operators: { id: string }[] }) => {
+            if (!response.operators) {
```
teogeb: If no operators is found, we receive an empty array from TheGraph:
```
{
  "data": {
    "operators": []
  }
}
```

We can just return it, and no need to have this extra check?

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
teogeb: Easy to confuse to the full ether. So this is essentially just converting a value between 0-1 to a percentage (0-100):
```
logger.trace(` -> is ${fraction / ONE_ETHER * 100}% > ${safeUnwithdrawnEarningsFraction / ONE_ETHER * 100}% ?`)
```

packages/broker/test/integration/plugins/operator/MaintainOperatorValueService.test.ts
samigabor: I see your point. Maybe those comments aren't describing much related to start. Removed them.

packages/broker/src/plugins/operator/MaintainOperatorValueService.ts
samigabor: Updated

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
samigabor: Updated

packages/broker/test/integration/plugins/operator/deploySponsorshipContract.ts
```diff
@@ -1,39 +1,45 @@
...
 import type { Sponsorship, SponsorshipFactory } from '@streamr/network-contracts'
 
 const { parseEther } = utils
 
+const logger = new Logger(module)
+
 export async function deploySponsorship(
     // eslint-disable-next-line max-len
-    chainConfig: { contracts: { SponsorshipFactory: string, SponsorshipStakeWeightedAllocationPolicy: string, SponsorshipDefaultLeavePolicy: string, SponsorshipVoteKickPolicy: string } },
+    chainConfig: { contracts: { SponsorshipFactory: string, SponsorshipStakeWeightedAllocationPolicy: string, SponsorshipDefaultLeavePolicy: string, SponsorshipVoteKickPolicy: string, StreamRegistry: string } },
```
teogeb: Why this is added? It is not used

packages/broker/test/integration/plugins/operator/deploySponsorshipContract.ts
```diff
@@ -1,39 +1,45 @@
...
+    const { contracts } = chainConfig
+    logger.debug('deploySponsorship', { contracts, deployerBalance: await deployer.getBalance() })
+
+    const sponsorshipFactory = new Contract(contracts.SponsorshipFactory, sponsorshipFactoryABI, deployer) as unknown as SponsorshipFactory
 
-    // console.log("Chain config: %o", chainConfig)
-    const sponsorshipFactory =
-        new Contract(chainConfig.contracts.SponsorshipFactory, sponsorshipFactoryABI, deployer) as unknown as SponsorshipFactory
-    // console.log("deployer balance", await deployer.getBalance())
+    logger.debug('Stream to be sponsored', { streamId })
```
teogeb:
```suggestion
```

packages/broker/test/integration/plugins/operator/deploySponsorshipContract.ts
```diff
@@ -1,39 +1,45 @@
...
         metadata = '{}',
         minimumStakeWei = parseEther('60'),
         minHorizonSeconds = 0,
         minOperatorCount = 1,
+        earningsPerSecond = parseEther('0.01'),
     } = {},
 ): Promise<Sponsorship> {
+    const { contracts } = chainConfig
+    logger.debug('deploySponsorship', { contracts, deployerBalance: await deployer.getBalance() })
+
```
teogeb: The deployer should always have enough balance in the test environment? (Or if not in some test, the test will fail)

```suggestion
```

packages/broker/test/integration/plugins/operator/operatorValueUtils.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const logger = new Logger(module)
+
+export async function getTotalUnwithdrawnEarnings(operatorContract: Operator): Promise<bigint> {
+    const { earnings } = await operatorContract.getEarningsFromSponsorships()
+    let unwithdrawnEarnings = BigInt(0)
+    for (const e of earnings) {
+        unwithdrawnEarnings += e.toBigInt()
+    }
+    logger.debug(`Total unwithdrawn earnings: ${formatEther(unwithdrawnEarnings.toString())} (t = ${Date.now()})`)
```
teogeb: Each log message will usually include a timestamp automatically

```suggestion
    logger.debug(`Total unwithdrawn earnings: ${formatEther(unwithdrawnEarnings.toString())}`)
```

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
```diff
@@ -1,33 +1,117 @@
...
+                        operators {
+                            id
+                        }
+                    }
+                    `
+            }
+        }
+        const parseItems = (response: { operators: { id: string }[] }) => response.operators
+        this.theGraphClient.updateRequiredBlockNumber(requiredBlockNumber)
+        const queryResult = this.theGraphClient.queryEntities<{ id: string }>(createQuery, parseItems)
```
teogeb: The default parseItems works ok as there is single root level item which is an array:
https://github.com/streamr-dev/network/blob/NET-537_operator_value_breach_watcher/packages/utils/src/TheGraphClient.ts#L61-L62

```suggestion
        const queryResult = this.theGraphClient.queryEntities<{ id: string }>(createQuery)
```

packages/broker/test/integration/plugins/operator/deploySponsorshipContract.ts
samigabor: Was used previously and got left there by mistake. Removed it.

packages/broker/test/integration/plugins/operator/deploySponsorshipContract.ts
samigabor: Removed the log. The contracts const is needed further down

packages/broker/src/plugins/operator/MaintainOperatorValueHelper.ts
samigabor: Cool, thanks for pointing that out

teogeb (approved):

PR #1643, opened 2023-07-06
feat: add load balacing for `MaintainTopologyService`

harbu:
## Summary

Add sharding support to `MaintainTopologyService`.

Before, all the nodes of an operator  would join all staked stream partitions. Now, the stream partition space is sharded / partitioned based on consistent hashing and nodes join only those stream partitions that are assigned to them.

## Changes
- Bring in 3rd party package `consistent-hash` for consistent hashing.
- Create wrapper class `ConstHash` to address the shortcomings of `consistent-hash` package (regarding our use case).
- Existing class `MaintainTopologyService`got simplified down a lot since now it receives `StreamPartID` instead of `StreamID` in the events it listens for, removing the need to find out how many stream partitions a specific stream has.
- Class `StreamAssignmentLoadBalancer` follows node joins and leaves as well as stream stakes and unstakes and by calculating the consistent hashes, provides `MaintainTopologyService` with knowledge of what streams to join and leave.

- Use different consistent hash library?


packages/broker/test/integration/plugins/operator/MaintainTopologyService.test.ts
```diff
@@ -48,31 +59,34 @@ describe('MaintainTopologyService', () => {
...
+            return containsAll(await getSubscribedStreamPartIds(client), [
+                ...stream1.getStreamParts(),
+                ...stream2.getStreamParts()
+            ])
+        }, 10000, 1000)
 
         await (await operatorContract.unstake(sponsorship1.address)).wait()
-        await waitForCondition(async () => (await client.getSubscriptions()).length === 3)
-        expect(await getSubscribedStreamPartIds(client)).toEqual(stream2.getStreamParts())
+        await waitForCondition(async () => {
```
harbu, line 102: Also check stream1.getStreamParts() _are not_ included!

packages/broker/src/plugins/operator/OperatorPlugin.ts
```diff
@@ -45,10 +48,20 @@ export class OperatorPlugin extends Plugin<OperatorPluginConfig> {
...
+        const operatorFleetState = new OperatorFleetState(
+            this.streamrClient,
+            toStreamID('/operator/coordination', serviceHelperConfig.operatorContractAddress)
+        )
         this.maintainTopologyService = new MaintainTopologyService(
             this.streamrClient,
-            new MaintainTopologyHelper(
-                serviceHelperConfig
+            new StreamAssignmentLoadBalancer(
+                'myNodeId', // TODO: real nodeId here
```
harbu: Should maybe use `setUpAndStartMaintainTopologyService` here but difficulty due to asynchronicity.

packages/broker/src/plugins/operator/ConstHash.ts
```diff
@@ -0,0 +1,63 @@
...
+import { StreamPartID, StreamPartIDUtils } from '@streamr/protocol'
+
+/**
+ * Slight variations at the very end of a string will not result in the keys
+ * being sufficiently assigned around. By having the partition number first,
+ * we properly randomize the assignments of stream parts of a stream.
+ */
+function formKey(streamPartId: StreamPartID): string {
+    const [streamId, partition] = StreamPartIDUtils.getStreamIDAndPartition(streamPartId)
+    return partition + `#` + streamId
```
teogeb:
```suggestion
    return `${partition}#${streamId}`
```

packages/broker/src/plugins/operator/ConstHash.ts
```diff
@@ -0,0 +1,63 @@
...
+        if (this.consistentHash === undefined) {
+            this.consistentHash = new ConsistentHash({
+                distribution: 'uniform'
+            })
+            this.nodes.sort()
+            for (const nodeId of this.nodes) {
+                this.consistentHash.add(nodeId)
+            }
+        }
+        if (this.replicationFactor > 1) {
```
teogeb: Would it make sense to cap the `replicationFactor` to `1` already in constructor so that we don't need this check?

packages/broker/src/plugins/operator/MaintainTopologyService.ts
```diff
@@ -1,68 +1,110 @@
...
+import { OperatorServiceConfig } from './OperatorPlugin'
+import { OperatorFleetState } from './OperatorFleetState'
 
 const logger = new Logger(module)
 
+/**
+ * Helper function for setting up and starting a MaintainTopologyService along
+ * with all its dependencies.
+ */
+export async function setUpAndStartMaintainTopologyService({
```
teogeb, line 16: Is this a helper method `MaintainTopologyService.test.ts` as this is not called from anywhere else? Could move there?

packages/broker/src/plugins/operator/MaintainTopologyService.ts
```diff
@@ -1,68 +1,110 @@
...
-        const subscriptions = this.subscriptions.get(streamId)
-        this.subscriptions.removeAll(streamId, subscriptions)
-        await Promise.all(subscriptions.map((sub) => sub.unsubscribe())) // TODO: rejects?
-    })
-
-    private async addStream(streamId: StreamID): Promise<void> {
-        let stream: Stream
+    private onAddStakedStreamPart = this.concurrencyLimiter(async (streamPartId: StreamPartID): Promise<void> => {
+        const id = StreamPartIDUtils.getStreamID(streamPartId)
+        const partition = StreamPartIDUtils.getStreamPartition(streamPartId)
```
teogeb:
```suggestion
        const [id, partition] = StreamPartIDUtils.getStreamIDAndPartition(streamPartId)
```

packages/broker/src/plugins/operator/MaintainTopologyService.ts
```diff
@@ -1,68 +1,110 @@
...
     async start(): Promise<void> {
-        this.maintainTopologyHelper.on('addStakedStream', this.onAddStakedStreams)
-        this.maintainTopologyHelper.on('removeStakedStream', this.onRemoveStakedStream)
-        await this.maintainTopologyHelper.start()
+        this.streamAssignmentLoadBalancer.on('assigned', this.onAddStakedStreamPart)
+        this.streamAssignmentLoadBalancer.on('unassigned', this.onRemoveStakedStreamPart)
         logger.info('Started')
     }
 
+    // eslint-disable-next-line class-methods-use-this
```
teogeb: Remove the method as it doesn't do anything?

packages/broker/src/plugins/operator/OperatorPlugin.ts
```diff
@@ -45,10 +49,21 @@ export class OperatorPlugin extends Plugin<OperatorPluginConfig> {
...
```
teogeb: Can we get the real nodeId synchronously, or do we need something like `(await streamrClient.getNode()).getNodeId()`?

Maybe it would make sense to move this the load balancer initialization to into `MaintainTopologyService#start `anyway?

packages/broker/src/plugins/operator/ConstHash.ts
harbu: Not sure I understand? We are not validating the value here but instead branching different logic depending on whether the factor is 1 or above?

packages/broker/src/plugins/operator/MaintainTopologyService.ts
harbu: Removed in https://github.com/streamr-dev/network/pull/1643/commits/af15d795c74b3596be525459ca469f501ae2c137

packages/broker/src/plugins/operator/MaintainTopologyService.ts
harbu: The idea would be that `setUpAndStartMaintainTopologyService` would also be used in source code. Specifically from [packages/broker/src/plugins/operator/OperatorPlugin.ts](https://github.com/streamr-dev/network/pull/1643/files/08246eadfa7e1ec3aad8137765e57ce979f252c8#diff-c73edeaf0a46f53e011c4cd619ea4e2e3fc7c13c5c7ae571b0899c6153f3be4a) here.

packages/broker/src/plugins/operator/config.schema.json
```diff
@@ -12,6 +12,11 @@
       "type": "string",
       "description": "Operator contract Ethereum address",
       "format": "ethereum-address"
+    },
+    "replicationFactor": {
+      "type": "integer",
+      "description": "How many nodes a stream partition should be assigned to",
+      "default": 1
```
teogeb:
```suggestion
      "default": 1,
      "minimum": 1
```

harbu:
TODO: extract helper for `toStreamID('/operator/coordination', serviceHelperConfig.operatorContractAddress)`

teogeb (approved):

PR #1728, opened 2023-08-04
feat(network): NET-1001 Simplified Inspect

juslesan:
## Summary

New feature for nodes to `Inspect` other nodes in stream partitions. The inspecting nodes will monitor that the inspected node is propagating data by opening a connection to the inspected node. The inspection will succeed once any message is received from the rest of the stream and the inspected node. Inspection will also succeed if the inspection timeouts and there were no messages received during the inspection.

## Future improvements

- Client integration
- Run FindPeer and Inspect in the same operation?
- Limit the size of the `inspectionMessages` map in InspectSession.ts
- Optimize calling of `Inspect.markMessage()`, currently it is called whenever the StreamNodeServer receives a new stream message.


packages/trackerless-network/src/logic/StreamrNode.ts
```diff
@@ -341,6 +341,14 @@ export class StreamrNode extends EventEmitter<Events> {
         })
     }
 
+    async inspect(peerDescriptor: PeerDescriptor, streamPartId: string): Promise<boolean> {
+        if (this.streams.get(streamPartId)?.type === StreamNodeType.RANDOM_GRAPH) {
+            const streamNode = this.streams.get(streamPartId)!.layer2 as RandomGraphNode
```
harbu, line 350: nitpick: casting could be removed with stricter typing

packages/trackerless-network/src/logic/RandomGraphNode.ts
```diff
@@ -264,6 +291,9 @@ export class RandomGraphNode extends EventEmitter<Events> implements IStreamNode
                 propagationTargets = propagationTargets.concat(this.config.proxyConnectionServer!.getSubscribers())
             }
         }
+
+        propagationTargets = propagationTargets.filter((target) => !this.config.inspector.isInspected(target as PeerIDKey))
```
harbu, line 277: nitpick: casting could be removed with more prominent use of branded string

packages/trackerless-network/src/logic/RandomGraphNode.ts
```diff
@@ -243,17 +264,23 @@ export class RandomGraphNode extends EventEmitter<Events> implements IStreamNode
...
         this.config.neighborUpdateManager.stop()
         this.config.proxyConnectionServer?.stop()
+        this.config.inspector.stop()
     }
 
     broadcast(msg: StreamMessage, previousPeer?: string): void {
         if (!previousPeer) {
             this.markAndCheckDuplicate(msg.messageRef!, msg.previousMessageRef)
+        } else {
+            this.config.inspector.markMessage(previousPeer as PeerIDKey, msg.messageRef!)
```
harbu: nitpick: can casting be removed here?

packages/trackerless-network/src/logic/inspect/InspectSession.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    constructor(config: InspectSessionConfig) {
+        super()
+        this.inspectedPeer = config.inspectedPeer
+    }
+
+    markMessage(senderId: PeerIDKey, messageRef: MessageRef): void {
+        const messageKey = createMessageKey(messageRef)
+        if (!this.inspectionMessages.has(messageKey)) {
+            this.inspectionMessages.set(messageKey, senderId === this.inspectedPeer)
```
harbu, line 30: Should this be also done if `senderId === this.inspectedPeer` is true? Now it seems as if a second call to markMessage is required?

packages/trackerless-network/src/logic/inspect/Inspector.ts
```diff
@@ -0,0 +1,89 @@
...
+        } catch (err) {
+            logger.warn('Inspect session timed out, removing')
+            this.sessions.delete(nodeId)
+            this.connectionLocker.unlockConnection(peerDescriptor, lockId)
+            return session.getInspectedMessageCount() < 1
+        }
+    }
+
+    markMessage(sender: PeerIDKey, messageId: MessageRef): void {
+        this.sessions.forEach((session) => session.markMessage(sender, messageId))
```
harbu, line 75: A bit unclear to me: why wouldn't only we pick from the `sessions` map the session that corresponds to `sender` and mark that one?

packages/trackerless-network/src/logic/inspect/RemoteInspectionServiceServer.ts
```diff
@@ -0,0 +1,27 @@
...
+import { Logger } from "@streamr/utils"
+
+const logger = new Logger(module)
+
+export class RemoteInspectionServiceServer extends Remote<IInspectionServiceClient> {
+
+    async inspectConnection(ownPeerDescriptor: PeerDescriptor): Promise<boolean> {
+        const options: DhtRpcOptions = {
+            sourceDescriptor: ownPeerDescriptor as PeerDescriptor,
+            targetDescriptor: this.remotePeerDescriptor as PeerDescriptor,
```
harbu: casts not needed?
```suggestion
            sourceDescriptor: ownPeerDescriptor,
            targetDescriptor: this.remotePeerDescriptor,
```

packages/trackerless-network/test/integration/Inspect.test.ts
```diff
@@ -0,0 +1,122 @@
...
+        inspectorNode = new NetworkStack({
+            layer0: {
+                entryPoints: [publisherDescriptor],
+                peerDescriptor: inspectorPeerDescriptor,
+                transportLayer: inspectorTransport
+            },
+            networkNode: {}
+        })
+        await inspectorNode.start()
+        await inspectorNode.getLayer0DhtNode().joinDht(publisherDescriptor)
```
harbu: Could be extracted to a helper function to avoid repetition.

packages/trackerless-network/src/logic/inspect/Inspector.ts
juslesan: The sender can be either the inspected node or a regular neighbor. This part should definitely be optimised but I left any type of mapping or filtering of sessions for now. The sessions could be filtered based on ie. stream partitions.

packages/trackerless-network/src/logic/inspect/InspectSession.ts
juslesan: Yes, the sessions are finished once any message is marked twice. Once by the inspected peer and once by any other peer in the stream.

packages/trackerless-network/src/logic/RandomGraphNode.ts
juslesan: This entire block turns out to be unnecessary so removing it.

packages/trackerless-network/src/logic/RandomGraphNode.ts
juslesan: Yeah the use of ie. `PeerIDKey` should definitely be increased and I have plans to make its use more prominent

packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -24,6 +24,10 @@ service NeighborUpdateRpc {
   rpc neighborUpdate (NeighborUpdate) returns (NeighborUpdate);
 }
 
+service InspectionService {
```
teogeb: To unify the naming of services:

```suggestion
service InspectionRpc {
```

Or if we want to start to use `Service` suffix for all service, then add a `TODO` comment to other services?

packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -24,6 +24,10 @@ service NeighborUpdateRpc {
...
+  rpc inspectConnection (InspectConnectionRequest) returns (InspectConnectionResponse);
```
teogeb: What this call does in practice: does it inspect a Node (or start inspecting it), or open a connection? (Maybe could rename it as a connection is not the target of inspection)

If it opens a connection, would it make sense to rename the method to something which represents the type of the node (i.e. `openLockedConnection`, `openDirectConnection` or something). In that way we could see a concrete attribute of that node (i.e. how it differs from other connections). 

(If we don't specify that the connection is just for inspection, it is also more difficult for an inspected node to see that it is under inspection. Currently maybe too trivial to find it out?)

packages/trackerless-network/src/NetworkNode.ts
```diff
@@ -27,6 +27,10 @@ export class NetworkNode {
         this.stack.getStreamrNode().setExtraMetadata(metadata)
     }
 
+    async inspect(peerDescriptor: PeerDescriptor, streamPartId: StreamPartID): Promise<boolean> {
+        return this.stack.getStreamrNode().inspect(peerDescriptor, streamPartId)
```
teogeb:
```suggestion
    async inspect(node: PeerDescriptor, streamPartId: StreamPartID): Promise<boolean> {
        return this.stack.getStreamrNode().inspect(node, streamPartId)
```

packages/trackerless-network/src/logic/RandomGraphNode.ts
```diff
@@ -58,6 +62,8 @@ export interface StrictRandomGraphNodeConfig {
     nodeName: string
     acceptProxyConnections: boolean
     neighborUpdateInterval: number
+    inspector: IInspector
+    inspectingConnections: PeerList
```
teogeb: Would it make sense if we could query the connections from `inspector` and remove this field?

packages/trackerless-network/src/logic/StreamNodeServer.ts
```diff
@@ -22,8 +27,9 @@ export class StreamNodeServer implements INetworkRpc {
     }
 
     async sendData(message: StreamMessage, context: ServerCallContext): Promise<Empty> {
+        const previousPeer = keyFromPeerDescriptor((context as DhtCallContext).incomingSourceDescriptor!)
+        this.config.markForInspection(previousPeer, message.messageRef!)
```
teogeb, line 31: Are we able to deliver the info about the message to Inspector via standard `this.config.broadcast` callback?

packages/trackerless-network/src/logic/createRandomGraphNode.ts
```diff
@@ -38,7 +41,7 @@ const createConfigWithDefaults = (config: RandomGraphNodeConfig): StrictRandomGr
         minPropagationTargets,
         randomGraphId: config.randomGraphId,
         sendToNeighbor: async (neighborId: string, msg: StreamMessage): Promise<void> => {
-            const remote = targetNeighbors.getNeighborWithId(neighborId)
+            const remote = targetNeighbors.getNeighborWithId(neighborId) || inspectingConnections.getNeighborWithId(neighborId)
```
teogeb:
```suggestion
            const remote = targetNeighbors.getNeighborWithId(neighborId) ?? inspectingConnections.getNeighborWithId(neighborId)
```

packages/trackerless-network/src/logic/createRandomGraphNode.ts
```diff
@@ -29,6 +30,8 @@ const createConfigWithDefaults = (config: RandomGraphNodeConfig): StrictRandomGr
     const nearbyContactPool = config.nearbyContactPool ?? new PeerList(peerId, numOfTargetNeighbors + 1)
     const randomContactPool = config.randomContactPool ?? new PeerList(peerId, maxNumberOfContacts)
     const targetNeighbors = config.targetNeighbors ?? new PeerList(peerId, maxNumberOfContacts)
+    const inspectingConnections = config.inspectingConnections ?? new PeerList(peerId, 10)
```
teogeb: E.g.
```suggestion
    const inspectingNodes = config.inspectingNodes ?? new PeerList(peerId, 10)
```

As peer list is a list of nodes, not connections?

packages/trackerless-network/src/logic/inspect/InspectSession.ts
```diff
@@ -0,0 +1,49 @@
...
+    inspectedPeer: PeerIDKey
+}
+
+const createMessageKey = (messageRef: MessageRef): string => {
+    return `${messageRef.messageChainId}:${messageRef.timestamp}:${messageRef.sequenceNumber}`
+}
+export class InspectSession extends EventEmitter<Events> {
+    
+    // Boolean indicates if the message has been received by the inspected node
+    private readonly inspectionMessages: Map<string, boolean> = new Map()
```
teogeb, line 19: Could use set of strings? If there is a string, it means the message has been received by the inspected node

packages/trackerless-network/src/logic/inspect/Inspector.ts
```diff
@@ -0,0 +1,89 @@
...
+    private readonly inspectionTimeout: number
+    private readonly openInspectConnection: (peerDescriptor: PeerDescriptor, lockId: string) => Promise<void>
+
+    constructor(config: InspectorConfig) {
+        this.graphId = config.graphId
+        this.ownPeerDescriptor = config.ownPeerDescriptor
+        this.client = toProtoRpcClient(new InspectionServiceClient(config.rpcCommunicator.getRpcClientTransport()))
+        this.connectionLocker = config.connectionLocker
+        this.inspectionTimeout = config.inspectionTimeout || DEFAULT_TIMEOUT
+        this.openInspectConnection = config.openInspectConnection || this.defaultOpenInspectConnection
```
teogeb:
```suggestion
        this.inspectionTimeout = config.inspectionTimeout ?? DEFAULT_TIMEOUT
        this.openInspectConnection = config.openInspectConnection ?? this.defaultOpenInspectConnection
```

packages/trackerless-network/src/logic/inspect/Inspector.ts
```diff
@@ -0,0 +1,89 @@
...
+            inspectedPeer: nodeId
+        })
+        const lockId = `inspector-${this.graphId}`
+        this.sessions.set(nodeId, session)
+        await this.openInspectConnection(peerDescriptor, lockId)
+        try {
+            await waitForEvent3<InspectSessionEvents>(session, 'done', this.inspectionTimeout)
+            this.connectionLocker.unlockConnection(peerDescriptor, lockId)
+            this.sessions.delete(nodeId)
+            return true
```
teogeb: Move to `finally` as done also in the `catch` block?

packages/trackerless-network/src/logic/StreamNodeServer.ts
juslesan: Not really since `broadcast` is only called on non-duplicate messages

packages/trackerless-network/src/logic/RandomGraphNode.ts
juslesan: moved to InspectionRpcServer

packages/trackerless-network/protos/NetworkRpc.proto
juslesan: It is used to open a connection to the inspected peer. When the server receives the connection it will start to propagate messages to the inspecting peer. In the simplified inspection it was decided that the inspection connections will have their own operational pipelines to not interfere the with the graph formation algorithm for a short lived functionality.

packages/trackerless-network/src/logic/inspect/InspectSession.ts
juslesan: We need to keep track of if a message have been received from the inspected peer and any of the stream neighbors. The session is done after any message has been seen from any of the inspectors stream neighbors and the inspected peer.

packages/trackerless-network/protos/NetworkRpc.proto
teogeb: Ok then the method could be something like `open*Connection`. E.g. `openDirectConnection` or something that describes the attributes and behavior of the connection. (The inspection word tells for what the connection is used at the moment, but it would be more descriptive to know how that connection differs from other connections). 

Also the inspection is a higher level concept, so it shouldn't be on protocol level if possible.

hpihkala: I see this as just a minor naming issue, but I'm pitching in as I was pinged for ideas. How about calling it a `temporary` or `adhoc` connection? Maybe one of those captures the correct semantic? 

Teo's point is valid in that the connection shouldn't care what it's being used for, as long as the other end knows how to handle the connection.

Temporary connectivity could have other uses in the future, beyond inspection?

harbu (approved):

packages/trackerless-network/src/logic/inspect/InspectSession.ts
```diff
@@ -0,0 +1,49 @@
...
+export interface Events {
+    done: () => void
+}
+
+interface InspectSessionConfig {
+    inspectedPeer: PeerIDKey
+}
+
+const createMessageKey = (messageRef: MessageRef): string => {
+    return `${messageRef.messageChainId}:${messageRef.timestamp}:${messageRef.sequenceNumber}`
```
teogeb: Need to include also publisherId as the messageChainId is unique only within one publisher. 

Is one session handling only messages of one stream, or should we also include streamPartId?

juslesan: Added `publisherId` to the key.

The `RandomGraphNode` and all of its components are created per stream partition so there is no need to include stream partition in the sessions

harbu:
Feel free to land @juslesan

PR #1734, opened 2023-08-09
feat(broker): [NET-981] Operator plugin accepts proxy connections

teogeb:
Set `acceptProxyConnections` client config option to `true` by default when running `operator` plugin. If the client configuration contains an explicit `false` value, the Broker initialization throws an error.

## Implementation

There is a new feature in Broker which allows plugins to define config defaults for the client. A plugin can implement `getClientConfig()` method to define a list of config options which are requited for that plugin. The method checks whether there is a user-defined config value for that option and sets a value if there is no user-defined value. Broker throws an error If there is a different user-defined value (e.g. `false` when a plugin requires `true`).

## Refactoring

As the config for `StreamrClient` needs info from plugins, it is no longer possible to pass `StreamrClient` instance to `Plugin` constructor. Now the `StreamrClient` is a parameter of `Plugin#start` method instead.

## Future improvements

The list returned by `getClientConfig()` is not type-safe. The config option is a string path and value is `any`. It could be possible to add some type-safe there.


packages/broker/src/broker.ts
```diff
@@ -19,19 +24,28 @@ export interface Broker {
     stop: () => Promise<unknown>
 }
 
+const applyPluginClientConfigs = (plugins: Plugin<any>[], clientConfig: StrictConfig['client']) => {
```
harbu:
```suggestion
const applyPluginClientConfigs = (plugins: Plugin<unknown>[], clientConfig: StrictConfig['client']) => {
```

packages/broker/src/broker.ts
```diff
@@ -19,19 +24,28 @@ export interface Broker {
...
+    plugins.forEach((plugin) => {
+        plugin.getClientConfig().forEach((item) => {
```
harbu: I would prefer for-loops.
```suggestion
    plugins.forEach((plugin) => {
        plugin.getClientConfig().forEach((item) => {
```

packages/broker/src/broker.ts
harbu: Could use unit tests this function perhaps (since there is a bit of `any`  and `isEqual`) (Would be maybe good to test `isEqual` special cases: array / object).

packages/broker/test/integration/plugins/operator/OperatorPlugin.test.ts
```diff
@@ -1,29 +1,84 @@
...
-        broker?.stop()
+        await broker?.stop()
     })
 
-    it('can start broker with operator plugin', async () => {
+    it('accepts proxy connections', async () => {
+        const subscriber = createClient(await fetchPrivateKeyWithGas())
+        const stream = await createTestStream(subscriber, module)
+
+        const sponsorer = await generateWalletWithGasAndTokens()
```
harbu, line 34: I would do all sponsoring related stuff first and then go to delegating and staking

packages/broker/test/integration/plugins/operator/OperatorPlugin.test.ts
```diff
@@ -1,29 +1,84 @@
...
                 }
             }
         })
+        const brokerDescriptor = await broker.getStreamrClient().getPeerDescriptor()
+        await subscriber.setProxies({ id: stream.id }, [brokerDescriptor], ProxyDirection.SUBSCRIBE)
+        const receivedMessages: any[] = []
+        await subscriber.subscribe(stream.id, (content: any) => {
+            receivedMessages.push(content)
+        })
+        await waitForCondition(() => receivedMessages.length > 0)
```
harbu: Could use `collect` ?

packages/broker/src/broker.ts
teogeb: `unknown` type can't be used as the generics in Plugin is defined as `T extends `object

teogeb: Added unit tests

harbu (approved):

PR #1736, opened 2023-08-10
ETH-554 inspect random node helper

samt1803:
Added InspectRandomNodeHelper and test
not breaking change


packages/broker/src/plugins/operator/InspectRandomNodeHelper.ts
```diff
@@ -0,0 +1,112 @@
...
+        }
+        const parseItems = (response: any) => {
+            if (!response.operator) {
+                logger.error('Unable to find operator in The Graph', { operatorContractAddress: this.operatorContractAddress })
+                return []
+            }
+            return response.operator.stakes
+        }
+        this.theGraphClient.updateRequiredBlockNumber(requiredBlockNumber)
+        const queryResult = this.theGraphClient.queryEntities<any>(createQuery, parseItems) // TODO: add type
```
samt1803: forgot these types, will add them

packages/broker/src/plugins/operator/InspectRandomNodeHelper.ts
```diff
@@ -0,0 +1,112 @@
...
+        }
+        const parseItems = (response: any) => {
+            if (!response.sponsorship) {
+                logger.error('Unable to find sponsorship in The Graph', { sponsorshipAddress })
+                return []
+            }
+            return response.sponsorship.stakes
+        }
+        this.theGraphClient.updateRequiredBlockNumber(requiredBlockNumber)
+        const queryResult = this.theGraphClient.queryEntities<any>(createQuery, parseItems) // TODO: add type
```
samt1803: will add types here

packages/broker/src/plugins/operator/InspectRandomNodeHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+                            }
+                        }
+                    }
+                    `
+            }
+        }
+        const parseItems = (response: { operator: { stakes: { id: string, sponsorship: { id: string } }[] } }): 
+        { id: string, sponsorship: { id: string } }[] => {
+            if (!response.operator) {
+                logger.error('Unable to find operator in The Graph', { operatorContractAddress: this.operatorContractAddress })
```
harbu:
```suggestion
                logger.warn('Unable to find operator in The Graph', { operatorContractAddress: this.operatorContractAddress })
```

packages/broker/src/plugins/operator/InspectRandomNodeHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+                return []
+            }
+            return response.operator.stakes
+        }
+        this.theGraphClient.updateRequiredBlockNumber(requiredBlockNumber)
+        const queryResult = this.theGraphClient.queryEntities<{ id: string, sponsorship: { id: string } }>(createQuery, parseItems)
+        const sponsorshipIds = new Set<EthereumAddress>()
+        for await (const stake of queryResult) {
+            const sponsorshipId = stake.sponsorship?.id
+            if (sponsorshipId) {
```
harbu: We have a convention of preferring explicit undefined checks (so that the semantics are more clear)
```suggestion
            if (sponsorshipId !== undefined) {
```

packages/broker/src/plugins/operator/InspectRandomNodeHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+                            }
+                        }
+                    }
+                    `
+            }
+        }
+        const parseItems = (response: { sponsorship: { stakes: { id: string, operator: { id: string } }[] } } ):
+        { id: string, operator: { id: string } }[] => {
+            if (!response.sponsorship) {
+                logger.error('Unable to find sponsorship in The Graph', { sponsorshipAddress })
```
harbu:
```suggestion
                logger.warn('Unable to find sponsorship in The Graph', { sponsorshipAddress })
```

packages/broker/src/plugins/operator/InspectRandomNodeHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+                return []
+            }
+            return response.sponsorship.stakes
+        }
+        this.theGraphClient.updateRequiredBlockNumber(requiredBlockNumber)
+        const queryResult = this.theGraphClient.queryEntities<{ id: string, operator: { id: string } }>(createQuery, parseItems)
+        const operatorIds = new Set<EthereumAddress>()
+        for await (const stake of queryResult) {
+            const operatorId = stake.operator?.id
+            if (operatorId) {
```
harbu:
```suggestion
            if (operatorId !== undefined) {
```

harbu (approved):

packages/broker/src/plugins/operator/InspectRandomNodeHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+                            number
+                            }
+                        }
+                    }
+                    `
+            }
+        }
+        const parseItems = (response: { operator: { stakes: { id: string, sponsorship: { id: string } }[] } }): 
+        { id: string, sponsorship: { id: string } }[] => {
+            if (!response.operator) {
```
teogeb, line 45: Explicit null check, and update the type definition at line to allow null?

packages/broker/src/plugins/operator/InspectRandomNodeHelper.ts
teogeb: Is this check needed? If a steak always belongs to some sponsorship?

packages/broker/src/plugins/operator/InspectRandomNodeHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+                    {
+                        operator(id: "${operatorAddress}") {
+                            stakes(where: {id_gt: "${lastId}"}, first: ${pageSize}) {
+                                id
+                                sponsorship {
+                                    id
+                                }
+                            }
+                        }
+                        _meta {
```
teogeb: Is the _meta block needed? Also at line 83

packages/broker/src/plugins/operator/InspectRandomNodeHelper.ts
teogeb: Is there always an operator for a stake, i.e. null check not needed? Or should we update the type definition to allow null?

packages/broker/test/integration/plugins/operator/InspectRandomNodeHelper.test.ts
```diff
@@ -0,0 +1,167 @@
...
+        const timeString = (new Date()).getTime().toString()
+        const streamPath1 = "/operatorclienttest-1-" + timeString
+        const streamPath2 = "/operatorclienttest-2-" + timeString
+        streamId1 = adminWallet.address.toLowerCase() + streamPath1
+        streamId2 = adminWallet.address.toLowerCase() + streamPath2
+        const streamRegistry = new Contract(config.contracts.StreamRegistry, streamRegistryABI, adminWallet) as unknown as StreamRegistry
+        logger.debug(`creating stream with streamId1 ${streamId1}`)
+        await (await streamRegistry.createStream(streamPath1, "metadata")).wait()
+        logger.debug(`creating stream with streamId2 ${streamId2}`)
+        await (await streamRegistry.createStream(streamPath2, "metadata")).wait()
```
teogeb: Could use StreamrClient to create the streams?
```
const client = createClient(await fetchPrivateKeyWithGas())
const stream1 = await createTestStream(client, module)
const stream2 = await createTestStream(client, module)
await client.destroy()
```

packages/broker/test/integration/plugins/operator/InspectRandomNodeHelper.test.ts
```diff
@@ -0,0 +1,167 @@
...
+        })
+
+        it("works to flag through the inspectRandomNodeHelper", async () => {
+            const flagger = await createWalletAndDeployOperator(provider, config, theGraphUrl)
+            logger.trace('deployed flagger contract ' + flagger.operatorConfig.operatorContractAddress)
+            const target = await createWalletAndDeployOperator(provider, config, theGraphUrl)
+            logger.trace('deployed target contract ' + target.operatorConfig.operatorContractAddress)
+
+            // await wait(5000) // wait for events to be processed // wait for events to be processed
+
```
teogeb:
```suggestion
```

packages/broker/test/integration/plugins/operator/InspectRandomNodeHelper.test.ts
```diff
@@ -0,0 +1,167 @@
...
+                streamId: streamId1 })
+            const sponsorship2 = await deploySponsorship(config, operatorWallet, {
+                streamId: streamId2
+            })
+
+            logger.debug(`Sponsorship deployed at ${sponsorship.address}, delegating...`)
+            await (await token.connect(operatorWallet).transferAndCall(operatorContract.address, parseEther("200"), operatorWallet.address)).wait()
+
+            logger.debug("Staking to sponsorship...")
+            await (await operatorContract.stake(sponsorship.address, parseEther("100"))).wait()
```
teogeb: Could use constant here instead of magic number? E.g. STAKE1 = 123, STAKE2 = 234 and at line then STAKE1+STAKE2

packages/broker/test/integration/plugins/operator/InspectRandomNodeHelper.test.ts
```diff
@@ -0,0 +1,167 @@
...
+            const target = await createWalletAndDeployOperator(provider, config, theGraphUrl)
+            logger.trace('deployed target contract ' + target.operatorConfig.operatorContractAddress)
+
+            // await wait(5000) // wait for events to be processed // wait for events to be processed
+
+            logger.trace('deploying sponsorship contract')
+            const sponsorship = await deploySponsorship(config, adminWallet, {
+                streamId: streamId1 })
+            logger.trace('sponsoring sponsorship contract')
+            await (await token.connect(flagger.operatorWallet).approve(sponsorship.address, parseEther('500'))).wait()
```
teogeb: Constant for this? Or is it calculatable from other values, e.g. line 114?

packages/broker/test/integration/plugins/operator/InspectRandomNodeHelper.test.ts
```diff
@@ -0,0 +1,167 @@
...
+            await (await sponsorship.connect(flagger.operatorWallet).sponsor(parseEther('500'))).wait()
+
+            logger.trace('each operator delegates to its operactor contract')
+            logger.trace('delegating from flagger: ' + flagger.operatorWallet.address)
+            await (await token.connect(flagger.operatorWallet).transferAndCall(flagger.operatorContract.address,
+                parseEther('200'), flagger.operatorWallet.address)).wait()
```
teogeb: Constant here and at line 118 as the values need to be equal?

packages/broker/test/integration/plugins/operator/InspectRandomNodeHelper.test.ts
```diff
@@ -0,0 +1,167 @@
...
+                `
+                })
+                return result.operator.flagsOpened.length === 1
+            }, 10000, 1000)
+
+            waitForCondition(async (): Promise<boolean> => {
+                const result = await graphClient.queryEntity<{ operator: { flagsTargeted: any[] } }>({ query: `
+                {
+                    operator(id: "${target.operatorContract.address.toLowerCase()}") {
+                        id
```
teogeb, line 155:
```suggestion
```

packages/broker/test/integration/plugins/operator/InspectRandomNodeHelper.test.ts
```diff
@@ -0,0 +1,167 @@
...
+
+            logger.trace('flagging target operator')
+            const inspectRandomNodeHelper = new InspectRandomNodeHelper(flagger.operatorConfig)
+            await inspectRandomNodeHelper.flag(toEthereumAddress(sponsorship.address), toEthereumAddress(target.operatorContract.address))
+
+            waitForCondition(async (): Promise<boolean> => {
+                const result = await graphClient.queryEntity<{ operator: { flagsOpened: any[] } }>({ query: `
+                {
+                    operator(id: "${flagger.operatorContract.address.toLowerCase()}") {
+                        id
```
teogeb, line 140:
```suggestion
```

packages/broker/test/integration/plugins/operator/InspectRandomNodeHelper.test.ts
```diff
@@ -0,0 +1,167 @@
...
+            await (await token.connect(flagger.operatorWallet).transferAndCall(flagger.operatorContract.address,
+                parseEther('200'), flagger.operatorWallet.address)).wait()
+            logger.trace('delegating from target: ' + target.operatorWallet.address)
+            await (await token.connect(target.operatorWallet).transferAndCall(target.operatorContract.address,
+                parseEther('200'), target.operatorWallet.address)).wait()
+        
+            await wait(3000) // sometimes these stake fail, possibly when they end up in the same block
+            logger.trace('staking to sponsorship contract from flagger and target and voter')
+            logger.trace('staking from flagger: ' + flagger.operatorContract.address)
+            await (await flagger.operatorContract.stake(sponsorship.address, parseEther('150'))).wait()
```
teogeb: Constant here and at line 126

packages/broker/src/plugins/operator/InspectRandomNodeHelper.ts
teogeb: Maybe it doesn't need to log anything? An empty list is a valid (and sometimes the expected) answer to the getSponsorshipsOfOperator() call. Also at line 95

harbu: If that is the case, yeah then no logging required. So we should remove logging.

samt1803: I'll remove it, since the method is somewhat general, but the usecase "get sponsorships I am staked in" should maybe throw or log something, because if my own operatorcontract is not found something definitely is not right

packages/broker/test/integration/plugins/operator/InspectRandomNodeHelper.test.ts
samt1803: the test does not test any logic, like "+"

packages/broker/test/integration/plugins/operator/InspectRandomNodeHelper.test.ts
samt1803: they don't need to be equal

packages/broker/test/integration/plugins/operator/InspectRandomNodeHelper.test.ts
samt1803: again they don't need to be equal, and making a const for every number, that is then used once in the next line just makes it less readable

packages/broker/test/integration/plugins/operator/InspectRandomNodeHelper.test.ts
samt1803: This works and after fiddling with the many config types I couldn't get the client to connect to the right chain, so I am leaving it as is.

packages/broker/src/plugins/operator/InspectRandomNodeHelper.ts
```diff
@@ -0,0 +1,98 @@
...
+import type { Operator } from "@streamr/network-contracts"
+import { EthereumAddress, Logger, TheGraphClient, toEthereumAddress } from "@streamr/utils"
+import { OperatorServiceConfig } from "./OperatorPlugin"
+import fetch from 'node-fetch'
+
+const logger = new Logger(module)
+
+export class InspectRandomNodeHelper {
+
+    private readonly operatorContractAddress: EthereumAddress
```
teogeb:
```suggestion
```

packages/broker/test/integration/plugins/operator/InspectRandomNodeHelper.test.ts
```diff
@@ -0,0 +1,167 @@
...
+        await (await streamRegistry.createStream(streamPath2, "metadata")).wait()
+
+        graphClient = new TheGraphClient({
+            serverUrl: theGraphUrl,
+            fetch,
+            logger: logger
+        })
+    })
+
+    describe("InspectRandomNodeHelper methods", () => {
```
teogeb: This wrapper is not needed?

packages/broker/test/integration/plugins/operator/InspectRandomNodeHelper.test.ts
```diff
@@ -0,0 +1,167 @@
...
+            logger.debug(`staked on sponsorship ${sponsorship.address}`)
+            await (await operatorContract.stake(sponsorship2.address, parseEther("100"))).wait()
+            logger.debug(`staked on sponsorship ${sponsorship2.address}`)
+
+            await waitForCondition(async (): Promise<boolean> => {
+                const res = await inspectRandomNodeHelper.getSponsorshipsOfOperator(toEthereumAddress(operatorContract.address), 0)
+                return res.length === 2
+            }, 10000, 1000)
+
+            await waitForCondition(async (): Promise<boolean> => {
```
teogeb: Could assert something after we've waited for conditions. Ideally just one `waitForCondition` and then all asserting something like:
```
expect(await inspectRandomNodeHelper.getSponsorshipsOfOperator()).toBe(...)
expect(await inspectRandomNodeHelper.getOperatorsInSponsorship()).toBe(...)
```

teogeb (approved):

PR #1742, opened 2023-08-14
refactor(client): Use mono subgraph

teogeb:
Use the new `network-subgraphs` instead of `network-contracts`. 

The structure of the subgraphs mostly equal for all queries that client uses. Changes relevant to client:
- `permissions` root level query renamed to `streamPermissions`

**TODO update also production value (`config.schema.json`) when the mono graph is in production. Do not merge this to main before that as we can't make client releases from main until the mono graph has been deployed to production...**

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -72,7 +72,7 @@ jobs:
     uses: ./.github/workflows/test-setup.yml
     with:
       package: client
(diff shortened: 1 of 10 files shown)


harbu:
We are apparently not going to use mono subgraph in brubeck but only in 1.0.

PR #1752, opened 2023-08-15
feat(client): NET-1001, client integration

juslesan:
## Summary

Added integration for inspection to the client


packages/client/src/StreamrClient.ts
```diff
@@ -564,6 +564,11 @@ export class StreamrClient {
         return this.node.getNode()
     }
 
+    async inspect(node: NetworkPeerDescriptor, streamDefinition: StreamDefinition): Promise<boolean> {
```
harbu, line 567: This will return `true` if `node` is in the stream and propagating messages and `false` otherwise, right?

Is there a circumstance where this will reject, perhaps on timeout?

harbu (approved):

packages/client/src/StreamrClient.ts
juslesan: You can never be sure if a rejection happens for some reason, but the function will return false (or true if 0 messages were received in the stream) on inspection timeouts.

PR #1757, opened 2023-08-18
Fix deterministic operator

samt1803:
small fix in OperatorPlugin.test.ts

### DIFF
--- a/packages/broker/test/integration/plugins/operator/OperatorPlugin.test.ts
+++ b/packages/broker/test/integration/plugins/operator/OperatorPlugin.test.ts
@@ -23,7 +23,7 @@ describe('OperatorPlugin', () => {
             privateKey: brokerWallet.privateKey,
             extraPlugins: {
                 operator: {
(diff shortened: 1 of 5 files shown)


packages/broker/test/integration/plugins/operator/OperatorPlugin.test.ts
```diff
@@ -23,7 +23,7 @@ describe('OperatorPlugin', () => {
             privateKey: brokerWallet.privateKey,
             extraPlugins: {
                 operator: {
-                    operatorContractAddress: '0x4A5C0EC07F7ddBd4B6050181638e24b0153991b2'
+                    operatorContractAddress: '0x139dfa493a45364b598f2f98e504192819082c85'
```
teogeb, line 26: Do we have a private key for this address? If we could calculate the address from it?

samt1803: its an operator contracts address, not a wallet (contract addresses don't have private keys)

jtakalai (approved):

samt1803:
strange broken commit and trivial change, closing it, the change will come in with other PRs

PR #1758, opened 2023-08-18
refactor(network): NET-1009 Protobuf definition cleanup with binary formats

juslesan:
## Summary

Cleaned up protobuf definitions regarding StreamMessages. Pass messages and data such as groupKeys and Ethereum addresses as binary. Added functions to translate binary data to UTF8 and UTF8 to binary.

Next step is to add binary support to the client as per [NET-1009](https://linear.app/streamr/issue/NET-1009/binary-support)


packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -119,7 +111,7 @@ message LeaveStreamNotice {
 }
 
 message NeighborUpdate {
-  string senderId = 1;
+  string senderId = 1; // TODO: remove redundant info
```
teogeb: Is there a ticket about this? Add link to that ticket here?

packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -34,11 +34,7 @@ message MessageRef {
   string messageChainId = 3;
```
teogeb: Could put the fields in the order in which they are in MessageID. I.e.

```
  string streamId = 1;
  int32 streamPartition = 2;
  int64 timestamp = 3;
  int32 sequenceNumber = 4;
  bytes publisherId = 5;
  string messageChainId = 6;
```

packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -47,48 +43,44 @@ enum StreamMessageType {
...
-  MessageRef messageRef = 5;
-  optional MessageRef previousMessageRef = 6;
-  optional string groupKeyId = 7;
-  optional EncryptedGroupKey newGroupKey = 8;
+  ContentType contentType = 2;
+  EncryptionType encryptionType = 3;
+  bytes content = 4;
+  bytes signature = 5;
+  MessageRef messageRef = 6;
+  optional MessageRef previousMessageRef = 7;
```
harbu, line 68: Should've noticed this during our call but... we only need to store `timestamp` and `sequenceNumber` for `previousMessageRef`. The other fields should not be stored.

packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -47,48 +43,44 @@ enum StreamMessageType {
...
-  string signature = 4;
-  MessageRef messageRef = 5;
-  optional MessageRef previousMessageRef = 6;
-  optional string groupKeyId = 7;
-  optional EncryptedGroupKey newGroupKey = 8;
+  ContentType contentType = 2;
+  EncryptionType encryptionType = 3;
+  bytes content = 4;
+  bytes signature = 5;
+  MessageRef messageRef = 6;
```
harbu: Keeping to existing terminology I would rename the type to `MessageID` and field name to `messageId` unless there is a specific reason to deviate from this naming convention.

packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -29,12 +29,12 @@ service TemporaryConnectionRpc {
 }
 
 message MessageRef {
```
teogeb: And entity which has these fields is called `MessageID` in our current protocol. And `MessageRef` is a simpler entity which contains just `timestamp` + `number` (usually a pointer to previous message of some message, i.e. we already know the other fields which are equal to that message).

Is in intentional that we have only one type? Maybe we'd need both type? (If we choose use only one type, we should maybe rename this to `MessageID`?)

packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -47,48 +43,44 @@ enum StreamMessageType {
...
-  optional string groupKeyId = 7;
-  optional EncryptedGroupKey newGroupKey = 8;
+  ContentType contentType = 2;
+  EncryptionType encryptionType = 3;
+  bytes content = 4;
+  bytes signature = 5;
+  MessageRef messageRef = 6;
+  optional MessageRef previousMessageRef = 7;
+  optional string groupKeyId = 8;
+  optional EncryptedGroupKey newGroupKey = 9;
```
harbu: We could improve the naming here for clarity (I know this mimics the existing `StreamMessage` structure).

Should it be
```
optional string requestedGroupKeyId = 8
```
to differentiate between what fields `groupKeyId` and `newGroupKey.groupKeyId`  refer to?

@teogeb do you remember the difference between these two fields?

packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -29,16 +29,12 @@ service TemporaryConnectionRpc {
...
-  string body = 1;
+  string streamId = 1;
+  int32 streamPartition = 2;
+  int64 timestamp = 3;
+  int32 sequenceNumber = 4;
+  bytes publisherId = 5;
+  string messageChainId = 6;
 }
 
 enum StreamMessageType {
```
teogeb, line 45: At some point we could have e.g. `ProtocolMessage` main type, which is either `ContentMessage`, or `GroupKeyRequest`, or `GroupKeyResponse`? But not part of this PR, maybe.

packages/trackerless-network/src/logic/protocol-integration/stream-message/GroupKeyRequestTranslator.ts
```diff
@@ -1,25 +1,26 @@
...
+            rsaPublicKey: toBinary(msg.rsaPublicKey),
             groupKeyIds: msg.groupKeyIds
         }
         return translated
     }
 
     static toClientProtocol(msg: GroupKeyRequest): OldGroupKeyRequest {
         const translated = new OldGroupKeyRequest({
-            recipient: msg.recipient as EthereumAddress,
+            recipient: toUTF8(msg.recipient) as EthereumAddress,
```
harbu: You need to use `toEthereumAddress` to ensure that the resulting `EthereumAddress` is lower-cased.

packages/trackerless-network/src/logic/protocol-integration/stream-message/GroupKeyResponseTranslator.ts
```diff
@@ -1,37 +1,36 @@
...
+        const encryptedGroupKeys = msg.groupKeys.map((groupKey: EncryptedGroupKey) => new OldEncryptedGroupKey(
             groupKey.groupKeyId,
-            groupKey.encryptedGroupKeyHex,
-            groupKey.serialized
+            toUTF8(groupKey.data),
         ))
         return new OldGroupKeyResponse({
             requestId: msg.requestId,
-            recipient: msg.recipient as EthereumAddress,
+            recipient: toUTF8(msg.recipient) as EthereumAddress,
```
harbu: Requires `toEthereumAddress` to ensure lower casing of address.

packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -47,48 +43,44 @@ enum StreamMessageType {
...
+  optional MessageRef previousMessageRef = 7;
+  optional string groupKeyId = 8;
+  optional EncryptedGroupKey newGroupKey = 9;
 }
 
 message GroupKeyRequest {
   string requestId = 1;
-  string recipient = 2;
-  string rsaPublicKey = 3;
+  bytes recipient = 2;
```
teogeb: We seem to have a convention that when we refer to id of some entity, we suffix the field with `Id` suffix:
- rename this to `recipientId`? (also in `GroupKeyResponse`)
- and `neighbors` to `neighborIds` at line 91?

packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -47,48 +43,44 @@ enum StreamMessageType {
...
+  repeated EncryptedGroupKey groupKeys = 3;
 }
 
-message Layer2Message {
-  Layer2Type type = 1;
-}
-
-enum Layer2Type {
-  Data = 0;
+message EncryptedGroupKey {
```
teogeb:
```suggestion
message GroupKey {
```

packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -47,48 +43,44 @@ enum StreamMessageType {
...
+  string groupKeyId = 1;
+  bytes data = 2;
 }
 
 message StreamHandshakeRequest {
```
teogeb, line 91: Rename all `randomGraphId` fields to `graphId`s? If there are no other graphs types? (And if there are other types, those other graphs would just have separate id space.)

packages/trackerless-network/src/logic/protocol-integration/stream-message/StreamMessageTranslator.ts
```diff
@@ -7,30 +7,56 @@ import {
...
     StreamMessage,
     StreamMessageType
 } from '../../../proto/packages/trackerless-network/protos/NetworkRpc'
-import { ContentMessageTranslator } from './ContentMessageTranslator'
 import { EthereumAddress } from '@streamr/utils'
 import { GroupKeyRequestTranslator } from './GroupKeyRequestTranslator'
 import { GroupKeyResponseTranslator } from './GroupKeyResponseTranslator'
+import { toBinary, toUTF8 } from '../../utils'
+
+const oldToNewEnryptionType = (type: OldEncryptionType): EncryptionType => {
```
harbu: When we do the follow-up PR to this of adding binary payload support to `streamr-client` we should think about removing the the `RSA` from the `OldEncryptionType` there as well.

packages/trackerless-network/src/logic/protocol-integration/stream-message/StreamMessageTranslator.ts
```diff
@@ -67,25 +93,25 @@ export class StreamMessageTranslator {
...
+                data: toBinary(msg.getNewGroupKey()!.encryptedGroupKeyHex),
+                groupKeyId: msg.getNewGroupKey()!.groupKeyId
             }
         }
         const translated: StreamMessage = {
-            content: content,
-            encryptionType: msg.encryptionType,
+            content,
+            contentType: oldToNewContentType(contentType),
+            encryptionType: oldToNewEnryptionType(msg.encryptionType),
```
harbu:
```suggestion
            encryptionType: oldToNewEncryptionType(msg.encryptionType),
```

packages/trackerless-network/src/logic/protocol-integration/stream-message/StreamMessageTranslator.ts
```diff
@@ -112,7 +139,7 @@ export class StreamMessageTranslator {
             msg.messageRef!.streamPartition,
             Number(msg.messageRef!.timestamp),
             msg.messageRef!.sequenceNumber,
-            msg.messageRef!.publisherId as EthereumAddress,
+            toUTF8(msg.messageRef!.publisherId) as EthereumAddress,
```
harbu: `toEthereumAddress`

packages/trackerless-network/protos/NetworkRpc.proto
teogeb: Also what kind of content is in `StreamHandshakeRequest#interleavingFrom`? That info could maybe part of the field name as it is not trivial. (the `interleaveTarget` fields in other entities are `PeerDescriptors`, but this field is not?)

harbu (changes requested):

packages/trackerless-network/src/logic/RandomGraphNode.ts
```diff
@@ -269,7 +269,7 @@ export class RandomGraphNode extends EventEmitter<Events> implements IStreamNode
         let propagationTargets = this.config.targetNeighbors.getStringIds()
         if (this.config.proxyConnectionServer) {
             const proxyTargets = (msg.messageType === StreamMessageType.GROUP_KEY_REQUEST)
-                ? this.config.proxyConnectionServer.getPeerKeysForUserId(GroupKeyRequest.fromBinary(msg.content).recipient)
+                ? this.config.proxyConnectionServer.getPeerKeysForUserId(toUTF8(GroupKeyRequest.fromBinary(msg.content).recipient))
```
teogeb: Maybe best to pass the recipient as-is to the component which uses the info, and it internally decodes it (if needed). 

In this case the `ProxyConnection#userId` could be just byte array and therefore we could just compare two byte arrays at line `ProxyStreamConnectionServer#72` (with some utility function).

packages/trackerless-network/src/logic/protocol-integration/stream-message/StreamMessageTranslator.ts
```diff
@@ -7,30 +7,56 @@ import {
...
+    return OldEncryptionType.NONE
+}
+
+const oldToNewContentType = (_type: OldContentType): ContentType => {
+    return ContentType.JSON
+}
+
+const newToOldContentType = (_type: ContentType): OldContentType => {
+    return OldContentType.JSON
+}
```
teogeb: Not needed?

packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -47,48 +43,44 @@ enum StreamMessageType {
...
 }
 
-message Layer2Message {
-  Layer2Type type = 1;
-}
-
-enum Layer2Type {
-  Data = 0;
+message EncryptedGroupKey {
+  string groupKeyId = 1;
```
teogeb: No need for prefix when storing an id of the entity itself.

```suggestion
  string id = 1;
```

packages/trackerless-network/src/logic/protocol-integration/stream-message/StreamMessageTranslator.ts
```diff
@@ -67,25 +93,25 @@ export class StreamMessageTranslator {
...
         if (msg.getNewGroupKey()) {
             newGroupKey = {
-                encryptedGroupKeyHex: msg.getNewGroupKey()!.encryptedGroupKeyHex,
-                groupKeyId: msg.getNewGroupKey()!.groupKeyId,
-                serialized: msg.getNewGroupKey()!.serialized ?? undefined
+                data: toBinary(msg.getNewGroupKey()!.encryptedGroupKeyHex),
+                groupKeyId: msg.getNewGroupKey()!.groupKeyId
             }
         }
         const translated: StreamMessage = {
```
teogeb, line 95: It is easiet to read the code (and not forget any field) if the fields are in the same order as in StreamMessage class itself.

packages/trackerless-network/src/logic/utils.ts
```diff
@@ -15,3 +15,14 @@ export const markAndCheckDuplicate = (
...
+const textEncoder = new TextEncoder()
+const textDecoder = new TextDecoder()
+
+export const toUTF8 = (bytes: Uint8Array): string => {
+    return textDecoder.decode(bytes)
+}
+
+export const toBinary = (utf8: string): Uint8Array => {
+    return textEncoder.encode(utf8)
+}
```
harbu: Hmmm now that I think of this we need separate encoders / decoders for hexadecimal strings (such as Ethereum addresses, and I _think_, the signatures) if we want to have the most efficient byte representation:

```ts
const ethAddress = '0x7dbA44371D5D0386b4bd4F5C5d98b578092e117a'
const encoded = Buffer.from(ethAddress.slice(2), 'hex')
encoded.length // == 20 bytes
```

vs

```ts
const ethAddress = '0x7dbA44371D5D0386b4bd4F5C5d98b578092e117a'
const encoded = new TextEncoder().encode(ethAddress)
encoded.length // = 42 bytes
```

packages/trackerless-network/protos/NetworkRpc.proto
teogeb: The `groupKeyId` id the group key used to encrypt this message. 

The `newGroupKey` is a new key which is delivered here for future use. In practice it is used in key rotation so that all subscribers get the key automatically. Usually we store it locally and use it to decode the next message (as the `groupKeyId` of that message will be the value of this message's `newGroupKey.id`)

packages/trackerless-network/src/logic/protocol-integration/stream-message/StreamMessageTranslator.ts
```diff
@@ -123,17 +150,17 @@ export class StreamMessageTranslator {
...
             newGroupKey = new OldEncryptedGroupKey(
                 msg.newGroupKey!.groupKeyId,
-                msg.newGroupKey!.encryptedGroupKeyHex,
-                msg.newGroupKey!.serialized
+                toUTF8(msg.newGroupKey!.data),
             )
         }
         const translated = new OldStreamMessage<T>({
-            signature: msg.signature,
+            signature: toUTF8(msg.signature),
```
teogeb: Should be `toHex`? Also at line `153` and e.g. in `utils.ts:9`

Maybe should double-check cases when we convert from/to a binary id like this? Usually that binary is just bytes which we represent in hex, not UTF8 chars.

packages/trackerless-network/protos/NetworkRpc.proto
harbu: Aahh never mind my comment then.

harbu: Let's keep it as is then.

packages/trackerless-network/src/logic/utils.ts
juslesan: Yeah I pretty much had the same thought. Might as well do the change here already instead of the client side changes.

juslesan: Added hex encoding/decoding functions

packages/trackerless-network/protos/NetworkRpc.proto
juslesan: Renamed recipient. Not making changes to messages that are not realted to StreamMessages in this PR.

packages/trackerless-network/protos/NetworkRpc.proto
juslesan: Only making changes related to StreamMessages in this PR.

packages/trackerless-network/src/logic/RandomGraphNode.ts
juslesan: Only making changes related to StreamMessages in this PR. This change requires a Proxy logic refactor.

harbu (approved):
lgtm!

PR #1756, opened 2023-08-18
refactor(protocol): Remove old network protocol messages

juslesan:
## Summary

Removed old network layer messages from protocol

PR #1761, opened 2023-08-18
refactor: remove unneeded code from protocol

harbu:
## Summary
Continuing from Santeri's work here https://github.com/streamr-dev/network/pull/1756 remove some more unneeded stuff from protocol package.

*Note*: broker plugins no longer support version parameter for outputting raw stream messages.


packages/protocol/src/protocol/message_layer/StreamMessage.ts
```diff
@@ -1,18 +1,16 @@
...
 import MessageRef from './MessageRef'
 import MessageID from './MessageID'
 import EncryptedGroupKey from './EncryptedGroupKey'
-import { Serializer } from '../../Serializer'
 import { StreamID } from '../../utils/StreamID'
 import { StreamPartID } from '../../utils/StreamPartID'
 import { EthereumAddress } from '@streamr/utils'
+import { fromArray, toArray } from './streamMessageSerialization'
 
-const serializerByVersion: Record<string, Serializer<StreamMessage>> = {}
```
teogeb, line 15: Maybe now we could have only one place where the version number `32` is defined.

E.g. static LATEST_VERSION = 32 at line 54 and use that static variable in `streamMessageSerialization.ts#8` and `streamMessageSerialization.test.ts#11`. Or alternatively we could hide the static version number field from public API?

teogeb (approved):

packages/protocol/src/protocol/message_layer/StreamMessage.ts
harbu: Feel free to add commit with this change

PR #1773, opened 2023-08-23
NET-1000: InspectRandomNodeService

harbu:
## Summary

Implement `InspectRandomNodeService` which on an interval selects a random operator's node for inspection. The random operator is picked from the set of sponsorships being worked on by "my" operator's node.

Full description of service is in the working document.

- Creating the service (and starting it) is commented out in `OperatorPlugin`. It will be a lot of code re-organization to get it properly initialized so I think that is better served by a 2nd PR after this one. 
- There are some rare edge cases that could use unit tests (notice the leftover TODO for this)
- Can we do the `operatorCount >= 2` filtering already in GraphQL side (minor optimization)? 
- We will need to extract part of the inspection logic for `VoteOnSuspectNodeService` later (see NET-1059)
- Break up `InspectRandomNodeService` functions into their own files?


packages/broker/src/helpers/weightedSample.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @returns The sampled item, or undefined if the array is empty
+ *
+ */
+export function weightedSample<T>(items: T[], weight: (t: T) => number, sampleFn = random): T | undefined {
+    if (items.length === 0) {
+        return undefined
+    }
+
+    const cumulativeWeights = [weight(items[0])]
+    for (let i = 1; i < items.length; ++i) {
```
teogeb, line 19: Are the some benefits to use `++i` instead of `i++`?

packages/broker/src/plugins/operator/InspectRandomNodeService.ts
```diff
@@ -1,15 +1,211 @@
...
+
+export async function findTarget(
+    myOperatorContractAddress: EthereumAddress,
+    helper: InspectRandomNodeHelper,
+    loadBalancer: StreamAssignmentLoadBalancer
+): Promise<Target | undefined> {
+    // choose sponsorship
+    const sponsorships = await helper.getSponsorshipsOfOperator(myOperatorContractAddress)
+    const suitableSponsorships = sponsorships
+        .filter(({ operatorCount }) => operatorCount >= 2)  // TODO: could this filtering be done in the graphql query?
```
teogeb: add a comment why this filter: because we don't want sponsorships where this operator is the only operator

packages/broker/src/plugins/operator/InspectRandomNodeService.ts
```diff
@@ -1,15 +1,211 @@
...
+        targetOperator: target.operatorAddress,
+        maxWait
+    })
+    const targetOperatorFleetState = new OperatorFleetState(
+        streamrClient,
+        toStreamID('/operator/coordination', target.operatorAddress)
+    )
+    try {
+        await targetOperatorFleetState.start()
+        await Promise.race([
```
teogeb, line 103: use `withTimeout` from utils package

packages/broker/src/plugins/operator/InspectRandomNodeService.ts
```diff
@@ -1,15 +1,211 @@
...
+        )
+
+        logger.info('Inspecting nodes of operator', {
+            targetOperator: target.operatorAddress,
+            targetStreamPart: target.streamPart,
+            targetNodes: targetPeerDescriptors.map(({ id }) => id),
+            targetSponsorship: target.sponsorshipAddress
+        })
+
+        for (const descriptor of targetPeerDescriptors) {
```
teogeb, line 207: In which order the nodes are in this array? E.g. in the order which the heartbeat were received? If that's the case, should we randomize the order?

packages/broker/src/plugins/operator/StreamAssignmentLoadBalancer.ts
```diff
@@ -44,6 +44,25 @@ export class StreamAssignmentLoadBalancer extends EventEmitter3<StreamAssignment
...
+
+    getPartitionsOfStreamAssignedToMe(streamId: StreamID): StreamPartID[] {
+        const streamParts: StreamPartID[] = []
+        for (const streamPart of this.myStreamParts.keys()) {
+            if (StreamPartIDUtils.getStreamID(streamPart) === streamId) {
+                streamParts.push(streamPart)
+            }
+        }
+        return streamParts
+    }
```
teogeb:
```suggestion
    isAnyPartitionOfStreamAssignedToMe(streamId: StreamID): boolean {
        return Array.from(this.myStreamParts.keys()).some((streamPart) => {
            return (StreamPartIDUtils.getStreamID(streamPart) === streamId)
        })
    }

    getPartitionsOfStreamAssignedToMe(streamId: StreamID): StreamPartID[] {
        return Array.from(this.myStreamParts.keys()).filter((streamPart) => {
            return (StreamPartIDUtils.getStreamID(streamPart) === streamId)
        })
    }
```

packages/broker/test/unit/plugins/operator/InspectRandomNodeService.test.ts
```diff
@@ -0,0 +1,281 @@
...
+import { EthereumAddress, wait, waitForCondition } from '@streamr/utils'
+import { MessageListener, NetworkPeerDescriptor, StreamrClient, Subscription } from 'streamr-client'
+import { createHeartbeatMessage } from '../../../../src/plugins/operator/heartbeatUtils'
+
+const MY_OPERATOR_ADDRESS = randomEthereumAddress()
+const OTHER_OPERATOR_ADDRESS = randomEthereumAddress()
+const SPONSORSHIP_ADDRESS = randomEthereumAddress()
+const STREAM_ID = toStreamID('streamId')
+
+describe(findTarget, () => {
```
teogeb, line 23: Add test case where this operator is the only operator

packages/broker/test/unit/plugins/operator/InspectRandomNodeService.test.ts
```diff
@@ -0,0 +1,281 @@
...
+            streamPart: expect.stringMatching(/^streamId#\d$/)
+        })
+    })
+
+    // TODO: few edge-cases where state changes during asynchronicity
+})
+
+const PEER_DESCRIPTOR_ONE = { id: 'node-1' }
+const PEER_DESCRIPTOR_TWO = { id: 'node-2' }
+const PEER_DESCRIPTOR_THREE = { id: 'node-3' }
```
teogeb: Not relevant in this test, but maybe peer descriptors could be hex values: easier to see that these values are node ids

```suggestion
const PEER_DESCRIPTOR_ONE = { id: '0x1111' }
const PEER_DESCRIPTOR_TWO = { id: '0x2222' }
const PEER_DESCRIPTOR_THREE = { id: '0x3333' }
```

packages/broker/src/plugins/operator/InspectRandomNodeService.ts
harbu: `withTimeout` throws on timeout so didn't use that here

packages/broker/src/plugins/operator/InspectRandomNodeService.ts
harbu: Maybe randomization would be good here.

(We could also consider sorting the array by latest heartbeat, i.e., the node with the latest heartbeat is most likely alive, but this would open a door to "game" the order.)

packages/broker/src/plugins/operator/StreamAssignmentLoadBalancer.ts
harbu: I will do this methods but in a bit of a different way.

I will add a general accessor to StreamAssignmentLoadBalancer for getting stream parts `getMyStreamParts(): StreamPartID[]` and then implement these methods as helper functions since we noticed code duplication in tests.

packages/broker/test/unit/plugins/operator/InspectRandomNodeService.test.ts
harbu: renamed test

packages/broker/test/unit/plugins/operator/InspectRandomNodeService.test.ts
harbu: fixed

teogeb (approved):

PR #1777, opened 2023-08-24
ci: remove release script

teogeb:
Remove `release.yml` which currently doesn't have any jobs. GitHub actions sends us notification mail about the empty file, and removing the file resolves that issue.


harbu:
Or should we just fix it for 1.0 era?

harbu:
Implement NET-1058 instead

PR #1778, opened 2023-08-24
chore: consistent nodeId hex-format

harbu:
## Summary

NodeId is now always represented in the same Hex format be it from `getNodeId()` or `getPeerDescriptor()`.

## Notable details
- Entrypoint node id (in broker tests) is `eeeeeeeeee`
- Broker ids (in client tests) are `b1`, `b2`, and `b3`
- Production entrypoint does not yet exist, use `aaaaaaaaaa` as placeholder for the time being


packages/client/src/utils/utils.ts
```diff
@@ -131,7 +131,7 @@ export function convertPeerDescriptorToNetworkPeerDescriptor(descriptor: PeerDes
     }
     return {
         ...omit(descriptor, 'kademliaId'),
-        id: PeerID.fromValue(descriptor.kademliaId).toString(),
+        id: PeerID.fromValue(descriptor.kademliaId).toKey(),
```
teogeb: The original implementation which called `toString` _should_ be the correct one. When we want to get the string representation of a `PeerID` and there is an implemented method `PeerID#toString`, it make a lot of sense to use that method.

I'd suggest that we correct the `PeerID` class to work correctly. (Maybe the class doesn't even need two representations of the data, i.e. we could remove `PeerIDKey`?)

packages/client/test/end-to-end/StreamrClient.test.ts
```diff
@@ -20,11 +18,11 @@ describe('StreamrClient', () => {
     })
 
     it('getPeerDescriptor', async () => {
-        await client.subscribe('foobar')
         const descriptor = await client.getPeerDescriptor()
         expect(descriptor).toMatchObject({
-            id: expect.stringMatching(UUID_REGEX),
+            id: expect.toBeString(),
```
teogeb, line 23: This checks the implementation detail of `PeerID`. Maybe any string is a valid `NetworkPeerDescriptor#id` value, as there is no specific format that the data is? (It it is maybe just semi-random bytes)

teogeb (changes requested):

packages/client/test/end-to-end/StreamrClient.test.ts
harbu: It used to check the implementation detail (i.e. that it is an UUID). Now it just checks that it is any string, which is less of an implementation detail?

> as there is no specific format that the data is?

Maybe there is or isn't. But it isn't a concern of the client or the end-user. For them it's just a string that represents a network node id.

packages/client/src/utils/utils.ts
harbu: Thoughts on this @juslesan 

Also is there a reason we need two different string representations?

juslesan: The key representation is absolutely there intentionally. For example the keys are used in map structures as the UTF8 string outputs have a danger colliding in cases where PeerIDs are generated from anything other than a string.

The safest way to ensure uniqueness of the PeerIDs is to use them as keys. If anything the toString representation should be removed, although it can be a useful tool in development

juslesan:
To make the use of hex representation in the NetworkPeerDescriptors consistent, the configuration of entry points and explicit PeerDescriptors in the client should also start using hex as the id field.

packages/client/src/utils/utils.ts
harbu: So If I understood correctly @juslesan :
- toKey() returns a unique identifier
- toString() returns a human-readable identifier that is _usually_ unique but has the possibility of conflict?

The primary use of `toString` is for development logs etc.? Whilst `toKey()` should be used as the true identifier.

teogeb: If possible it would be best to have only one string representation. E.g. if we see from a log that we receive a message from peer `foo` (`peerId.toString()`), but from a debugger we see that a message is in a map with key `bar` (`peerId.toKey()`), it is very misleading and slows down our debugging etc.

teogeb: As we landed some recent PR's to `streamr-1.0`, it is now trivial to convert between `kademliaId` and `NetworkPeerDescriptor#id`.

The `id` field is always a hex string of the node id bytes. The `kademliaId` has the same bytes in `Uint8Array`. Therefore the correct way convert between the data types is to use `binaryToHex` and `hexToBinary` utility functions from `@streamr/utils` package.

packages/broker/bin/entry-point.ts
```diff
@@ -1,9 +1,9 @@
 #!/usr/bin/env node
-import { DhtNode, NodeType, PeerID } from '@streamr/dht'
+import { DhtNode, NodeType, PeerID, PeerIDKey } from '@streamr/dht'
 
 const main = async () => {
     const peerDescriptor = {
-        kademliaId: PeerID.fromString('entrypoint').value,
+        kademliaId: PeerID.fromKey('eeeeeeeeee' as PeerIDKey).value,
```
teogeb: All conversions can use `hexToBinary` and `binaryToHex` instead of `PeerID` methods

```suggestion
        kademliaId: hexToBinary('eeeeeeeeee')
```

harbu: So should we use that conversion here?

teogeb (approved):

PR #1779, opened 2023-08-24
feat(broker): Expose `StreamrClient` instead of `NetworkNodeStub`

teogeb:
The `NetworkNodeStub` is an internal class. Therefore it is better to expose `StreamrClient` instead of that.

The client is currently used only in a test (`broker-subscriptions.test.ts`). Moved also a helper method from utils to the test and simplified it.

### DIFF
--- a/packages/broker/test/integration/broker-subscriptions.test.ts
+++ b/packages/broker/test/integration/broker-subscriptions.test.ts
@@ -28,7 +28,7 @@ const grantPermissions = async (streams: Stream[], brokerUsers: Wallet[]) => {
 
 export const getStreamParts = async (broker: Broker): Promise<StreamPartID[]> => {
     const node = await (broker.getStreamrClient().getNode())
-    return Array.from(node.getStreamParts())
+    return node.getStreamParts()
 }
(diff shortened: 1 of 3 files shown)


packages/broker/test/integration/broker-subscriptions.test.ts
```diff
@@ -26,6 +26,11 @@ const grantPermissions = async (streams: Stream[], brokerUsers: Wallet[]) => {
     }
 }
 
+export const getStreamParts = async (broker: Broker): Promise<StreamPartID[]> => {
+    const node = await (broker.getStreamrClient().getNode())
+    return node.getStreamParts()
+}
```
harbu: Not necessarily within the scope of this PR (up to you): but could we get this info thru `client.getSubscriptions()` as well?

harbu (approved):

packages/broker/test/integration/broker-subscriptions.test.ts
teogeb: Good point 👍

PR #1780, opened 2023-08-24
refactor(network): Binary utils to utils

juslesan:
## Summary

Move binary utils to the `utils` package. The binary utils will be needed in the client package in the `NET-1009` changes.


packages/utils/src/binaryUtils.ts
```diff
@@ -0,0 +1,28 @@
...
+    }
+    return Buffer.from(bytes).toString('hex')
+}
+
+export const hexToBinary = (hex: string): Uint8Array => {
+    if (hex.startsWith('0x')) {
+        hex = hex.slice(2)
+    }
+    return Buffer.from(hex, 'hex')
+}
```
harbu: Add check to prevent "partial" results due to [data trunctation](https://nodejs.org/api/buffer.html#buffers-and-character-encodings). Return undefined or throw error.

```suggestion
export const hexToBinary = (hex: string): Uint8Array | undefined => {
    if (hex.startsWith('0x')) {
        hex = hex.slice(2)
    }
    const result = Buffer.from(hex, 'hex')
    return hex.length === result.length * 2 ? result : undefined
}
```

harbu (commented):
lgtm; add check to disallow what node.js refers to as "data truncation"

packages/utils/test/binaryUtils.test.ts
```diff
@@ -0,0 +1,17 @@
...
+        const utf8 = 'Hello, world!'
+        const binary = utf8ToBinary(utf8)
+        expect(binaryToUtf8(binary)).toEqual(utf8)
+    })
+
+    it('can translate hex', () => {
+        const hex = '0x1234567890abcdef'
+        const binary = hexToBinary(hex)
+        expect(binaryToHex(binary, true)).toEqual(hex)
+    })
```
teogeb, line 15: Also maybe test cases:
- malformed hex input
- which do not contain the hex prefix

packages/utils/src/binaryUtils.ts
```diff
@@ -16,11 +16,15 @@ export const binaryToHex = (bytes: Uint8Array, addPrefix = false): string => {
...
+export const hexToBinary = (hex: string): Uint8Array | undefined => {
     if (hex.startsWith('0x')) {
         hex = hex.slice(2)
     }
-    return Buffer.from(hex, 'hex')
+    if (hex.length % 2 !== 0) {
+        return undefined
+    }
+    const result = Buffer.from(hex, 'hex')
+    return (hex.length - (hex.length % 2)) === result.length * 2 ? result : undefined
```
teogeb: `hex.length % 2` is zero (per line `23`), no need to subtract it

teogeb (approved):

harbu (approved):

PR #1781, opened 2023-08-24
revamped await timings, conservative estimate for CI, removed unneede…

samt1803:
no breaking changes, fixes flaky MaintainTopologyHelper test


samt1803:
it got fixed one day earlier in the refactor

PR #1782, opened 2023-08-24
refactor(dht): Simplify promise handling

teogeb:
These changes resolved eslint `promise/always-return` warnings.

## Other changes

- change log level of timeout error and simplified logging in `RoutingSession`
- disabled the warning in two benchmark tests


juslesan:
Disable virtual time completely in browser tests via webpack to avoid the observed setImmediate issue with virtual time in browsers.

packages/dht/src/dht/routing/Router.ts
```diff
@@ -122,15 +122,20 @@ export class Router implements IRouter {
...
-            raceEvents3<RoutingSessionEvents>(session, ['routingSucceeded', 'partialSuccess', 'routingFailed', 'stopped', 'noCandidatesFound'], 10000)
-                .then(() => {
+            const eventReceived = raceEvents3<RoutingSessionEvents>(
+                session,
+                ['routingSucceeded', 'partialSuccess', 'routingFailed', 'stopped', 'noCandidatesFound'],
+                null
+            )
+            setImmediate(async () => {
+                try {
+                    await withTimeout(eventReceived, 10000)
```
juslesan, line 135: Awaiting a promise inside a setImmediate block is a dangarous patter as the promise could be rejected before the await listener is set.

teogeb:
> Disable virtual time completely in browser tests via webpack to avoid the observed setImmediate issue with virtual time in browsers.

Implemented in https://github.com/streamr-dev/network/pull/1850

packages/dht/src/dht/routing/Router.ts
teogeb: Using `executeSafePromise` wrapper to underline the fact that this promise can't reject (as it only waits events and never timeouts)

juslesan (approved):

PR #1784, opened 2023-08-25
ci: ignore missing log file warning

teogeb:
Currently we always upload CI log artifacts from GitHub actions (see commit https://github.com/streamr-dev/network/commit/e6349ec77eb9b0cfa2830e46445e93cfc383d954). Modified job definition so that it doesn't output a warning if there is no log file.


harbu:
Nice one 👍

harbu (approved):

PR #1785, opened 2023-08-25
feat(broker): collect peer descriptors in `OperatorFleetState`

harbu:
## Summary

Collect the `NetworkPeerDescriptor` sent in heartbeats in the `OperatorFleetState` and make them queryable by the user.

This is required to implement one of the steps of the `InspectRandomNodeService` (wip, #1773), from our planning doc:

> Create a new instance of OperatorFleetState for the selected Operator, and use it to find out which node(s) in the fleet are working on the stream-partition - those are the ones that should be inspected

To be able to inspect those nodes, we need to know their peer descriptors and I guess this is the most obvious way of acquiring them?

## Changes
- Collect the `NetworkPeerDescriptor` included in heartbeats in `OperatorFleetState`.
- Add validation library `zod` to broker and use it to create a validation function for the incoming heartbeat mesages.
- Create helper method `createHeartbeatMessage` as we create heartbeat messages in many places in the code.


packages/broker/src/plugins/operator/OperatorFleetState.ts
```diff
@@ -37,6 +31,7 @@ export class OperatorFleetState extends EventEmitter<OperatorFleetStateEvents> {
     private readonly heartbeatIntervalInMs: number
     private readonly latencyExtraInMs: number
     private readonly heartbeatTimestamps = new Map<NodeId, number>()
```
teogeb: Could simplify the state handling by combining `heartbeatTimestamps` and `peerDescriptors` to a single private `readonly latestHeartbeats = new Map<NodeId, number>()`

```
interface Heartbeat {
    timestamp: number
    peerDescriptor: NetworkPeerDescriptor
}
```

packages/broker/src/plugins/operator/OperatorFleetState.ts
```diff
@@ -103,6 +103,10 @@ export class OperatorFleetState extends EventEmitter<OperatorFleetStateEvents> {
         return [...this.heartbeatTimestamps.keys()]
     }
 
+    getPeerDescriptorOf(nodeId: NodeId): NetworkPeerDescriptor | undefined {
```
teogeb:
```suggestion
    getPeerDescriptor(nodeId: NodeId): NetworkPeerDescriptor | undefined {
```

harbu: It's not really a pure getter though since it requires an argument?

packages/broker/src/plugins/operator/OperatorFleetState.ts
harbu: Fixed in d2887938881ae8400c0d35b0e33de7ecce795b39

PR #1786, opened 2023-08-25
refactor(client, protocol): pass signature as binary

juslesan:
## Summary

`@streamr/protocol` StreamMessage's signature field is passed as binary in the client.

## Changes

- signature is now of type `Uint8Array`
- `signingUtils.ts` functions now output `Uint8Array` and take signature inputs as `Uint8Array`
- StreamMessage `toArray` and `fromArray` translate the binary field to hex. Used in storage and resend related logic.

PR #1787, opened 2023-08-25
refactor(trackerless-network): Proxy connection userId type

teogeb:
Changed the protocol field from string to byte array as it is currently an Ethereum address (like e.g. `MessageID#publisherId`).

Introduced a helper type `UserId` to `identifiers.ts`. 

Currently just this one filed has been annotated with the type. But we could create a separate PR in which we annotate all fields which represents user's identity (i.e. `client.getAddress()`) with that type. Those types include at least `MessageID#publisherId`, `GroupKeyRequest#recipientId`.


harbu:
lgtm; but santeri should double-check

harbu (approved):

packages/trackerless-network/src/logic/utils.ts
```diff
@@ -40,3 +41,15 @@ export const hexToBinary = (hex: string): Uint8Array => {
     }
     return Buffer.from(hex, 'hex')
 }
+
+export const areEqualUsers = (userId1: UserID, userId2: UserID): boolean => {
+    return areEqualByteArrays(userId1, userId2)
+}
+
+const areEqualByteArrays = (arr1: Uint8Array, arr2: Uint8Array): boolean => {
```
juslesan, line 49: This function can just be:
`return Buffer.compare(a, b) === 0`
as in #1786 [@streamr/utils](https://github.com/streamr-dev/network/blob/ecc46c1564c2896bcc4e9a5b6d7f0eeced299b88/packages/utils/src/binaryUtils.ts#L26)

teogeb: Fixed 👍

juslesan (approved):

PR #1791, opened 2023-08-25
refactor(trackerless-network): Rename `NodeId`

teogeb:
Unify type naming: rename `NodeId` to `NodeID`.


packages/trackerless-network/src/logic/utils.ts
```diff
@@ -40,3 +41,15 @@ export const hexToBinary = (hex: string): Uint8Array => {
...
+    return areEqualByteArrays(userId1, userId2)
+}
+
+const areEqualByteArrays = (arr1: Uint8Array, arr2: Uint8Array): boolean => {
+    if (arr1.byteLength === arr2.byteLength) {
+        return arr1.every((value, index) => value === arr2[index])
+    } else {
+        return false
+    }
+}
```
harbu: There is a better way to compare with `Buffer`?

harbu (approved):
Maybe description and title should also include that the type of userId field changes in protobuf specification?

packages/trackerless-network/src/logic/utils.ts
teogeb: Yes, fixed that in the base PR (https://github.com/streamr-dev/network/pull/1787)

teogeb:
> Maybe description and title should also include that the type of userId field changes in protobuf specification?

That change was mentioned in the base PR (https://github.com/streamr-dev/network/pull/1787). This PR is only about the `NodeID` type.

PR #1792, opened 2023-08-27
refactor(client, protocol) GroupKey binary

juslesan:
## Summary

Group key data is now passed as already binary inside the client and protocol.

## Changes

- `@streamr/protocol` EncryptedGroupKey field `encryptedGroupKeyHex` is renamed to `data`
- the `data` field is now of type `Uint8Array` instead of a hex string
- RSA key pair encryption and decryption now only accept binary inputs and always output binary


packages/client/src/encryption/EncryptionUtil.ts
```diff
@@ -20,25 +20,17 @@ export class EncryptionUtil {
     }
 
     /**
-     * Returns a Buffer or a hex String
+     * Returns a Buffer
```
harbu: Remove comment

packages/client/src/encryption/GroupKey.ts
```diff
@@ -68,22 +68,22 @@ export class GroupKey {
 
     /** @internal */
     encryptNextGroupKey(nextGroupKey: GroupKey): EncryptedGroupKey {
-        return new EncryptedGroupKey(nextGroupKey.id, EncryptionUtil.encryptWithAES(nextGroupKey.data, this.data))
+        return new EncryptedGroupKey(nextGroupKey.id, hexToBinary(EncryptionUtil.encryptWithAES(nextGroupKey.data, this.data)))
```
teogeb, line 71: The `encryptWithAES` method could return `Buffer` or `Uint8Array` instead of string? And similarly `decryptWithAES` could get the binary input instead of string?

packages/client/test/unit/MessageFactory.test.ts
```diff
@@ -116,7 +116,7 @@ describe('MessageFactory', () => {
         expect(msg.groupKeyId).toBe(GROUP_KEY.id)
         expect(msg.newGroupKey).toMatchObject({
             groupKeyId: nextGroupKey.id,
-            encryptedGroupKeyHex: expect.any(String)
+            data: expect.any(Buffer)
```
teogeb: Should it be exactly `Uint8Array`?

```suggestion
            data: expect.any(Uint8Array)
```

teogeb (changes requested):

packages/client/src/encryption/GroupKey.ts
juslesan: AES encryption changes are here. I could extract them to a new PR. https://github.com/streamr-dev/network/pull/1827/files#diff-12fc8d398bfd4ecaddfae2289da7958c5d3c67c51a9d8f61524c9a043ee6e4b4

teogeb: Could cherry-pick to this PR as needed here. Separate PR is also good but maybe that requires some redundant extra work.

juslesan: It is only used for group keys here though. Adding it here will require binary translation for regular messages so I don't see it as much of a win

harbu (approved):

teogeb (approved):

PR #1794, opened 2023-08-28
refactor(trackerless-network): Use `StreamPartID`

teogeb:
Use `StreamPartID` type instead of plain `string`.


packages/trackerless-network/src/logic/StreamEntryPointDiscovery.ts
```diff
@@ -66,7 +67,8 @@ interface StreamEntryPointDiscoveryConfig {
 export class StreamEntryPointDiscovery {
     private readonly abortController: AbortController
     private readonly config: StreamEntryPointDiscoveryConfig
-    private readonly servicedStreams: Map<string, NodeJS.Timeout>
+    // TODO rename: contains stream parts, not streams
+    private readonly servicedStreams: Map<StreamPartID, NodeJS.Timeout>
```
harbu:
```suggestion
    private readonly servicedStreamParts: Map<StreamPartID, NodeJS.Timeout>
```
Why not rename in this PR?

harbu (approved):

juslesan (approved):

packages/trackerless-network/src/logic/StreamEntryPointDiscovery.ts
teogeb: Fixed this. This is only one of the mismatch in the codebase, it would make sense go through these in a separate PR. Also the this could snowball e.g. if we'd consider that the class name is not optimal either.

PR #1796, opened 2023-08-28
NET-1020: refactor MaintainTopologyHelper and attempt to fix flakyness

harbu:
## Summary

Refactor MaintainTopologyHelper.test.ts and increase waits for reduce flakyness. Also fix the stop method of class `MaintainTopologyHelper` to properly remove listeners.


packages/broker/test/integration/plugins/operator/MaintainTopologyHelper.test.ts
```diff
@@ -1,17 +1,29 @@
...
-describe('MaintainTopologyHelper', () => {
+import { delegate, deploySponsorshipContract, setupOperatorContract, stake, unstake } from './contractUtils'
+import { OperatorServiceConfig } from '../../../../src/plugins/operator/OperatorPlugin'
 
+const TIMEOUT = 30 * 1000
+const WAIT_FOR_EVENTS_IN_MS = 10 * 1000
+const NO_EVENTS_FIRED_WAIT_IN_MS = 4000
+const WAIT_FOR_EVENT_HANDLERS_TO_REGISTER_IN_MS = 4000
+
+async function ensureMinElapsedTime<T>(task: () => Promise<T>, minElapsedTimeInMs: number): Promise<T> {
```
teogeb: Rename e.g. to `runWhileWaiting`?

packages/broker/test/integration/plugins/operator/MaintainTopologyHelper.test.ts
```diff
@@ -20,166 +32,144 @@ describe('MaintainTopologyHelper', () => {
...
+            }, WAIT_FOR_EVENT_HANDLERS_TO_REGISTER_IN_MS)
 
             await stake(operatorContract, sponsorship1.address, 100)
-            await waitForCondition(() => receivedAddStreams === 1, 10000, 1000)
-            await stake(operatorContract, sponsorship2.address, 100)
-            await waitForCondition(() => receivedAddStreams === 1, 10000, 1000)
-
-            await wait(10000) // wait for events to be processed
+            await waitForCondition(() => events.length >= 1, WAIT_FOR_EVENTS_IN_MS)
+            expect(events.length).toEqual(1)
```
teogeb, line 145: Should we check that the event contains info about the both streams? (also lines `158` and `172`)

packages/broker/test/integration/plugins/operator/MaintainTopologyHelper.test.ts
harbu: Fixed in edbe5d9264c86787379a9d93534c9b693c923032

packages/broker/test/integration/plugins/operator/MaintainTopologyHelper.test.ts
harbu: In this case it is actually just one stream since both sponsorships point to the same string.

If you're okay with it, I'll leave the value check out for now. We should most likely anyway turn most of these cases into unit tests in the future and keep the main happy path as an integration test for example.

teogeb (approved):

PR #1799, opened 2023-08-28
Update network-contracts to 5.0.0

jtakalai:
This should be merged to `streamr-1.0` ~at the same time as the new Docker images are pushed, maybe schedule a video call.

So:
1) push new Docker images
2) re-run the tests of this PR
3) merge
4) everywhere other open PR, merge from `streamr-1.0`

### DIFF
--- a/packages/broker/test/integration/plugins/operator/contractUtils.ts
+++ b/packages/broker/test/integration/plugins/operator/contractUtils.ts
@@ -12,15 +12,15 @@ import { OperatorServiceConfig } from '../../../../src/plugins/operator/Operator
 import { range } from 'lodash'
 
 const TEST_CHAIN = 'dev2'
-// TODO read from config when https://github.com/streamr-dev/network-contracts/pull/604 
+// TODO read from config when https://github.com/streamr-dev/network-contracts/pull/604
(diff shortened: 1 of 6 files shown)


harbu (approved):

teogeb (approved):

packages/broker/test/integration/plugins/operator/contractUtils.ts
```diff
@@ -12,15 +12,15 @@ import { OperatorServiceConfig } from '../../../../src/plugins/operator/Operator
 import { range } from 'lodash'
 
 const TEST_CHAIN = 'dev2'
-// TODO read from config when https://github.com/streamr-dev/network-contracts/pull/604 
+// TODO read from config when https://github.com/streamr-dev/network-contracts/pull/604
```
samigabor, line 15: This PR has beed merged. Could now get the url from the config

samigabor (approved):

packages/broker/test/integration/plugins/operator/contractUtils.ts
jtakalai: there's no graph url in https://www.npmjs.com/package/@streamr/config/file/24f3973f2a7a388aa79d2edc869eddb81bbe84b93c1e3a99913ff4279ca203b7 which is ~same file as https://github.com/streamr-dev/network-contracts/blob/master/packages/config/src/index.ts

PR #1801, opened 2023-08-28
feat(network): Websocket port range

juslesan:
## Summary

The network layer WebSocket servers are now started in a given port range. By default this range is between 32200 and 32250 for production values. For development the port range is set between the ports 32400 and 32800.

When starting a node the range of ports is iterated through to find an open port. Once a open port is found the node will initiate connectivity checks as before. If an open port is not found the node will continue to operate without a WS server and will be dependant on connecting on WS client and WebRTC connections.

Starting the servers in avoided in browser environments to avoid crashing behaviour.

## Additional Changes

- Fixed a leaking test case in Broker where the `stop()` was never called.
- Fixed leaking test where a client was never destroyed.
- Added safe guards to WebSocket connection cases where leaks might occure (The observed WS related leaks are likely caused by stop not being called in all tests in the broker and client projects.)

- A new network control layer option to completely avoid starting a WebSocket server could be added.

- (The client and broker tests the `afterEach` blocks often call stop/destroy on the client or broker with the `?` operator. In cases where timeouts occure during start this will cause stop not to be called. This causes logging to leak on top of other tests and in the worst cases other test cases will fail because of it. It would be better to use the `!` operator to cause these test cases to fail and tear down their environments in the process instead of allowing failed test cases to leak on top of others.)


packages/broker/test/integration/plugins/operator/AnnounceNodeToStreamService.test.ts
```diff
@@ -38,5 +38,6 @@ describe(AnnounceNodeToStreamService, () => {
             msgType: 'heartbeat',
             peerDescriptor: await client.getPeerDescriptor()
         })
+        await anonymousClient.destroy()
```
harbu, line 41: thx good catch 👍

packages/dht/test/unit/WebSocketServer.test.ts
```diff
@@ -6,34 +6,45 @@ describe('WebSocketServer', () => {
...
     })
 
-    it('throws if port is already in use', async () => {
+    it('Starts server in next port if first one is already in use', async () => {
+        const server1 = new WebSocketServer()
+        await server1.start({ min: 19792, max: 19793 })
+        
+        const server2 = new WebSocketServer()
+        await server2.start({ min: 19792, max: 19793 })
+
```
harbu: You could check the return values of the `start` promises for good measure.

packages/dht/src/dht/DhtNode.ts
```diff
@@ -202,14 +209,17 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
                 maxConnections: this.config.maxConnections
             }
             // If own PeerDescriptor is given in config, create a ConnectionManager with ws server
-            if (this.config.peerDescriptor && this.config.peerDescriptor.websocket) {
+            if (this.config.peerDescriptor && this.config.peerDescriptor.websocket && isNodeJS()) {
```
harbu: Slightly shorter
```suggestion
            if (this.config.peerDescriptor?.websocket && isNodeJS()) {
```

packages/dht/src/dht/DhtNode.ts
```diff
@@ -202,14 +209,17 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
...
+                    max: this.config.peerDescriptor.websocket.port
+                }
             } else {
                 // If webSocketPort is given, create ws server using it, webSocketHost can be undefined
-                if (this.config.webSocketPort) {
+                if (this.config.webSocketPortRange) {
                     connectionManagerConfig.webSocketHost = this.config.webSocketHost
-                    connectionManagerConfig.webSocketPort = this.config.webSocketPort
+                    connectionManagerConfig.webSocketPortRange = this.config.webSocketPortRange
                 }
```
harbu: I'm a bit confused by this else-block.

1. Shouldn't `isNodeJS()` also be checked for truthiness here?

2. What is `this.config.webSocketHost` and how does it relate to `this.config.peerDescriptor.websocket.ip`?

packages/client/src/Config.ts
```diff
@@ -76,12 +76,12 @@ export interface ControlLayerConfig {
...
     /**
-     * The port to use for the client's Network Node WebSocket server.
-     * If not specified, the server will not be started.
+     * The port range used to find a free port for the client's network layer WebSocket server.
+     * If not specified, a server will not be started.
      * The server is used by the network layer to accept incoming connections
      * over the public internet to improve the network node's connectivity.
      */
-    webSocketPort?: number
+    webSocketPortRange?: PortRange
```
teogeb: Is it possible to disable the websocket server in Broker's config file?

teogeb: The lowercase naming convention is usually used in the public API (e.g in `NetworkPeerDescriptor` and in Broker plugin name).

```suggestion
    websocketPortRange?: PortRange
```

packages/dht/src/connection/WebSocket/WebSocketConnector.ts
```diff
@@ -31,32 +31,31 @@ export class WebSocketConnector implements IWebSocketConnectorService {
     private readonly rpcCommunicator: ListeningRpcCommunicator
     private readonly canConnectFunction: (peerDescriptor: PeerDescriptor, _ip: string, port: number) => boolean
     private readonly webSocketServer?: WebSocketServer
-    private readonly connectivityChecker: ConnectivityChecker
+    private connectivityChecker?: ConnectivityChecker
     private readonly ongoingConnectRequests: Map<PeerIDKey, ManagedConnection> = new Map()
     private incomingConnectionCallback: (connection: ManagedConnection) => boolean
-    private webSocketPort?: number
+    private webSocketPortRange?: PortRange
```
teogeb:
```suggestion
    private portRange?: PortRange
```

(could drop the prefix from `webSocketServer` field)

packages/dht/src/connection/WebSocket/WebSocketConnector.ts
```diff
@@ -102,18 +101,22 @@ export class WebSocketConnector implements IWebSocketConnectorService {
                     this.attachHandshaker(connection)
                 }
             })
-            await this.webSocketServer.start(this.webSocketPort!, this.webSocketHost)
+            const port = await this.webSocketServer.start(this.webSocketPortRange!, this.webSocketHost)
+            this.selectedPort = port
+            this.connectivityChecker = new ConnectivityChecker(this.selectedPort!)
```
teogeb, line 106: What is the difference between `this.connectivityChecker `and the `checkConnectivity` method? Do we need both?

packages/dht/src/dht/DhtNode.ts
```diff
@@ -181,6 +182,12 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
         logger.trace(`Starting new Streamr Network DHT Node with serviceId ${this.config.serviceId}`)
         this.started = true
 
+        if (!isNodeJS()) {
```
teogeb, line 185: Would it make sense if we'd set these values outside of `DhtNode`? Now the object receives a config value in constructor but doesn't always apply that exact kind of configuration.  I.e. we could have a method `createDhtConfigDefaults` which would check the environment and pass in valid defaults?

(In that case we can drop also the check in at line `212`?)

packages/dht/src/dht/DhtNode.ts
juslesan: Making some clean ups related to this, but the NodeJS does not need to be explicitly checked here as the WS related configs are nullified at line 185.

packages/client/src/Config.ts
juslesan: This is listed as a future improvement in the PR description.

packages/dht/src/connection/WebSocket/WebSocketConnector.ts
juslesan: checkConnectivity uses the ConnectivityChecker-class.

packages/dht/src/dht/DhtNode.ts
juslesan: Could refactor the DhtNode's default config logic but not in this PR.

harbu (approved):

packages/dht/src/connection/WebSocket/WebSocketConnector.ts
```diff
@@ -102,39 +101,39 @@ export class WebSocketConnector implements IWebSocketConnectorService {
...
                     return preconfiguredConnectivityResponse
                 } else {
-                    // Do real connectivity checking
-                    
+                    // Do real connectivity checking     
                     let response = noServerConnectivityResponse
-
-                    response = await this.connectivityChecker.sendConnectivityRequest(this.entrypoints[0])
-
+                    response = await this.connectivityChecker!.sendConnectivityRequest(sample(this.entrypoints)!)
```
teogeb: Do we need `let response = noServerConnectivityResponse`? We could just `return await this.connectivityChecker!.sendConnectivityRequest(sample(this.entrypoints)!)` in this block?

packages/dht/src/connection/WebSocket/WebSocketConnector.ts
```diff
@@ -102,39 +101,39 @@ export class WebSocketConnector implements IWebSocketConnectorService {
...
-                openInternet: false,
-                ip: '127.0.0.1',
-                natType: NatType.UNKNOWN
-            }
-
             if (!this.webSocketServer) {
                 // If no websocket server, return openInternet: false
                 return noServerConnectivityResponse
             } else {
                 if (!this.entrypoints || this.entrypoints.length < 1) {
```
teogeb, line 124: It would maybe make sense to normalize the entrypoints array so that we can treat empty array and missing array equally. The normalization should be done already in `DhtNodeConfig`. Maybe a separate PR about this?

teogeb (approved):

PR #1803, opened 2023-08-29
Refactor DHT: move peer management to a separate class in DHT

ptesavol:
* Moved peer management to separate class in DHT
* Added getTestInterface() function to test-utils for extracting testing interfaces from classes (typesafe alternative to ts-ignore private field)
* increased timeout of getClosestPeers() in RemoteDhtNode from 5s to 10s 

Future improvements:

* Hide everything kademlia-specific inside PeerManager
* remove eslint rule 'promise/always-return' as it is not relevant in typescipt  
* Overall discussion about naming should be held, a tentative list of terms to be discussed:
  - the class name: `PeerManager` vs `ContactManager`, as the methods of the class seems to handle contacts https://github.com/streamr-dev/network/pull/1803#discussion_r1371424419
  - "active contacts" vs "neighbors" https://github.com/streamr-dev/network/pull/1803#discussion_r1377313695


packages/dht/src/dht/DhtNode.ts
```diff
@@ -213,10 +211,10 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
                 }
             }
 
-            const connectionManager = new ConnectionManager(connectionManagerConfig)
-            await connectionManager.start(this.generatePeerDescriptorCallBack)
-            this.connectionManager = connectionManager
-            this.transportLayer = connectionManager
+            this.connectionManager = new ConnectionManager(connectionManagerConfig)
+            await this.connectionManager.start(this.generatePeerDescriptorCallBack)
+            //this.connectionManager = connectionManager
```
juslesan: Remove commented line

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,308 @@
...
+        if (this.bucket!.count() === 0) {
+            this.emit('kBucketEmpty')
+        }
+    }
+
+    private onKBucketAdded(contact: DhtPeer): void {
+        if (this.stopped) {
+            return
+        }
+        // this.contactOnAddedCounter++
```
juslesan: remove commented line

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,308 @@
...
+            logger.trace(' ' + this.config.nodeName + ' connectionschange add ' + this.connections.size)
+        } else {
+            logger.trace('new connection not set to connections, there is already a connection with the peer ID')
+        }
+        if (this.config.nodeName === 'entrypoint') {
+            logger.trace('connected: ' + this.config.nodeName + ', ' + peerDescriptor.nodeName + ' ' + this.connections.size)
+        }
+    }
+
+    public handleDisconnected(peerDescriptor: PeerDescriptor, dicsonnectionType: DisconnectionType, isLayer0 = false): void {
```
juslesan: typo in `dicsonnectionType`

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,308 @@
...
+            const peerId = peerIdFromPeerDescriptor(contact)
+            if (!peerId.equals(this.config.ownPeerId!)) {
+                logger.trace(`Adding new contact ${contact.kademliaId.toString()}`)
+                const dhtPeer = this.config.createDhtPeer(contact)
+                if (!this.bucket!.get(contact.kademliaId) && !this.neighborList!.getContact(peerIdFromPeerDescriptor(contact))) {
+                    this.neighborList!.addContact(dhtPeer)
+                    if (contact.openInternet) {
+                        this.openInternetPeers!.addContact(dhtPeer)
+                    }
+                    //this.contactAddCounter++
```
juslesan: remove commented line

packages/dht/src/dht/DhtNode.ts
```diff
@@ -659,23 +461,21 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
...
     public getK(): number {
         return this.config.numberOfNodesPerKBucket
     }
 
-    public getKBucketPeers(): PeerDescriptor[] {
-        return this.bucket!.toArray().map((dhtPeer: DhtPeer) => dhtPeer.getPeerDescriptor())
-    }
-
+    /*
     public getOpenInternetPeerDescriptors(): PeerDescriptor[] {
```
juslesan: remove commented code block

juslesan:
There seem to be some changes to the behaviour of the DhtNode that seem to cause some new tests to fail especially in the network and client

packages/dht/src/connection/Simulator/Simulator.ts
```diff
@@ -407,4 +407,53 @@ export class Simulator extends EventEmitter<ConnectionSourceEvents> {
...
+        let count = 0
+
+        for (const node of graph.keys()) {
+            if (!visited.get(node)) {
+                dfs(node)
+                count++
+            }
+        }
+
+        function dfs(node: PeerIDKey) {
```
teogeb: What `dfs` means? Could maybe use longer method name

packages/dht/src/dht/DhtNode.ts
```diff
@@ -300,125 +312,73 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
...
+        if (this.connectionManager! && this.config.entryPoints && this.config.entryPoints.length > 0
             && !isSamePeerDescriptor(this.config.entryPoints[0], this.ownPeerDescriptor!)) {
             this.connectToEntryPoint(this.config.entryPoints[0])
         }
     }
 
-    private initKBuckets = (selfId: PeerID) => {
-        this.bucket = new KBucket<DhtPeer>({
-            localNodeId: selfId.value,
+    private initPeerManager = (selfId: PeerID) => {
```
teogeb: Rename parameter to `ownPeerId` for consistency

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,339 @@
...
+    numberOfNodesPerKBucket: number
+    maxNeighborListSize: number
+    getClosestContactsLimit: number
+    ownPeerId: PeerID
+    connectionManager: ConnectionManager
+    nodeName: string
+    createDhtPeer: (peerDescriptor: PeerDescriptor) => DhtPeer
+}
+
+export interface PeerManagerEvents {
```
teogeb, line 27: `PeerManager` seems to emit events about contacts. Are _peer_ and _contact_ synonyms?

packages/dht/src/dht/DhtNode.ts
```diff
@@ -224,20 +239,19 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
...
             if (this.config.peerDescriptor?.websocket) {
                 connectionManagerConfig.websocketHost = this.config.peerDescriptor.websocket.host
-                connectionManagerConfig.websocketPortRange = { 
+                connectionManagerConfig.websocketPortRange = {
                     min: this.config.peerDescriptor.websocket.port,
                     max: this.config.peerDescriptor.websocket.port
                 }
-            // If websocketPortRange is given, create ws server using it, websocketHost can be undefined
-            } else if (this.config.websocketPortRange) { 
+                // If websocketPortRange is given, create ws server using it, websocketHost can be undefined
```
```suggestion
            // If websocketPortRange is given, create ws server using it, websocketHost can be undefined
```

packages/dht/src/dht/IPeerManager.ts
```diff
@@ -0,0 +1,13 @@
...
+import { DhtPeer } from './DhtPeer'
+
+export interface IPeerManager {
+    getClosestPeersTo(kademliaId: Uint8Array, limit?: number, excludeSet?: Set<PeerIDKey>): DhtPeer[]
+    getNumberOfPeers(excludeSet?: Set<PeerIDKey>): number
+    handlePeerActive(peer: DhtPeer): void
+    handlePeerUnresponsive(peer: DhtPeer): void
+    handleNewPeers(peerDescriptors: PeerDescriptor[], setActive?: boolean): void
+    getDistance(kademliaId1: Uint8Array, kademliaId2: Uint8Array): number
+    getKBucketSize(): number
```
teogeb: The handle methods could maybe renamed as it is but unclear what the methods do. Also method names don't need to contain the word peer as by default all methods of `PeerManager` operate on peers.

Some ideas, not sure if correct:

```suggestion
    getClosestTo(kademliaId: Uint8Array, limit?: number, excludeSet?: Set<PeerIDKey>): DhtPeer[]
    getCount(excludeSet?: Set<PeerIDKey>): number
    setActive(peer: DhtPeer): void
    onUnresponsive(peer: DhtPeer): void
    add(peerDescriptors: PeerDescriptor[], setActive?: boolean): void
    getDistance(kademliaId1: Uint8Array, kademliaId2: Uint8Array): number
    getKBucketSize(): number
```

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,339 @@
...
+        const numClosest = closest.getClosestContacts().filter((contact) => {
+            if (!excludeSet) {
+                return true
+            } else {
+                return !excludeSet.has(contact.getPeerId().toKey())
+            } 
+        }).length
+
+        return numClosest
+    }
```
teogeb:
```suggestion
    public getNumberOfPeers = (excludeSet?: Set<PeerIDKey>): number => {
        return this.getClosestPeersTo(this.config.ownPeerId!).length
    }
```

packages/dht/src/dht/contact/SortedContactList.ts
```diff
@@ -42,7 +42,7 @@ export class SortedContactList<C extends { getPeerId: () => PeerID }> extends Co
             return
         }
         if (!this.contactsById.has(contact.getPeerId().toKey())) {
-            if (this.contactIds.length < this.maxSize) {
+            if (!this.maxSize || this.contactIds.length < this.maxSize) {
```
teogeb, line 45:
```suggestion
            if ((this.maxSize === undefined) || (this.contactIds.length < this.maxSize)) {
```

packages/dht/src/dht/IPeerManager.ts
```diff
@@ -0,0 +1,13 @@
+import { PeerIDKey } from '../helpers/PeerID'
+import { PeerDescriptor } from '../proto/packages/dht/protos/DhtRpc'
+import { DhtPeer } from './DhtPeer'
+
+export interface IPeerManager {
```
teogeb, line 5: Are we planning that there will be multiple `PeerManager` implementations? Or why an interface is needed?

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,339 @@
...
+            localNodeId: selfId.value,
+            numberOfNodesPerKBucket: this.config.numberOfNodesPerKBucket,
+            numberOfNodesToPing: this.config.numberOfNodesPerKBucket
+        })
+
+        this.bucket.on('ping', (oldContacts: DhtPeer[], newContact: DhtPeer) => this.onKBucketPing(oldContacts, newContact))
+        this.bucket.on('removed', (contact: DhtPeer) => this.onKBucketRemoved(contact))
+        this.bucket.on('added', (contact: DhtPeer) => this.onKBucketAdded(contact))
+        this.bucket.on('updated', (_oldContact: DhtPeer, _newContact: DhtPeer) => {
+            // TODO: Update contact info to the connection manager and reconnect
```
teogeb, line 74: If there is a linear ticket, we could add link to that?

packages/test-utils/src/testInterface.ts
```diff
@@ -0,0 +1,10 @@
+interface HavingTestInterface { testInterfaceType?: any }
+
+/**
+ * Gets the Testing Interface
+ * @param obj The object to get the testing interface from
+ * @returns The testing interface
+ */
+export function getTI<T extends HavingTestInterface> (obj: T): NonNullable<T['testInterfaceType']> {
```
teogeb: Rename to `getTestInterface` as TI is not a well-known acronym (and therefore not known to those who read our tests)

packages/test-utils/src/testInterface.ts
```diff
@@ -0,0 +1,10 @@
...
+    return (obj as unknown as any)['testInterface']!
```
teogeb, line 9: Could maybe rename the field to `_testInterface` (and the type to `_TestInterfaceType`) so that IDE/eslint know that it is ok to have the "unused local varible". 

(Eslint seems to be allow it also without underscore, but good to follow the convention anyway)

https://github.com/streamr-dev/network/blob/7eb22634afa250917420a432320484f2d65e4a3c/.eslintrc.js#L59

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,339 @@
...
+                    this.config.connectionManager?.weakUnlockConnection(contact.getPeerDescriptor())
+                    this.removeContact(contact.getPeerDescriptor())
+                    this.addClosestContactToBucket()
+                })
+            }
+        }
+    }
+
+    private addClosestContactToBucket(): void {
+        if ( this.stopped ) {
```
teogeb:
```suggestion
        if (this.stopped) {
```

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,339 @@
...
+        }
+    }
+
+    public handlePeerLeaving(peerDescriptor: PeerDescriptor, removeFromOpenInternetPeers = false): void {
+        logger.trace('peer leaving: ' + this.config.nodeName + ', ' + peerDescriptor.nodeName + ' ')
+        this.removeContact(peerDescriptor, removeFromOpenInternetPeers)
+    }
+
+    private removeContact(contact: PeerDescriptor, removeFromOpenInternetPeers = false): void {
+        if ( this.stopped ) {
```
teogeb:
```suggestion
        if (this.stopped) {
```

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,339 @@
...
+
+    public handlePeerUnresponsive(peer: DhtPeer): void { 
+        this.bucket!.remove(peer.getPeerId().value)
+        this.neighborList!.removeContact(peer.getPeerId())
+        this.openInternetPeers!.removeContact(peer.getPeerId())
+    }
+
+    public handleNewPeers(peerDescriptors: PeerDescriptor[], setActive?: boolean): void { 
+        peerDescriptors.forEach((contact) => {
+            if ( this.stopped ) {
```
teogeb:
```suggestion
            if (this.stopped) {
```

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,339 @@
...
+import { DisconnectionType } from '../transport/ITransport'
+
+const logger = new Logger(module)
+
+export interface PeerManagerConfig {
+    numberOfNodesPerKBucket: number
+    maxNeighborListSize: number
+    getClosestContactsLimit: number
+    ownPeerId: PeerID
+    connectionManager: ConnectionManager
```
teogeb, line 22: The `connectionManager` is used to unlock connections, but the locking is done elsewhere? Should we move the unlocking also to that same place? (The connection management seems to a separate concern, not part of the functionality of this class.)

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,339 @@
...
+
+    private onKBucketPing(oldContacts: DhtPeer[], newContact: DhtPeer): void {
+        if (this.stopped) {
+            return
+        }
+        const sortingList: SortedContactList<DhtPeer> = new SortedContactList(this.config.ownPeerId!, 100)
+        sortingList.addContacts(oldContacts)
+        const sortedContacts = sortingList.getAllContacts()
+        this.config.connectionManager?.weakUnlockConnection(sortedContacts[sortedContacts.length - 1].getPeerDescriptor())
+        this.bucket?.remove(sortedContacts[sortedContacts.length - 1].getPeerId().value)
```
teogeb, line 104: Do we know that `sortedContacts` can't be empty?

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,339 @@
...
+        if ( this.stopped ) {
+            return
+        }
+        const closest = this.getClosestActiveContactNotInBucket()
+        if (closest) {
+            this.handleNewPeers([ closest.getPeerDescriptor() ])
+        }
+    }
+
+    private getClosestActiveContactNotInBucket(): DhtPeer | undefined {
```
teogeb: Are _closest active contacts_ and _neighbors_ synonyms?

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,339 @@
...
+            logger.trace(' ' + this.config.nodeName + ' connectionschange add ' + this.connections.size)
+        } else {
+            logger.trace('new connection not set to connections, there is already a connection with the peer ID')
+        }
+        if (this.config.nodeName === 'entrypoint') {
+            logger.trace('connected: ' + this.config.nodeName + ', ' + peerDescriptor.nodeName + ' ' + this.connections.size)
+        }
+    }
+
+    public handleDisconnected(peerDescriptor: PeerDescriptor, disconnectionType: DisconnectionType, isLayer0 = false): void {
```
teogeb: Maybe good to have the `isLayer0` as a required parameter as it is not trivial to the caller what would be the default. Alternatively this could be moved to a field of `PeerManager` if the value never changes during the lifecycle of the instance.

```suggestion
    public handleDisconnected(peerDescriptor: PeerDescriptor, disconnectionType: DisconnectionType, isLayer0: boolean): void {
```

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,339 @@
...
+        this.neighborList!.stop()
+        this.randomPeers!.stop()
+        this.openInternetPeers!.stop()
+        this.connections.clear()
+    }
+
+    // IPeerManager implementation start
+
+    public getClosestPeersTo = (kademliaId: Uint8Array, limit?: number, excludeSet?: Set<PeerIDKey>): DhtPeer[] => {
+
```
teogeb:
```suggestion
```

teogeb (commented):
First round of comments and questions

packages/dht/src/dht/PeerManager.ts
ptesavol: done

packages/dht/src/dht/DhtNode.ts
ptesavol: done

packages/dht/src/connection/Simulator/Simulator.ts
ptesavol: renamed to depthFirstSearch

packages/dht/src/dht/DhtNode.ts
ptesavol: removed the function parameter completely, and use this.ownPeerId! directly

packages/dht/src/dht/PeerManager.ts
ptesavol: "contact" implies that we do not know whether the we have a connection to the node or not, we just know its contact information (in most cases we do not have a connection to it). "peer" on the other hand implies that there probaly is (or has been) a connection

packages/dht/src/dht/DhtNode.ts
ptesavol: done

packages/dht/src/dht/IPeerManager.ts
ptesavol: using "handle" is intetentional here, we do not want the caller to know what callee does when the event occurres. We just want that the callee to do something appropriate with it. This pattern is also used in some other places of the codebase, "handle" is kind of shorthand for handleXXEvent()

packages/dht/src/dht/PeerManager.ts
ptesavol: done

packages/dht/src/dht/contact/SortedContactList.ts
ptesavol: !this.maxSize allows using null "0" and "undefined" as indicator for "no limit", which I think is a good feature. I we only would allow "undefined as the indicator", we would have to throw if 0 or null was given. This would make the code more cumbersome.

packages/dht/src/dht/IPeerManager.ts
ptesavol: Yes, the plan is to have multiple ones at least in the development phase, and to compare which implementations give the best benchmark results.

packages/dht/src/dht/PeerManager.ts
ptesavol: @juslesan ?

packages/test-utils/src/testInterface.ts
ptesavol: done

packages/dht/src/dht/PeerManager.ts
ptesavol: done

packages/dht/src/dht/PeerManager.ts
ptesavol: done

packages/test-utils/src/testInterface.ts
ptesavol: I would not do this, because these variables are not unused in reality.

packages/dht/src/dht/PeerManager.ts
ptesavol: I would agree with putting the locking functionality in same place. @juslesan opinions about this?

packages/dht/src/dht/PeerManager.ts
ptesavol: "Active contacts" is a term from the kbucket implementation that is not strictly synonymous with neighbors. I think at some point we should make a list of all the vague terms, and have a meeting about renaming them.

packages/dht/src/dht/PeerManager.ts
ptesavol: yes we know, because this event handler only fires if there are too many contacts in a k-bucket and we need to choose which ones to keep and which one to discard

packages/dht/src/dht/PeerManager.ts
ptesavol: done

packages/dht/src/dht/PeerManager.ts
ptesavol: moved to PeerManagerConfig

packages/dht/src/dht/PeerManager.ts
teogeb: Is the `PeerManager` managing just peers, or also contacts? (if I understood each peer is a contact, but not vice versa)

packages/dht/src/dht/contact/SortedContactList.ts
teogeb: If `maxSize` would be `0`, it would pass this if statement? That would be incorrect, as we want to execute this branch only if `this.contactIds.length` is smaller than the limit.

(The value can't be `null` if we trust the type system, and we of course should trust it.)

packages/dht/src/dht/PeerManager.ts
ptesavol: it is managing all Peer-related, also contacts, because contacts are potential Peers

packages/dht/src/dht/IPeerManager.ts
teogeb: Maybe good to add a comment about that fact, and `TODO` comment so that we can remove the interface when the development of this feature has been completed.

packages/test-utils/src/testInterface.ts
teogeb: Adding the prefix would communicate exactly that: i.e. that the variable is not unused even when IDE/`eslint` doesn't find any usages for it.

packages/dht/src/dht/contact/SortedContactList.ts
ptesavol: In general my principle is that "only compare something with undefined if there is some good reason for doing so, otherwise always test for falsiness". From my experience following this principle helps to avoid a lot of hard-to-debug errors at runtime. In this case, would you say there is some good reason for explicitly testing for equality with undefined?

packages/dht/src/dht/IPeerManager.ts
teogeb: We currently seem to use both `handleX` and `onX` naming conventions for event observers. Currently most of the classes in `dht` use the onX naming convention (`DhtNode`, `DiscoverySession`, `RecursiveFindSession`, `ConnectionManager`, `ManagedConnection`, `NodeWebRtcConnection`, `WebRtcConnector`) and in some classes the `handleX` convention is also used in one or two methods.

I'd prefer that we use the `onX` convention here and maybe migrate other usages to that convention in a separate PR later. (Or if we want to choose the other naming convention we can migrate all other use cases to that. But good to have just one naming convention.)

The `peer` part in the names is anyway redundant.

packages/dht/src/dht/PeerManager.ts
teogeb: Maybe the class should be renamed to `ContactManager` then. The methods also imply that the class handles contacts in general, not just peers:
```
    newContact: (peerDescriptor: PeerDescriptor, closestPeers: PeerDescriptor[]) => void
    contactRemoved: (peerDescriptor: PeerDescriptor, closestPeers: PeerDescriptor[]) => void
    newKbucketContact: (peerDescriptor: PeerDescriptor, closestPeers: PeerDescriptor[]) => void
    kbucketContactRemoved: (peerDescriptor: PeerDescriptor) => void
    newOpenInternetContact: (peerDescriptor: PeerDescriptor, closestPeers: PeerDescriptor[]) => void
    openInternetContactRemoved: (peerDescriptor: PeerDescriptor, closestPeers: PeerDescriptor[]) => void
    newRandomContact: (peerDescriptor: PeerDescriptor, closestPeers: PeerDescriptor[]) => void
    randomContactRemoved: (peerDescriptor: PeerDescriptor, closestPeers: PeerDescriptor[]) => void
```

packages/dht/src/dht/PeerManager.ts
juslesan: Made a ticket https://linear.app/streamr/issue/NET-1141/kbucket-on-updated

packages/dht/src/dht/PeerManager.ts
juslesan: Seems that the PeerManager is only calling `weakLockConnection` and `weakUnlockConnection` from the ConnectionManager. The PeerManager could only simply take the `ConnectionLocker`-interface as a parameter since other ConnectionManager logic seems to not be relevant here.

packages/dht/src/dht/PeerManager.ts
ptesavol: Well, ContactManager sounds like a synonym of AddressBook, which the class is certainly not about. I'd leave the name as-is.

packages/dht/src/dht/IPeerManager.ts
ptesavol: The logic behind using the name prefixes "on" and "handle" are as follows: methods starting with "on" are event listeners that get invoked by eventemitter3, and they should not be called directly. Methods starting with "handle" are event handlers that can be called directly.

packages/dht/src/dht/PeerManager.ts
ptesavol: Good idea, I'll switch to passing the interface

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,320 @@
...
+}
+
+interface IPeerManagerTest {
+    getNeighborList: () => SortedContactList<RemoteDhtNode>
+    getKBucketSize: () => number
+    getKBucketPeers: () => PeerDescriptor[]
+}
+
+export class PeerManager extends EventEmitter<PeerManagerEvents> implements IPeerManager {
+    private _jee: number = 0
```
teogeb, line 42:
```suggestion
```

packages/dht/src/dht/PeerManager.ts
teogeb: As the term `PeerManager`/`ContactManager` is quite abstract, maybe it would makes sense to write a short comment to the class which describes what the core functionality of this class is.

One option is to revisit the naming question when we have the glossary (https://linear.app/streamr/issue/NET-1128/add-to-internal-docs-a-glossary-of-network-level-terms), and therefore also have clear definitions for each name.

packages/dht/src/dht/IPeerManager.ts
teogeb: Why a method should care who calls it? If it is a public method of `PeerManager`, any external component can call it. The `onFooChanged` pattern would clearly communicate that the external component should call this method whenever it changes the value of foo.

packages/dht/src/dht/contact/SortedContactList.ts
teogeb: If `maxSize` is be 0, the `addContact` should not do anything, right? I.e. it should not execute lines 46-48 in that case.  But as `!0` is `true` at line 45, it actually executes the line and adds a contact to the list.

The meaning of 0 is not always very clear if a variable has type `number | undefined`. Therefore I like to be very explicit about it when doing the comparisons:
- `if ((foo !== undefined) || (foo === 0))` tells that both `undefined` and `0` are the exceptional case
-  `if (foo !== undefined)` tells that only undefined is the exceptional case, `0` is normal valid value
- `const maxRetries = config.maxRetries ?? 123` tells that `0` is normal valid value

On the other hand these lines don't tell that information, and are buggy if `0` is actually just a normal valid value:
- `if (foo)`
- `const maxRetries = config.maxRetries || 123`

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,320 @@
...
+            this.connections.set(PeerID.fromValue(RemoteDhtNode.id).toKey(), RemoteDhtNode)
+        } else {
+            logger.trace('new connection not set to connections, there is already a connection with the peer ID')
+        }
+    }
+
+    public handleDisconnected(peerDescriptor: PeerDescriptor, disconnectionType: DisconnectionType): void {
+        this.connections.delete(keyFromPeerDescriptor(peerDescriptor))
+        
+        // only remove from bucket if we are on layer 0
```
teogeb, line 182: Maybe a comment _why_ it is removed only if we are on layer 0? Also the layer seems to affect also whether we call `this.removeContact(peerDescriptor, true)`

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,320 @@
...
+    getKBucketSize: () => number
+    getKBucketPeers: () => PeerDescriptor[]
+}
+
+export class PeerManager extends EventEmitter<PeerManagerEvents> implements IPeerManager {
+    private _jee: number = 0
+    private bucket?: KBucket<RemoteDhtNode>
+    private neighborList?: SortedContactList<RemoteDhtNode>
+    private openInternetPeers?: SortedContactList<RemoteDhtNode>
+    private randomPeers?: RandomContactList<RemoteDhtNode>
```
teogeb: All these fields are initialized in the constructor (as it calls `initKBuckets()`), and won't be `undefined`.
```suggestion
    private bucket: KBucket<RemoteDhtNode>
    private neighborList: SortedContactList<RemoteDhtNode>
    private openInternetPeers: SortedContactList<RemoteDhtNode>
    private randomPeers: RandomContactList<RemoteDhtNode>
```

packages/dht/src/dht/DhtNode.ts
teogeb: The same issue seems to be in `PeerManager#initKBuckets` (should we even inline that method as it does other things than just init the KBuckets?)

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,320 @@
...
+                            'newKbucketContact',
+                            contact.getPeerDescriptor(),
+                            this.neighborList!.getClosestContacts(this.config.getClosestContactsLimit).map((peer) => peer.getPeerDescriptor())
+                        )
+                    } else {
+                        this.config.connectionManager?.weakUnlockConnection(contact.getPeerDescriptor())
+                        this.removeContact(contact.getPeerDescriptor())
+                        this.addClosestContactToBucket()
+                    }
+                    return
```
teogeb, line 137:
```suggestion
```

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,320 @@
...
+
+    public handleConnected(peerDescriptor: PeerDescriptor): void {
+        if (this.config.ownPeerId!.equals(PeerID.fromValue(peerDescriptor.kademliaId))) {
+            logger.error('handleConnected() to self')
+            return
+        }
+
+        const RemoteDhtNode = this.config.createRemoteDhtNode(peerDescriptor)
+        if (!this.connections.has(PeerID.fromValue(RemoteDhtNode.id).toKey())) {
+            this.connections.set(PeerID.fromValue(RemoteDhtNode.id).toKey(), RemoteDhtNode)
```
teogeb:
```suggestion
        const remoteDhtNode = this.config.createRemoteDhtNode(peerDescriptor)
        if (!this.connections.has(PeerID.fromValue(remoteDhtNode.id).toKey())) {
            this.connections.set(PeerID.fromValue(remoteDhtNode.id).toKey(), remoteDhtNode)
```

packages/dht/src/connection/Simulator/Simulator.ts
```diff
@@ -400,4 +400,53 @@ export class Simulator extends EventEmitter<ConnectionSourceEvents> {
...
+                    if (!ret.get(sourceKey)!.includes(targetKey)) {
+                        ret.get(sourceKey)!.push(targetKey)
+                    }
+                }
+            }
+        })
+        return ret
+    }
+
+    public getNumberOdConnectedComponents(): number {
```
teogeb:
```suggestion
    public getNumberOfConnectedComponents(): number {
```

teogeb: Which component uses this method? (Or was it used earlier, but is now redundant?)

packages/dht/src/connection/Simulator/Simulator.ts
```diff
@@ -400,4 +400,53 @@ export class Simulator extends EventEmitter<ConnectionSourceEvents> {
...
+        const graph = this.getAssociationsAsUndirectedGraph()
+        const visited: Map<string, boolean> = new Map()
```
teogeb:
```suggestion
        const visited: Set<PeerIDKey> = new Set()
```

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,320 @@
...
+        this.randomPeers.on('newContact', (newContact: RemoteDhtNode, closestContacts: RemoteDhtNode[]) =>
+            this.emit('newRandomContact', newContact.getPeerDescriptor(), closestContacts.map((peer) => peer.getPeerDescriptor()))
+        )
+    }
+
+    private onKBucketPing(oldContacts: RemoteDhtNode[], newContact: RemoteDhtNode): void {
+        if (this.stopped) {
+            return
+        }
+        const sortingList: SortedContactList<RemoteDhtNode> = new SortedContactList(this.config.ownPeerId!, 100)
```
teogeb: `this.config.ownPeerId` is never `undefined`

```suggestion
        const sortingList: SortedContactList<RemoteDhtNode> = new SortedContactList(this.config.ownPeerId, 100)
```

(also 4 other places in this file)

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,320 @@
...
+                contact.ping().then((result) => {
+                    if (result) {
+                        logger.trace(`Added new contact ${contact.getPeerId().value.toString()}`)
+                        this.emit(
+                            'newKbucketContact',
+                            contact.getPeerDescriptor(),
+                            this.neighborList!.getClosestContacts(this.config.getClosestContactsLimit).map((peer) => peer.getPeerDescriptor())
+                        )
+                    } else {
+                        this.config.connectionManager?.weakUnlockConnection(contact.getPeerDescriptor())
```
teogeb, line 133: It is never `undefined`

```suggestion
                        this.config.connectionManager.weakUnlockConnection(contact.getPeerDescriptor())
```
(also 4 other places in this file)

packages/dht/src/dht/PeerManager.ts
```diff
@@ -0,0 +1,320 @@
...
+                        this.removeContact(contact.getPeerDescriptor())
+                        this.addClosestContactToBucket()
+                    }
+                    return
+                }).catch((_e) => {
+                    this.config.connectionManager?.weakUnlockConnection(contact.getPeerDescriptor())
+                    this.removeContact(contact.getPeerDescriptor())
+                    this.addClosestContactToBucket()
+                })
```
teogeb: As the case for `false` result and promise error are the same, the logic could be simplified like this:

```suggestion
                setImmediate(async () => {
                    let result
                    try {
                        result = await contact.ping()
                    } catch {
                        result = false
                    }
                    if (result) {
                        logger.trace(`Added new contact ${contact.getPeerId().value.toString()}`)
                        this.emit(
                            'newKbucketContact',
                            contact.getPeerDescriptor(),
                            this.neighborList!.getClosestContacts(this.config.getClosestContactsLimit).map((peer) => peer.getPeerDescriptor())
                        )
                    } else {
                        this.config.connectionManager?.weakUnlockConnection(contact.getPeerDescriptor())
                        this.removeContact(contact.getPeerDescriptor())
                        this.addClosestContactToBucket()
                    }
                })```

File: packages/dht/src/dht/PeerManager.ts [hunk 35]
```diff
@@ -0,0 +1,320 @@
...
+                logger.trace(`Adding new contact ${contact.kademliaId.toString()}`)
+                const RemoteDhtNode = this.config.createRemoteDhtNode(contact)
+                if (!this.bucket!.get(contact.kademliaId) && !this.neighborList!.getContact(peerIdFromPeerDescriptor(contact))) {
+                    this.neighborList!.addContact(RemoteDhtNode)
+                    if (contact.openInternet) {
+                        this.openInternetPeers!.addContact(RemoteDhtNode)
+                    }
+                    this.bucket!.add(RemoteDhtNode)
+                } else {
+                    this.randomPeers!.addContact(RemoteDhtNode)
```
teogeb:
```suggestion
                const remoteDhtNode = this.config.createRemoteDhtNode(contact)
                if (!this.bucket!.get(contact.kademliaId) && !this.neighborList!.getContact(peerIdFromPeerDescriptor(contact))) {
                    this.neighborList!.addContact(remoteDhtNode)
                    if (contact.openInternet) {
                        this.openInternetPeers!.addContact(remoteDhtNode)
                    }
                    this.bucket!.add(remoteDhtNode)
                } else {
                    this.randomPeers!.addContact(remoteDhtNode)
```

packages/dht/src/dht/discovery/DiscoverySession.ts
```diff
@@ -97,16 +83,18 @@ export class DiscoverySession {
...
 
     private findMoreContacts(): void {
         if (this.stopped) {
             return
         }
-        const uncontacted = this.config.neighborList.getUncontactedContacts(this.config.parallelism)
+        const uncontacted = this.config.peerManager.getClosestPeersTo(this.config.targetId, this.config.parallelism, this.contactedPeers)
+
         if (uncontacted.length < 1 || this.noProgressCounter >= this.config.noProgressLimit) {
+            logger.trace( this.config.nodeName + ' discoveryCompleted in findMoreContacts, path: ' + this.joiningPath.join(', '))
```
teogeb: Use node id instead of `nodeName`, see https://github.com/streamr-dev/network/pull/1967

(also at line 133, and remove the field from `DiscoverySessionConfig`)

packages/dht/src/dht/discovery/DiscoverySession.ts
```diff
@@ -37,33 +30,18 @@ export class DiscoverySession {
     private noProgressCounter = 0
     private ongoingClosestPeersRequests: Set<string> = new Set()
     private readonly config: DiscoverySessionConfig
-    private readonly ownPeerId: PeerID
+    private contactedPeers: Set<PeerIDKey> = new Set()
+    private joiningPath: Array<string> = []
```
teogeb: Do we add items to this field somewhere? Seems that it is always empty?

packages/dht/src/dht/discovery/DiscoverySession.ts
```diff
@@ -117,32 +105,32 @@ export class DiscoverySession {
             }
             this.ongoingClosestPeersRequests.add(nextPeer.getPeerId().toKey())
             // eslint-disable-next-line promise/catch-or-return
-            this.getClosestPeersFromContact(nextPeer)
-                .then((contacts) => this.onClosestPeersRequestSucceeded(nextPeer.getPeerId(), contacts))
-                .catch((err) => this.onClosestPeersRequestFailed(nextPeer, err))
+            this.getClosestPeersFromContact(nextPeer!)
+                .then((contacts) => {
```
teogeb:
```suggestion
                // eslint-disable-next-line promise/always-return
                .then((contacts) => {
```

packages/dht/src/dht/discovery/DiscoverySession.ts
```diff
@@ -117,32 +105,32 @@ export class DiscoverySession {
...
         }
     }
 
-    public async findClosestNodes(timeout: number): Promise<SortedContactList<RemoteDhtNode>> {
-        if (this.config.neighborList.getUncontactedContacts(this.config.parallelism).length < 1) {
-            logger.trace('getUncontactedContacts length was 0 in beginning of discovery, this.neighborList.size: '
-                + this.config.neighborList.getSize())
-            return this.config.neighborList
+    public async findClosestNodes(timeout: number): Promise<void> {
+        if (this.config.peerManager.getNumberOfPeers(this.contactedPeers) < 1) {
```
teogeb:
```suggestion
        if (this.config.peerManager.getNumberOfPeers(this.contactedPeers) === 0) {
```

packages/dht/src/dht/discovery/PeerDiscovery.ts
```diff
@@ -139,11 +119,10 @@ export class PeerDiscovery {
...
         }
-        await Promise.allSettled(this.config.bucket.closest(this.config.ownPeerId.value, this.config.parallelism).map(async (peer: RemoteDhtNode) => {
-            const contacts = await peer.getClosestPeers(this.config.ownPeerDescriptor.kademliaId)
-            contacts.forEach((contact) => {
-                this.config.addContact(contact)
-            })
+        await Promise.allSettled(this.config.peerManager.getClosestPeersTo(
+            this.config.ownPeerDescriptor.kademliaId, this.config.parallelism).map(async (peer: RemoteDhtNode) => {
+            const contacts = await peer.getClosestPeers(this.config.ownPeerDescriptor.kademliaId!)
+            this.config.peerManager.handleNewPeers(contacts)
```
teogeb: Just an indentation fix

```suggestion
        await Promise.allSettled(
            this.config.peerManager.getClosestPeersTo(
                this.config.ownPeerDescriptor.kademliaId,
                this.config.parallelism
            ).map(async (peer: RemoteDhtNode) => {
                const contacts = await peer.getClosestPeers(this.config.ownPeerDescriptor.kademliaId!)
                this.config.peerManager.handleNewPeers(contacts)
            })
        )
```

packages/dht/src/dht/discovery/PeerDiscovery.ts
```diff
@@ -47,7 +36,7 @@ export class PeerDiscovery {
         this.abortController = new AbortController()
     }
 
-    async joinDht(entryPointDescriptor: PeerDescriptor, doRandomJoin = true, retry = true): Promise<void> {
+    async joinDht(entryPointDescriptor: PeerDescriptor, doRandomJoin = true, _retry = true): Promise<void> {
```
teogeb: The unused parameter can be removed

ptesavol:
> There seem to be some changes to the behaviour of the DhtNode that seem to cause some new tests to fail especially in the network and client

These problems have been resolved

packages/dht/src/dht/discovery/DiscoverySession.ts
ptesavol: done

packages/dht/src/dht/discovery/DiscoverySession.ts
ptesavol: removed file

packages/dht/src/dht/PeerManager.ts
ptesavol: Made a point in future improvements about naming

packages/dht/src/dht/DhtNode.ts
```diff
@@ -246,37 +255,39 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
...
             localPeerDescriptor: this.localPeerDescriptor!,
-            bucket: this.bucket!,
-            neighborList: this.neighborList!,
             joinNoProgressLimit: this.config.joinNoProgressLimit,
             peerDiscoveryQueryBatchSize: this.config.peerDiscoveryQueryBatchSize,
             joinTimeout: this.config.dhtJoinTimeout,
             serviceId: this.config.serviceId,
             parallelism: this.config.joinParallelism,
-            addContact: this.addNewContact.bind(this),
-            connectionManager: this.connectionManager,
```
teogeb, line 261: We should pass `connectionManager` to `PeerDiscovery`

teogeb:
Closing this PR as functionality changes have been cherry-picked to other PRs. The `getTestInterface()` can be crerry-picked from this later if needed.

PR #1806, opened 2023-08-29
refactor(trackerless-network): Type-safe node ID

teogeb:
We now use `NodeID` annotation instead of `string` in all places where `trackerless-network` operates on node ids. 

Also unified naming: in `trackerless-network` we handle _nodes_, not _peers_ (that is a `dht` level concept).
- An exception to that rule is the `PeerDescriptor`: didn't rename that to `NodeDescriptor` as it refers to the `dht` level entity

## Changes

- Introduced `getNodeIdFromPeerDescriptor` helper method which is just a call to `keyFromPeerDescriptor` but casts the return type to `NodeID`
- Menamed some classes, methods, fields and variables from `peer*` to `node*`, e.g.:
  - class `PeerList` -> `NodeList` (and some methods of the class)
  - methods of `ProxyStreamConnectionClient`, `ProxyStreamConnectionServer` and `TemporaryConnectionServer`
- Renamed all `stringId` -> `nodeId`
  - remove obsolete `NetworkNode#getNodeStringId `and `StreamrNode#getNodeStringId`
- Also using `NodeID` annotation in `client` in `broker` (at least most of the places)
 
## Notes

- `NodeIDs` are actually byte arrays, but in practice it is convenient that we handle them as strings (e.g. to use as map keys). In that sense it is very similar to e.g. how we handle `EthereumAddress` objects.

## Future improvements

- Use `bytes` type for node id in the protobuf definitions (as we decided that it is just a hex-representation of byte array)
- Maybe we don't need `PeerID` class in `trackerless-network` node if get the node id as string in that level
- The contact term is also from `dht` level, maybe should use `node` or `neighbor` instead?
- Maybe could simplify some naming, e.g. `targetNeighbor` could be just a `neighbor`? 
- Maybe could shorten the `randomGraphId` to `graphId` (or maybe a separate term, which tells that this is a the id in the stream delivery network: e.g. `streamrDeliveryId` or `deliveryId`?)


juslesan (approved):

packages/client/test/end-to-end/ProxyPublishSubscribe.test.ts
```diff
@@ -61,8 +61,8 @@ describe('PubSub with proxy connections', () => {
         const proxyUser1 = await proxyClient1.getAddress()
         const proxyUser2 = await proxyClient2.getAddress()
 
-        proxyPeerKey1 = await proxyClient1.getNodeId()
-        proxyPeerKey2 = await proxyClient2.getNodeId()
+        proxyPeerKey1 = await proxyClient1.getNodeId() as NodeID
+        proxyPeerKey2 = await proxyClient2.getNodeId() as NodeID
```
harbu: Shouldn't the result of `getNodeId()` already be a `NodeID`, i.e. why re-cast here?

packages/client/test/integration/Subscriber.test.ts
```diff
@@ -47,7 +48,7 @@ describe('Subscriber', () => {
     
             const sub = await subscriber.subscribe(stream.id)
     
-            const publisherNode = environment.startNode(publisherWallet.address)
+            const publisherNode = environment.startNode(publisherWallet.address as NodeID)
```
harbu, line 51: With NodeID no longer being (necessarily) an Ethereum address, would it be better for the tests to reflect this fact? Right now the tests may give the impression that Wallet has something to do with NodeID while in reality they are separate concepts.

packages/trackerless-network/src/NetworkNode.ts
```diff
@@ -116,20 +117,20 @@ export class NetworkNode {
 
     getNeighborsForStreamPart(streamPartId: StreamPartID): ReadonlyArray<string> {
```
harbu: Should this be?
```suggestion
    getNeighborsForStreamPart(streamPartId: StreamPartID): ReadonlyArray<NodeID> {
```

harbu (approved):

packages/client/test/end-to-end/ProxyPublishSubscribe.test.ts
harbu: Maybe `Client#getNodeId` should return `NodeID` if it isn't already.

teogeb: Changed the return type

packages/client/test/integration/Subscriber.test.ts
teogeb: I'll make a separate PR about this

PR #1810, opened 2023-09-01
Updated partition docs

yaruno:
## Summary

Expanded the partition docs to cover partition key selection behaviour when multiple partitions exist but no partition key is given for publishing.

### DIFF
--- a/docs/docs/usage/streams/partitioning.md
+++ b/docs/docs/usage/streams/partitioning.md
@@ -46,7 +46,7 @@ console.log(
 ## Publish to partitioned streams
 In most use cases, a user wants **related messages (e.g. messages from a particular device) to be assigned to the same partition**, so that the messages retain a deterministic order and reach the same subscriber(s) to allow them to compute stateful aggregates correctly.
 
-If no partition is specified, the **data goes to partition 0 by default**.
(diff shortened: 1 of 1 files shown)


docs/docs/usage/streams/partitioning.md
```diff
@@ -46,7 +46,7 @@ console.log(
 ## Publish to partitioned streams
 In most use cases, a user wants **related messages (e.g. messages from a particular device) to be assigned to the same partition**, so that the messages retain a deterministic order and reach the same subscriber(s) to allow them to compute stateful aggregates correctly.
 
-If no partition is specified, the **data goes to partition 0 by default**.
+If stream has been setup to have only one partition and no partition key is specified while publishing, the **data goes to partition 0 by default**. However, if stream has been set to have multiple partitions and no partition key is given the partition will be selected randomly from the range between 0 and partition count minus 1 .
```
teogeb, line 49: Isn't it obvious that if is only one partition, the data is published to that partition? (and it goes there also if partition key is used)

yaruno: For us yes, but for a random user not necessarily. Also, the starting number of partitions is not given knowlegde, one could expect it to start from partition 0 or from partition 1, if not explicitly defined in documentation.

yaruno: moved to https://github.com/streamr-dev/network/pull/1815

PR #1811, opened 2023-09-01
clean: remove extraneous package.json in dist folders

harbu:
## Summary

Remove extraneous package.json files in dist folders. The only package that requires the package.json to be in the dist folder is the client package. Also gets rid of jest warning:

```
jest-haste-map: Haste module naming collision: streamr-broker
  The following files share their name; please adjust your hasteImpl:
    * <rootDir>/package.json
    * <rootDir>/dist/package.json
```


harbu:
Right. It needs to exist in `dist` because it is used in code.

PR #1813, opened 2023-09-01
NET-1054: fetch `redundancyFactor` from operator contract

harbu:
## Summary

Fetch `redundancyFactor` from operator contract (instead of config file).

## Limitations / Future improvements

- The operator contract metadata can be updated at any time and thus the value of `redundancyFactor` can change at any time. The broker will fetch the value just at the beginning in this implementation. A more sophisticated implementation would additionally react to changes in the smart contract.


packages/broker/src/plugins/operator/OperatorPlugin.ts
```diff
@@ -83,9 +83,10 @@ export class OperatorPlugin extends Plugin<OperatorPluginConfig> {
     }
 
     async start(): Promise<void> {
+        const redundancyFactor = await fetchRedundancyFactor(this.serviceConfig)
         this.maintainTopologyService = await setUpAndStartMaintainTopologyService({
             streamrClient: this.streamrClient,
-            redundancyFactor: this.pluginConfig.redundancyFactor,
+            redundancyFactor: redundancyFactor,
```
teogeb:
```suggestion
            redundancyFactor,
```

packages/broker/src/plugins/operator/fetchRedundancyFactor.ts
```diff
@@ -0,0 +1,24 @@
...
+    let metadata: Record<string, unknown> = {}
+    if (metadataAsString.length > 0) {
+        try {
+            metadata = JSON.parse(metadataAsString)
+        } catch (e) {
+            logger.warn('Encountered malformed metadata', { operatorContractAddress, metadataAsString })
+        }
+    }
+    const redundancyFactor = Number(metadata.redundancyFactor)
+    return !isNaN(redundancyFactor) ? Math.max(redundancyFactor, 1) : 1
```
teogeb: Would it make sense that we'd throw if there is a invalid data? E.g. if a user configures redundancy factor to `50` meaning 50% redundancy, it would be good to throw show an error/warning?

harbu: Hmm.. good question. It actually got me thinking and to write this test https://github.com/streamr-dev/network/pull/1820.

The redundancy factor here isn't a percentage but number of nodes to replicate to.

jtakalai (approved):

harbu:
@teogeb Made the changes we discussed yesterday.

packages/broker/src/plugins/operator/fetchRedundancyFactor.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    if (metadataAsString.length === 0) {
+        return 1
+    }
+
+    let metadata: Record<string, unknown>
+    try {
+        metadata = JSON.parse(metadataAsString)
+    } catch {
+        logger.warn('Encountered malformed metadata', { operatorContractAddress, metadataAsString })
```
teogeb, line 30: Why we both log and throw? Maybe the caller should catch the error, and logging info about the error is just one of the actions the caller does in that case.

harbu: I want a log message with details (i.e. { operatorContractAddress, metadataAsString }) and a proper description. Doing so at a higher level would require the caller to know about details that aren't their responsibility.

One option could be to have a special error type for this?
```ts
class ErrorWithLogMessage extends Error {
    private readonly message: string
    private readonly context: Record<string, unknown>
    private readonly logLevel: LogLevel
    
    construtor(message: string, context: Record<string, unknown>, logLevel: LogLevel) {
        // set values
    }
}
```

harbu: ping @teogeb

harbu: Let's just return `undefined`.

teogeb (approved):

PR #1816, opened 2023-09-01
NET-1034: expose `Signer` of client and use in operator plugin

harbu:
## Summary

Add method `getSigner` to streamr-client public API. Use this method in operator plugin to fetch the appropriate signer.


packages/client/src/StreamrClient.ts
```diff
@@ -546,6 +547,13 @@ export class StreamrClient {
     // Authentication
     // --------------------------------------------------------------------------------------------
 
+    /**
+     * Gets the Signer associated with the current {@link StreamrClient} instance.
+     */
+    getSigner(): Promise<Signer> {
```
teogeb, line 553: It is not self-explanatory that this will return a signer that works only in the sidechain. 

Should this have a parameter about the chain? If we may want to do signing in main chain or in side chain in production? Or if we don't really need support for signing in the main chain, should we rename this e.g. to `getSidechainSigner()`?

harbu: The mainchain stuff is going away in 1.0 since we no longer have trackers and hence have no need for tracker registries. So we will only have "sidechain" as a single chain. So in that sense there is no need for distinction.

harbu: Which probably means we should get rid of the sidechain / mainchain distinction elsewhere in the code including the public API configs?

teogeb (approved):

PR #1817, opened 2023-09-01
test(client): [NET-1022]: Fix flaky `multiple-clients` test

teogeb:
The test was flaky, because we have concurrency bug (or missing feature) in encryption handling: if two clients open the local DB at the same time, it is possible that one of the client fails to open the DB as it has not "migrated" yet (i.e. the tables haven't been created). There is a separate ticket about this bug NET-1057, and in this PR we added a new test which focuses to this bug. The new test is not flaky as it always fails (and therefore it is currently skipped). 

The actual flaky test (`multiple-clients.test.ts`) has been resolved by using separate private keys for each subscriber. The feature under test here is more abstract, and the private keys of the subscribers are not relevant to this test. The main idea in the flaky test is to check whether there can be multiple publishers on a single stream. It could be implement with just with one subscriber, or like we do now, with two subscribers having separate private keys.

### DIFF
--- a/packages/client/test/unit/ServerPersistence.test.ts
+++ b/packages/client/test/unit/ServerPersistence.test.ts
@@ -1,5 +1,7 @@
-import { Database } from 'sqlite'
 import { randomEthereumAddress } from '@streamr/test-utils'
+import range from 'lodash/range'
+import { join } from 'path'
+import { Database } from 'sqlite'
 import ServerPersistence from '../../src/utils/persistence/ServerPersistence'
 import { mockLoggerFactory } from '../test-utils/utils'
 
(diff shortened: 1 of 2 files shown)


packages/client/test/unit/ServerPersistence.test.ts
```diff
@@ -49,4 +51,22 @@ describe('ServerPersistence', () => {
...
+    // enable when NET-1057 done
+    it.skip('concurrency', async () => {
+        const instanceCount = 10
+        const clientId = randomEthereumAddress()
+        const values = await Promise.all(range(instanceCount).map(async (i: number) => {
+            const instance = await ServerPersistence.createInstance({
+                loggerFactory: mockLoggerFactory(),
+                clientId,
+                namespaces: [NAMESPACE],
+                migrationsPath: join(__dirname, '../../src/encryption/migrations')
```
harbu: Migration files and `NAMESPCE` mismatch

teogeb: Not relevant in this test, but better to unify

harbu (approved):

PR #1824, opened 2023-09-05
refactor(dht): Remove unused fields

teogeb:
Refactored `ExternalApi` to `registerExternalApiRpcMethod` function and removed obsolete `DhtNode#externalApi` field. 
- also simplified findData by removing the obsolete `context` parameter

Removed `RemoteConnectionLocker#peerId`
- the implementation doesn’t need to store `peerId` field as it stores `ownPeerDescriptor`

Added link to issue for unused `NodeWebRtcConnection#bufferThresholdHigh`l


packages/dht/src/dht/registerExternalApiRpcMethod.ts
```diff
@@ -0,0 +1,26 @@
+import { DhtNode } from '../exports'
+import { FindDataRequest, FindDataResponse, FindMode } from '../proto/packages/dht/protos/DhtRpc'
+
+export const registerExternalApiRpcMethod = (thisNode: DhtNode): void => {
```
juslesan, line 4: I would rather have the method name as plural even if there is only a single function to register for now.

juslesan (approved):

PR #1830, opened 2023-09-05
refactor(trackerless-network): User id as Ethereum address

teogeb:
Handle the user id field similarly to other id fields: internally it is a string, so that we can use it e.g. as a map key. Before this PR it was stored as a byte array.

The protocol doesn't change: there it is transferred as a byte array, similarly to `NodeID` and others.

The user id is now annotated as `EthereumAddress` as it is that kind of string in practice. It would be possible to have it just as a branded string if we'd want to be more abstract about the content of the strings. (At `trackerless-network` level we don't care what kind of bytes it contains, it is just an ID of a user).


harbu (commented):
## If we want to keep the userId concept:

Let's take this even further.

We should have the following additional functions to keep the trackerless-network "from knowing" about Ethereum addresses:

```ts
export const toUserId = toEthereumAddress

export const binaryToUserId = (x) => toEthereumAddress(binaryToHex(x))
```

Of course the trackerless-network will still know about EthereumAddress and hex-format but it will be easy-to-change in future if we wanted due to the relationship between userId and Ethereum address being specificed in only a couple of places.

## If we don't want to keep the userId concept:

Remove it altogether. Just use `EthereumAddress` everywhere and forget about `UserID`.

harbu (approved):

PR #1833, opened 2023-09-06
feat(network): External ip config

juslesan:
## Summary

Add configuration option for external IP as in the old network.


packages/dht/src/connection/WebRTC/ParsedLocalCandidate.ts
```diff
@@ -0,0 +1,39 @@
+export enum CandidateType {
+    HOST = 'host',
+    SRFLX = 'srflx',
+    PRFLX = 'prflx',
+    RELAY = 'relay'
+}
+
+export class ParsedLocalCandidate {
```
harbu: I think a class here is a bit overblown since what we want to achieve is to replace the IP address when the candidateType is `host`?

Something like
```ts
replaceIpIfCandidateTypeIsHost(candidate: string, replacementIp: string): string {
    if (candidate.endsWith('host')) {
        const [_ip, ...rest] = candidate.split(' ')
        return [replacementIp, ...rest].join(' ')
    }
    return candidate
}
```

packages/dht/src/connection/WebRTC/WebRtcConnector.ts
```diff
@@ -128,7 +130,15 @@ export class WebRtcConnector implements IWebRtcConnectorService {
...
-            remoteConnector.sendIceCandidate(this.ownPeerDescriptor!, candidate, mid, connection.connectionId.toString())
+            const parsedCandidate = new ParsedLocalCandidate(candidate)
+            if (this.config.externalIp && parsedCandidate.getType() === CandidateType.HOST) {
+                parsedCandidate.setIp(this.config.externalIp)
+                const injectedCandidate = parsedCandidate.toString()
+                logger.error(`onLocalCandidate injected external ip ${injectedCandidate} ${mid}`)
+                remoteConnector.sendIceCandidate(this.ownPeerDescriptor!, injectedCandidate, mid, connection.connectionId.toString())
+            } else {
+                remoteConnector.sendIceCandidate(this.ownPeerDescriptor!, candidate, mid, connection.connectionId.toString())
+            }
```
harbu: Code repetition here, both if and else statements should end with the same call to (with only the 2nd argument varying) `remoteConnector.sendIceCandidate`.

harbu (changes requested):

harbu (approved):

packages/dht/src/connection/WebRTC/WebRtcConnector.ts
```diff
@@ -28,6 +28,15 @@ import { PortRange } from '../ConnectionManager'
 
 const logger = new Logger(module)
 
+const replaceIpIfCandidateTypeIsHost = (candidate: string, ip: string) => {
+    const parsed = candidate.split(' ')
+    const type = parsed[7]
```
teogeb, line 33: What are these magic number 7 and 4 in this function? (maybe could provide the example string content of `candidate` string which we are splitting here)

packages/dht/src/connection/WebRTC/WebRtcConnector.ts
```diff
@@ -28,6 +28,15 @@ import { PortRange } from '../ConnectionManager'
 
 const logger = new Logger(module)
 
+const replaceIpIfCandidateTypeIsHost = (candidate: string, ip: string) => {
```
teogeb: What it means if the type is `host`? E.g. that we have an IP address, but it is an internal IP address?

If that's the case, maybe this method could be called `withExternalpAdress`?

packages/dht/src/connection/WebRTC/WebRtcConnector.ts
harbu: Hmm.. actually the unit test for this got lost at some point in the PR (https://github.com/streamr-dev/network/pull/1833/commits/ba230d1aa9639499fb73a464ecf8f282d719ee1a). Can we return that unit test @juslesan? (it also acted a good example as well)

packages/dht/src/connection/WebRTC/WebRtcConnector.ts
juslesan: When ICE connectivity checks are started, the peers will find any IP addresses that the host machine has on its interfaces. The IceCandidates created with these IP addresses always have the type host.

teogeb: If we use terminology of internal/external:
- the "IP addresses that the host machine has on its interface" are "internal" IPs?
- and in this function we rename internal IP with an external IP?

packages/dht/test/unit/replaceIpIfCandidateTypeIsHost.test.ts
```diff
@@ -0,0 +1,18 @@
+import { replaceIpIfCandidateTypeIsHost } from '../../src/connection/WebRTC/WebRtcConnector'
+
+describe('replaceIpIfCandidateTypeIsHost', () => {
+
+    const hostCandidate = 'candidate:1 1 UDP 2013266431 127.0.0.1 30000 typ host'
+    const srflxCandidate = 'candidate:1 1 UDP 2013266431 127.0.0.1 30000 typ srflx'
```
teogeb: Is this a typical external IP? as it is pointing to localhost 127.0.0.1

PR #1834, opened 2023-09-06
refactor(broker): Refactor `MaintainOperatorPoolValue` service

teogeb:
Extracted the main functionality of `MaintainOperatorPoolValueService` to separate function. The scheduler is run inside `OperatorPlugin`. This is similar change than what was implemented for other services in https://github.com/streamr-dev/network/pull/1843.

The test executes the method instead of polling the state modified by the scheduler.
- in previous implementation we got timeout errors because we polled once a second whether the total earnings have gone below 3, but at the same the sponsorships was generating 3 tokens of earnings per second
- the test uses only one sponsorship

## Refactoring
- use `multiply()` helper function in `maintainOperatorPoolValue`
- inline and simplify helper function of `operatorPoolValueUtils.ts`
- authenticate as a node (instead of an operator) in `checkOperatorPoolValueBreach` test


packages/broker/src/helpers/multiply.ts
```diff
@@ -0,0 +1,6 @@
+// this precision is big enough as we typically handle bigints which are in weis (the smallest denomination of Ethereum)
+const PRECISION = 1e18
+
+export const multiply = (val1: bigint, val2: number): bigint => {
+    return val1 * BigInt(PRECISION * val2) / BigInt(PRECISION)
+}
```
harbu, line 6: Name is a bit generic (especially since the function is in its own file)

`multiplyBigIntAndNumber` or `multiplyBigIntByNumber`

harbu: On second thought this seems a bit specific? Should it just be included in "maintainOperatorPoolValue.ts" as opposed to being in its own file.

harbu (approved):

packages/broker/src/helpers/multiply.ts
teogeb: The short name tells what the function does, and we can see the parameter types from the type information. When we read the code where the function is used, the short name tells tells accurately what action will be done (it just multiplies two numbers).

PR #1836, opened 2023-09-06
refactor(broker): Rename operator value

teogeb:
Rename `OperatorValue` to `OperatorPoolValue`  so that the terminology is consistent between Broker and the contracts.

Also renamed `MaintainOperatorPoolValueHelper#getPenaltyLimitFraction` to `getDriftLimitFraction`. There are `poolValueDriftLimitFraction` and `poolValueDriftPenaltyFraction` fields `StreamrConfig.sol`, and therefore the previous name was bit misleading.  
- https://github.com/streamr-dev/network-contracts/blob/3bcb08d5be82624323e5912fb4ecb543aa3b9b59/packages/network-contracts/contracts/OperatorTokenomics/StreamrConfig.sol#L59


harbu (approved):
Nice, good to naming fixed 👍

PR #1838, opened 2023-09-07
refactor(trackerless-network): Remove `PeerId`

teogeb:
Do not use `PeerID` class in `trackerless-network`. Use the utilities which handle `NodeID` instead (e.g. `getNodeIdFromPeerDescriptor`).


packages/trackerless-network/src/logic/RandomGraphNode.ts
```diff
@@ -110,11 +110,21 @@ export class RandomGraphNode extends EventEmitter<Events> implements IStreamNode
     async start(): Promise<void> {
         this.started = true
         this.registerDefaultServerMethods()
-        this.config.layer1.on('newContact', (peerDescriptor, closestPeers) => this.newContact(peerDescriptor, closestPeers))
-        this.config.layer1.on('contactRemoved', (peerDescriptor, closestPeers) => this.removedContact(peerDescriptor, closestPeers))
-        this.config.layer1.on('newRandomContact', (peerDescriptor, randomPeers) => this.newRandomContact(peerDescriptor, randomPeers))
-        this.config.layer1.on('randomContactRemoved', (peerDescriptor, randomPeers) => this.removedRandomContact(peerDescriptor, randomPeers))
-        this.config.P2PTransport.on('disconnected', (peerDescriptor: PeerDescriptor) => this.onNodeDisconnected(peerDescriptor))
+        this.config.layer1.on('newContact', (peerDescriptor: PeerDescriptor, closestNodes: PeerDescriptor[]) => {
```
juslesan: Remove curly brackets from on listeners. On and off listeners should contain the same callback function.

juslesan (approved):

packages/trackerless-network/src/logic/RandomGraphNode.ts
teogeb: Fixed in https://github.com/streamr-dev/network/pull/1839

PR #1840, opened 2023-09-07
fix(network): Fix joining on streams when all entry points queried from DHT are offline.

juslesan:
## Summary

Fix bug where if all stream entrypoints stored in the DHT are offline, the stream would be inaccessible until the stored items' TTLs run out.

## Additional changes
- `joinDHT()` operation no longer throws on time outs. Instead a rejoin is attempted as long as the new `doRejoin` is `true` (it is true by default). Previously this kind of rejoining was sometimes an infinite loop.
- layer1 joins only do DHT rejoins in cases where the stream entry points are explicitly set.
- the scheduler of `PeerDiscovery#getClosestPeersFromBucket` invocation is only started once.


packages/trackerless-network/src/logic/StreamEntryPointDiscovery.ts
```diff
@@ -196,8 +197,8 @@ export class StreamEntryPointDiscovery {
             if (this.config.streams.has(streamPartId)) {
                 const stream = this.config.streams.get(streamPartId)
                 const rediscoveredEntrypoints = await this.discoverEntryPoints(streamPartId)
-                await stream!.layer1!.joinDht(rediscoveredEntrypoints, false)
-                if (stream!.layer1!.getBucketSize() < 4) {
+                await stream!.layer1!.joinDht(rediscoveredEntrypoints, false, false)
+                if (stream!.layer1!.getBucketSize() < this.config.networkSplitAvoidanceLimit) {
                     throw new Error(`Network split is still possible`)
                 }
```
harbu: Should there be a check here of whether `stream.layer1 !== undefined`

packages/trackerless-network/src/logic/StreamEntryPointDiscovery.ts
```diff
@@ -196,8 +197,8 @@ export class StreamEntryPointDiscovery {
             if (this.config.streams.has(streamPartId)) {
                 const stream = this.config.streams.get(streamPartId)
```
harbu: I would move the non-null assertion `!` here for clarity (we just checked with `has` that it does exist -> no need to go forward with a `stream` of type `StreamObject | unefined`).

```suggestion
                const stream = this.config.streams.get(streamPartId)!
```

packages/trackerless-network/src/logic/StreamrNode.ts
```diff
@@ -216,10 +220,10 @@ export class StreamrNode extends EventEmitter<Events> {
             forwardingPeer
         )
         entryPoints = knownEntryPoints.concat(discoveryResult.discoveredEntryPoints)
-        await layer1.joinDht(sampleSize(entryPoints, 4))
+        await layer1.joinDht(sampleSize(entryPoints, NETWORK_SPLIT_AVOIDANCE_LIMIT), false, enableRejoins)
         await this.streamEntryPointDiscovery!.storeSelfAsEntryPointIfNecessary(
             streamPartId,
-            discoveryResult.joiningEmptyStream,
+            layer1.getBucketSize() < NETWORK_SPLIT_AVOIDANCE_LIMIT,
```
harbu: Is `discoveryResult.joiningEmptyStream` used elsewhere? If not it could be removed along with the logic to calculate it.

packages/trackerless-network/test/end-to-end/proxy-connections.test.ts
```diff
@@ -205,7 +205,7 @@ describe('Proxy connections', () => {
         await waitForCondition(() => 
             proxiedNode.hasProxyConnection(streamPartId, getNodeIdFromPeerDescriptor(proxyNodeDescriptor1), ProxyDirection.SUBSCRIBE)
         , 25000)
-        expect(proxyNode1.hasProxyConnection(streamPartId, proxiedPeerId.toKey() as unknown as NodeID, ProxyDirection.SUBSCRIBE)).toBe(false)
+        expect(proxyNode1.hasProxyConnection(streamPartId, proxiedPeerId.toKey() as unknown as NodeID, ProxyDirection.SUBSCRIBE)).toBe(true)
```
harbu: Was this test assertion wrong before? Seems like now it properly works? (i.e. the proxy connection is recovered one the proxy node comes online?)

juslesan: Yeah there seemed to be a bug with the test.

packages/trackerless-network/src/logic/StreamEntryPointDiscovery.ts
juslesan: As the stream object is held in memory before await call the worst case here is that the layer1-node is stopped which is fine as the AbortController should have canceled the avoidNetworkSplit loop if the node has left the stream.

packages/dht/src/dht/DhtNode.ts
```diff
@@ -620,12 +620,12 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
         await this.router!.send(msg, reachableThrough)
     }
 
-    public async joinDht(entryPointDescriptors: PeerDescriptor[], doRandomJoin?: boolean): Promise<void> {
+    public async joinDht(entryPointDescriptors: PeerDescriptor[], doRandomJoin?: boolean, doRejoin?: boolean): Promise<void> {
```
```suggestion
    public async joinDht(entryPointDescriptors: PeerDescriptor[], doRandomJoin?: boolean, retry?: boolean): Promise<void> {
```

(also in other places where rejoin flag is used)

packages/dht/src/dht/discovery/PeerDiscovery.ts
```diff
@@ -90,16 +90,19 @@ export class PeerDiscovery {
...
             }
+        } catch (_e) {
+            logger.debug(`DHT join on ${this.config.serviceId} timed out`)
+        } finally {
             if (!this.stopped) {
                 if (this.config.bucket.count() === 0) {
-                    this.rejoinDht(entryPointDescriptor).catch(() => {})
-                } else {
+                    if (doRejoin) {
+                        setAbortableTimeout(() => this.rejoinDht(entryPointDescriptor), 1000, this.abortController.signal)
```
teogeb, line 99: Could simplify the waiting?

packages/dht/src/dht/discovery/PeerDiscovery.ts
```diff
@@ -90,16 +90,19 @@ export class PeerDiscovery {
...
+                    }
+                } else if (!this.recoveryIntervalStarted) {
+                    this.recoveryIntervalStarted = true
```
teogeb: getClosestPeersFromBucket -> fetchClosestPeersFromBucket

packages/dht/src/dht/discovery/PeerDiscovery.ts
```diff
@@ -90,16 +90,19 @@ export class PeerDiscovery {
...
                     await scheduleAtInterval(() => this.getClosestPeersFromBucket(), 60000, true, this.abortController.signal)
```
teogeb: Is this something like `ensureClosestPeersFooRunning`

packages/trackerless-network/src/logic/StreamEntryPointDiscovery.ts
```diff
@@ -138,11 +134,11 @@ export class StreamEntryPointDiscovery {
...
     async storeSelfAsEntryPointIfNecessary(
         streamPartId: StreamPartID,
-        joiningEmptyStream: boolean,
+        avoidNetworkSplit: boolean,
         entryPointsFromDht: boolean,
         currentEntrypointCount: number
     ): Promise<void> {
-        if (joiningEmptyStream) {
+        if (avoidNetworkSplit) {
             await this.storeSelfAsEntryPoint(streamPartId)
```
teogeb, line 144: This is done in both if branches, and we catching errors in only one of them? Maybe we can remove the try-catch?

packages/trackerless-network/src/logic/StreamEntryPointDiscovery.ts
```diff
@@ -138,11 +134,11 @@ export class StreamEntryPointDiscovery {
 
     async storeSelfAsEntryPointIfNecessary(
         streamPartId: StreamPartID,
-        joiningEmptyStream: boolean,
+        avoidNetworkSplit: boolean,
```
teogeb: the method can analyze `stream.layer1!.getBucketSize() < this.config.networkSplitAvoidanceLimit` without us passing the booleann

packages/trackerless-network/src/logic/StreamrNode.ts
```diff
@@ -95,6 +95,8 @@ export interface StreamrNodeConfig {
     acceptProxyConnections?: boolean
 }
 
+const NETWORK_SPLIT_AVOIDANCE_LIMIT = 4
```
teogeb: This constant could be in `StreamEntryPointDiscovery` and we don't need `this.config.networkSplitAvoidanceLimit`

packages/trackerless-network/src/logic/StreamrNode.ts
```diff
@@ -205,6 +208,7 @@ export class StreamrNode extends EventEmitter<Events> {
         }
         logger.debug(`Joining stream ${streamPartId}`)
         const knownEntryPoints = this.knownStreamEntryPoints.get(streamPartId) ?? []
+        const enableRejoins = knownEntryPoints.length > 0
```
teogeb: Could inline this

packages/trackerless-network/src/logic/StreamrNode.ts
```diff
@@ -216,10 +220,10 @@ export class StreamrNode extends EventEmitter<Events> {
             forwardingNode
         )
         entryPoints = knownEntryPoints.concat(discoveryResult.discoveredEntryPoints)
-        await layer1.joinDht(sampleSize(entryPoints, 4))
+        await layer1.joinDht(sampleSize(entryPoints, NETWORK_SPLIT_AVOIDANCE_LIMIT), false, enableRejoins)
```
teogeb: If we move the constant and export it, we can use it here

packages/trackerless-network/test/end-to-end/proxy-connections.test.ts
```diff
@@ -206,8 +206,7 @@ describe('Proxy connections', () => {
         await waitForCondition(() => 
             proxiedNode.hasProxyConnection(streamPartId, getNodeIdFromPeerDescriptor(proxyNodeDescriptor1), ProxyDirection.SUBSCRIBE)
         , 25000)
-        expect(proxyNode1.hasProxyConnection(streamPartId, proxiedNodeId, ProxyDirection.SUBSCRIBE)).toBe(false)
-
+        expect(proxyNode1.hasProxyConnection(streamPartId, proxiedNodeId, ProxyDirection.SUBSCRIBE)).toBe(true)
```
teogeb, line 208: This is a bug fix?

packages/trackerless-network/test/integration/joining-streams-on-offline-peers.test.ts
```diff
@@ -0,0 +1,104 @@
...
+    })
+
+    it('should recover if discovered nodes are offline', async () => {
+        let numOfMessages = 0
+
+        // store invalid peer descriptors to DHT
+        await entryPoint.getLayer0DhtNode().storeDataToDht(streamPartIdToDataKey(streamPartId), Any.pack(offlineDescriptor1, PeerDescriptor))
+        await entryPoint.getLayer0DhtNode().storeDataToDht(streamPartIdToDataKey(streamPartId), Any.pack(offlineDescriptor2, PeerDescriptor))
+        
+        await node1.getStreamrNode().subscribeToStream(streamPartId)
```
teogeb, line 94: Could rename this to `joinToStream`? (in a separate PR)

packages/trackerless-network/test/integration/joining-streams-on-offline-peers.test.ts
```diff
@@ -0,0 +1,104 @@
...
+        await node1.getStreamrNode().on('newMessage', () => {numOfMessages += 1})
```
teogeb:
```suggestion
        await node1.getStreamrNode().on('newMessage', () => { messageReceived = true )
```

packages/trackerless-network/test/end-to-end/proxy-connections.test.ts
juslesan: yes

packages/trackerless-network/src/logic/StreamEntryPointDiscovery.ts
juslesan: This change turns out to require a larger refactor related to adding an interface such as `ILayer1` which the layer1 DhtNodes implement. Keeping the code as is makes it more testable for now.

packages/dht/src/dht/discovery/PeerDiscovery.ts
```diff
@@ -128,7 +130,14 @@ export class PeerDiscovery {
         }
     }
 
-    private async getClosestPeersFromBucket(): Promise<void> {
+    private async ensureClosestPeersIntervalIsRunning(): Promise<void> {
```
teogeb: What kind of recovery the we refer here (in `recoveryIntervalStarted` field name)? Should we rename this method something like `ensureFooRecoveryRunning`?

packages/trackerless-network/src/logic/StreamEntryPointDiscovery.ts
```diff
@@ -138,19 +134,17 @@ export class StreamEntryPointDiscovery {
 
     async storeSelfAsEntryPointIfNecessary(
         streamPartId: StreamPartID,
-        joiningEmptyStream: boolean,
         entryPointsFromDht: boolean,
         currentEntrypointCount: number
     ): Promise<void> {
-        if (joiningEmptyStream) {
+        if (!this.config.streams.has(streamPartId)) {
```
teogeb, line 140: Should this method do antything if `(entryPointsFromDht && currentEntrypointCount < ENTRYPOINT_STORE_LIMIT)` is `true`, but `this.config.streams.has(streamPartId)` is `false`? Or is that combination possible?

packages/trackerless-network/src/logic/StreamrNode.ts
```diff
@@ -217,10 +217,9 @@ export class StreamrNode extends EventEmitter<Events> {
             forwardingNode
         )
         entryPoints = knownEntryPoints.concat(discoveryResult.discoveredEntryPoints)
-        await layer1.joinDht(sampleSize(entryPoints, 4))
+        await layer1.joinDht(sampleSize(entryPoints, NETWORK_SPLIT_AVOIDANCE_LIMIT), false, knownEntryPoints.length > 0)
```
teogeb: Is it intentional that we no longer use `randomJoin` (we used the default `value` of true before this PR and now explicit `false`)

packages/trackerless-network/test/integration/joining-streams-on-offline-peers.test.ts
```diff
@@ -0,0 +1,104 @@
...
+    beforeEach(async () => {
+        simulator = new Simulator(LatencyType.RANDOM)
+        const entryPointTransport = new SimulatorTransport(entryPointPeerDescriptor, simulator)
+        entryPoint = new NetworkStack({
+            layer0: {
+                transportLayer: entryPointTransport,
+                peerDescriptor: entryPointPeerDescriptor,
+                entryPoints: [entryPointPeerDescriptor]
+            },
+            networkNode: {}
```
teogeb:
```suggestion
```

also at lines `66` and `75`

packages/trackerless-network/test/integration/joining-streams-on-offline-peers.test.ts
```diff
@@ -0,0 +1,104 @@
...
+        await entryPoint.stop()
+        await node1.stop()
+        await node2.stop()
+        simulator.stop()
+    })
+
+    it('should recover if discovered nodes are offline', async () => {
+        let messageReceived = false
+
+        // store invalid peer descriptors to DHT
```
teogeb: Are these invalid descriptors (e.g. malformed), or just descriptors of nodes which are offline?

packages/trackerless-network/test/unit/StreamEntrypointDiscovery.test.ts
```diff
@@ -71,10 +74,21 @@ describe('StreamEntryPointDiscovery', () => {
 
     const fakeDeleteEntryPointData = async (_key: Uint8Array): Promise<void> => {}
 
+    const addNodesToStream = (layer1: MockLayer1, count: number) => {
+        range(count).forEach(() => {
+            layer1.addNewRandomPeerToKBucket()
```
teogeb, line 79: Would it make sense to rename `addNewRandomPeerToKBucket` e.g. to `addRandomNode()`?

juslesan: We might want to add tests where we add peers to the `neightborList` of `MockLayer1` nodes so it makes sense to differentiate between which struct is being used.

packages/trackerless-network/test/integration/joining-streams-on-offline-peers.test.ts
juslesan: offline

packages/trackerless-network/src/logic/StreamEntryPointDiscovery.ts
juslesan: If the node has left the stream before this method is called for whatever reason we should not store the node as an entrypoint

packages/trackerless-network/src/logic/StreamrNode.ts
```diff
@@ -217,7 +217,7 @@ export class StreamrNode extends EventEmitter<Events> {
             forwardingNode
         )
         entryPoints = knownEntryPoints.concat(discoveryResult.discoveredEntryPoints)
-        await layer1.joinDht(sampleSize(entryPoints, NETWORK_SPLIT_AVOIDANCE_LIMIT), false, knownEntryPoints.length > 0)
+        await layer1.joinDht(sampleSize(entryPoints, NETWORK_SPLIT_AVOIDANCE_LIMIT), true, knownEntryPoints.length > 0)
```
teogeb, line 220: Could write a test for this as none of our test caught this bug

teogeb (approved):

packages/trackerless-network/src/logic/StreamrNode.ts
juslesan: The random joins are not strictly necessary when joining streams as we care more about knowledge of nearby peers but it does improve the reliability of the stream if nodes know of more candidates to connect to

teogeb: https://linear.app/streamr/issue/NET-1078/test-case-for-random-join

PR #1843, opened 2023-09-08
refactor(broker): Some services as functions

teogeb:
We may want to merge https://github.com/streamr-dev/network/pull/1844 first to avoid merge conflicts in that PR.

Extracted the main functionality of these services to separate functions, and run a scheduler in `OperatorPlugin`:
- `AnnounceNodeToContractService`
- `AnnounceNodeToStreamService`
- `OperatorPoolValueBreachWatcher`

The `maintainOperatorPoolValue` was not changed in this PR. There is a separate PR for that service: https://github.com/streamr-dev/network/pull/1834.

The `OperatorPlugin#start()` starts some of the tasks in background so that the `OperatorPlugin#start()` is not blocked by those task initializations (Broker is able to start e.g. before the fleet state is ready).

Also simplified the related test to call the function instead of starting/stopping the service. Note that the main scheduler logic and leader checks are no longer covered by these tests. We'll implement a separate end-to-end test for that in a separate PR.
- also removed `announceToStream` unit  test: the same functionality is already covered by the integration test

## Other changes

- call `OperatorFleetState#destroy()` on `OperatorPlugin#stop()`

## Future improvements

- Unify error handling?
- Unify scheduling: could use `scheduleAtInterval` instead of `setAbortableInterval`?
- Confiugrable poll intervals


harbu (approved):
lgtm; we lost some test coverage due to the scheduling and leader check happening at the caller. Need to remember to add that back later.

PR #1859, opened 2023-09-12
feat(broker): Get `theGraphUrl` from client

teogeb:
Add new internal `getConfig()` to `StreamrClient`. Use that method in `OperatorPlugin` to get the url of The Graph. 

The config object always contains a value: if user has not specified any value (in `StreamrClient` construtor / in Broker's config file) the value is the production default value.

Note that the current production default (`https://api.thegraph.com/subgraphs/name/streamr-dev/streams` in `config.schema.json` at line 278) value doesn't work as that is not the mono subgraph URL. We have not yet deployed the mono subgraph to production, but when we do and update the correct value to `config.schema.json,` Broker starts to use it automatically.

### DIFF
--- a/packages/client/src/StreamrClient.ts
+++ b/packages/client/src/StreamrClient.ts
@@ -659,6 +659,13 @@ export class StreamrClient {
         return (await this.node.getNode()).getDiagnosticInfo()
     }
 
+    /**
+     * @deprecated This in an internal method
+     */
+    getConfig(): StrictStreamrClientConfig {
+        return this.config
+    }
+
     // --------------------------------------------------------------------------------------------
     // Events
     // --------------------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


packages/client/src/StreamrClient.ts
```diff
@@ -659,6 +659,13 @@ export class StreamrClient {
         return (await this.node.getNode()).getDiagnosticInfo()
     }
 
+    /**
+     * @deprecated This in an internal method
```
hpihkala, line 663: Is there a reason why it needs to be internal?

hpihkala (approved):

packages/client/src/StreamrClient.ts
teogeb: It can be public if we want. 

Earlier we've annotated `StreamrClient#getNode` as internal because we don't want end-users to access the network via the low level methods of `NetworkNodeStub`. Similarly the config defaults (e.g. the url the of The Graph) are very low level details, which end-users should never need.
https://github.com/streamr-dev/network/blob/6fde5ebde7af6c382a6d343a307da09f2d4bd28a/packages/client/src/StreamrClient.ts#L571

juslesan (approved):

PR #1860, opened 2023-09-12
fix network split issue in small networks

juslesan:
## Summary

Fixed a stream split issue discovered in smaller networks with random kademlia ids. 

Finding closest peers from k-bucket during join can leave out online peers during parellel DHT joins in smaller networks. Adding peers to the `nearbyNodeView` from the layer1 k-bucket when the layer1 contacts are updated if the size of the `nearbyNodeView` is smaller than the configured `nodeViewSize` fixes the issue.

When combined with #1840 https://github.com/streamr-dev/network/actions/runs/6158897945/job/16712688016 passed 100 times.


packages/trackerless-network/src/logic/RandomGraphNode.ts
```diff
@@ -200,7 +190,16 @@ export class RandomGraphNode extends EventEmitter<Events> implements IStreamNode
...
             return
         }
-        this.config.nearbyNodeView.replaceAll(closestNodes.map((descriptor) =>
+        this.updateNearbyNodeView(closestNodes)
+    }
+
+    private updateNearbyNodeView(nodes: PeerDescriptor[]) {
+        if (nodes.length < this.config.nodeViewSize) {
+            this.config.layer1.getKBucketPeers().forEach((descriptor) => {
+                nodes.push(descriptor)
```
teogeb: Maybe should not mutate the parameter value

packages/trackerless-network/src/logic/RandomGraphNode.ts
```diff
@@ -200,13 +190,29 @@ export class RandomGraphNode extends EventEmitter<Events> implements IStreamNode
...
+    private updateNearbyNodeView(nodes: PeerDescriptor[]) {
+        this.config.nearbyNodeView.replaceAll(Array.from(nodes).map((descriptor) =>
             new RemoteRandomGraphNode(
                 descriptor,
                 this.config.randomGraphId,
                 toProtoRpcClient(new NetworkRpcClient(this.config.rpcCommunicator.getRpcClientTransport()))
             )
         ))
+        for (const descriptor of this.config.layer1.getKBucketPeers()) {
+            if (this.config.nearbyNodeView.size() < this.config.nodeViewSize) {
```
teogeb: Is this correct? We break out of loop (i.e. don't add the item) if there are are less items than what we want to be in `this.config.nodeViewSize`?

juslesan: fixed

teogeb (approved):

PR #1546, opened 2023-06-16
Streamr 1.0

juslesan:
## Summary

Base branch for Streamr 1.0

### Future improvements

- Interface `JsonPeerDescriptor` is bound to change in future. E.g. NodeType may become optional or be removed.

PR #1866, opened 2023-09-13
refactor: [NET-990] rm `NodeType`

teogeb:
Remove `PeerDescription#type` as it is no longer needed (see PR description in https://github.com/streamr-dev/network/pull/1546). The value of the field was not read anywhere in the code.

## Related changes

- changed ids in two `getMockPeers` functions so that descriptors are kept unique ([aaa8a3a](https://github.com/streamr-dev/network/pull/1866/commits/aaa8a3ab03316311a0f9411e60272c8905d1667c))


teogeb:
The node type is needed in https://linear.app/streamr/issue/NET-1067/peerdescriptor-and-entrypoint-config-needs-schema-hostname-and-ssl

PR #1872, opened 2023-09-15
fix(network): Fix flaky browser test RandomGraphNode-Layer1Node

juslesan:
## Summary

64 node runs `RandomGraphNode-Layer1Node` tests now use `waitForCondition` for all connections to be bidirectional.

### DIFF
--- a/packages/trackerless-network/test/integration/RandomGraphNode-Layer1Node-Latencies.test.ts
+++ b/packages/trackerless-network/test/integration/RandomGraphNode-Layer1Node-Latencies.test.ts
@@ -137,6 +137,7 @@ describe('RandomGraphNode-DhtNode-Latencies', () => {
                     }
                 })
             })
+            // NET-1074 Investigate why sometimes unidirectional connections remain.
             return mismatchCounter <= 2
         }, 20000, 1000)
     }, 90000)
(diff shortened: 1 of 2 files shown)


packages/trackerless-network/test/integration/RandomGraphNode-Layer1Node-Latencies.test.ts
```diff
@@ -124,19 +124,20 @@ describe('RandomGraphNode-DhtNode-Latencies', () => {
...
+                        if (!neighbor!.getTargetNeighborIds().includes(nodeId)) {
+                            mismatchCounter += 1
+                        }
                     }
-                }
+                })
             })
-        })
-        expect(mismatchCounter).toBeLessThanOrEqual(2)
+            return mismatchCounter <= 2
```
teogeb, line 141: Does this test allow that there are 2 neighbors, which shouldn't be neighbors? Or what exactly the `mismatchCounter` represents? Why that kind of mismatch is allowed?

juslesan: There's a known issue in low latency (simulator) environments where sometimes nodes do not seem to be able to form bideirectional connections. This is quite rare and debugging it in the past has been quite the rabbit hole

teogeb: Ok could add a link to that Liner ticket and TODO that we should change the test to not allow mismatch when the issue have been fixed

teogeb (approved):

PR #1873, opened 2023-09-17
feat(broker) [NET-1046]: implement inspection in `VoteOnSuspectNodeService`

harbu:
## Summary

- Refactor inspecting logic into inspectUtils.ts file.
- Add inspection to voting

- Voting needs end-to-end testing


packages/broker/src/plugins/operator/InspectRandomNodeService.ts
```diff
@@ -1,136 +1,15 @@
...
-        await targetOperatorFleetState.destroy()
-    }
-}
-
 export type FindTargetFn = typeof findTarget
-export type FindNodesForTargetFn = typeof findNodesForTarget
+export type InspectTargetFn = typeof inspectTarget
 export type FetchRedundancyFactorFn = typeof fetchRedundancyFactor
 
 export class InspectRandomNodeService {
```
teogeb: Could move the scheduler to `OperatorPlugin` (similarly to `announceNodeToStream`, `announceNodeToContract` and `maintainOperatorPoolValue`).

Then we don't need this class and this file can just contain the `inspectRandomNode` function (i.e. the current `inspect` method),.

Maybe it would even make sense to move the function to `inspectionUtils`. In that case we could also move the `inspectSuspectNode` function to that file.

packages/broker/src/plugins/operator/inspectionUtils.ts
```diff
@@ -0,0 +1,183 @@
...
+        operatorAddress: targetOperatorAddress,
+        streamPart: targetStreamPart
+    }
+}
+
+export async function findNodesForTarget(
+    target: Target,
+    streamrClient: StreamrClient,
+    fetchRedundancyFactorFn: FetchRedundancyFactorFn,
+    maxWait: number,
```
teogeb:
```suggestion
    timeout: number,
```

packages/broker/src/plugins/operator/inspectionUtils.ts
```diff
@@ -0,0 +1,183 @@
...
+        sponsorshipAddress: targetSponsorship.sponsorshipAddress,
+        operatorAddress: targetOperatorAddress,
+        streamPart: targetStreamPart
+    }
+}
+
+export async function findNodesForTarget(
+    target: Target,
+    streamrClient: StreamrClient,
+    fetchRedundancyFactorFn: FetchRedundancyFactorFn,
```
teogeb: Would it make sense if this would be a simple function which has only one argument: 
```
getRedundancyFactor: (operatorContractAddress: EthereumAddress) => Promise<number | undefined>
```

That way `findNodesForTarget` could just call it with some operator contract address and get the result back. In `OperatorPlugin` we'd curry the `fetchRedundancyFactor` to get that function:
```
const getRedundancyFactor = (address) => fetchRedundancyFactor(address, this.serviceConfig.signer)
```

(When we combine all helpers to a single `OperatorRegistry` / `OperatorContractFacade` class this will simplify anyway, so maybe we could consider this refactoring then)

packages/broker/test/unit/plugins/operator/inspectionUtils.test.ts
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    beforeEach(() => {
+        helper = mock<InspectRandomNodeHelper>()
+        loadBalancer = mock<StreamAssignmentLoadBalancer>()
+    })
+
+    it('returns undefined if no sponsorships are found', async () => {
+        setupEnv([])
+        setStreamPartsAssignedToMe([])
```
teogeb:
```suggestion
```

packages/broker/test/unit/plugins/operator/inspectionUtils.test.ts
```diff
@@ -0,0 +1,231 @@
...
+        expect(result).toBeUndefined()
+    })
+
+    it('returns undefined if only finds sponsorship with my operator as only operator', async () => {
+        setupEnv([{
+            address: SPONSORSHIP_ADDRESS,
+            operators: [MY_OPERATOR_ADDRESS],
+            streamId: STREAM_ID,
+        }])
+        setStreamPartsAssignedToMe([])
```
teogeb:
```suggestion
```

packages/broker/test/unit/plugins/operator/inspectionUtils.test.ts
```diff
@@ -0,0 +1,231 @@
...
+
+    it('returns two of the nodes if multiple nodes found (replicationFactor=2)', async () => {
+        fetchRedundancyFactorFn.mockResolvedValueOnce(2)
+        comeOnline(PEER_DESCRIPTOR_ONE)
+        comeOnline(PEER_DESCRIPTOR_TWO)
+        comeOnline(PEER_DESCRIPTOR_THREE)
+        const result = await resultPromise
+        expect(result.length).toEqual(2)
+        expect(result).toIncludeAnyMembers([PEER_DESCRIPTOR_ONE, PEER_DESCRIPTOR_TWO, PEER_DESCRIPTOR_THREE])
+        expect(result[0]).not.toEqual(result[1])
```
teogeb: We always omit one of the three nodes, but maybe it it an internal implementation detail which node we omit? The `findNodesForTarget` just promises to provide some two online nodes if `replicationFactor` is 2.

packages/broker/src/plugins/operator/inspectSuspectNode.ts
```diff
@@ -1,16 +1,36 @@
...
-    targetOperator: string,
+    sponsorship: EthereumAddress,
+    targetOperator: EthereumAddress,
     partition: number,
-    voteOnSuspectNodeHelper: VoteOnSuspectNodeHelper
+    voteOnSuspectNodeHelper: VoteOnSuspectNodeHelper,
+    streamrClient: StreamrClient,
+    abortSignal: AbortSignal,
+    fetchRedundancyFactor = _fetchRedundancyFactor,
+    heartbeatLastResortTimeoutInMs = 60 * 1000,
```
teogeb: What does the "last resort" part of the parameter name reflect?

packages/broker/src/plugins/operator/inspectionUtils.ts
```diff
@@ -0,0 +1,183 @@
...
+    heartbeatLastResortTimeoutInMs,
+    abortSignal,
+    findNodesForTargetFn = findNodesForTarget
+}: {
+    target: Target
+    streamrClient: StreamrClient
+    fetchRedundancyFactor: FetchRedundancyFactorFn
+    heartbeatLastResortTimeoutInMs: number
+    abortSignal: AbortSignal
+    findNodesForTargetFn?: typeof findNodesForTarget
```
teogeb: Maybe this would simpler if we'd pass the `targetPeerDescriptors` as a parameter instead of `findNodesForTargetFn`? This function just calls the `findNodesForTargetFn` once.

packages/broker/src/plugins/operator/inspectSuspectNode.ts
harbu: Rename to `heartbeatTimeoutInMs`

teogeb (approved):

PR #1876, opened 2023-09-18
refactor(network): ILayer1

juslesan:
## Summary

Added a new interface `ILayer1` that is used for Layer1 nodes to improve the testability of the `RandomGraphNode` and `StreamEntryPointDiscovery` classes.

### DIFF
--- a/packages/trackerless-network/src/logic/ILayer1.ts
+++ b/packages/trackerless-network/src/logic/ILayer1.ts
@@ -1,5 +1,4 @@
 import { PeerDescriptor, SortedContactList, DhtPeer } from '@streamr/dht'
-import EventEmitter from 'eventemitter3'
 
 export interface ILayer1Events {
     newContact: (peerDescriptor: PeerDescriptor, closestPeers: PeerDescriptor[]) => void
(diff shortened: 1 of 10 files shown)


packages/trackerless-network/src/logic/ILayer1.ts
```diff
@@ -0,0 +1,19 @@
...
+import EventEmitter from 'eventemitter3'
+
+export interface ILayer1Events {
+    newContact: (peerDescriptor: PeerDescriptor, closestPeers: PeerDescriptor[]) => void
+    contactRemoved: (peerDescriptor: PeerDescriptor, closestPeers: PeerDescriptor[]) => void
+    newRandomContact: (peerDescriptor: PeerDescriptor, randomPeers: PeerDescriptor[]) => void
+    randomContactRemoved: (peerDescriptor: PeerDescriptor, randomPeers: PeerDescriptor[]) => void
+}
+
+export interface ILayer1 extends EventEmitter<ILayer1Events> {
```
teogeb: Should we use this event style also on `ITransport`? (a separate PR?)

packages/trackerless-network/src/logic/StreamrNode.ts
```diff
@@ -249,10 +250,10 @@ export class StreamrNode extends EventEmitter<Events> {
             rpcRequestTimeout: 15000,
             dhtJoinTimeout: 60000,
             nodeName: this.config.nodeName + ':layer1'
-        })
+        }) as ILayer1
```
teogeb: Why casting is needed? (also in tests)

teogeb (commented):
Maybe this PR could be about adding those tests? It would give the context to the `ILayer1` interface usage.

packages/trackerless-network/src/logic/StreamrNode.ts
juslesan: It is likely because the `DhtNode` implements the `ITransport` interface. Because of this DhtNodes cannot be directly assigned to `ILayer1` without casting. We also don't want ILayer1 to expose the `ITransport`-methods as they are not needed when interacting with Layer1 nodes.

teogeb (approved):

PR #1880, opened 2023-09-20
refactor(trackerless-network): Rename `subscribe`/`unsubscribe` to `join`/`leave`

teogeb:
Renames in `trackerless-network:`
- `subscribe` -> `join`
- `unsubscribe` -> `leave`
- `publish` -> `broadcast`

The `join()` method can't be called for proxied streams. The client which uses a proxy is not really part of the network for the given stream part (it just connects to the network, doesn't join to it), and therefore it makes sense that we don't call `join()` for such streams.
- There is now an explicit check for in `SubscriberSession.ts`, which checks if a stream is a proxied stream, and doesn't call `join()` if it is. Exposed a helper function `isProxiedStreamPart` from NetworkNode (and StreamrNode) for that.

Removed `subscribeAndWaitForJoin` and `waitForJoinAndPublish`. The `join` method now has an additional optional parameter to control the waiting of neighbors.

Note that the names of two metrics changed:
- `publishMessagesPerSecond` -> `broadcastMessagesPerSecond`
- `publishBytesPerSecond` -> `broadcastBytesPerSecond`

Also:
- moved some logic from `NetworkNode` to `NetworkStack`
- fine-tuned proxy connection checks (the `isProxiedConnection` calls which ensures that e.g. joinStreamPart is not called for a proxied stream).

## Future improvements

- The `joinStream` method should be renamed to `joinStreamPart` as it operator stream parts, not on streams. Also all other similar methods and entities: `leaveStream`, `LeaveStreamNotice`, `createStream`... (also note that the createStream doesn't create the stream, but DHT layers)
- Broadcast automatically joins the network. It could make sense that we'd require explicit join call before broadcasts can be done? But maybe the current API which does the join is easier to use?
- Currently we need to call joinStream in `StreamrNode#broadcast `in a way which part of the method is executed immediately (the initialization which is needed for `this.streams.set` call), but it is not good to await the whole method (as the node may become available as a side-effect of some other call). Maybe we should divide the `joinStream` method so that there is explicit sync part and explicit async part?


packages/trackerless-network/src/NetworkStack.ts
```diff
@@ -68,6 +71,34 @@ export class NetworkStack extends EventEmitter<NetworkStackEvents> {
...
     }
 
+    async joinStreamPart(streamPartId: StreamPartID, neighborRequirement?: { minCount: number, timeout: number }): Promise<void> {
+        if (this.getStreamrNode().isProxiedStreamPart(streamPartId)) {
+            throw new Error(`Cannot join to ${streamPartId} as proxy connections have been set`)
+        }
+        await this.joinLayer0IfRequired(streamPartId)
+        setImmediate(async () => {
+            try {
+                await this.getStreamrNode().joinStream(streamPartId)
```
juslesan, line 81: If make the `createStream` method in `StreamrNode` a public method we can separate the state setting and joining logic. Then the `StreamrNode#joinStream` can only be called in cases where the stream object is newly created. Then `joinStream()` can always be called in parallel when necessary.

packages/trackerless-network/src/logic/StreamrNode.ts
```diff
@@ -162,32 +124,17 @@ export class StreamrNode extends EventEmitter<Events> {
...
-        }
-    }
-
-    publishToStream(msg: StreamMessage): void {
+    broadcast(msg: StreamMessage): void {
         const streamPartId = toStreamPartID(msg.messageId!.streamId as StreamID, msg.messageId!.streamPartition)
-        if (this.streams.has(streamPartId)) {
-            this.streams.get(streamPartId)!.layer2.broadcast(msg)
-        } else {
+        if (!this.streams.has(streamPartId)) {
```
juslesan, line 129: If the previously suggested `createStream` change is implemented, `this.joinStream` can be called here inside `setImmediate`

juslesan (approved):

packages/trackerless-network/src/NetworkStack.ts
teogeb: Will be implemented in a separate PR

PR #1886, opened 2023-09-21
refactor(broker): Read Docker dev entry point from config

teogeb:
Read Docker dev entry point from `@streamr/contracts`.  

Use new `entry-point` streamr-docker-dev service in `broker-docker-image` CI job.
- configure `entry-point.ts` so that it is possible to run it inside Docker container

Broker, client and cli-tools tests no longer use Brokers as their entry-points. Instead they use `run-entry-point: true`, i.e. they run the `run-entry-point.ts` script.

Note that all `development*` and `broker*` Broker config files use the new entry point.


Dockerfile.broker
```diff
@@ -5,6 +5,7 @@ RUN --mount=type=cache,target=/root/.npm \
 	npm run bootstrap-pkg --package=streamr-client && \
 	npm run bootstrap-pkg --package=streamr-broker && \
 	npm run prune-pkg -- streamr-broker
+RUN chmod +x /usr/src/network/packages/broker/dist/bin/*.js
```
juslesan: Should we investigate why this is needed?

Dockerfile.broker
```diff
@@ -25,6 +26,7 @@ EXPOSE 7170/tcp
 EXPOSE 7171/tcp
 
 WORKDIR /home/streamr/network/packages/broker
+# is this needed? (if it is, why it has worked ok previously?)
```
juslesan: Remove unnecessary comment

juslesan (approved):

PR #1888, opened 2023-09-22
feat(network): Advertised Websocket  server host

juslesan:
## Summary

Added support for configuring an advertised websocket host name. This will be needed in the docker environment and likely in cases where nodes are ran behind a reverse-proxy or load balancer.

## Changes

- The `websocketHost` configuration is now used to advertise a host name that the machine it self might not know of.

- WebSocket servers are always started in `0.0.0.0`

The configuration of `externalIp` (for webrtc) and the advertised `websocketHost` could be unified in some way, but it will require thought on how to do it exactly. The details of starting websocket servers will change after the auto-certifier is completed.


packages/dht/src/connection/WebSocket/WebSocketServer.ts
```diff
@@ -90,7 +90,7 @@ export class WebSocketServer extends EventEmitter<ConnectionSourceEvents> {
             })
 
             try {
-                this.httpServer.listen(port, host)
+                this.httpServer.listen(port, '0.0.0.0')
```
teogeb, line 94: Add a comment that listens all IPv4 interfaces, but not IPv6 interfaces (as it would do if we don't specify any host)

teogeb (approved):

packages/client/src/Config.ts
```diff
@@ -90,6 +90,13 @@ export interface ControlLayerConfig {
      */
     websocketPortRange?: PortRange
 
+    /**
+     * The host name or IP address of the WebSocket server used to connect to it over the internet.
+     * If not specified, the host name will be auto-detected. 
+     * Can be useful in situations where the host is running begind a reverse-proxy or load balancer.
```
```suggestion
     * Can be useful in situations where the host is running behind a reverse-proxy or load balancer.
```

PR #1889, opened 2023-09-22
refactor(trackerless-network): Simplify `StreamrNode#publishToStream`

teogeb:
Remove obsolete `streamPartId` parameter as we can get the value from the message.

Also added comment about required fields which are incorrectly annotated as optional in `NetworkRpc.ts`


juslesan (approved):

packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -64,6 +64,7 @@ message StreamMessage {
   EncryptionType encryptionType = 3;
   bytes content = 4;
   bytes signature = 5;
+  // this is a required field but in generated NetworkRpc.ts it is incorrectly annotated as optional (NET-1082)
```
juslesan, line 67: This is a known issue with protobuf after the `required` annotation was removed. Validation of messages inside messages has to be done on a higher level.

PR #1892, opened 2023-09-25
Net 1041 autocertifier

ptesavol:
## Summary

AutoCertifier client and server to be integrated into the Streamr DHT

* Checking whether Streamr node is running in a certain port needs to be moved to use WSS
* Needs more testing


juslesan:
Logic loved to #2009

PR #1906, opened 2023-09-29
fix(network): Avoid offline peers in network split avoidance

juslesan:
## Summary

If peers queried from the DHT are detected as offline network avoidance will no longer attempt to reach them.

### DIFF
--- a/packages/trackerless-network/src/logic/StreamEntryPointDiscovery.ts
+++ b/packages/trackerless-network/src/logic/StreamEntryPointDiscovery.ts
@@ -9,6 +9,7 @@ import { Any } from '../proto/google/protobuf/any'
 import { Logger, setAbortableTimeout, wait } from '@streamr/utils'
 import { StreamObject } from './StreamrNode'
 import { StreamPartID } from '@streamr/protocol'
+import { NodeID, getNodeIdFromPeerDescriptor } from '../identifiers'
 
 export const streamPartIdToDataKey = (streamPartId: StreamPartID): Uint8Array => {
     return new Uint8Array(createHash('md5').update(streamPartId).digest())
@@ -69,6 +70,7 @@ export class StreamEntryPointDiscovery {
     private readonly config: StreamEntryPointDiscoveryConfig
     private readonly servicedStreamParts: Map<StreamPartID, NodeJS.Timeout>
     private readonly cacheInterval: number
+    private readonly networkSplitAvoidedNodes: Map<StreamPartID, Set<NodeID>> = new Map()
 
     constructor(config: StreamEntryPointDiscoveryConfig) {
         this.config = config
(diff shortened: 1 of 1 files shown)


packages/trackerless-network/src/logic/StreamEntryPointDiscovery.ts
```diff
@@ -69,6 +70,7 @@ export class StreamEntryPointDiscovery {
     private readonly config: StreamEntryPointDiscoveryConfig
     private readonly servicedStreamParts: Map<StreamPartID, NodeJS.Timeout>
     private readonly cacheInterval: number
+    private readonly networkSplitAvoidedNodes: Map<StreamPartID, Set<NodeID>> = new Map()
```
teogeb, line 73: Is entry point discovery for each stream a separate task? I.e. to discover an entry point for stream part S1, do we need any info we possible gathered previously when entry points for stream part S2 was discovered?

If that's the case, we could make `StreamEntryPointDiscovery` to operate just with an explicit stream part and create a separate instance each time we create a new `StreamObject` in `StreamrNode`. The `knownStreamEntryPoints` could be also a field in `StreamObject`, or embedded to the `StreamEntryPointDiscovery` class?

packages/trackerless-network/src/logic/StreamEntryPointDiscovery.ts
```diff
@@ -100,8 +102,17 @@ export class StreamEntryPointDiscovery {
...
         const dataKey = streamPartIdToDataKey(streamPartId)
-        return forwardingNode ? 
-            this.queryEntryPointsViaNode(dataKey, forwardingNode) : await this.queryEntrypoints(dataKey)
+        let discoveredEntryPoints = forwardingNode ? 
+            await this.queryEntryPointsViaNode(dataKey, forwardingNode) : await this.queryEntrypoints(dataKey)
+    
+        if (this.networkSplitAvoidedNodes.has(streamPartId)) {
+            const filtered = discoveredEntryPoints.filter((node) => 
+                !this.networkSplitAvoidedNodes.get(streamPartId)!.has(getNodeIdFromPeerDescriptor(node)))
+            if (filtered.length > 0) {
```
teogeb, line 112: add comment why we don't return empty list

packages/trackerless-network/src/logic/StreamEntryPointDiscovery.ts
```diff
@@ -192,13 +203,23 @@ export class StreamEntryPointDiscovery {
                 const rediscoveredEntrypoints = await this.discoverEntryPoints(streamPartId)
                 await stream.layer1!.joinDht(rediscoveredEntrypoints, false, false)
                 if (stream.layer1!.getBucketSize() < NETWORK_SPLIT_AVOIDANCE_LIMIT) {
+                    const nodesToAvoid = rediscoveredEntrypoints.filter((peer) => !stream.layer1!.getKBucketPeers().includes(peer))
```
teogeb, line 208: Add comment: the node is not online?

packages/trackerless-network/src/logic/StreamEntryPointDiscovery.ts
```diff
@@ -192,13 +203,23 @@ export class StreamEntryPointDiscovery {
...
+                    this.setAvoidedNodes(streamPartId, nodesToAvoid)
```
teogeb: Does this overwrite as the `set` prefix tells? Or could rename to `addAvoidedNodes`?

packages/trackerless-network/src/logic/StreamEntryPointDiscovery.ts
```diff
@@ -192,13 +203,23 @@ export class StreamEntryPointDiscovery {
...
                     throw new Error(`Network split is still possible`)
                 }
+                this.networkSplitAvoidedNodes.delete(streamPartId)
```
teogeb: Could move this to outer block (before `logger.trace)`?

packages/trackerless-network/src/logic/StreamEntryPointDiscovery.ts
juslesan: Yes, the class could be refactored to be per stream. However this might make future features like caching previously found stream entry points more difficult. This could be useful in cases where a node leaves and rejoins a stream in a quick succession.

teogeb (approved):

PR #1909, opened 2023-10-02
fix(trackerless-network): [NET-1093] `setProxies` doesn’t remove proxies when `connectionCount` is 0

teogeb:
If the `connectionCount` parameter for `StreamrNode#setProxies `call is `0`, the existing proxy is removed.

Also the proxy is not created if empty `nodes` array or `connectionCount=0` is used when the setProxies is called for the first time.

Also renamed the nodes `parameter` in `setProxies`.


packages/trackerless-network/test/unit/StreamrNode.test.ts
```diff
@@ -33,25 +27,75 @@ describe('StreamrNode', () => {
...
+        it('can leave', async () => {
+            await node.joinStream(streamPartId)
+            node.leaveStream(streamPartId)
+        })
     })
 
-    it('can leave', async () => {
-        await node.joinStream(streamPartId)
-        node.leaveStream(streamPartId)
+    describe('proxied stream', () => {
```
juslesan, line 61: Add async/await for setting proxies in this test block.

juslesan (changes requested):

packages/trackerless-network/test/unit/StreamrNode.test.ts
teogeb: Fixed

juslesan (approved):

PR #1914, opened 2023-10-02
refactor(trackerless-network): `Remote#formDhtRpcOptions` helper method

teogeb:
Added `Remote#formDhtRpcOptions` helper method as all Remote calls use `sourceDescriptor` and `targetDescriptor` fields.

Also removed `Remote` class in `trackerless-network` and using `dht`'s version there, too. The both classes had the same implementation. Also unified the constructor parameter order. Also changed the visibility of the fields to `private`

## Future improvements

Make methods of Remote `protected` instead of `public`? (serviceId is currently needed in `RoutingSesstion` and, `peerDescriptor` and peerId currently needed because `IContact`; maybe we could remove `IContact`)


packages/trackerless-network/src/logic/Remote.ts
```diff
@@ -1,4 +1,4 @@
-import { PeerDescriptor } from '@streamr/dht'
+import { DhtRpcOptions, PeerDescriptor } from '@streamr/dht'
 import { ProtoRpcClient } from '@streamr/proto-rpc'
 
 export abstract class Remote<T> {
```
juslesan: Add ownPeerDescriptor to `Remote` instead of passing as parameter in every call

juslesan:
Trackerless and DHT could likely use the same remote class. There might be some slight differences in how they operate at the moment

packages/dht/src/dht/contact/Remote.ts
```diff
@@ -3,38 +3,54 @@ import { ProtoRpcClient } from '@streamr/proto-rpc'
...
-        return this.peerId
+        return this.remotePeerId
     }
 
     getPeerDescriptor(): PeerDescriptor {
-        return this.peerDescriptor
+        return this.remotePeerDescriptor
+    }
+
+    getOwnPeerDescriptor(): PeerDescriptor {
```
juslesan: getLocalPeerDescriptor?

packages/dht/src/dht/contact/Remote.ts
```diff
@@ -3,38 +3,54 @@ import { ProtoRpcClient } from '@streamr/proto-rpc'
...
+        return this.ownPeerDescriptor
     }
 
     getServiceId(): string {
         return this.serviceId
     }
 
+    getClient(): ProtoRpcClient<T> {
```
juslesan, line 45: Are all new remote methods protected?

juslesan (approved):

PR #1916, opened 2023-10-04
fix(dht): [NET-1099] Incorrect event payload `SortedContactList#onContactRemoved`

teogeb:
The event payload of `contactRemoved` was incorrect when it is emitted from `SortedContactList#addContact`. The first argument should be the removed contact but is the added contact.

That parameter is currently not used in any of the event listeners. Maybe good to keep the parameter for consistency, as similar parameter is also `newContact` event.

### DIFF
--- a/packages/dht/src/dht/contact/SortedContactList.ts
+++ b/packages/dht/src/dht/contact/SortedContactList.ts
@@ -56,13 +56,14 @@ export class SortedContactList<Contact extends IContact> extends EventEmitter<Ev
                 this.contactIds.sort(this.compareIds)
             } else if (this.compareIds(this.contactIds[this.maxSize - 1], contact.getPeerId()) > 0) {
                 const removed = this.contactIds.pop()
+                const removedDescriptor = this.contactsById.get(removed!.toKey())!.contact.getPeerDescriptor()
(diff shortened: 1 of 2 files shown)


packages/dht/src/dht/contact/SortedContactList.ts
```diff
@@ -56,13 +56,14 @@ export class SortedContactList<Contact extends IContact> extends EventEmitter<Ev
...
                 const removed = this.contactIds.pop()
+                const removedDescriptor = this.contactsById.get(removed!.toKey())!.contact.getPeerDescriptor()
                 this.contactsById.delete(removed!.toKey())
                 this.contactsById.set(contact.getPeerId().toKey(), new ContactState(contact))
                 this.contactIds.push(contact.getPeerId())
                 this.contactIds.sort(this.compareIds)
                 this.emit(
                     'contactRemoved',
-                    contact.getPeerDescriptor(),
+                    removedDescriptor,
```
juslesan, line 66: We currently do not use the removeDescriptor for anything which is why nothing failed. Perhaps it can be removed from the emitted event entirely

juslesan (approved):

PR #1921, opened 2023-10-05
fix(trackerless-network): [NET-1106] Proxy connections counted as neighbors

teogeb:
Fixed `getNeighors` to exclude proxy connections. (Neighbors are the connected nearby-nodes if a node is part of that network. A client which uses a proxy is not part of that stream-network, it just connects to it.)

Also renamed `NetworkNode#getNeighborsForStreamPart` -> `getNeighbors(streamPart)` and it now just delegates to the existing getNeighbors method in `StreamrNode`.


juslesan:
It might be valid for `getNeighbors` to return the proxy connections if the node is running a `ProxyStreamConnectionClient` on the stream

juslesan (approved):

PR #1929, opened 2023-10-06
refactor(trackerless-network): Remove `IStreamNode`

teogeb:
Removed `IStreamNode` interface. It had two implementations: `RandomGraphNode` and `ProxyStreamConnectionClient`. The proxy client is not a node, and therefore the interface was a bit misleading.

Refactored `StreamObject` structure so that it doesn't need `IStreamNode`. Also simplified the type narrowing: there is now `proxied: boolean` field which indicates whether this stream is proxied or not. If a stream is proxied, we store a proxy client, and if it is a non-proxied stream we store the node and the associated layer1 instance.

There are also two functions: `broadcast()` and `stop()`, as those are the only functions which are needed to operate a `StreamObject` instance in `StreamrNode`. The `layer1` field is currently needed in `StreamEntryPointDiscovery` and other fields are needed in tests. 
- currently the usage in `StreamEntryPointDiscovery` needs casting, but actually we should do checks about the stream state there, too (similar check what we do e.g. in `StreamrNode.ts#176`)
- but if we'll refactor the `StreamEntryPointDiscovery` to operate on only one stream part (so that there is one discovery instance per stream part), that state handling will be a lot simpler and no checks maybe needed at all
- it may be possible to refactor the structure so that there are no public fields, just the `broadcast()` and `stop()` functions (and possibly some extra functions)

Also simplified some test to use `StreamrNode#getNeighbors()` instead of querying the node directly. And removed some neighbor checks from `proxy-connections.test.ts` which queried neighbors of proxied nodes (the checks were redundant as we already checked that the node correct connections).

## Future improvements

Rename `ProxyStreamConnectionClient#getTargetNeighborIds` as the target nodes of proxied nodes aren't neighbors. That method is no longer needed to have that name as `ProxyStreamConnectionClient` doesn't implement `IStreamNode`.


juslesan (commented):
Why not just rename `IStreamNode` to ie. `IStreamParticipant` if node is confusing?

teogeb:
> Why not just rename `IStreamNode` to ie. `IStreamParticipant` if node is confusing?

That was one possibility, but the interface had many methods which we don't need in `StreamrNode`. And the interface is kind of internal to `StreamrNode` as no other classes need to operate on both the `RandomGraphNode` and `ProxyStreamConnectionClient`. Therefore it is good to define the interface in `StreamrNode` and not in the global scope.

Currently the interface if called `StreamrObject` (as it was already before this PR) but will renamed to `StreamPartDelivery` in https://github.com/streamr-dev/network/pull/1928.

juslesan (approved):

PR #1931, opened 2023-10-07
NET-1110: hotfix: add msg handler to MaintainTopologyService

hpihkala:
Add a msg handler to the `streamrClient.subscribe` call in `MaintainTopologyService`, because without it, message pipeline buffers get full and prevent other subscriptions for the same streams from getting messages.


hpihkala:
Merging as trivial. More elaborate fixing TBD in [NET-1115](https://linear.app/streamr/issue/NET-1115/prevent-problems-when-nothing-consumes-from-a-subscription).

PR #1933, opened 2023-10-09
refactor(dht): `ContactList` item type

teogeb:
Removed `{ getPeerDescriptor: () => PeerDescriptor }` from `ContactList` item type. The class and the subclasses operate on the peer id information, and therefore the peer descriptor information is redundant for `ContactList`'s point of view. Changed the events signature emitted from `ContactList` so that the event parameters are the generic type (the `C` of `ContactList<C>`. This allows the event consumers to access the peer descriptor data type-safely.

## Open questions

- Do we want to hide the exact contact type from `ILayer1`? There was already earlier the function `getNeighborList: () => SortedContactList<DhtPeer>`, which exposed that neighbors are `DhtPeer` instances.
- What is the difference between terms `contact`, `peer` and `node`? E.g. the entry point nodes are defined using `PeerDescriptor` interface, so are the words peer and node interchangeable?
- The 2nd parameter of the events is sometimes has name `activeContacts` instead of `closestContacts`. What does an active contact mean? Maybe the 2nd event parameter of newRandomContact/randomContactRemoved should be `closestContacts`/`closesPeers` instead of `randomContacts`?

## Future improvements

- Could make the event naming consistent: e.g. `contactAdded`+`contactRemoved` instead of `newContact`+`contactRemoved`
- Would it make sense to change data type of `ContactList` just simple `{ peerId: PeerID }`, i.e. not use a function to fetch the value?


packages/dht/src/dht/DhtNode.ts
```diff
@@ -46,15 +46,15 @@ import { isNodeJS } from '../helpers/browser/isNodeJS'
 import { sample } from 'lodash'
 
 export interface DhtNodeEvents {
-    newContact: (peerDescriptor: PeerDescriptor, closestPeers: PeerDescriptor[]) => void
-    contactRemoved: (peerDescriptor: PeerDescriptor, closestPeers: PeerDescriptor[]) => void
+    newContact: (contact: DhtPeer, closestContacts: DhtPeer[]) => void
```
juslesan: Better to emit just the `PeerDescriptor `here , or something more abstract like `Remote`

teogeb: Reverted to use `PeerDescriptor`

juslesan (approved):

PR #1935, opened 2023-10-09
refactor(broker): `ContractFacade` utility

teogeb:
Combine operator and sponsorship contract helpers to new `ContractFacade` class. Moved some of tests to new `ContractFacade.test.ts` files (separate test files for unit and integration tests).

## Other changes

- `MaintainTopologyHelper` doesn't remove all event listeners (only listeners which were added by it)
- stricter type in `MainainTopologyHelper` when using data from `pullStakedStreams`
- In `ContractFacade#getProvider` we get the provider directly from `config.signer` (more straightforward than querying it via contract object)
- in the queries to The Graph we use `getOperatorContractAddress()` instead of `operatorContract.address` to ensure that the address is in lowercase:
  - The Graph is case-senstive
  - our indexer (in `network-contracts` repo) stores all addresses in lower case
  - documentation indicates that the address could be in checksum case: https://docs.ethers.org/v5/api/utils/address/#address (but we don't use function but a field of `Contract` instance)
  - in practice the `operatorContract.address` field in ContractFacade is always in lower case (as we pass the address in the lowercased form to the `Contract` constructor)


packages/broker/test/integration/plugins/operator/announceNodeToContract.test.ts
```diff
@@ -1,35 +1,37 @@
...
 
     it('read empty heartbeat, then write heartbeat then read timestamp', async () => {
-        expect(await helper.getTimestampOfLastHeartbeat()).toBeUndefined()
+        expect(await contractFacade.getTimestampOfLastHeartbeat()).toBeUndefined()
 
-        await helper.writeHeartbeat({
-            id: 'foobar'
-        })
+        const streamrClient = {
+            getPeerDescriptor: () => 'foobar' as any
```
harbu: better to use real PeerDescriptor imo

packages/broker/test/integration/plugins/operator/contractUtils.ts
```diff
@@ -91,6 +94,7 @@ interface DeployOperatorContractOpts {
  * @returns Operator
  */
 export async function deployOperatorContract(opts: DeployOperatorContractOpts): Promise<Operator> {
+    logger.debug('deploying OperatorContract')
```
harbu:
```suggestion
    logger.debug('Deploying OperatorContract')
```

packages/broker/test/integration/plugins/operator/contractUtils.ts
```diff
@@ -114,6 +118,7 @@ export async function deployOperatorContract(opts: DeployOperatorContractOpts):
     )).wait()
     const newOperatorAddress = operatorReceipt.events?.find((e) => e.event === 'NewOperator')?.args?.operatorContractAddress
     const newOperator = new Contract(newOperatorAddress, operatorABI, opts.deployer) as unknown as Operator
+    logger.debug(`deployed OperatorContract ${newOperator.address}`)
```
harbu:
```suggestion
    logger.debug('Deployed OperatorContract', { address: newOperator.address })
```

packages/broker/test/integration/plugins/operator/contractUtils.ts
```diff
@@ -134,6 +139,7 @@ export interface DeploySponsorshipContractOpts {
 }
 
 export async function deploySponsorshipContract(opts: DeploySponsorshipContractOpts): Promise<Sponsorship> {
+    logger.debug('deploying SponsorshipContract')
```
harbu: Same changes here as above

packages/broker/test/integration/plugins/operator/contractUtils.ts
```diff
@@ -197,21 +212,25 @@ export async function generateWalletWithGasAndTokens(opts?: GenerateWalletWithGa
 }
 
 export const delegate = async (delegator: Wallet, operatorContractAddress: string, amount: number, token?: TestToken): Promise<void> => {
+    logger.debug(`delegate ${amount}`)
```
harbu:
```suggestion
    logger.debug('Delegate', { amount })
```

packages/broker/test/integration/plugins/operator/contractUtils.ts
```diff
@@ -197,21 +212,25 @@ export async function generateWalletWithGasAndTokens(opts?: GenerateWalletWithGa
...
     // onTokenTransfer: the tokens are delegated on behalf of the given data address
     // eslint-disable-next-line max-len
     // https://github.com/streamr-dev/network-contracts/blob/01ec980cfe576e25e8c9acc08a57e1e4769f3e10/packages/network-contracts/contracts/OperatorTokenomics/Operator.sol#L233
     await transferTokens(delegator, operatorContractAddress, amount, delegator.address, token)
 }
 
 export const stake = async (operatorContract: Operator, sponsorshipContractAddress: string, amount: number): Promise<void> => {
+    logger.debug(`stake ${amount}`)
```
harbu: same as above here

packages/broker/test/unit/plugins/operator/ContractFacade.test.ts
```diff
@@ -0,0 +1,69 @@
...
+        })
+    
+        it('throws given valid json with field "partition" but outside integer range', () => {
+            expect(() => parsePartitionFromReviewRequestMetadata('{ "partition": -50 }')).toThrowError(ParseError)
+        })
+    
+        it('returns partition given valid json with field "partition" within integer range', () => {
+            expect(parsePartitionFromReviewRequestMetadata('{ "partition": 50 }')).toEqual(50)
+        })
+    })
```
harbu: I would move this outside `ContractFacade` describe-block

packages/broker/src/plugins/operator/ContractFacade.ts
```diff
@@ -0,0 +1,371 @@
...
+
+const logger = new Logger(module)
+
+export interface SponsorshipResult {
+    sponsorshipAddress: EthereumAddress
+    streamId: StreamID
+    operatorCount: number
+}
+
+export class ContractFacade {
```
mondoreale, line 73: This

```ts
export class ContractFacade {
    private readonly operatorContract: any
    private readonly theGraphClient: any
    private readonly config: any

    constructor(operatorContract: any, theGraphClient: any, config: any) {
        this.operatorContract = operatorContract
        this.theGraphClient = theGraphClient
        this.config = config
    }
}
```
is effectively this

```ts
export class ContractFacade {
    constructor(
        private readonly operatorContract: any,
        private readonly theGraphClient: any,
        private readonly config: any,
    ) {}
}
```

packages/broker/src/plugins/operator/ContractFacade.ts
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    try {
+        ensureValidStreamPartitionIndex(partition)
+    } catch {
+        throw new ParseError('invalid partition numbering')
+    }
+
+    return partition
+}
```
mondoreale, line 57: Maybe zoddified?

```ts
import { z } from 'zod'

export function parsePartitionFromReviewRequestMetadata(metadataAsString: unknown) {
    return z
        .string()
        .transform((json) => JSON.parse(json))
        .pipe(
            z.object({
                partition: z
                    .number()
                    .min(0)
                    .max(MAX_PARTITION_COUNT)
                    .refine(Number.isSafeInteger),
            }),
        )
        .transform(({ partition }) => partition)
        .parse(metadataAsString)
}
```

`zod` can be used for any kind of parsing, gives nicely typed results, and accurate issue explanations.

Just FYI. No pressure.

packages/broker/src/plugins/operator/MaintainTopologyHelper.ts
```diff
@@ -82,68 +67,28 @@ export class MaintainTopologyHelper extends EventEmitter<MaintainTopologyHelperE
...
                 this.streamIdOfSponsorship.set(sponsorshipId, streamId)
                 const sponsorshipCount = (this.sponsorshipCountOfStream.get(streamId) || 0) + 1
                 this.sponsorshipCountOfStream.set(streamId, sponsorshipCount)
             }
         }
-        return Array.from(this.sponsorshipCountOfStream.keys())
+        const initialStreams = Array.from(this.sponsorshipCountOfStream.keys())
+        if (initialStreams.length > 0) {
+            this.emit('addStakedStreams', initialStreams)
+        }
```
mondoreale: Or

```ts
if (this.sponsorshipCountOfStream.size) {
    this.emit('addStakedStreams', [...this.sponsorshipCountOfStream.keys()])
}
```
which saves you a redundant keys-to-array conversion for each zero-sized map.

packages/broker/src/plugins/operator/ContractFacade.ts
```diff
@@ -0,0 +1,371 @@
...
+        this.operatorContract.on(eventName, listener)
+    }
+
+    removeOperatorContractStakeEventListener(eventName: 'Staked' | 'Unstaked', listener: (sponsorship: string) => unknown): void {
+        this.operatorContract.off(eventName, listener)
+    }
+
+    getProvider(): Provider {
+        return this.config.signer.provider!
+    }
```
mondoreale, line 370: `!` means trouble, sooner or later. It's better to face the facts.

```ts
getProvider(): Provider {
    const { provider } = this.config.signer

    if (!provider) {
        throw new Error('I splode here so others dont have to')
    }

    return provider
}

```

packages/broker/src/plugins/operator/ContractFacade.ts
mondoreale: Of course you'd have to convert that `ZodError` into `ParseError` but this could be a decent entry point into a fruitful adventure of a `toParseError(e: ZodError)` utility. 😉

mondoreale (commented):
My "cosmetic" 2¢.

packages/broker/src/plugins/operator/ContractFacade.ts
teogeb: Yes. We chose not to use that pattern so that each class has all of its fields listed in one place. Then it is very easy to check what fields any class has. Usually the complete list of fields in a class tells very much about a class, and is an efficient way to get familiar what a class does.

E.g. here someone could easily misread that the class has only one field:
```
class Foo {

     private foo: number = 0
     
     constructor(private bar: number) {}

}
```

packages/broker/src/plugins/operator/MaintainTopologyHelper.ts
teogeb: Using helper variable makes it bit more readable. Converting an empty set to an empty array should be quite efficient anyway.

harbu (approved):

packages/broker/src/plugins/operator/ContractFacade.ts
harbu: We have an eslint rule in place where the preference is to declare the fields explicitly instead of using the short-hand you suggested.

packages/broker/src/plugins/operator/ContractFacade.ts
harbu: Looks actually pretty nice with `zod`, I'll keep this in mind. But maybe this change can be made later.

packages/broker/src/plugins/operator/MaintainTopologyHelper.ts
harbu: I suppose you could combine both points like:
```ts
if (this.sponsorshipCountOfStream.size > 0) {
    const initialStreams = Array.from(this.sponsorshipCountOfStream.keys()) // or [...keys] notations
    this.emit('addStakedStreams', initialStreams)
}
```

packages/broker/src/plugins/operator/ContractFacade.ts
mondoreale: Cool.

packages/broker/src/plugins/operator/ContractFacade.ts
harbu: I would actually check for this in the constructor of the class. That way we fail even faster and basically make it clear that "we only accept signers that have an associated provider".

harbu (approved):

packages/broker/src/plugins/operator/MaintainTopologyHelper.ts
mondoreale: Fair point @harbu.

@teogeb, your point about readability is fair, too. I def can't tell what are the keys of `sponsorshipCountOfStream` by the name of the map itself w/o looking it up. And what makes them "initial", eh? For readability why not just

```ts
const queryResult = this.contractFacade.pullStakedStreams(latestBlock)

// this is new
const initialStreamIds: string[] = []

for await (const stake of queryResult) {
    const sponsorshipId = toEthereumAddress(stake.sponsorship.id)
    const streamId = toStreamID(stake.sponsorship.stream.id)
    if (this.streamIdOfSponsorship.get(sponsorshipId) !== streamId) {
        this.streamIdOfSponsorship.set(sponsorshipId, streamId)
        const sponsorshipCount = (this.sponsorshipCountOfStream.get(streamId) || 0) + 1
        this.sponsorshipCountOfStream.set(streamId, sponsorshipCount)

        // and this is new
        initialStreamIds.push(streamId)
    }
}

// readabilitiest ever!
if (initialStreamIds.length) {
    this.emit('addStakedStreams', initialStreamIds)
}
```

packages/broker/src/plugins/operator/ContractFacade.ts
teogeb: Created a PR about this: https://github.com/streamr-dev/network/pull/1954. The client always provides a signer with `Provider`, and after this PR our type definitions also ensures it.

packages/broker/src/plugins/operator/MaintainTopologyHelper.ts
teogeb: Applied a simplification

PR #1950, opened 2023-10-10
feat: make operator plugin configurable

harbu:
## Summary

Make timers and other key values of `OperatorPlugin` configurable. 


packages/broker/src/plugins/operator/ContractFacade.ts
```diff
@@ -70,20 +70,25 @@ export interface SponsorshipResult {
     operatorCount: number
 }
 
+export type ExtendedOperatorServiceConfig = OperatorServiceConfig & {
```
teogeb: These values are used only in the `getEarningsOf`, and that method is used only from `checkOperatorValueBreach` function. 

Maybe it would make sense to give these as parameters to the `checkOperatorValueBreach` function? That function is used in `OperatorPlugin`, i.e. it can easily plug in the correct values from the `OperatorPluginConfig` object. The `checkOperatorValueBreach` function would then pass the values to `ContractFacade#getEarningsOf` when it calls it.

packages/broker/src/plugins/operator/OperatorFleetState.ts
```diff
@@ -41,14 +35,33 @@ export class OperatorFleetState extends EventEmitter<OperatorFleetStateEvents> {
...
-    constructor(
+    static createOperatorFleetStateBuilder(
+        streamrClient: StreamrClient,
+        heartbeatIntervalInMs: number,
+        pruneAgeInMs: number,
+        pruneIntervalInMs: number,
+        latencyExtraInMs: number,
+        timeProvider = Date.now
+    ): CreateOperatorFleetStateFn {
+        return (coordinationStreamId) => new OperatorFleetState(
```
teogeb, line 46: Is this actually a factory? It creates new instances with same config values. In builders the focus is usually to control what kind of configs we apply for each generated object?

packages/broker/src/plugins/operator/OperatorPlugin.ts
```diff
@@ -19,19 +19,35 @@ import { MaintainTopologyHelper } from './MaintainTopologyHelper'
 import { inspectRandomNode } from './inspectRandomNode'
 import { ContractFacade } from './ContractFacade'
 
-export const DEFAULT_MAX_SPONSORSHIP_IN_WITHDRAW = 20 // max number to loop over before the earnings withdraw tx gets too big and EVM reverts it
-export const DEFAULT_MIN_SPONSORSHIP_EARNINGS_IN_WITHDRAW = 1 // token value, not wei
-
 export interface OperatorPluginConfig {
     operatorContractAddress: string
+    heartbeatUpdateIntervalInMs: number // 10 secs
```
teogeb: Could remove the values from the comments so they won't get out-of-sync with `config.schema.json`

packages/broker/src/plugins/operator/ContractFacade.ts
harbu: Sounds better yeah, will refactor

packages/broker/src/plugins/operator/OperatorFleetState.ts
harbu: The concept that varies here is the `coordinationStreamId`. We want to create instances of `OperatorFleetState` for different `coordinationStreamId` with the other config values being held constant.

packages/broker/src/plugins/operator/OperatorPlugin.ts
harbu: Agree'd, will remove.

harbu: Fixed

packages/broker/src/plugins/operator/ContractFacade.ts
harbu: Fixed

packages/broker/src/plugins/operator/OperatorFleetState.ts
harbu: For me factory sounds like a class so I avoided using that name.

teogeb: Yeah, for me it doesn't make a difference. Maybe this use case a kind of hybrid.

Usually it think factory like this:
```
const factory = ...
const instance1 = factory()
const instence2 = factory()
```

And builder something like this:
```
const instance = builder()
    .withFoo(123)
    .withBar(234)
    .build()    
```

teogeb (approved):

PR #1955, opened 2023-10-12
refactor(trackerless-network): [NET-1116] `StreamEntryPointDiscover` instances are stream part specific 

teogeb:
Refactored `StreamEntryPointDiscover` to be stream part specific, which simplifies the state handling significantly. Also fixes NET-1116 as the discovery instance is always stopped when a stream is left.

Simplified also scheduling of `keepSelfAsEntryPoint` to use scheduler instead of recursive timers.

### DIFF
--- a/packages/trackerless-network/src/logic/StreamPartEntryPointDiscovery.ts
+++ b/packages/trackerless-network/src/logic/StreamPartEntryPointDiscovery.ts
@@ -1,15 +1,14 @@
-import { createHash } from 'crypto'
 import {
-    isSamePeerDescriptor,
+    DataEntry,
     PeerDescriptor,
     RecursiveFindResult,
-    DataEntry
+    isSamePeerDescriptor
 } from '@streamr/dht'
-import { Any } from '../proto/google/protobuf/any'
-import { Logger, setAbortableTimeout, wait } from '@streamr/utils'
-import { StreamPartDelivery } from './StreamrNode'
 import { StreamPartID } from '@streamr/protocol'
+import { Logger, scheduleAtInterval, wait } from '@streamr/utils'
+import { createHash } from 'crypto'
 import { NodeID, getNodeIdFromPeerDescriptor } from '../identifiers'
+import { Any } from '../proto/google/protobuf/any'
 import { ILayer1 } from './ILayer1'
 
 export const streamPartIdToDataKey = (streamPartId: StreamPartID): Uint8Array => {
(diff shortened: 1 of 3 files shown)


packages/trackerless-network/src/logic/StreamPartEntryPointDiscovery.ts
```diff
@@ -57,31 +56,29 @@ const ENTRYPOINT_STORE_LIMIT = 8
...
 interface StreamPartEntryPointDiscoveryConfig {
-    streamParts: Map<string, StreamPartDelivery>
+    streamPartId: StreamPartID
     ownPeerDescriptor: PeerDescriptor
+    layer1: ILayer1
     getEntryPointData: (key: Uint8Array) => Promise<RecursiveFindResult>
     getEntryPointDataViaNode: (key: Uint8Array, node: PeerDescriptor) => Promise<DataEntry[]>
     storeEntryPointData: (key: Uint8Array, data: Any) => Promise<PeerDescriptor[]>
-    deleteEntryPointData: (key: Uint8Array) => Promise<void>
+    deleteEntryPointData: (key: Uint8Array) => Promise<unknown>
```
juslesan: Should be possible keep the return type as `Promise<void>`

packages/trackerless-network/src/logic/StreamPartEntryPointDiscovery.ts
```diff
@@ -101,20 +98,19 @@ export class StreamPartEntryPointDiscovery {
...
-        if (this.networkSplitAvoidedNodes.has(streamPartId)) {
-            const filtered = discoveredEntryPoints.filter((node) => 
-                !this.networkSplitAvoidedNodes.get(streamPartId)!.has(getNodeIdFromPeerDescriptor(node)))
-            // If all discovered entry points have previously beed detected as offline, try again
-            if (filtered.length > 0) {
-                discoveredEntryPoints = filtered
-            }
+        const filtered = discoveredEntryPoints.filter((node) => 
+            !this.networkSplitAvoidedNodes.has(getNodeIdFromPeerDescriptor(node)))
+        // If all discovered entry points have previously beed detected as offline, try again
```
juslesan: Typo
// If all discovered entry points have previously `beed` detected as offline, try again

// If all discovered entry points have previously `been` detected as offline, try again

juslesan (approved):

PR #1959, opened 2023-10-13
refactor(trackerless-network): Remove obsolete config fields

teogeb:
Removed fields which were not used by any component:

- `HandshakerServerConfig`: `ownPeerDescriptor`
- `ProxyStreamConnectionClientConfig`: `userId`, `nodeName`
  - these values are used via `ProxyDefinition` interface
- `StrictRandomGraphNodeConfig`: `name`
  - also moved `maxNumberOfContacts`, `minPropagationTargets`, `acceptProxyConnections` and `neighborUpdateInterval` to `RandomGraphNodeConfig` as the fields are used by the `createConfigWithDefaults` helper method  but not directly by `RandomGraphNode`

Renamed:

- `StreamrNodeConfig#firstConnectionTimeout` renamed and moved into `DhtNodeOptions`
- `StreamPartEntryPointDiscoveryConfig#cacheInterval` -> `storeInterval`
- `HandshakerServerConfig#N` -> `maxNeighborCount`
- `FindNeighborsSessionConfig#N` -> `minCount`
  
## Other changes

- simplified joining to the control layer in `NetworkStack`
  - added `ILaye0#waitForNetworkConnectivity()`
- do not rename `ControlLayerConfig` interface in client's `exports.ts`
- changed visibility of `StreamrNode#config` to `private`
- removed obsolete `public` annotations from public methods


packages/client/src/Config.ts
```diff
@@ -111,6 +111,12 @@ export interface ControlLayerConfig {
      * port mappings on the public side.
     */
     externalIp?: string
+
+    /**
+     * The maximum time to wait when forming a connection to the control layer. If the connection
+     * is not formed within this time, the client's network node will throw an error.
+     */
+    connectionTimeout?: number
```
juslesan: This should more clearly indicate that the configuration is only used when opening the first connection to reach the network as before.

packages/client/src/Config.ts
```diff
@@ -111,6 +111,12 @@ export interface ControlLayerConfig {
      * port mappings on the public side.
     */
     externalIp?: string
+
+    /**
+     * The maximum time to wait when forming a connection to the control layer. If the connection
```
juslesan:
```suggestion
     * The maximum time to wait when establishing connectivity to the control layer. If the connection
```

packages/trackerless-network/src/NetworkStack.ts
```diff
@@ -100,38 +71,24 @@ export class NetworkStack extends EventEmitter<NetworkStackEvents> {
...
-        }
+    private async ensureConnectedToControlLayer(): Promise<void> {
         // TODO we could wrap joinDht with pOnce and call it here (no else-if needed in that case)
         if (!this.layer0DhtNode!.hasJoined()) {
-            await this.joinDht()
-        } else if (this.layer0DhtNode!.getNumberOfConnections() < 1) {
-            await this.waitForFirstConnection()
+            setImmediate(async () => {
+                if (this.options.layer0?.entryPoints !== undefined) {
+                    // TODO should catch possible rejection?
```
juslesan, line 86: Good point but currently the joinDht call should never reject

juslesan (approved):

PR #1961, opened 2023-10-13
refactor(network): NET-1122, the DhtNode now uses the externalApi when necessary for DHT queries.

juslesan:
The DhtNode now uses the external API for getting dat from DHT when necessary instead of delegating the responsibily for a higher layer.

Removed unused `isJoinOngoing` from ilayer0


packages/trackerless-network/src/logic/ILayer0.ts
```diff
@@ -1,13 +1,11 @@
-import { DataEntry, ITransport, PeerDescriptor, RecursiveFindResult } from '@streamr/dht'
+import { DataEntry, ITransport, PeerDescriptor } from '@streamr/dht'
```
teogeb, line 1: Does the `dht` package need to export `RecursiveFindResult` class anymore?

packages/dht/src/dht/DhtNode.ts
```diff
@@ -657,8 +657,12 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
...
     }
 
-    public async getDataFromDht(idToFind: Uint8Array): Promise<RecursiveFindResult> {
-        return this.recursiveFinder!.startRecursiveFind(idToFind, FindMode.DATA)
+    public async getDataFromDht(idToFind: Uint8Array): Promise<DataEntry[]> {
+        if (this.peerDiscovery!.isJoinOngoing() && this.config.entryPoints && this.config.entryPoints.length > 0) {
+            return this.findDataViaPeer(idToFind, sample(this.config.entryPoints)!)
+        }
+        const result = await this.recursiveFinder!.startRecursiveFind(idToFind, FindMode.DATA)
+        return result.dataEntries || []
```
```suggestion
        return result.dataEntries ?? []
```

packages/trackerless-network/src/logic/StreamPartEntryPointDiscovery.ts
```diff
@@ -116,24 +111,9 @@ export class StreamPartEntryPointDiscovery {
...
-
-    // TODO remove this method in NET-1122
-    private async queryEntryPointsViaNode(key: Uint8Array, node: PeerDescriptor): Promise<PeerDescriptor[]> {
-        logger.trace(`Finding data via node ${this.config.ownPeerDescriptor.nodeName}`)
-        try {
-            const results = await this.config.getEntryPointDataViaNode(key, node)
-            if (results) {
-                return parseEntryPointData(results)
+            const result = await this.config.getEntryPointData(key)
+            if (result) {
```
teogeb: The `result` can't be undefined by the type definition. Could maybe remove if statement?

teogeb (approved):

PR #1965, opened 2023-10-13
added keywords to docs pages

yaruno:
## Summary

Added keywords to individual documentation pages to increase SEO.


lelas:
@yaruno Wouldn't deploy this. Meta keywords hold no value. Points for proactiveness though!

harbu:
Seems like this won't get deployed after all? Closing. Re-open if still needed.

PR #1973, opened 2023-10-18
refactor(trackerless-network): Rename `stream` -> `streamPart` when operating on stream partitions

teogeb:
Continue the work started on https://github.com/streamr-dev/network/pull/1928.

Renamed:
- `ProxyStreamConnectionClient` -> `ProxyClient`, `ProxyStreamConnectionServer` -> `ProxyServer`
- `StreamPartEntryPointDiscovery` -> `EntryPointDiscovery`
- `StreamrNode#hasStream`
- `StreamrNode#getStream` -> `getStreamPartDelivery`

Also:
- stricter parameter type for `StreamrNode#isProxiedStreamPart` parameter type
- stricter `StreamrNode` field types
- inline `StreamrNode#createProxyStream`
- simplify tests
- use `NodeID` type in `Handshaker#getOngoingHandshakes`
- changed proxy client service id, and unified some service ids

## Future improvements

- Rename `StreamNodeServer`. That will be done in a separate PR where we unify RPC server & client naming.


packages/trackerless-network/src/logic/proxy/ProxyClient.ts
```diff
@@ -58,19 +58,21 @@ interface ProxyDefinition {
 
 const logger = new Logger(module)
 
-export class ProxyStreamConnectionClient extends EventEmitter {
+const SERVICE_ID = 'system/proxy-client'
```
juslesan, line 62: Better to use `layer2-${config.streamPartId}`. The `layer2` prefix need s to be used here as the proxy clients use the same service as random graph nodes for message propagation. The prefix could be an exported constant such as `STREAM_PART_DELIVERY_SERVICE_ID_PREFIX`

juslesan: The value can also be changed `layer2` -> `stream-part-delivery`

juslesan (approved):

PR #1985, opened 2023-10-25
refactor(dht, trackerless-network): Remove obsolete parameters

teogeb:
Removed unused placeholder parameters from methods.

Methods like this:
```
public hasLocalLockedConnection(peerDescriptor: PeerDescriptor, _serviceId?: ServiceId): boolean {
```

Were changed to:
```
public hasLocalLockedConnection(peerDescriptor: PeerDescriptor): boolean {
```

In many cases the unused parameter was optional and callers didn't provide any value for it. But in some cases the unused parameter was a "required" parameter, and therefore the caller needed to construct some redundant value in order to call the method.


packages/dht/src/connection/WebRTC/WebRtcConnector.ts
```diff
@@ -149,11 +148,11 @@ export class WebRtcConnector implements IWebRtcConnectorService {
         })
 
         if (offering) {
-            connection.once('localDescription', (description: string, _type: string) => {
+            connection.once('localDescription', (description: string) => {
```
juslesan, line 152: Might be useful to check the type field here to ensure that the type of the description is `offer` or `answer` when appropriate

juslesan (approved):

PR #1991, opened 2023-10-27
test(dht): Merge bucket size tests

teogeb:
Merged 3 very similar tests into one test. Each test case is just a test run with different latency type. 

Changes:
- assertions were unified to expect contacts length / bucket size to be `>= getK() / 2`.
- all entry points use random region


juslesan (approved):
The test name could be related to Dht Joining ie. `DhtJoinPeerDiscovery.test.ts`

PR #1996, opened 2023-10-30
refactor(dht): Rename `isSamePeerDescriptor`

teogeb:
Rename to `areEqualPeerDescriptors`.


packages/dht/src/helpers/peerIdFromPeerDescriptor.ts
```diff
@@ -9,6 +9,6 @@ export const keyFromPeerDescriptor = (peerDescriptor: PeerDescriptor): PeerIDKey
     return PeerID.fromValue(peerDescriptor.kademliaId).toKey()
 }
 
-export const isSamePeerDescriptor = (peerDescriptor1: PeerDescriptor, peerDescriptor2: PeerDescriptor): boolean => {
+export const areEqualPeerDescriptors = (peerDescriptor1: PeerDescriptor, peerDescriptor2: PeerDescriptor): boolean => {
```
juslesan, line 12: As we only compare the `kademliaIds` here and do not check any other values, calling the function `areEqualPeerDescriptors` might be a bit misleading

juslesan (approved):

PR #2007, opened 2023-10-30
refactor(dht): Remove `protocolVersion`

teogeb:
Removed `protocolVersion` fields from RPC messages.

We may later add some kind of versioning in NET-1149, but in that case we could inject the information e.g. to `Message` entity.


packages/dht/protos/DhtRpc.proto
```diff
@@ -294,12 +290,10 @@ enum DisconnectMode {
 }
 
 message DisconnectNotice {
-  string protocolVersion = 1;
-  DisconnectMode disconnecMode = 2;
+  DisconnectMode disconnecMode = 1;
```
juslesan: disconnecMode -> disconnectMode

juslesan (approved):

packages/dht/protos/DhtRpc.proto
teogeb: fixed

PR #2008, opened 2023-10-31
refactor(dht): Refactor `RecursiveFindReport`

teogeb:
Rename RPC message entities and fields:
- `RecursiveFindRequest` -> `FindRequest`
- `RecursiveFindReport` -> `FindResponse`
- `RecursiveFindSessionService#reportRecursiveFindResult` -> `sendFindResponse`
- `FindRequest#findMode` -> `fetchData`

## Future improvements

- What would be good name for `noCloserNodesFound` field: is it also about connected peers?
- Rename also classes and methods so that they don't contain the term "recursive"
- Do we need `sessionId` or could we get the info from `serviceId`?
  - or would it be better to keep the `sessionId` and just use some static `serviceId` for the RPC? Then we'd add sessionId also to `FindResponse`
  - f we don't do that, we could maybe prefix sessionId with `system/find-session` so that it is easy to see what service it is


packages/dht/protos/DhtRpc.proto
```diff
@@ -27,7 +27,7 @@ service StoreService {
 }
 
 service RecursiveFindSessionService {
```
juslesan, line 29: Rename this as well when changing class names

juslesan (approved):

PR #2012, opened 2023-10-31
refactor(dht): Rename some `DhtNode` options

teogeb:
Renamed:
- `storeNumberOfCopies` -> `storageRedundancyFactor`
- `getClosestContactsLimit` -> `peerDiscoveryQueryBatchSize`

Some of the usages of `getClosestContactsLimit` will be removed in https://linear.app/streamr/issue/NET-1157/remove-4-unused-events-from-the-dhtnode

### DIFF
--- a/packages/dht/src/dht/DhtNode.ts
+++ b/packages/dht/src/dht/DhtNode.ts
@@ -71,7 +71,7 @@ export interface DhtNodeOptions {
     storeHighestTtl?: number
     storeMaxTtl?: number
     networkConnectivityTimeout?: number
-    storeNumberOfCopies?: number  // TODO better name?
+    storageRedundancyFactor?: number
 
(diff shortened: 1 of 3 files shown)


juslesan:
> What is a good name for getClosestContactsLimit. It seems to very important config option, which controls how many closest contacts we try to be aware of?
After NET-1157 the config option will only be required by PeerDiscovery logic. The option could be called something like `PeerDiscoveryQueryBatchSize`

> Should DhtNode#getClosestContacts use that value as default if maxCount is not given? Currently it use defaultContactQueryLimit implicitly. Typically those two have different values: 5 vs 20.
this function can just return the default 20 as it is a public getter and not related to PeerDiscovery

juslesan (approved):

PR #2014, opened 2023-11-01
refactor(dht): Extract `FindRpc`

teogeb:
Created new `FindRpc` RPC service by moving `findRecursively` method from `RouterRpc` to this service.

Also renamed:
- service `RecursiveFindSessionService` -> `FindSessionRpc`
- remote class `RemoteRecursiveFindSession` -> `FindSessionRpcRemote`

## Future improvements

- extract`RouterRpcLocal` from `Router`
- rename all `Recursive*` classes
- if we refactor `FindSessionRpc` to be singleton service (instead of session-specific service), the we can maybe move `FindSessionRpc#sendFindResponse` to `FindRpc` and remove the `FindSessionRpc`.


packages/dht/src/dht/DhtNode.ts
```diff
@@ -256,7 +256,7 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
             addContact: this.addNewContact.bind(this),
             connectionManager: this.connectionManager
         })
-        this.router = new Router({
+        this.routerRpcLocal = new RouterRpcLocal({
```
juslesan: This is not necessarily a good renaming since we need to bring back a class called `Router` which hadles routing logic. RouterRpcLocal should not be the class which handles the routing logic, it should only have the local rpc methods bound to the Rpc server.

packages/dht/src/dht/routing/FindRpcRemote.ts
```diff
@@ -0,0 +1,38 @@
...
+import { keyFromPeerDescriptor } from '../../helpers/peerIdFromPeerDescriptor'
+import { Remote } from '../contact/Remote'
+import { Logger } from '@streamr/utils'
+import { IFindRpcClient } from '../../proto/packages/dht/protos/DhtRpc.client'
+
+const logger = new Logger(module)
+
+export class FindRpcRemote extends Remote<IFindRpcClient> {
+
+    async findRecursively(params: RouteMessageWrapper): Promise<boolean> {
```
juslesan: findRecursively can simply be called find

teogeb: Renamed

packages/dht/src/dht/DhtNode.ts
teogeb: Reverted back to `Router`. Let's extract the  LocalRpc functionality in a separate PR.

packages/dht/protos/DhtRpc.proto
```diff
@@ -17,7 +17,10 @@ service DhtNodeRpc {
 service RouterRpc {
   rpc routeMessage (RouteMessageWrapper) returns (RouteMessageAck);
   rpc forwardMessage (RouteMessageWrapper) returns (RouteMessageAck);
-  rpc findRecursively (RouteMessageWrapper) returns (RouteMessageAck);
+}
+
+service FindRpc {
+  rpc find (RouteMessageWrapper) returns (RouteMessageAck);
```
juslesan: this could be called `routeFindRequest` to better explain what is happening.

juslesan (approved):

packages/dht/protos/DhtRpc.proto
teogeb: Renamed

PR #2016, opened 2023-11-02
refactor(dht, trackerless-network): Rename `ownPeerDescriptor`

teogeb:
Rename `ownPeerDescriptor` to `localPeerDescritor` so that the naming is in sync with the RPC naming (where we a have local RPC and a remote RPC).

## Future improvements

- Rename some `peerDescriptor` fields/methods to `localPeerDescriptor`, e.g. `StreamrNode#getPeerDescriptor()`?


packages/dht/src/connection/ConnectorFacade.ts
```diff
@@ -37,13 +37,13 @@ export interface DefaultConnectorFacadeConfig {
     webrtcPortRange?: PortRange
     maxMessageSize?: number
     tlsCertificate?: TlsCertificate
-    createOwnPeerDescriptor: (connectivityResponse: ConnectivityResponse) => PeerDescriptor
+    createLocalPeerDescriptor: (connectivityResponse: ConnectivityResponse) => PeerDescriptor
```
juslesan, line 40: This could be called `createPeerDescriptor` since `ITransport` has `getPeerDescriptor` for accessing the local PeerDescriptor

juslesan: or rename ITransports getter as well

juslesan (approved):

packages/dht/src/connection/ConnectorFacade.ts
teogeb: Renamed

PR #2030, opened 2023-11-03
fix(DHT): NET-1140 fix critical memory leak in ListeningRpcCommunicator

juslesan:
## Summary

The `ListeningRpcCommunicator` had a critical bug where it set en event listener to the `ITransport` passed to it, but it never cleaned it up. This caused all `ListeningRpcCommunicators` created for RecursiveFindSessions and RecursiveFindResponses to leak. Issue likely also existed for `RandomGraphNodes` in the trackerless-network. If a node a stream partition the `ListeningRpcCommunicator` created for it would leak.

`ListeningRpcCommunicator` has now a new method called `destroy` which ensures that the listener is cleaned up.
Also after a `RoutingSession` is completed the router will call stop on in


packages/dht/src/transport/ListeningRpcCommunicator.ts
```diff
@@ -4,11 +4,19 @@ import { RpcCommunicatorConfig } from '@streamr/proto-rpc'
...
             this.handleMessageFromPeer(msg)
-        })
+        }
+        this.transport = transport
+        transport.on('message', this.handler) 
+    }
+
+    destroy(): void {
+        this.transport!.off('message', this.handler)
+        this.stop()
```
teogeb, line 21: Could rename `stop` to `destroy` in `RpcCommunicator` (maybe in a separate PR)

packages/dht/src/transport/ListeningRpcCommunicator.ts
```diff
@@ -4,11 +4,19 @@ import { RpcCommunicatorConfig } from '@streamr/proto-rpc'
 import { Message } from '../proto/packages/dht/protos/DhtRpc'
 
 export class ListeningRpcCommunicator extends RoutingRpcCommunicator {
+    private readonly transport: ITransport | undefined
+    private readonly handler: (msg: Message) => void
     constructor(ownServiceId: string, transport: ITransport, config?: RpcCommunicatorConfig) {
         super(ownServiceId, transport.send, config)
-        
-        transport.on('message', (msg: Message) => {
+        this.handler = (msg: Message) => {
```
teogeb:
```suggestion
        this.listener = (msg: Message) => {
```

teogeb (approved):

packages/dht/src/transport/ListeningRpcCommunicator.ts
juslesan: Yeah, at the moment the ListeningRcCommunicator calls stop from the super class in destroy. Would probably make sense to get rid of inheritance here

PR #2031, opened 2023-11-06
refactor(dht): `ServiceID` type

teogeb:
New `ServiceID` type, which is just an alias to `string`. Also new `LockID` type for `ConnectionLockHandler`.

## Future improvements

- `ServiceID` could be a `BrandedString` if we want stricter type safety


packages/dht/src/connection/ConnectionLockHandler.ts
```diff
@@ -2,13 +2,12 @@
 // that is why this class is needed
 
 import { PeerIDKey } from '../helpers/PeerID'
-
-type ServiceName = string
+import { ServiceID } from '../types/ServiceID'
```
juslesan: This should be LockID

teogeb: Fixed

juslesan (approved):

PR #2034, opened 2023-11-06
refactor(dht): Extract `DhtNodeRpcLocal`

teogeb:
Extract  `DhtNodeRpcLocal` class.

### DIFF
--- a/packages/dht/src/dht/DhtNodeRpcLocal.ts
+++ b/packages/dht/src/dht/DhtNodeRpcLocal.ts
@@ -1,0 +1,68 @@
+import { ServerCallContext } from '@protobuf-ts/runtime-rpc'
+import { Logger } from '@streamr/utils'
+import KBucket from 'k-bucket'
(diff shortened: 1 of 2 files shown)


juslesan (approved):

packages/dht/src/dht/DhtNodeRpcLocal.ts
```diff
@@ -0,0 +1,69 @@
...
+        const response: PingResponse = {
+            requestId: request.requestId
+        }
+        return response
+    }
+
+    async leaveNotice(request: LeaveNotice, context: ServerCallContext): Promise<Empty> {
+        // TODO check signature??
+        // TODO is the serviceId check needed (it is defined DhtNode where RoutingRpcCommunicator is created)
+        if (request.serviceId === this.config.serviceId) {
```
juslesan: These kind of sanity checks can be removed. Any time a request is received we can assume that it was routed to the component correctly

PR #2036, opened 2023-11-06
refactor(dht): Extract `RouterRpcLocal`

teogeb:
Extract `RouterRpcLocal` class.

Refactored unit test so that we provide the incoming messages via new `FakeRpcCommunicator` class instead of calling `Router` methods directly.

## Future improvements

- Use `FakeRpcCommunicator` in some other tests, too?
- Add better typing for `FakeRpcCommunicator` (i.e. don't use `any`)


packages/dht/src/dht/routing/Router.ts
```diff
@@ -67,13 +55,37 @@ export class Router implements IRouter {
...
-            (routedMessage: RouteMessageWrapper) => this.routeMessage(routedMessage))
+        const rpcLocal = new RouterRpcLocal({
+            doRouteMessage: (routedMessage: RouteMessageWrapper, mode?: RoutingMode) => this.doRouteMessage(routedMessage, mode),
+            addContact: (contact: PeerDescriptor, setActive: boolean) => config.addContact(contact, setActive),
+            setForwardingEntries: (routedMessage: RouteMessageWrapper) => this.setForwardingEntries(routedMessage),
+            duplicateRequestDetector: this.duplicateRequestDetector,
+            localPeerDescriptor: this.localPeerDescriptor,
+            connectionManager: config.connectionManager
+        })
+        this.rpcCommunicator.registerRpcMethod(
```
juslesan, line 72: Add a `registerLocalRpcMethods` function to handle RPC server side binding

juslesan (approved):

PR #2049, opened 2023-11-10
refactor(dht): Do not store `ConnectionType`

teogeb:
We don't need to store a `connectionType` value for each `ManagedConnection` instance as the information can be queried  by calling the `expectedConnectionType` helper function. The helper was introduced in https://github.com/streamr-dev/network/pull/2043.

Removed obsolete `ConnectionType#SIMULATOR_*` enum values.

## Future improvements

`ConnectionType` could be string literals instead of enum values as we prefer those nowadays.


packages/dht/src/connection/ConnectionManager.ts
```diff
@@ -362,7 +362,7 @@ export class ConnectionManager extends EventEmitter<TransportEvents> implements
     private onConnected(connection: ManagedConnection) {
         const peerDescriptor = connection.getPeerDescriptor()!
         this.emit('connected', peerDescriptor)
-        logger.trace(getNodeIdFromPeerDescriptor(peerDescriptor) + ' onConnected() ' + connection.connectionType)
+        logger.trace(getNodeIdFromPeerDescriptor(peerDescriptor) + ' onConnected()')
```
harbu, line 352: I would keep the connectionType in the logging statements when it is relevant.

packages/dht/test/end-to-end/Layer0Webrtc.test.ts
```diff
@@ -70,10 +69,6 @@ describe('Layer0 with WebRTC connections', () => {
 
         expect((node1.getTransport() as ConnectionManager).hasConnection(node2.getLocalPeerDescriptor())).toEqual(true)
         expect((node2.getTransport() as ConnectionManager).hasConnection(node1.getLocalPeerDescriptor())).toEqual(true)
-        expect((node1.getTransport() as ConnectionManager).getConnection(node2.getLocalPeerDescriptor())!.connectionType)
-            .toEqual(ConnectionType.WEBRTC)
-        expect((node2.getTransport() as ConnectionManager).getConnection(node1.getLocalPeerDescriptor())!.connectionType)
```
harbu: These assertions related to connectionType are important to validate?

harbu (commented):
Sounds to me (on the surface level) like we should keep the logging statements and tests assertions related to connectionType. But instead of storing connectionType we can infer it statically with a helper function?

PR #2051, opened 2023-11-10
refactor(dht): Extract `WebrtcConnector`

teogeb:
Extract `WebrtcConnector` from `WebrtcConnectorRpcLocal`.

## Small changes

Removed obsolete `ongoingConnectAttempts.has(peerKey)` check.

## Future improvements

- `ongoingConnectAttempts`: pass accessor methods instead of passing a mutable entity
- check "`TODO remove redundant check?`" comments


juslesan (approved):
ongoingConnectAttempts.has() can likely be removed as delete can be called despite nothing being there.

PR #2056, opened 2023-11-14
refactor(dht): Extract `WebsocketConnector`

teogeb:
Extract `WebsocketConnector` from `WebsocketConnectorRpcLocal`.

## Small changes

- removed redundant `this.destroyed` check (the state is checked just before executing the method)
- use `abortController`

### DIFF
--- a/packages/dht/src/connection/websocket/WebsocketConnectorRpcLocal.ts
+++ b/packages/dht/src/connection/websocket/WebsocketConnectorRpcLocal.ts
@@ -1,292 +1,41 @@
-import { ClientWebsocket } from './ClientWebsocket'
-import { IConnection, ConnectionType } from '../IConnection'
-import { ITransport } from '../../transport/ITransport'
-import { ListeningRpcCommunicator } from '../../transport/ListeningRpcCommunicator'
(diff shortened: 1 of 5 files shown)


packages/dht/src/connection/websocket/WebsocketConnectorRpcLocal.ts
```diff
@@ -1,292 +1,37 @@
...
         const senderPeerDescriptor = (context as DhtCallContext).incomingSourceDescriptor!
-        if (!this.destroyed && this.canConnectFunction(senderPeerDescriptor)) {
+        if (this.config.canConnect(senderPeerDescriptor)) {
             setImmediate(() => {
-                if (this.destroyed) {
-                    return
-                }
-                const connection = this.connect(senderPeerDescriptor)
-                this.onIncomingConnection(connection)
+                const connection = this.config.connect(senderPeerDescriptor)
```
juslesan, line 33: Would be good to have some sort of check here for cases where the connector is destroyed and the setImmediate block is executed. The setImmediate call might leak now if it is executed after stop and the timings are worst case

teogeb: Fixed

juslesan (approved):

PR #2057, opened 2023-11-14
test(dht): Memory leak test

teogeb:
Add memory leak test. Sends a RPC message from a node to another.

## Open questions

- Check "`TODO should this work?`" for alternative way for initializing the sender and the receiver. Should that approach work?

## Future improvements

- More test cases

### DIFF
--- a/packages/dht/test/end-to-end/memory-leak.test.ts
+++ b/packages/dht/test/end-to-end/memory-leak.test.ts
@@ -1,0 +1,84 @@
+import LeakDetector from 'jest-leak-detector'
+import { binaryToHex, waitForCondition } from '@streamr/utils'
+import { randomBytes } from 'crypto'
+import { DhtNode } from '../../src/dht/DhtNode'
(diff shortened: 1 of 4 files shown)


packages/dht/test/end-to-end/memory-leak.test.ts
```diff
@@ -0,0 +1,84 @@
...
+                min: entryPointDescriptor.websocket.port,
+                max: entryPointDescriptor.websocket.port
+            },
+            entryPoints: [entryPointDescriptor]
+        })
+        await entryPoint.start()
+        await entryPoint.joinDht([entryPointDescriptor])
+        let sender: DhtNode | undefined = new DhtNode({})
+        let receiver: DhtNode | undefined = new DhtNode({})
+        /*TODO should this work? await Promise.all([
```
juslesan, line 35: I don't really see a reason why this should not work. Might be worth adding a new test case for this

juslesan (approved):

PR #2058, opened 2023-11-15
test(broker): [NET-1158, NET-1162] Nonce handling in `generateWalletWithGasAndTokens`

teogeb:
Add a `retry` utility to `@streamr/utils` and use it in Broker's `generateWalletWithGasAndTokens` test utility. There is only one adminWallet and therefore it is possible that multiple callers execute a transaction concurrently for that wallet. 

Also remove redundant explicit nonce handling in that method (if no `nonce` is defined, the functionality which we implemented is the default functionality).

### DIFF
--- a/packages/utils/src/retry.ts
+++ b/packages/utils/src/retry.ts
@@ -2,7 +2,7 @@ import { wait } from './wait'
 
 export const retry = async <T>(
     task: () => Promise<T>,
-    onFailure: (message: string, error: any) => void,
+    onRetryableFailure: (message: string, error: any) => void,
     description: string,
     maxCount: number,
     delay: number,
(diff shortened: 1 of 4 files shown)


packages/utils/src/retry.ts
```diff
@@ -0,0 +1,23 @@
...
+        try {
+            const result = await task()
+            return result
+        } catch (err: any) {
+            if (i < (maxCount - 1)) {
+                const message = `${description} failed, retrying in ${delay} ms`
+                onRetryableFailure(message, err)
+            }
+        }
+        await wait(delay)
```
harbu: Skip wait on last round?
```suggestion
        if (i < (maxCount - 1)) {
            await wait(delay)
        }
```

harbu (changes requested):
nice 👍

packages/utils/src/retry.ts
teogeb: Fixed

harbu (approved):

PR #2063, opened 2023-11-15
fix(DHT): Disable WS TLS after failed post-autocertification connectivity checks

juslesan:
## Summary

If the connectivity checks fail after autocertification, the WS server is now restarted with TLS disabled.

## Other changes

No longer log the full content of the autocertifier config file when doing updateSubdomainIp() in AutocertifierClient

### DIFF
--- a/packages/autocertifier-client/src/AutoCertifierClient.ts
+++ b/packages/autocertifier-client/src/AutoCertifierClient.ts
@@ -161,7 +161,7 @@ export class AutoCertifierClient extends EventEmitter<AutoCertifierClientEvents>
         }
         // TODO: use async fs methods?
         const oldSubdomain = JSON.parse(fs.readFileSync(this.configFile, 'utf8')) as CertifiedSubdomain
-        logger.info('updateSubdomainIp() called for ' + JSON.stringify(oldSubdomain))
+        logger.info('updateSubdomainIp() called for ' + oldSubdomain.fqdn)
         const sessionId = await this.restClient.createSession()
         this.ongoingSessions.add(sessionId)
         await this.restClient.updateSubdomainIp(
--- a/packages/dht/src/connection/ConnectorFacade.ts
+++ b/packages/dht/src/connection/ConnectorFacade.ts
@@ -97,45 +97,45 @@ export class DefaultConnectorFacade implements ConnectorFacade {
         // LocalPeerDescriptor could be stored in one place and passed from there to the connectors
         const temporarilySelfSigned = (!this.config.tlsCertificate && this.config.websocketServerEnableTls === true)
         const connectivityResponse = await this.websocketConnector.checkConnectivity(temporarilySelfSigned)
-        let localPeerDescriptor = this.config.createLocalPeerDescriptor(connectivityResponse)
-        this.localPeerDescriptor = localPeerDescriptor
-        this.websocketConnector.setLocalPeerDescriptor(localPeerDescriptor)
+        const localPeerDescriptor = this.config.createLocalPeerDescriptor(connectivityResponse)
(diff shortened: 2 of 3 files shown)


packages/autocertifier-client/src/AutoCertifierClient.ts
```diff
@@ -161,7 +161,7 @@ export class AutoCertifierClient extends EventEmitter<AutoCertifierClientEvents>
         }
         // TODO: use async fs methods?
         const oldSubdomain = JSON.parse(fs.readFileSync(this.configFile, 'utf8')) as CertifiedSubdomain
-        logger.info('updateSubdomainIp() called for ' + JSON.stringify(oldSubdomain))
+        logger.info('updateSubdomainIp() called for ' + oldSubdomain.fqdn)
```
teogeb, line 164: This is a separate fix, maybe worth mentioning in the PR description

packages/dht/src/connection/ConnectorFacade.ts
```diff
@@ -106,26 +106,37 @@ export class DefaultConnectorFacade implements ConnectorFacade {
                 const connectivityResponse = await this.websocketConnector!.checkConnectivity(false)
                 localPeerDescriptor = this.config.createLocalPeerDescriptor(connectivityResponse)
                 this.localPeerDescriptor = localPeerDescriptor
+                this.websocketConnector!.setLocalPeerDescriptor(localPeerDescriptor)
                 if (localPeerDescriptor.websocket === undefined) {
-                    logger.warn('ConnectivityCheck failed after autocertification, websocket server connectivity disabled')
+                    logger.warn('ConnectivityCheck failed after autocertification, disabling websocket server TLS')
```
teogeb: Warnings are visible to the end-users, and therefore could avoid internal names (i.e. `ConnectivityCheck` component)

```suggestion
                    logger.warn('Connectivity check failed after auto-certification, disabling WebSocket server TLS')
```

(similar at line 118)

packages/dht/src/connection/ConnectorFacade.ts
```diff
@@ -106,26 +106,37 @@ export class DefaultConnectorFacade implements ConnectorFacade {
...
+        await this.websocketConnector!.destroy()
+        this.websocketConnector = new WebsocketConnector({
+            ...webSocketConnectorConfig,
+            serverEnableTls: false,
+        })
+        await this.websocketConnector.start()
+        const connectivityResponse = await this.websocketConnector.checkConnectivity(false)
+        const localPeerDescriptor = this.config.createLocalPeerDescriptor(connectivityResponse)
+        this.localPeerDescriptor = localPeerDescriptor
+        this.websocketConnector.setLocalPeerDescriptor(localPeerDescriptor)
```
teogeb: We could have a helper method `setLocalPeerDescriptor` in `ConnectorFacade`, which would set all three fields. That way the values are always in sync, and the correct value of `localPeerDescriptor` will be updated to `webrtcConnector` too.

We'd use the method here an at line ~108. But wouldn't call it at line 125, i.e the `webrtcConnector` would receive the `peerDescriptor` a bit later (at the same time when websocket connector receives it). 

(Alternatively, if it is important that `webrtcConnector` can start asap, already before the `websocketConnector` is ready, we should update the correct `peerDescritor` also to that component.)

```

private setLocalPeerDescriptor(peerDescriptor: PeerDescriptor) {
    this.localPeerDescriptor = peerDescriptor
    this.websocketConnector!.setLocalPeerDescriptor(peerDescriptor)
    this.webrtcConnector!.setLocalPeerDescriptor(peerDescriptor)
}
```

teogeb (approved):

PR #2065, opened 2023-11-16
fix(dht): Improve `disconnected` event propagation

juslesan:
## Summary

Fix issues with 'disconnected' event propagation in duplicate connection cases.

In ManagedConnection if a connection is lost before it is fully opened an error is thrown. This ensures that Rpc Requests are  rejected faster in many cases instead of waiting for timeouts.

If a ManagedConnection detects errors it will set the `doNotEmitDisconnected` flag to false.

## Future improvements
This is a temporary fix, for the proper fix:
- Refactor the ManagedConnection class entirely to remove the need to use the `doNotEmitDisconnect` and `firstSend` flags.


juslesan:
Proper fixes were merged in #2074 and #1904

PR #2071, opened 2023-11-17
ETH-701 expired flag closer broker operator plugin

samt1803:
Added an operator plugin, that looks for flags in the sponsorships, that my operator is staked in, and that are expired, and then closes those flags by calling the voteOnFlag method on the sponsorship through the own operator contract.

Added one unittest (pretty pointless one), integrationtest would require either interfering with the predeployed set of contracts' config (and possibly breaking the env for other tests) OR deploying a new set of contracts with different config values, and then deploying a new subgraph that indexes the new contracts..


.vscode/launch.json
```diff
@@ -5,7 +5,23 @@
     "version": "0.2.0",
     "configurations": [
         {
-            "type": "pwa-node",
+            "name": "Launch via NPM",
```
teogeb: These changes are not part of this PR?

packages/broker/src/plugins/operator/ContractFacade.ts
```diff
@@ -72,10 +72,21 @@ export interface SponsorshipResult {
...
+    sponsorship: {
+        id: string
+    }
+}
+
 export class ContractFacade {
 
     private readonly operatorContract: Operator
-    private readonly theGraphClient: TheGraphClient
+    readonly theGraphClient: TheGraphClient
```
teogeb: The test could use the `ContractFacade` constructor which allows us to pass a custom `TheGraphClient` and keep this private

packages/broker/src/plugins/operator/ContractFacade.ts
```diff
@@ -171,6 +182,41 @@ export class ContractFacade {
         return results
     }
 
+    async getExpiredRelevantFlags(sponsorships: EthereumAddress[], maxFlagAgeSec: number): Promise<Flag[]> {
```
teogeb: Maybe could add `result` filter as a parameter, so that the module boundary is clear. In that case we could rename the method to `getExpiredFlags`.

packages/broker/src/plugins/operator/ContractFacade.ts
```diff
@@ -171,6 +182,41 @@ export class ContractFacade {
...
+                            id
+                        }
+                    }
+                }`
+            }
+        }
+        const parseItems = (response: { flags: Flag[] }): Flag[] => {
+            return response.flags ?? []
+        }
+        const flagsResults = this.theGraphClient.queryEntities<Flag>(createQuery, parseItems)
```
teogeb: Maybe rename as `result` has a explicit meaning in `Flag` (the status of the flag).

```suggestion
        const entities = this.theGraphClient.queryEntities<Flag>(createQuery, parseItems)
```

packages/broker/src/plugins/operator/closeExpiredFlags.ts
```diff
@@ -0,0 +1,33 @@
+import { EthereumAddress, Logger } from '@streamr/utils'
+import { ContractFacade, SponsorshipResult } from './ContractFacade'
+
+const logger = new Logger(module)
+
+export const closeExpiredFlags = async (
+    flagLifetime: number,
```
teogeb: To be in sync with the `ContractFacade` method

```suggestion
    maxFlagAgeSec: number,
```

packages/broker/src/plugins/operator/closeExpiredFlags.ts
```diff
@@ -0,0 +1,33 @@
...
+    operatorContractAddress: EthereumAddress,
+    contractFacade: ContractFacade
+): Promise<void> => {
+    logger.info('running closeExpiredFlags')
+
+    const minFlagStartTime = Math.floor(Date.now() / 1000) - flagLifetime
+
+    const sponsorships = (await contractFacade.getSponsorshipsOfOperator(operatorContractAddress))
+        .map((sponsorship: SponsorshipResult) => sponsorship.sponsorshipAddress)
```
teogeb, line 14: This mapping is needed only at line 22, could move it there

packages/broker/test/integration/plugins/operator/OperatorPlugin.test.ts
```diff
@@ -66,7 +66,7 @@ describe('OperatorPlugin', () => {
         expect(receivedMessages[0].content).toEqual({ foo: 'bar' })
         await subscriber.destroy()
         await publisher.destroy()
-    }, 30 * 1000)
+    }, 60 * 1000)
```
teogeb: Not part of this PR?

teogeb (commented):
First round of comments

.vscode/launch.json
samt1803: They are changes needed to easily run the tests in VSCode in debug-mode with breakpoints. I left them in because they could be helpful to other VSCode users. Should I remove them (then probably the whole file should be removed)

packages/broker/src/plugins/operator/ContractFacade.ts
samt1803: Actually the "Relevant" only stands for "in sponsorships I am staked in" and the sponsorships are a parameter anyway, so it is clear already. I'll just rename it to `getExpiredFlags`.

packages/broker/src/plugins/operator/closeExpiredFlags.ts
samt1803: I'll leave it, since it won't save anything, because in the length=0 case the mapping won't run anyway.

packages/broker/test/integration/plugins/operator/OperatorPlugin.test.ts
samt1803: reverted

packages/broker/src/plugins/operator/OperatorPlugin.ts
```diff
@@ -42,6 +43,10 @@ export interface OperatorPluginConfig {
     inspectRandomNode: {
         intervalInMs: number
     }
+    closeExpiredFlags: {
+        intervalInMs: number
+        maxFlagAgeSec: number
```
harbu: (nitpick) I would keep naming consistent with above
```suggestion
        maxFlagAgeInSec: number
```

packages/broker/src/plugins/operator/closeExpiredFlags.ts
```diff
@@ -0,0 +1,33 @@
...
+import { ContractFacade, SponsorshipResult } from './ContractFacade'
+
+const logger = new Logger(module)
+
+export const closeExpiredFlags = async (
+    maxFlagAgeSec: number,
+    operatorContractAddress: EthereumAddress,
+    contractFacade: ContractFacade
+): Promise<void> => {
+    logger.info('running closeExpiredFlags')
```
harbu: The module name (i.e. file name) will already be included in the log statement so could shorten to 
```suggestion
    logger.info('Start')
```

packages/broker/src/plugins/operator/closeExpiredFlags.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    const minFlagStartTime = Math.floor(Date.now() / 1000) - maxFlagAgeSec
+
+    const sponsorships = (await contractFacade.getSponsorshipsOfOperator(operatorContractAddress))
+        .map((sponsorship: SponsorshipResult) => sponsorship.sponsorshipAddress)
+    if (sponsorships.length === 0) {
+        logger.info('no sponsorships found')
+        return
+    }
+    logger.info(`found ${sponsorships.length} sponsorships`)
```
harbu: I would log this at `debug` level. Most of the time user doesn't need to know this, only when something is wrong and they want to look into it (hence they enable `debug` level)

```suggestion
    logger.debug(`Found ${sponsorships.length} sponsorships`)
    if (sponsorships.length === 0) {
        return
    }
```

packages/broker/src/plugins/operator/closeExpiredFlags.ts
```diff
@@ -0,0 +1,33 @@
...
+    const flags = await contractFacade.getExpiredFlags(sponsorships, maxFlagAgeSec)
+    logger.info(`found ${flags.length} flags`)
```
harbu: Again would go for `debug` level here and add "expired" as adjective.
```suggestion
    const flags = await contractFacade.getExpiredFlags(sponsorships, maxFlagAgeSec)
    logger.debug(`Found ${flags.length} expired flags`)
```

packages/broker/src/plugins/operator/closeExpiredFlags.ts
```diff
@@ -0,0 +1,33 @@
...
+    for (const flag of flags) {
+        const operatorAddress = flag.target.id
+        const sponsorship = flag.sponsorship.id
+        if (flag.flaggingTimestamp < minFlagStartTime) {
```
harbu: Doesn't the Graph query already exclude flags that would not be expired? Hence this check is not needed?

packages/broker/src/plugins/operator/closeExpiredFlags.ts
```diff
@@ -0,0 +1,33 @@
...
+            // voteOnFlag calls on expired flags will close the flag, ignoring the vote data
+            // also anyone can clal this function
```
harbu:
```suggestion
            // also anyone can call this function
```

packages/broker/src/plugins/operator/closeExpiredFlags.ts
```diff
@@ -0,0 +1,33 @@
...
+        return
+    }
+    logger.info(`found ${sponsorships.length} sponsorships`)
+    const flags = await contractFacade.getExpiredFlags(sponsorships, maxFlagAgeSec)
+    logger.info(`found ${flags.length} flags`)
+    for (const flag of flags) {
+        const operatorAddress = flag.target.id
+        const sponsorship = flag.sponsorship.id
+        if (flag.flaggingTimestamp < minFlagStartTime) {
+            // voteOnFlag calls on expired flags will close the flag, ignoring the vote data
```
harbu:
```suggestion
            // voteOnFlag is not used to vote here but to close the expired flag. The vote data gets ignored.
```

packages/broker/src/plugins/operator/closeExpiredFlags.ts
```diff
@@ -0,0 +1,33 @@
...
+            // also anyone can clal this function
+            await contractFacade.voteOnFlag(sponsorship, operatorAddress, false)
```
harbu: Here I would add an `info` level log to inform user that an action was taken
```suggestion
            logger.info('Close expired flag', { flag })
            await contractFacade.voteOnFlag(sponsorship, operatorAddress, false)
```

packages/broker/test/unit/plugins/operator/closeExpiredFlags.test.ts
```diff
@@ -0,0 +1,41 @@
+import { closeExpiredFlags } from '../../../../src/plugins/operator/closeExpiredFlags'
+import { ContractFacade, SponsorshipResult } from '../../../../src/plugins/operator/ContractFacade'
+import { mock } from 'jest-mock-extended'
+import { toEthereumAddress } from '@streamr/utils'
+
+const testAddress = toEthereumAddress('0xb85ea99a770d3d79d03855045cad830ef028a024') // some random address
```
harbu: I would differentiate between sponsorship address and operator address to make test assertion a bit stronger.

```
const sponsorshipAddress = toEthereumAddress('0xb85ea99a770d3d79d03855045cad830ef028a024')
const operatorAddress = toEthereumAddress('0x9b9dA31a92fA066A193fc78aE55Ee3eafa865aC7')
```

packages/broker/test/unit/plugins/operator/closeExpiredFlags.test.ts
```diff
@@ -0,0 +1,41 @@
...
+    ]
+})
+
+describe('closeExpiredFlags', () => {
+    
+    const flagLifetime = 1000
+
+    test('closes expired flags', async () => {
+        await closeExpiredFlags(flagLifetime, testAddress, contractFacadeMock)
+        expect(contractFacadeMock.voteOnFlag).toHaveBeenCalledTimes(1)
```
harbu:
```suggestion
        expect(contractFacadeMock.voteOnFlag).toHaveBeenCalledTimes(1)
        expect(contractFacadeMock.voteOnFlag).toHaveBeenCalledWith(sponsorshipAddress, operatorAddress, false) // 3rd boolean param is ignored in actual usage
```

harbu (changes requested):

packages/broker/src/plugins/operator/closeExpiredFlags.ts
```diff
@@ -0,0 +1,29 @@
...
+    }
+    const flags = await contractFacade.getExpiredFlags(sponsorships, maxFlagAgeSec)
+    logger.debug(`found ${flags.length} expired flags to close`)
+    for (const flag of flags) {
+        const operatorAddress = flag.target.id
+        const sponsorship = flag.sponsorship.id
+        // voteOnFlag is not used to vote here but to close the expired flag. The vote data gets ignored.
+        // Anyone can call this function at this point.
+        logger.info('Closing expired flag', { flag })
+        await contractFacade.voteOnFlag(sponsorship, operatorAddress, false)
```
teogeb: The purpose of the facade is to make the interface easy to use, and hide the implementation details:

```suggestion
        await contractFacade.closeVote(sponsorship, operatorAddress)
```

Also the comments can be moved there.

packages/broker/test/unit/plugins/operator/closeExpiredFlags.test.ts
```diff
@@ -0,0 +1,46 @@
+import { closeExpiredFlags } from '../../../../src/plugins/operator/closeExpiredFlags'
+import { ContractFacade, SponsorshipResult } from '../../../../src/plugins/operator/ContractFacade'
+import { mock } from 'jest-mock-extended'
+import { toEthereumAddress } from '@streamr/utils'
+
+const sponsorshipAddress = toEthereumAddress('0xb85ea99a770d3d79d03855045cad830ef028a024')
```
teogeb:
```
import { randomEthereumAddress } from '@streamr/test-utils'
const sponsorshipAddress = randomEthereumAddress()
...
```

If the addresses are just random addresses?

packages/broker/test/unit/plugins/operator/closeExpiredFlags.test.ts
```diff
@@ -0,0 +1,46 @@
...
+            sponsorshipAddress,
+            operatorCount: 2,
+            streamId: '0x123',
+        } as SponsorshipResult
+    ]
+})
+contractFacadeMock.getExpiredFlags.mockImplementation(async () => {
+    return [
+        {
+            id: 'flagid',
```
teogeb:
```suggestion
            id: 'flagId',
```

.vscode/launch.json
teogeb: Maybe could remove. We could add full launch.json in a separate PR if needed (this `launch.json` is just specific to `broker`). My suggestion would be to add this file to `.gitignore` so that everyone can easily setup their own launch configs.

packages/broker/src/plugins/operator/ContractFacade.ts
```diff
@@ -171,6 +182,41 @@ export class ContractFacade {
...
+                        sponsorship {
+                            id
+                        }
+                    }
+                }`
+            }
+        }
+        const parseItems = (response: { flags: Flag[] }): Flag[] => {
+            return response.flags ?? []
+        }
```
teogeb: This can be removed, as this is the default implementation of `parseItems`.

packages/broker/src/plugins/operator/OperatorPlugin.ts
```diff
@@ -42,6 +43,10 @@ export interface OperatorPluginConfig {
     inspectRandomNode: {
         intervalInMs: number
     }
+    closeExpiredFlags: {
+        intervalInMs: number
+        maxFlagAgeInSec: number
```
teogeb: The `flag` word is redundant, and I'd consider changing the data unit to milliseconds. For users perspective it is easier if all our config parameters have the same unit. (of course the field name tells the unit, but anyway easier to have just milliseconds everywhere)

```suggestion
        maxAgeInMs: number
```

packages/broker/src/plugins/operator/closeExpiredFlags.ts
```diff
@@ -0,0 +1,27 @@
...
+    logger.info('Start')
+
+    const sponsorships = (await contractFacade.getSponsorshipsOfOperator(operatorContractAddress))
+        .map((sponsorship: SponsorshipResult) => sponsorship.sponsorshipAddress)
+    logger.debug(`Found ${sponsorships.length} sponsorships`)
+    if (sponsorships.length === 0) {
+        return
+    }
+    const flags = await contractFacade.getExpiredFlags(sponsorships, maxFlagAgeSec)
+    logger.debug(`found ${flags.length} expired flags to close`)
```
harbu: Let's keep the capitalization here for consistency.
```suggestion
    logger.debug(`Found ${flags.length} expired flags to close`)
```

packages/broker/src/plugins/operator/closeExpiredFlags.ts
```diff
@@ -0,0 +1,27 @@
...
+    for (const flag of flags) {
+        const operatorAddress = flag.target.id
+        const sponsorship = flag.sponsorship.id
+        logger.info('Closing expired flag', { flag })
```
harbu: Let's use the base form verbs in this case, see: https://github.com/streamr-dev/network/blob/main/internal-docs/logging-practices.md#grammar-of-log-messages
```suggestion
        logger.info('Close expired flag', { flag })
```

packages/broker/src/plugins/operator/config.schema.json
```diff
@@ -127,6 +127,26 @@
...
+      "properties": {
+        "intervalInMs": {
+          "type": "integer",
+          "description": "How often to run (in milliseconds)",
+          "minimum": 0,
+          "default": 600000
+        },
+        "maxAgeInMs": {
+          "type": "integer",
+          "description": "The maximum age (in seconds) of a flag before it is expired",
```
teogeb:
```suggestion
          "description": "The maximum age (in milliseconds) of a flag before it is expired",
```

packages/broker/src/plugins/operator/config.schema.json
```diff
@@ -127,6 +127,26 @@
...
+          "minimum": 0,
+          "default": 86400
```
teogeb:
```suggestion
          "default": 86400000
```

packages/broker/src/plugins/operator/OperatorPlugin.ts
```diff
@@ -183,6 +188,18 @@ export class OperatorPlugin extends Plugin<OperatorPluginConfig> {
                 }
             }, this.pluginConfig.inspectRandomNode.intervalInMs, false, this.abortController.signal)
 
+            await scheduleAtInterval(async () => {
+                try {
+                    await closeExpiredFlags(
+                        this.pluginConfig.closeExpiredFlags.maxAgeInMs / 1000,
```
teogeb: Could do the unit transformation in the place where we need it in the different unit, i.e. in `getExpiredFlags` (these other components don't need to be aware of that implementation detail)

teogeb (approved):

harbu (approved):

PR #2075, opened 2023-11-19
feat(autocertifier): Kademliaid check support for autocertifier server

juslesan:
## Summary

The kademliaId of the peer using the autocertifier-client is now included to all rest requests to the autocertifier-server that require the use of the StreamrChallenger


juslesan:
Not needed as the StreamrChallenger does not need to set a presumed targetPeerDescriptor for the managed connection

PR #2079, opened 2023-11-20
NET-1159: avoid raising flag for operators that have already been flagged

harbu:

- Todo add unit tests (crunch)

### DIFF
--- a/packages/broker/src/plugins/operator/ContractFacade.ts
+++ b/packages/broker/src/plugins/operator/ContractFacade.ts
@@ -183,7 +183,6 @@ export class ContractFacade {
             id: string
             operator: {
                 id: string
-                flagsTargeted: Array<{ id: string }>
             }
         }
         const createQuery = (lastId: string, pageSize: number) => {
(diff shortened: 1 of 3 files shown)


packages/broker/src/plugins/operator/ContractFacade.ts
```diff
@@ -187,6 +195,12 @@ export class ContractFacade {
                                 id
                                 operator {
                                     id
+                                    flagsTargeted(where: {
```
teogeb: We could use hasOpenFlag for the selected node (and therefore no changes to this method)

packages/broker/src/plugins/operator/ContractFacade.ts
```diff
@@ -327,6 +343,33 @@ export class ContractFacade {
...
+        }
+        const queryResult = this.theGraphClient.queryEntities<{ id: string }>(createQuery)
+
+        const flags = await collect(queryResult, 1)
+        if (flags.length > 0) {
+            logger.debug('Found open flag', { flag: flags[0] })
+            return true
+        } else {
+            return false
+        }
```
teogeb, line 361: Always log the flags? the can simplify to this: `return (flags.length > 0)`

packages/broker/src/plugins/operator/ContractFacade.ts
harbu: Simplified to this in a4943bdd3f94fba185979350c37057b5070218de

packages/broker/src/plugins/operator/ContractFacade.ts
harbu: I'll keep it as it is for now

PR #2080, opened 2023-11-20
fix(dht): `WebsocketServer` memory leak

teogeb:
Improve `WebsocketServer` cleanup. When we explicitly remove `httpServer` listeners, the instance doesn't leak memory immediately after `WebsocketServer#stop` is called. 

There wasn't a permanent memory leak: e.g. `await wait(100)` before the check in the unit test would also make the test pass.

## Open questions

Why the change causes this tests to fail:

dht-browser:
```
1) starts and stops
     WebsocketServer
     ReferenceError: require is not defined
```


ptesavol (approved):
Agreed.

juslesan (approved):
What is the status of this?

PR #2088, opened 2023-11-20
test(dht): Fix `ReplicateData` checks

teogeb:
## Summary

Fixed some issues in `ReplicateData` integration test.

## Test functionality change

The "Data replicates to the last remaining node if all other nodes leave gracefully" test case has been changed (and renamed) so that it just stops the node which originally stored the data.

## Fixes

Some changes:
- fix data entry assertion
  - the `getEntry` returned a `Map` and was therefore always truthy
- await `joinDht` calls
  - there was no implications when this was not awaited?
- store random peer descriptor instead of entry point descriptor
  - just a separation of concerns
- remove logging


packages/dht/test/integration/ReplicateData.test.ts
```diff
@@ -67,6 +61,7 @@ describe('Replicate data from node to node in DHT', () => {
         await Promise.all([
             ...nodes.map(async (node) => await node.stop())
         ])
+        await entryPoint.stop()
```
ptesavol: This is not needed, on line 46 entrypoint is pushed among the nodes.

packages/dht/test/integration/ReplicateData.test.ts
```diff
@@ -109,20 +93,19 @@ describe('Replicate data from node to node in DHT', () => {
...
         })
 
         logger.info(NUM_NODES + ' nodes joining layer0 DHT')
         await Promise.all(
-            nodes.map((node) => {
-                if (keyFromPeerDescriptor(node.getLocalPeerDescriptor()) != '0') {
-                    node.joinDht([entrypointDescriptor])
+            nodes.map(async (node, i) => {
+                if (i !== 0) {
+                    await node.joinDht([entrypointDescriptor])
```
ptesavol: Wouldn't 'return' be more appropriate here since there is the promise.all()?

packages/dht/test/integration/ReplicateData.test.ts
```diff
@@ -136,32 +119,21 @@ describe('Replicate data from node to node in DHT', () => {
...
             }
-
             logger.info(getNodeIdFromPeerDescriptor(node.getLocalPeerDescriptor()) + hasDataMarker)
         })
 
         const closestNode = nodesById.get(PeerID.fromValue(closest[0].getPeerDescriptor().nodeId).toKey())!
 
-        // @ts-expect-error private field
-        expect(closestNode.localDataStore.getEntry(dataKey).size).toBeGreaterThanOrEqual(1)
+        // TODO assert the content?
```
ptesavol: Yeah, why not add a id comparison, since the data is a peerdescriptor?

ptesavol (changes requested):
Only a couple of comments, otherwise looks good to me.

packages/dht/test/integration/ReplicateData.test.ts
teogeb: Refactored so that `entryPoint` is kept separate from other nodes

packages/dht/test/integration/ReplicateData.test.ts
teogeb: `joinDht` doesn't return anything (i.e. just `Promise<void>`), so `await` communicates the intention better

harbu:
The test most likely fails because multiple nodes leave the network in a relatively small time. That leads to network splitting and therefore replicating data can't work like we assert in the test (we assert that after `n-1` nodes have have left, the last node contains the data).

Action points:
- Create a ticket for this issue.
- Part of that ticket would be to create an automated test that focuses on the downscale network partitioning issue.
- Then come up with a solution that makes that test pass (and thus fixes the issue)

teogeb:
The ticket: https://linear.app/streamr/issue/NET-1236/investigate-dht-splitting-when-the-vast-majority-of-the-nodes-leave-in

juslesan (approved):
LGTM, if the downward scaling test is still flaky it would make sense to increase the redundancy factor of replication when nodes are gracefully leaving

PR #1171, opened 2023-04-11
ci: Linting with Type Information

teogeb:
Utilize TypeScript type information in the lint process:
https://tseslint.com/linting/typed-linting/

Currently we get lot of errors:
| count | rule |
| ----- | ---- |
| 654 | [`no-unsafe-member-access`](https://typescript-eslint.io/rules/no-unsafe-member-access) |
| 458 | [`no-unsafe-assignment`](https://typescript-eslint.io/rules/no-unsafe-assignment) |
| 448 | [`no-unsafe-argument`](https://typescript-eslint.io/rules/no-unsafe-argument) |
| 298 | [`no-floating-promises`](https://typescript-eslint.io/rules/no-floating-promises) |
| 233 | [`require-await`](https://typescript-eslint.io/rules/require-await) |
| 226 | [`no-unsafe-call`](https://typescript-eslint.io/rules/no-unsafe-call) |
| 131 | [`restrict-template-expressions`](https://typescript-eslint.io/rules/restrict-template-expressions) |
| 121 | [`no-unsafe-return`](https://typescript-eslint.io/rules/no-unsafe-return) |
| 120 | [`unbound-method`](https://typescript-eslint.io/rules/unbound-method) |
|  83 | [`restrict-plus-operands`](https://typescript-eslint.io/rules/restrict-plus-operands) |
|  82 | [`no-misused-promises`](https://typescript-eslint.io/rules/no-misused-promises) |
|  81 | [`no-unnecessary-type-assertion`](https://typescript-eslint.io/rules/no-unnecessary-type-assertion) |
|  23 | [`await-thenable`](https://typescript-eslint.io/rules/await-thenable) |

At least the promise-related rules are most likely usable. Some of the no-unsafe rules are not maybe applicable if we want to continue to use the `any` type.

PR #2089, opened 2023-11-21
fix(dht, trackerless-network): Strict boolean expressions

teogeb:
Use strict boolean expressions, i.e. `if (foo !== undefined)` instead of `if (foo)`. 

Some expressions were fixed and some issues were annotated with TODO comments.

## Background

There has been several problems related to boolean expression which handle nullable values in checks/assertions:
- currently we have some boolean expressions which are unintentionally always `true`
  - see "`better check (currently this condition is always true)`" comments added in this PR
- there was recently a problem when we used nullable `HandshakeError` enum https://github.com/streamr-dev/network/pull/1904
  - in many enums on one of the values is `0`, and therefore lenient `null` checks can incorrectly ignore the value
 - the `MigrateData` test is broken https://github.com/streamr-dev/network/pull/2088

## Future improvements

- These warnings were found with https://typescript-eslint.io/rules/strict-boolean-expressions/ `eslint` rule. We could enable the rule if we add TypeScript support to our `eslint` setup (see https://github.com/streamr-dev/network/pull/1171). 
- Could also  run the `eslint` rule with `allowString=false` and `allowNumber=false` to get some more potential places to fix?
- Could also change some `if` statements which use objects to have explicit undefined check (i.e. `if (obj)` to `if (obj !== undefined`)) 
  - use `if \([!a-zA-Z0-9.]+\)` regexp to find those

### DIFF
--- a/packages/dht/src/connection/ConnectionLockHandler.ts
+++ b/packages/dht/src/connection/ConnectionLockHandler.ts
@@ -26,7 +26,7 @@ export class ConnectionLockHandler {
     }
 
     public isLocalLocked(id: PeerIDKey, lockId?: LockID): boolean {
-        if (!lockId) {
+        if (lockId === undefined) {
             return this.localLocks.has(id)
         } else {
             return this.localLocks.has(id) && this.localLocks.get(id)!.has(lockId)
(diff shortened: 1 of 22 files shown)


juslesan (approved):
Opening PR's for TODO comments ASAP

PR #2090, opened 2023-11-21
refactor(dht): Remove obsolete `setActive` parameters

teogeb:
Removed obsolete `setActive` parameter from `PeerDiscovery#addContact` and `Finder#addContact` callbacks.

## Open questions

- is "active" state needed in practice?
- what is "contacted" flag of `ContactState`


juslesan (approved):
Contacted = your node has atleast attempted to contact the node
Active = your node has successfully contacted the node

Contacted is important for 

- Joining, where nodes are set as contacted when trying to contacft a new node
- RoutingSessions, where the SortedContactList is used to sort and set nodes as contacted

Active is mostly important for when k-bucket pings are attempted. If the pings fail then the we cycle new nodes to the kBuckets.

PR #2092, opened 2023-11-21
ClientWebSocket connect selfSigned default false

juslesan:
## Summary

`selfSigned` in `AbstractClientWebsocket#connect` is now a required paramater. Renamed to `allowSelfSignedCertificate`.


packages/dht/src/connection/websocket/ClientWebsocket.ts
```diff
@@ -23,7 +23,7 @@ export class ClientWebsocket extends EventEmitter<ConnectionEvents> implements I
         this.connectionId = new ConnectionID()
     }
 
-    public connect(address: string, selfSigned?: boolean): void {
+    public connect(address: string, selfSigned = false): void {
```
teogeb: Is it important to know for each connection request whether the certificate is `selfSigned`? If so, this should be required parameter. (And also the parameter name should tell that this is about a certificate.)

juslesan: At the moment there is only one case where selfSigned is set to true. It happens on the first ConnectivityRequest to an entry point when a node is started with auto certification enabled

packages/dht/src/connection/websocket/ClientWebsocket.ts
```diff
@@ -27,8 +27,7 @@ export class ClientWebsocket extends EventEmitter<ConnectionEvents> implements I
         this.connectionId = new ConnectionID()
     }
 
-    // TODO explicit default value for "selfSigned" or make it required
-    public connect(address: string, selfSigned?: boolean): void {
+    public connect(address: string, selfSigned: boolean): void {
```
teogeb: Does this setting apply only if we are using secure connection (i.e. `wss`)? 

Does the `false` mean that we don't allow that the certificate is self-signed?  And if `true`, we allow any certificate (self-signed or normal)?
- if this is the case, could rename to `allowSelfSignedCertificate`?

harbu (approved):

teogeb (approved):

PR #2094, opened 2023-11-21
refactor(DHT): Router error enum typing

juslesan:
## Summary

Router related errors are now passed as protobuf enums between nodes in the network.

This is a breaking change.


packages/dht/src/dht/routing/FindRpcRemote.ts
```diff
@@ -22,7 +22,7 @@ export class FindRpcRemote extends Remote<IFindRpcClient> {
         const options = this.formDhtRpcOptions()
         try {
             const ack = await this.getClient().routeFindRequest(message, options)
-            if (ack.error.length > 0) {
+            if (ack.error !== undefined) {
```
teogeb, line 27: Should the log statement below (line 26) show a human readable error, or is the enum number ok? (Maybe it is as this is a `debug` level log)

packages/dht/protos/DhtRpc.proto
```diff
@@ -177,7 +177,15 @@ message RouteMessageWrapper {
...
 message RouteMessageAck {
   string requestId = 1;
-  string error = 2;
+  optional RouteMessageError error = 2;
+}
+
+enum RouteMessageError {
+  NO_TARGETS = 0;
+  DUPLICATE = 1;
+  // TODO: can this be removed? If DhtNode is already stopped the server side requests
```
teogeb, line 186: Should we decide what to do with this before we release the next version?

packages/dht/src/dht/routing/Router.ts
```diff
@@ -167,8 +162,8 @@ export class Router implements IRouter {
                     `Failed to send (routeMessage: ${this.serviceId}) to ${getNodeIdFromPeerDescriptor(routedMessage.destinationPeer!)}`
                 )
             }
-            logger.trace('noCandidatesFound sessionId: ' + session.sessionId)
-            return createRouteMessageAck(routedMessage, RoutingErrors.NO_CANDIDATES_FOUND)
+            logger.trace('no targets sessionId: ' + session.sessionId)
```
teogeb:
```suggestion
            logger.trace('no targets', { sessionId: session.sessionId })
```

packages/dht/src/dht/routing/FindRpcRemote.ts
juslesan: logging the error number should be fine. There's no need to necessarily log this even on the debug level so reducing to trace

packages/dht/protos/DhtRpc.proto
juslesan: Could try to remove the stopped guards, but it could lead to memory leaks etc. Might be worth a try.

packages/dht/protos/DhtRpc.proto
```diff
@@ -177,7 +177,15 @@ message RouteMessageWrapper {
 
 message RouteMessageAck {
   string requestId = 1;
-  string error = 2;
+  optional RouteMessageError error = 2;
+}
+
+enum RouteMessageError {
+  NO_TARGETS = 0;
```
teogeb, line 184: Does each message has one target? I.e. should this be `NO_TARGET`?

juslesan: Yes the message has a single target but from the perspective of a router it checks if it knows anyone it can propagate the message to. The source node will also route the message to two peers initially

teogeb (approved):

PR #2096, opened 2023-11-21
refactor(network): stricter typing for doParallelHandshakes

juslesan:
## Summary

`doParallelHandshakes` now takes in two `HandshakeRemoteRpc` objects as parameter instead of a list. This gets rid of the possibility of undefined values in promise handling


packages/trackerless-network/src/logic/neighbor-discovery/Handshaker.ts
```diff
@@ -80,7 +80,11 @@ export class Handshaker implements IHandshaker {
         const exclude = excludedIds.concat(this.config.targetNeighbors.getIds())
         const targetNeighbors = this.selectParallelTargets(exclude)
         targetNeighbors.forEach((contact) => this.ongoingHandshakes.add(getNodeIdFromPeerDescriptor(contact.getPeerDescriptor())))
-        return this.doParallelHandshakes(targetNeighbors, exclude)
+        if (targetNeighbors.length >= 2) {
```
teogeb, line 76: Does this change the handshake functionality in any way?

packages/trackerless-network/src/logic/neighbor-discovery/Handshaker.ts
```diff
@@ -94,17 +98,18 @@ export class Handshaker implements IHandshaker {
...
 
-    private async doParallelHandshakes(targets: HandshakeRpcRemote[], excludedIds: NodeID[]): Promise<NodeID[]> {
-        const results = await Promise.allSettled(
-            Array.from(targets.values()).map(async (target: HandshakeRpcRemote, i) => {
-                const otherNode = i === 0 ? targets[1] : targets[0]
-                const otherNodeId = otherNode ? getNodeIdFromPeerDescriptor(otherNode.getPeerDescriptor()) : undefined
-                return this.handshakeWithTarget(target, otherNodeId)
-            })
-        )
+    private async doParallelHandshakes(targetOne: HandshakeRpcRemote, targetTwo: HandshakeRpcRemote, excludedIds: NodeID[]): Promise<NodeID[]> {
```
teogeb: Could we get rid of the mutation of `excludedIds` parameter? What this and other `Promise<NodeID[]>` methos actually return?

PR #2101, opened 2023-11-22
fix(DHT): Ensure that Messages are routed to existing connections only

juslesan:
## Summary

The goal of the routing is to not open any connections for routing. Instead the message is passed over existing connections for faster propagation. This change ensures that new connections are not opened for the routing paths

Added improvements to the routing to ensure that routed messages are sent through existing connections.
Routed messages are passed through network faster as nodes will no longer attempt to open new connections in corner cases. Also if a routing node has left the network it is no longer possible to attempt to connect to it for routing. Instead the request fails fast with the doNotConnect flag and the message is then routed to another node.

## Changes
 
- Routing related RemoteRpc classes now use the doNotConnect flag
- RoutingSession removes unconnected nodes from the contactList

### DIFF
--- a/packages/dht/src/dht/routing/RoutingSession.ts
+++ b/packages/dht/src/dht/routing/RoutingSession.ts
@@ -116,7 +116,7 @@ export class RoutingSession extends EventEmitter<RoutingSessionEvents> {
         if (this.ongoingRequests.has(peerId.toKey())) {
             this.ongoingRequests.delete(peerId.toKey())
         }
-        const contacts = this.findMoreContacts()
+        const contacts = this.getRoutablePeers()
         if (contacts.length === 0 && this.ongoingRequests.size === 0) {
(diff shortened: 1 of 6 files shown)


packages/dht/src/dht/routing/RoutingSession.ts
```diff
@@ -172,8 +172,13 @@ export class RoutingSession extends EventEmitter<RoutingSessionEvents> {
 
     findMoreContacts(): RemoteContact[] {
         logger.trace('findMoreContacts() sessionId: ' + this.sessionId)
-        // the contents of the connections might have changed between the rounds
-        // addContacts() will only add new contacts that were not there yet
+        // Remove stale contacts that may have been removed from connections
```
teogeb, line 175: Maybe the method name should be changed to reflect the new functionality?

teogeb (approved):

PR #2103, opened 2023-11-22
NET-1188: Update node start logs for v1.0

mondoreale:
## Summary

I update copy and remove mentions of the Network Explorer.

### DIFF
--- a/packages/broker/src/broker.ts
+++ b/packages/broker/src/broker.ts
@@ -48,7 +48,6 @@ export const createBroker = async (configWithoutDefaults: Config): Promise<Broke
             const mnemonic = generateMnemonicFromAddress(toEthereumAddress(brokerAddress))
 
             logger.info(`Welcome to the Streamr Network. Your node's generated name is ${mnemonic}.`)
-            logger.info(`View your node in the Network Explorer: https://streamr.network/network-explorer/nodes/${encodeURIComponent(nodeId)}`)
             logger.info(`Network node ${nodeId} running`)
             logger.info(`Ethereum address ${brokerAddress}`)
 
--- a/packages/broker/src/broker.ts
+++ b/packages/broker/src/broker.ts
@@ -33,7 +33,7 @@ export const createBroker = async (configWithoutDefaults: Config): Promise<Broke
             return streamrClient
         },
         start: async () => {
-            logger.info(`Start broker version ${CURRENT_VERSION}`)
+            logger.info(`Start Streamr node version ${CURRENT_VERSION}`)
             await Promise.all(plugins.map((plugin) => plugin.start(streamrClient)))
             const httpServerEndpoints = plugins.flatMap((plugin: Plugin<any>) => {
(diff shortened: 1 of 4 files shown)


packages/broker/src/broker.ts
```diff
@@ -33,7 +33,7 @@ export const createBroker = async (configWithoutDefaults: Config): Promise<Broke
             return streamrClient
         },
         start: async () => {
-            logger.info(`Start broker version ${CURRENT_VERSION}`)
+            logger.info(`Start Streamr node version ${CURRENT_VERSION}`)
```
harbu, line 36: I guess `Streamr node` is the new name of broker? Wonder if we should still keep `broker` in parenthesis since the package name still has broker, i.e.
```suggestion
            logger.info(`Start Streamr (broker) node version ${CURRENT_VERSION}`)
```

packages/broker/src/broker.ts
```diff
@@ -48,9 +48,8 @@ export const createBroker = async (configWithoutDefaults: Config): Promise<Broke
             const mnemonic = generateMnemonicFromAddress(toEthereumAddress(brokerAddress))
 
             logger.info(`Welcome to the Streamr Network. Your node's generated name is ${mnemonic}.`)
-            logger.info(`View your node in the Network Explorer: https://streamr.network/network-explorer/nodes/${encodeURIComponent(nodeId)}`)
             logger.info(`Network node ${nodeId} running`)
-            logger.info(`Ethereum address ${brokerAddress}`)
+            logger.info(`Node address ${brokerAddress}`)
```
harbu, line 52: We used to have an emphasis on the wallet (Polygon) address being the identity of the node. This is no longer the case, since `nodeId` is no longer associated with the wallet address but is its own thing entirely. Thus I think we should add the word `wallet` here to make that distinction clear, dunno do what do you think?
```suggestion
            logger.info(`Node wallet address ${brokerAddress}`)
```

harbu (approved):

packages/broker/src/broker.ts
fonty1: Yes please. We should drop 'broker' at every opportunity in favour of Streamr node.

packages/broker/src/broker.ts
fonty1: It's all a bit subjective but I'll add my perspective:
In the Docs/Hub we've been using `Wallet` in the context of the Owner wallet, where the account has moreso of a value holding agenda. The Node address has moreso of a signing agenda, so I think `address` or `account` is more appropriate. Since the terminology we use in our Docs onboarding and the Hub is `Node address` I think we should adopt it here too.

harbu: Makes sense

harbu:
Feel free to merge as far as I'm concerned 👍

packages/broker/src/broker.ts
fonty1: Sorry Eric, I didn't read this properly. 

Lets walk away from the Broker terminology. It should be `Start Streamr node version`
Hopefully one day even the package name :P

PR #2106, opened 2023-11-22
fix(dht, trackerless-network): Fix some TypeScript `eslint` warnings

teogeb:
Needs https://github.com/streamr-dev/network/pull/2107 so that Broker tests pass?

Fixed or annotated with TODO comments these `eslint` warnings:
- `@typescript-eslint/require-await`
- `@typescript-eslint/no-unnecessary-type-assertion`
- `@typescript-eslint/no-floating-promises`
- `@typescript-eslint/no-unsafe-assignment`
- `@typescript-eslint/restrict-plus-operands`
- `@typescript-eslint/no-misused-promises`

These warnings were found with TypeScript eslint

## Future improvements

- Currently our `eslint` doesn't use TypeScript, and therefore these rules aren't used when we run `npm run eslint.` We should enable most of these rules when we add TypeScript support to our eslint setup (see https://github.com/streamr-dev/network/pull/1171).


harbu (approved):

harbu:
Those promises could use catches for sure

PR #2110, opened 2023-11-23
fix(DHT): Fix external store storer PeerDescriptor

juslesan:
## Summary

When storing data with the external store operation, the incorrect PeerDescriptor was assigned for the storer. This could create scenarios where the external node used to run the store operation would set itself as the storer of data. When multiple request would be handled for the same keys to store data the stored would be overwritten. Additionally, the original storer would not be able to delete the data with its own PeerDescriptor

## Future improvements

- Stop using setStreamPartEntryPoints in trackerless-network tests. There could be an explicit test that tests the feature but by the default behaviour should be used in all other tests

### DIFF
--- a/packages/dht/test/integration/Store.test.ts
+++ b/packages/dht/test/integration/Store.test.ts
@@ -1,7 +1,7 @@
 import { LatencyType, Simulator } from '../../src/connection/simulator/Simulator'
 import { DhtNode } from '../../src/dht/DhtNode'
 import { PeerDescriptor } from '../../src/proto/packages/dht/protos/DhtRpc'
-import { createMockConnectionDhtNode, waitConnectionManagersReadyForTesting } from '../utils/utils'
+import { createMockConnectionDhtNode, createMockPeerDescriptor, waitConnectionManagersReadyForTesting } from '../utils/utils'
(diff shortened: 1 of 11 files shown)


packages/dht/test/integration/Store.test.ts
```diff
@@ -47,23 +47,43 @@ describe('Storing data in DHT', () => {
...
-            const fetchedDescriptor = Any.unpack(entry.data!, PeerDescriptor)
-            expect(areEqualPeerDescriptors(fetchedDescriptor, entrypointDescriptor)).toBeTrue()
+            const foundData = Any.unpack(entry.data!, PeerDescriptor)
+            expect(areEqualPeerDescriptors(foundData, storedData)).toBeTrue()
         })
     }, 90000)
+
+    it('storing with explicit storer PeerDescriptor', async () => {
+        const storingNode = getRandomNode()
+        const dataKey = PeerID.fromString('3232323e12r31r3')
```
teogeb, line 74: Are these just random `kademliaIds`? I.e. could be created with `createRandomKademliaId()`

teogeb: We'll do this in a separate PR

teogeb (approved):

PR #2113, opened 2023-11-23
feat(dht): Improve `DhtNode#onKBucketRemoved`

teogeb:
A draft. Maybe better title which describes the improvement.


teogeb:
This was an attempt to fix network split issues related to flaky data replication

PR #2118, opened 2023-11-24
refactor(dht): Rename `DataEntry` fields

teogeb:
Renamed:
- `DataEntry#storer`, `StoreDataRequest#storer` -> `creator`
- `DataEntry#storerTime`, `StoreDataRequest#storerTime` -> `creatorTime`
- `DataEntry#kademliaId`, `StoreDataRequest#kademliaId,` `DeleteDataRequest#kademliaId`, `ExternalFindDataRequest#kademliaID` -> `key`

Also reordered `DataEntry` fields.

## Future improvements

- Simplify the code block at `StoreRpcLocal:129-135`


juslesan (approved):
The storer fields contains the list of nodes that the data was successfully stored at. It is still possible that the data has been replicated from the original storers to more appropriate nodes as the network lives on. 

Yes the comment is correct as in this cases the function inquestion returns `true`

PR #2119, opened 2023-11-24
refactor(dht): Rename `PeerDescriptor#kademliaId`

teogeb:
## Background

When we'll start to use hybrid-DHT, the `PeerDescritor` id is no longer a `kademliaId`. Instead the `PeerDescriptor` id is an abstract id, and the `kademliaId` is the hash of that id.

## Changes

Renamed:
- `PeerDescriptor#kademliaId` -> `nodeId`
- `NetworkPeerDescriptor#id` -> `nodeId`
- `ClosestPeersRequest#kademliaId` -> `nodeId`

## Compatibility

Added minimalistic compatibility layer, which allows users to define and use `NetworkPeerDescriptor` in the legacy format. See commit 767356e. We can revert that commit after testnet/mainnet starts (as we instruct the users to change their entry point configs anyway). It may be obsolete even sooner if we implement these, and therefore won't have legacy configs in practice:
 - https://linear.app/streamr/issue/NET-1133/update-config-wizard-to-create-10-configs
 - https://linear.app/streamr/issue/NET-1004/use-streamrconfig-for-streamr-client-production-values
    
## Future improvements

- rename `id` -> `nodeId` for each entry point in `@streamr/config`
- all `sourcePeer`/`targetPeer`/`destinationPeer` in `Router` (and related) classes should be renamed to just `source`/`target`


docs/docs/guides/become-an-operator.md
```diff
@@ -89,15 +89,15 @@ The only information they will need from you is your address. Please do not enga
...
+                        "nodeId": "e1",
                         "websocket": {
                             "host": "entrypoint-1.streamr.network",
                             "port": 40401,
                             "tls": true
                         }
                     },
                     {
-                        "id": "e2",
+                        "node": "e2",
```
harbu:
```suggestion
                        "nodeId": "e2",
```

packages/client/src/utils/utils.ts
```diff
@@ -139,8 +154,7 @@ export function convertPeerDescriptorToNetworkPeerDescriptor(descriptor: PeerDes
         throw new Error('nodeType "virtual" not supported')
     }
     return {
-        ...omit(descriptor, 'kademliaId'),
-        id: binaryToHex(descriptor.kademliaId),
+        nodeId: binaryToHex(descriptor.nodeId),
         type: descriptor.type === NodeType.NODEJS ? NetworkNodeType.NODEJS : NetworkNodeType.BROWSER
```
harbu: Doesn't necessarily include `websocket` field (or others)?

(Missing unit test for this?)

PR #2122, opened 2023-11-26
feat(DHT): NET-1192 DHT store delete operation within Find

juslesan:
## Summary

The delete operation is now run within a find operation. Previously a find operation was done to find the closest nodes and then individual direct requests were made to each node.

This has two important optimisations / improvememts:

- Data stored along the path to the node will be marked as deleted
- Network nodes can delete themselves as entry points when gracefully leaving the network #2124 

## Future improvements
- Rename fields / methods according to the new names given to the Rpc objects.


harbu:
Let's address naming questions and future improvements in ticket https://linear.app/streamr/issue/NET-1205/rename-find-to-recursiveoperation-in-dht-package (in testnet-2)

PR #2123, opened 2023-11-26
fix(dht): Self-connection guards in `DhtNode`

teogeb:
When we receive `connection` event from transport, we have been checking if the connection is from this node. We check that fact both when we receive `connected` event from `ITransport`, and when we query the initial connections.

If the connection is from this node we now return from the methods and don't  update the `connections` state.

This is a cherry-pick from https://github.com/streamr-dev/network/pull/1803, i.e. the original author is @ptesavol.

## Open question

- Is the error state possible in practice? If we receive a connection from self, is that a bug? (In that case we should fix the bug and could ignore this PR. We could also remove the related error logging.)


juslesan (approved):

ptesavol (approved):
LGTM

PR #2124, opened 2023-11-26
feat(trackerless-network): NET-1192 nodes delete stream entry point data on destroy

juslesan:
## Summary

Network nodes will now delete stream entry point data of themselves from the DHT when gracefully leaving the network
Fixes flaky client test `end-to-end/publish-subscribe.test.ts`

## Future improvements

- If waitForCompletion is set to false, the information could be set to the FindRequest. Then nodes along the path could avoid sending responses
- Add functionality for waiting for routed messages to be fanned out.


packages/dht/src/dht/find/Finder.ts
```diff
@@ -126,14 +127,21 @@ export class Finder implements IFinder {
...
+                    15000
+                )
+            } catch (err) {
+                logger.debug(`doRouteFindRequest failed with error ${err}`)
+            }
+        } else {
+            this.doRouteFindRequest(routeMessage, excludedPeer)
+            // Wait for delete operation to be fanned out
+            // TODO: Should wait for the delete operation to be fanned out?
+            await wait(50)
```
harbu:
```suggestion
            // Wait for delete operation to be sent out by the router
            // TODO: Add a feature to wait for the router to pass the message?
            await wait(50)
```

PR #2134, opened 2023-11-28
feat(trackerless-network): NET-1190 store new entry points for streams when entry points leave

juslesan:
## Summary

Non-entry point nodes store themselves as entry points for streams if neighboring entry points leave the stream. This change makes sure that entry points exist for streams in the long run. Before the change, nodes would store themselves as entry points on stream partitions if necessary only when initially joining the stream. This could lead to long running parts of the stream networks being split into their own stream that never finds newly joining nodes in the worst cases. 

## Changes

RandomGraphNode now takes in two new function parameters:

- `localNodeIsEntryPoint` -> returns true if self is acting as entry point in the stream partition, is passed to leaveNotices so receiveing nodes can react
- `onEntryPointLeaveDetected` -> called when receiving a leave notice from a stream partition entry point

The default behaviour of `onEntryPointLeaveDetected` is for the caller to store itself as an entry point for the stream partition.

## Future improvements

- Add more testing, currently only testing the worst path where all stream entry points leave at once
- Investigate why the test is running so slow in CI.


packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -111,6 +111,7 @@ message InterleaveNotice {
 
 message LeaveStreamPartNotice {
   string streamPartId = 1;
+  bool sourceIsStreamEntryPoint = 2;
```
teogeb, line 114:
```suggestion
  bool isEntryPoint = 2;
```

packages/trackerless-network/src/logic/EntryPointDiscovery.ts
```diff
@@ -119,11 +117,12 @@ export class EntryPointDiscovery {
     }
 
     async storeSelfAsEntryPointIfNecessary(currentEntrypointCount: number): Promise<void> {
-        if (this.abortController.signal.aborted) {
+        if (this.abortController.signal.aborted || this.localPeerDescriptorStored) {
```
teogeb, line 120: Would it be possible do the check in the caller? I.e. check whether the local node is an entry point, and call the method only if it is not?

packages/trackerless-network/src/logic/DeliveryRpcRemote.ts
```diff
@@ -19,9 +19,10 @@ export class DeliveryRpcRemote extends RpcRemote<IDeliveryRpcClient> {
         })
     }
 
-    leaveStreamPartNotice(): void {
+    leaveStreamPartNotice(localNodeIsEntryPoint: boolean): void {
```
teogeb, line 22: Would it be possible to pass a callback function to `DeliveryRpdRemote`? We'd query the `boolean` value by calling that method when needed.

PR #2137, opened 2023-11-29
refactor(dht): Simplify `StoreRpcLocal`

teogeb:
Some simplification for the implementation:
- use `findIndex` helper from `lodash`
- `shouldReplicateDataToNewNode` doesn't mutate anything, just answers to the question (the mutation moved to where the method is called)
- move `destroy()` to the end of the file
- `onNewContact` method instead of getting event emitter as a parameter

In `replicateDataToNeighborsIfNeeded` removed a redundant duplicated check and combined the two blocks execution branches.


juslesan (approved):
replicateDataToNeighborsIfNeeded could be renamed to something that more properly matches its functionality. For example just `replicateDataToNeighbors`

PR #2139, opened 2023-11-29
docs: use environment preset

teogeb:
Use new environment config feature of client to configure Mumbai defaults.

### DIFF
--- a/docs/docs/guides/become-an-operator.md
+++ b/docs/docs/guides/become-an-operator.md
@@ -144,52 +144,7 @@ Below is the template you can use to override and replace the contents of your c
         "auth": {
             "privateKey": "NODE_PRIVATE_KEY"
         },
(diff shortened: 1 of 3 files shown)


fonty1 (approved):
LGTM. @teogeb has this landed in a testnet release yet? 
Will the old way (the `controlLayer` approach) continue to work if `environment` is omitted?

teogeb:
> LGTM. @teogeb has this landed in a testnet release yet? Will the old way (the `controlLayer` approach) continue to work if `environment` is omitted?

@fonty1 This will be included in the `testnet2` release (which has not released yet, but will be released soon). Therefore would be good to merge this now so that this is included when we deploy docs for `testnet2`.

fonty1:
Great, thanks @teogeb

PR #2140, opened 2023-11-29
refactor(trackerless-network): Rename `StreamrNode`

teogeb:
Renamed `StreamrNode` class to `ContentDeliveryManager`. The instance is not a node in any of the networks. Instead it orchestrates the stream part delivery layer operations by managing `StreamPartDelivery` instances (each of which is a node in layer 2 or a proxy client).

## Other changes

Moved and rephrased the startup logging "Starting new StreamrNode..." from `StreamrNode` to `NetworkStack`.

## Future improvements

- Would it make sense if `NetworkStack` would provide the `createLayer1Node()` and/or `createRandomGraphNode()` functions?
- We could unify the naming in `NetworkOptions`? The fields and interfaces don't seem match. Actually after this PR there is negative correlation as the type of `layer0` is a `...Node`, and the type of `networkNode` is `...Manager`. (Also if there is no difference between `Options` and `Config`, we could maybe use only either of those). 
```ts
export interface NetworkOptions {
    layer0?: DhtNodeOptions
    networkNode?: ContentDeliveryManagerConfig
    ...
}
```


juslesan (commented):
So are we now calling the component that composes the joint operations of layer1 (stream part peer discovery) layer2 (stream part message delivery/transport) DeliveryLayer. This could be a confusing naming convention as Layer2 is what facilitates delivery of messages in streams

harbu:
What about `DeliveryLayerManager` or `DeliveryLayerState`? Thoughts on whether these would be more appropriate names @juslesan ?

juslesan:
> What about DeliveryLayerManager or DeliveryLayerState? Thoughts on whether these would be more appropriate names

I think we need to think about the vocabulary a little bit concerning the stream specific layers. Going a bit philosophical here...

The main idea here is the separation of the network control and content delivery related tasks.

The main purpose of the layer1 networks is to provide peer discovery for the streams. This is strictly a network control task for streams. (Additionally there is the StreamEntryPoint discovery component which is used to discover Layer1 networks)

The main purpose of the layer2 networks is to provide content delivery for the streams. This task has network control and content delivery related tasks. Creating the graph in layer2 is a control task, where as message propagation over the stream is a content delivery task.

The StreamrNode class composes the required components for a node to operate in a stream. From a users perspective operating in streams is just publishing and subscribing to data. (There is also inspection, but that is arguably a control task and users are not interested in manually invoking the operation)

What I'm getting at here is that don't like associating the word 'layer' as in `DeliveryLayerManager` or `DeliveryLayerState` to what the current StreamrNode does. The rational is to avoid confusion between adding another "layer" to act as an umbrella term for all stream operations. The goal of the class is not to create a new network layer. Rather, to participate in the required network layers to operate in streams.

The current StreamrNode class provides an interface for joining streams and to broadcast/listen to messages in the given streams. As the goal of all operations underneath is to provide content delivery over the network I would prefer a name such as `ContentDeliveryManager`.

juslesan (approved):

PR #2144, opened 2023-11-30
fix(dht): Use `storeMaxTtl` config option

teogeb:
Use the `storeMaxTtl` config option instead of hardcode `TTL_MAX`.

Before this PR the data storage TTL handling was done in two places: 
- in `LocalDataStore` using the hardcoded: it scheduled the auto-deletion to happen between `TTL_MIN` and `TTL_MAX`
- in `StoreRpcLocal` using the configured value:  it mutated the entry's `ttl` if it was greater than `storeMaxTtl`.

Now the TTL handling is done only in `LocalDataStore`. The`StoreRpcLocal` no longer mutates the entry. That way the `storeMaxTtl` option just controls how long this node keeps the entry, and the option doesn't affect the entries which are replicated to other nodes via this node.

Removed also hardcoded `TTL_MIN` limit. If we need the limit, we should add a `DhtNode` config option for that.

### DIFF
--- a/packages/dht/src/dht/store/StoreRpcLocal.ts
+++ b/packages/dht/src/dht/store/StoreRpcLocal.ts
@@ -29,7 +29,6 @@ interface DataStoreConfig {
     localPeerDescriptor: PeerDescriptor
     localDataStore: LocalDataStore
     serviceId: ServiceID
-    maxTtl: number
     highestTtl: number
     redundancyFactor: number
     getNodesClosestToIdFromBucket: (id: Uint8Array, n?: number) => DhtNodeRpcRemote[]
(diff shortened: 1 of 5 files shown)


packages/dht/src/dht/store/StoreRpcLocal.ts
```diff
@@ -184,8 +181,7 @@ export class StoreRpcLocal implements IStoreRpc {
 
     // RPC service implementation
     async storeData(request: StoreDataRequest): Promise<StoreDataResponse> {
-        const ttl = Math.min(request.ttl, this.maxTtl)
-        const { key, data, createdAt, creator } = request
+        const { key, data, creator, createdAt, ttl } = request
```
juslesan, line 184: Could be useful to allow users to set a max TTL for any data that they store as before. Now the requestor is in control

juslesan (approved):

PR #2145, opened 2023-11-30
refactor(dht, trackerless-network): Use config options and named constants

teogeb:
Use config options or named constants instead of magic numbers. 

## TODO

Currently the magic numbers are just annotated with TODO comments. We should go through all these comments and resolve them by:
- using a method to get the value (e.g. `getTimeout`)
- using a config option if that value should be configurable by the end-users
- referring to an existing constant
- introducing new constant

For some constants it could be good to comment how fine-tuned the value is. E.g.:
- it should have exactly that value (and why that is the case)
- we are currently using a valid value, but it would make sense to optimize it after we get some benchmarks results


ptesavol (approved):
Agreed

PR #2148, opened 2023-12-01
fix(dht): Recursive router can exclude peer

teogeb:
Re-added `excludedPeer` parameter to `Router#doRouteMessage`. It was incorrectly removed in https://github.com/streamr-dev/network/pull/2033.

The parameter is used at least in `RecursiveOperationManager#doRouteRequest`. That method is called e.g. from `ExternalApiRpcLocal#externalFindData` which sets `senderPeerDescriptor` as the excluded peer.


ptesavol (approved):
Please merge.

PR #2150, opened 2023-12-01
NET-998 query entrypoints from The Graph

hpihkala:
## Summary

`StreamrClient` can now query a set of recent operator heartbeats from the network subgraph, and pass their peer descriptors to the network node upon creation. The discovered entrypoints are added to the array of configured entrypoints. The entrypoint discovery can be disabled via config, but it is enabled by default. The default upper limit is 20 discovered entrypoints.

## Note about config and layers

Note the new config options are under `network`, as for the users it's clearer if entrypoint-related configs are in proximity of each other. However that's 'wrong' as the discovery feature is in the client layer.

As the client anyway transforms config in `NetworkNodeFacade` to something that the network layer understands, in the future it could be better to move the user-facing entrypoint configs to be under `client` instead of `client.network.controlLayer`. This config refactor was not done in this PR in order to avoid breaking changes.


packages/client/src/NetworkNodeFacade.ts
```diff
@@ -68,21 +70,25 @@ export class NetworkNodeFactory {
 export class NetworkNodeFacade {
 
     private cachedNode?: NetworkNodeStub
+    private cachedEntryPoints?: NetworkPeerDescriptor[]
```
teogeb: Caching is maybe not needed. The data is only queried once to initialize the node, and we never need to re-initialize the node again.

packages/client/src/NetworkNodeFacade.ts
```diff
@@ -236,7 +242,16 @@ export class NetworkNodeFacade {
         this.eventEmitter.once(eventName, listener as any)
     }
 
-    getEntryPoints(): PeerDescriptor[] {
-        return this.config.network.controlLayer.entryPoints!.map(peerDescriptorTranslator)
+    async getEntryPoints(): Promise<PeerDescriptor[]> {
```
teogeb: Could make this method `private`

packages/client/src/registry/OperatorRegistry.ts
```diff
@@ -0,0 +1,56 @@
...
+        loggerFactory: LoggerFactory
+    ) {
+        this.theGraphClient = theGraphClient
+        this.logger = loggerFactory.createLogger(module)
+    }
+
+    async findRandomNetworkEntrypoints(
+        maxEntryPoints: number = 5,
+        maxQueryResults: number = 50, 
+        maxHeartbeatAgeHours: number = 24,
```
teogeb: In the future these could be fields of `EntrypointDiscovery` and the defaults defined in `config.schema.json`.

PR #2153, opened 2023-12-03
Fix websocket requestConnection()

ptesavol:
~**Notice, includes a protobuf change, remember to run proto.sh in packages.dht**~

* Fixed WebSocketConnectionRequest to work regardless of the peerIDs of the requester/requestee
* added test case for verifying that the above mentioned fix works
* ~made WebSocketConnectionRequest into a notification~
* added handing for promise returned when calling requestConnection()

EDIT: breaking changes removed from this PR, to be done separately


packages/dht/src/connection/websocket/WebsocketConnector.ts
```diff
@@ -100,16 +99,19 @@ export class WebsocketConnector {
 
     private registerLocalRpcMethods(config: WebsocketConnectorConfig) {
         const rpcLocal = new WebsocketConnectorRpcLocal({
-            canConnect: (peerDescriptor: PeerDescriptor) => config.canConnect(peerDescriptor),
             connect: (targetPeerDescriptor: PeerDescriptor) => this.connect(targetPeerDescriptor),
+            hasConnection: (targetPeerDescriptor: PeerDescriptor): boolean => {
+                const peerKey = keyFromPeerDescriptor(targetPeerDescriptor)
+                return this.connectingConnections.has(peerKey)
```
teogeb, line 105: What is the difference between the `WebsocketConnector` fields `connectingConnections` and `ongoingConnectRequests`? It it ok that this checks only `connectingConnections`?

packages/dht/src/connection/websocket/WebsocketConnector.ts
```diff
@@ -293,12 +301,8 @@ export class WebsocketConnector {
         if (this.ongoingConnectRequests.has(peerId.toKey())) {
             const ongoingConnectRequest = this.ongoingConnectRequests.get(peerId.toKey())!
             ongoingConnectRequest.attachImplementation(serverWebsocket)
-            if (targetPeerDescriptor && !areEqualPeerDescriptors(this.localPeerDescriptor!, targetPeerDescriptor)) {
```
teogeb, line 296: Does this deletion bring back any problems which we fixed by implementing https://github.com/streamr-dev/network/pull/1904?

PR #2156, opened 2023-12-04
fix(dht): Update of data `stale` state in `StoreRpcLocal#onNewContact`

teogeb:
Before this PR we incorrectly updated the data entry state to true when `StoreRpcLocal#onNewContact` was called for the closest node. 

This bug was introduced in https://github.com/streamr-dev/network/pull/2137. In `StoreRpcLocal#shouldReplicateDataToNewNode` there is a guard at line `102`. After the refactoring we didn't have any special handling for the guard case, and therefore we incorrectly called `this.localDataStore.setStale(..., true)` if we returned from `StoreRpcLocal#shouldReplicateDataToNewNode` because of the guard.

Now we have a method `replicateAndUpdateStaleStateIfClosest`. If the local node is the closest node for the data:
- it always updates the stale state
- if the new node is quite close to the data (within `redundancyFactor`), it replicates the data to that new node

## TODO

Does this change do the same thing, or something better than https://github.com/streamr-dev/network/pull/2159? If something is missing, we should include that into this PR or cherry-pick that PR 2159 to `streamr-1.0`.

## Open questions

Why we update our `stale` state only if we are the closest node for the data? 

## Other changes

Also renamed the `getNodesClosestToIdFromBucket` to `getClosestNeighborsTo` in `StoreRpcLocal` as that is the name of `PeerManager` method it calls.

### DIFF
--- a/packages/dht/src/dht/DhtNode.ts
+++ b/packages/dht/src/dht/DhtNode.ts
@@ -281,7 +281,7 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
             highestTtl: this.config.storeHighestTtl,
             redundancyFactor: this.config.storageRedundancyFactor,
             localDataStore: this.localDataStore,
-            getNodesClosestToIdFromBucket: (id: Uint8Array, n?: number) => {
+            getClosestNeighborsTo: (id: Uint8Array, n?: number) => {
                 return this.peerManager!.getClosestNeighborsTo(id, n)
             },
             rpcRequestTimeout: this.config.rpcRequestTimeout
--- a/packages/dht/src/dht/store/StoreRpcLocal.ts
+++ b/packages/dht/src/dht/store/StoreRpcLocal.ts
@@ -31,7 +31,7 @@ interface DataStoreConfig {
     serviceId: ServiceID
     highestTtl: number
     redundancyFactor: number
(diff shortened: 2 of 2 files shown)


juslesan (changes requested):
Few comments related to my own debugging of the issue from yesterday. Setting staleness for data seems to be broken when scaling up in general:

1. `replicateAndUpdateStaleStateIfClosest` Should always set `stale = false` for locally stored data **if it is the closest node**. This is the expected correct behaviour.

2. If the local node is NOT the closest node to the data in `replicateAndUpdateStaleStateIfClosest` and is not even with the closest `redundancyFactor` to the data, the data should be set stale. This can be done in an else branch around line 107:
`this.localDataStore.setStale(dataEntry.key, peerIdFromPeerDescriptor(dataEntry.creator!), !this.selfIsOneOfClosestPeers(dataEntry.key))
`

 I would recommend implementing these changes before merging (you can also test with the benchmark should be quite a performance improvement)
 
3. Future consideration: Should take a closer look at setting staleness for when the network is scaling down.

juslesan:
Data is now set stale on newContact only in cases where the local node is no longer one of the `redundancyFactor` closest nodes. The primary storer will also no longer set any data as stale or not stale on new contacts. Instead data that has been received by replication or in another state of the network is kept as stale by the primary storer.

juslesan (approved):

PR #2162, opened 2023-12-04
fix(DHT) optimise parallel joining with the use of a shared contacted peers set

juslesan:
## Summary

Optimised the parallel join operations by giving parallel join operations a shared set of contacted peers.

This is an important optimisation especially on layer1 where all requests are routed through layer0. In the worst cases we would be sending 4x more routed messages during joins than necessary.

Saw significant reduction in CPU usage in `benchmark/first-message.ts`

## Future improvements

- Investigate suspected increased in CPU usage for individual join operations. For example the DHT integration test `RoutedMessages.test.ts` uses 150% of cpu in a rather small network of 33 peers.
- The router could be refactored to have its own contacted set similar to the `DiscoverySessions`. 
- Parallel rejoins should use a shared set of contactedPeers as well

### DIFF
--- a/packages/dht/src/dht/discovery/DiscoverySession.ts
+++ b/packages/dht/src/dht/discovery/DiscoverySession.ts
@@ -18,6 +18,7 @@ interface DiscoverySessionConfig {
     parallelism: number
     noProgressLimit: number
     peerManager: PeerManager
+    // Note that contacted peers will be mutated by the DiscoverySession or other parallel sessions
     contactedPeers: Set<PeerIDKey>
 }
 
--- a/packages/dht/src/dht/discovery/DiscoverySession.ts
+++ b/packages/dht/src/dht/discovery/DiscoverySession.ts
@@ -18,6 +18,7 @@ interface DiscoverySessionConfig {
     parallelism: number
     noProgressLimit: number
     peerManager: PeerManager
+    contactedPeers: Set<PeerIDKey>
 }
 
 export class DiscoverySession {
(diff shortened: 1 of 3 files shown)


packages/dht/src/dht/discovery/PeerDiscovery.ts
```diff
@@ -36,7 +37,12 @@ export class PeerDiscovery {
         this.abortController = new AbortController()
     }
 
-    async joinDht(entryPointDescriptor: PeerDescriptor, doAdditionalRandomPeerDiscovery = true, retry = true): Promise<void> {
+    async joinDht(
+        entryPointDescriptor: PeerDescriptor,
+        contactedPeers: Set<PeerIDKey>,
```
teogeb, line 57: Maybe add a comment that this `Set` will be mutated by `DiscoverySession` (and maybe TODO comment to implement it in the future without mutation)

packages/dht/src/dht/discovery/DiscoverySession.ts
```diff
@@ -48,7 +48,7 @@ export class DiscoverySession {
         }
         logger.trace(`Getting closest peers from contact: ${getNodeIdFromPeerDescriptor(contact.getPeerDescriptor())}`)
         this.outgoingClosestPeersRequestsCounter++
-        this.contactedPeers.add(contact.getPeerId().toKey())
+        this.config.contactedPeers.add(contact.getPeerId().toKey())
```
teogeb, line 52: In Router we use `setContacted` feature of `SortedContactList`, but maybe the same patterns is not applicable here we have separate lists for each session?

juslesan: The router could be refactored to have its own contacted set similar to the `DiscoverySessions`. 

As the SortedContactList in the RoutingSession is always per session the old setContacted works there as intended.

teogeb (approved):

PR #2164, opened 2023-12-04
feat(trackerless-network): NET-1190 store new entry points for streams when entry points leave (re-opened)

juslesan:
## Summary

Non-entry point nodes store themselves if necessary as entry points for streams if neighboring entry points leave the stream. This change makes sure that entry points exist for streams in the long run. Before the change, nodes would store themselves as entry points on stream partitions if necessary only when initially joining the stream. This could lead to long running parts of the stream networks being split into their own stream that never finds newly joining nodes in the worst cases. 

## Changes

RandomGraphNode now takes in a new function parameter:

- `isLocalNodeEntryPoint` -> returns true if self is acting as entry point in the stream partition, is passed to leaveNotices so receiveing nodes can react

RandomGraphNode now emits a new event when it detects that entry points leave a stream.
- `entryPointLeaveDetected`
- StreamrNode listens to the event and will store itself as an entry point if necessary.

## Future improvements

- Add more testing, currently only testing the worst path where all stream entry points leave at once
- Investigate why the test is running so slow in CI.
- Stream split detection and recovery could be moved to its own module outside of the EntryPointDiscovery class.


packages/trackerless-network/src/logic/EntryPointDiscovery.ts
```diff
@@ -69,16 +69,14 @@ export class EntryPointDiscovery {
     private readonly config: EntryPointDiscoveryConfig
     private readonly storeInterval: number
     private readonly networkSplitAvoidedNodes: Set<NodeID> = new Set()
-
+    private localPeerDescriptorStored = false
```
ptesavol: rename to make intention clear

packages/trackerless-network/src/logic/EntryPointDiscovery.ts
```diff
@@ -175,6 +174,10 @@ export class EntryPointDiscovery {
         logger.trace(`Network split avoided`)
     }
 
+    public localNodeIsEntryPoint(): boolean {
```
ptesavol: rename to same logic as line 72

packages/trackerless-network/src/logic/RandomGraphNode.ts
```diff
@@ -56,6 +56,8 @@ export interface StrictRandomGraphNodeConfig {
     numOfTargetNeighbors: number
     inspector: IInspector
     temporaryConnectionRpcLocal: TemporaryConnectionRpcLocal
+    localNodeIsEntryPoint: () => boolean
```
ptesavol: reaname to follow same logic as above

packages/trackerless-network/src/logic/RandomGraphNode.ts
```diff
@@ -279,7 +287,7 @@ export class RandomGraphNode extends EventEmitter<Events> {
         }
         this.abortController.abort()
         this.config.proxyConnectionRpcLocal?.stop()
-        this.config.targetNeighbors.getAll().map((remote) => remote.leaveStreamPartNotice())
+        this.config.targetNeighbors.getAll().map((remote) => remote.leaveStreamPartNotice(this.config.localNodeIsEntryPoint()))
```
ptesavol: reanme to follow same logic as above

ptesavol (approved):

PR #2166, opened 2023-12-04
fix(dht): Close connections after timeout

juslesan:
## Summary

Fixed a critical issue where in cases where the `ManagedConnection's` send time outs we do not close the connections in many cases. This was a major issue in all cases that required signalling to form the connection ie. requested ws connections and webrtc connections. In some cases the websocket client connect would also be stuck in connecting state forever. The clears up that issue as well.

## Future improvements

- Investigate why the `integration/store.test.ts` required an increased afterEach timeout after the fix.
- Are both the closing and stopped fields needed in `ManagedConnection`

### DIFF
--- a/packages/dht/src/connection/ManagedConnection.ts
+++ b/packages/dht/src/connection/ManagedConnection.ts
@@ -56,7 +56,8 @@ export class ManagedConnection extends EventEmitter<Events> {
     private localPeerDescriptor: PeerDescriptor
     protected outgoingConnection?: IConnection
     protected incomingConnection?: IConnection
-
+    private created = Date.now()
+    
     constructor(
         localPeerDescriptor: PeerDescriptor,
         connectionType: ConnectionType,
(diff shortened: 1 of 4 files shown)


packages/dht/src/connection/ManagedConnection.ts
```diff
@@ -57,6 +57,9 @@ export class ManagedConnection extends EventEmitter<Events> {
     protected outgoingConnection?: IConnection
     protected incomingConnection?: IConnection
 
+    // TODO: Temporary debug variable, should be removed in the future.
+    private created = Date.now()
+
```
teogeb, line 62: Do we need both `stopped` and `closing` fields?

teogeb (approved):

packages/dht/src/connection/ManagedConnection.ts
juslesan: Yeah could take a look at that in the future

PR #2172, opened 2023-12-06
fix(proto-rpc): Fix RpcCommunicator mem leak under timeouts

juslesan:
## Summary

If a request timed out it was not removed from the outgoingRequests map. This fix removes a major memory leak.

## Future improvements

- the new onTimeout handler could be made into a onCompleted function which would be a call back to remove OngoingRequests from the RpcCommunicator

### DIFF
--- a/packages/proto-rpc/src/RpcCommunicator.ts
+++ b/packages/proto-rpc/src/RpcCommunicator.ts
@@ -40,12 +40,12 @@ class OngoingRequest {
     private deferredPromises: ResultParts
     private timeoutRef: NodeJS.Timeout
 
-    constructor(deferredPromises: ResultParts,
-        timeout: number) {
+    constructor(deferredPromises: ResultParts, timeout: number, onTimeout: () => void) {
         this.deferredPromises = deferredPromises
         this.timeoutRef = setTimeout(() => {
             const error = new Err.RpcTimeout('Rpc request timed out', new Error())
             this.rejectDeferredPromises(error, StatusCode.DEADLINE_EXCEEDED)
+            onTimeout()
         }, timeout)
     }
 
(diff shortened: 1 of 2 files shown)


packages/proto-rpc/src/RpcCommunicator.ts
```diff
@@ -40,12 +40,12 @@ class OngoingRequest {
...
 
-    constructor(deferredPromises: ResultParts,
-        timeout: number) {
+    constructor(deferredPromises: ResultParts, timeout: number, onTimeout: () => void) {
         this.deferredPromises = deferredPromises
         this.timeoutRef = setTimeout(() => {
             const error = new Err.RpcTimeout('Rpc request timed out', new Error())
             this.rejectDeferredPromises(error, StatusCode.DEADLINE_EXCEEDED)
+            onTimeout()
         }, timeout)
```
teogeb:
```suggestion
    constructor(deferredPromises: ResultParts, timeout: number, onTimeout?: () => void) {
        this.deferredPromises = deferredPromises
        this.timeoutRef = setTimeout(() => {
            const error = new Err.RpcTimeout('Rpc request timed out', new Error())
            this.rejectDeferredPromises(error, StatusCode.DEADLINE_EXCEEDED)
            onTimeout?.()
        }, timeout)
```

packages/proto-rpc/src/RpcCommunicator.ts
```diff
@@ -188,7 +188,7 @@ export class RpcCommunicator extends EventEmitter<RpcCommunicatorEvents> {
     private onOutgoingMessage(rpcMessage: RpcMessage, deferredPromises?: ResultParts, callContext?: ProtoCallContext): void {
         if (this.stopped) {
             if (deferredPromises) {
-                const ongoingRequest = new OngoingRequest(deferredPromises, 1000)
+                const ongoingRequest = new OngoingRequest(deferredPromises, 1000, () => {})
```
teogeb: Why we timeout these requests after a second but don't add the `onTimeout` handler?

teogeb (approved):

PR #2175, opened 2023-12-07
fix(dht): `PeerManager` excluded nodes

teogeb:
Fix to `PeerManager#getClosestNeighborsTo` and `PeerManager#getClosestContactsTo`: we pass `excludedNodeIds` to the `SortedContactList` instead of filtering the `getClosestContacts` explicitly.
- As we filtered the items after the `getClosestContacts` query, we possibly reduced the item count below the limit even if there were enough nodes.
- Example: 
  - there are `20` contacts in total
  - we want `10` closest contacts and call `PeerManager#getClosestContacts(10)`
  - the `closest.getClosestContacts(limit)` returns `10` contacts
  - but we filtered out e.g. `2` excluded nodes
  - and therefore the `PeerManager` method returned only `8` closest nodes

## Other changes

Also added unit test cases for `SortedContactList`.

### DIFF
--- a/docs/docs/guides/how-to-run-streamr-node.md
+++ b/docs/docs/guides/how-to-run-streamr-node.md
@@ -244,7 +244,7 @@ If you're running a node to become an Operator, then you could now jump to [Step
 -   Run `streamr-broker` to start the node! You should start to see logging similar to this:
 
 ```
-INFO [2023-10-31T17:42:30.897] (broker              ): Start broker version ...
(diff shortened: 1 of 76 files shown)


juslesan (approved):
As the number of nodes returned by `PeerManager#getClosestNeighborsTo` is increased, the overall CPU usage of the network might have slightly increased. This was also observed by flakyness being reduced after the simulated time was turned off.

If this is the case we can reduce the limit parameter to adjust the CPU use.

PR #2176, opened 2023-12-07
refactor(dht): Use `sortedIndexBy` in `SortedContactList#removeContact`

teogeb:
Optimize `SortedContactList#removeContact` by using `sortedIndexBy` to find the index of removed item.

## Open questions

- Why `test/integration/Store.test.ts` is flaky after this change? (now it fails ~20-30% of runs, and in `streamr-1.0` it is not flaky)

### DIFF
--- a/packages/dht/src/dht/contact/SortedContactList.ts
+++ b/packages/dht/src/dht/contact/SortedContactList.ts
@@ -1,27 +1,29 @@
-import { PeerID, PeerIDKey } from '../../helpers/PeerID'
+import { PeerID } from '../../helpers/PeerID'
 import { ContactState, Events } from './ContactList'
 import { sortedIndexBy } from 'lodash'
 import EventEmitter from 'eventemitter3'
 import { getDistance } from '../PeerManager'
+import { NodeID, areEqualNodeIds } from '../../helpers/nodeId'
+import { hexToBinary } from '@streamr/utils'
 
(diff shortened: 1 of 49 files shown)


packages/dht/src/dht/contact/SortedContactList.ts
```diff
@@ -161,8 +161,7 @@ export class SortedContactList<C extends { getNodeId: () => NodeID }> extends Ev
     public removeContact(id: NodeID): boolean {
         if (this.contactsById.has(id)) {
             const removed = this.contactsById.get(id)!.contact
-            // TODO use sortedIndexBy?
-            const index = this.contactIds.findIndex((nodeId) => areEqualNodeIds(nodeId, id))
+            const index = sortedIndexBy(this.contactIds, id, (id: NodeID) => { return this.distanceToReferenceId(id) })
             this.contactIds.splice(index, 1)
```
ptesavol, line 165: Please add here a check that the contactsById[index] is indeed the "id" that we are trying to remove. If not, log an error or use a breakpoint to find out if the reason for the test becoming flaky is here.

ptesavol (changes requested):
Please debug the test that became flaky, hints for doing this in code comment.

teogeb:
The original implementation is most like more efficient than this implementation.

PR #2179, opened 2023-12-07
Bump gas price when voting

hpihkala:
Should we do something like this?


teogeb (approved):

jtakalai (approved):
I guess this wasn't the 30% bump?

hpihkala:
No, this can be closed. It's replaced on `testnet-one` by the 30% bump and on `streamr-1.0` by Teo's cleaner version that uses some gas utilities from `streamr-client`.

PR #2181, opened 2023-12-08
fix(network): Do not use virtual time

juslesan:
## Summary

Stop using virtual time in trackerless-network, virtual time causes significantly increased CPU use and causes increased test flakyness


teogeb (approved):
Could stop using the virtual time also in `dht` package for the same reason?

PR #2182, opened 2023-12-08
streamr-1.0 fix-start-and-stop-of-autocertifier-client

ptesavol:
This is the fix-start-and-stop-of-autocertifier-client pull request moved to base upon streamr-1.0 instead of testnet-one

* make test/end-to-end/RecoveryFromFailedAutoCertification.test.ts to stop the entrypoint last to prevent nodes from starting to reconnect through the stopped entrypoint at the end of the test
* rename WebsocketConnector's autoCertify() method to  startAutoCertifierClient() as this is what the method does
* stop AutoCertifierClientFacade in WebsocketConnector if it was started
* start AutoCertifierClient in AutoCertifierClientFacade using an abortable withTimeout() instead of waitForEvent3 (using a async function as waitForEvent3 predicate caused the timeout to keep running in case the predicate threw an exception, which was the reason for test/end-to-end/RecoveryFromFailedAutoCertification.test.ts to hang) 
* because of the above changes test/end-to-end/RecoveryFromFailedAutoCertification.test.ts does not hang anymore

### DIFF
--- a/packages/dht/src/connection/websocket/AutoCertifierClientFacade.ts
+++ b/packages/dht/src/connection/websocket/AutoCertifierClientFacade.ts
@@ -1,14 +1,14 @@
 import {
     AutoCertifierClient,
     HasSessionRequest,
-    HasSessionResponse, 
+    HasSessionResponse,
     CertifiedSubdomain,
     SERVICE_ID as AUTO_CERTIFIER_SERVICE_ID,
     HasSession
 } from '@streamr/autocertifier-client'
 import { ListeningRpcCommunicator } from '../../transport/ListeningRpcCommunicator'
-import { Logger, waitForEvent3 } from '@streamr/utils'
-import { ITransport } from '../../transport/ITransport' 
+import { Logger, withTimeout } from '@streamr/utils'
+import { ITransport } from '../../transport/ITransport'
 
 const START_TIMEOUT = 60 * 1000
 
@@ -20,14 +20,14 @@ const defaultAutoCertifierClientFactory = (
 ) => new AutoCertifierClient(
     configFile,
     wsServerPort,
-    autoCertifierUrl, 
+    autoCertifierUrl,
     (_serviceId: string, rpcMethodName: string, method: HasSession) => {
         autoCertifierRpcCommunicator.registerRpcMethod(
             HasSessionRequest,
             HasSessionResponse,
             rpcMethodName,
             method
-        )                       
+        )
     }
 )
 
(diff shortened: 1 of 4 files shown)


packages/dht/src/connection/websocket/AutoCertifierClientFacade.ts
```diff
@@ -36,7 +36,6 @@ export interface IAutoCertifierClient {
     stop(): void
     on(eventName: string, cb: (subdomain: CertifiedSubdomain) => void): void
 }
-
```
teogeb, line 39: Keep this line

packages/dht/src/connection/websocket/WebsocketConnector.ts
```diff
@@ -334,10 +334,12 @@ export class WebsocketConnector {
 
     public async destroy(): Promise<void> {
         this.abortController.abort()
-        this.rpcCommunicator.destroy()
 
-        const requests = Array.from(this.ongoingConnectRequests.values())
-        await Promise.allSettled(requests.map((conn) => conn.close(false)))
+        if (this.autoCertifierClient) {
+            await this.autoCertifierClient.stop()
```
teogeb, line 339:
```suggestion
            this.autoCertifierClient.stop()
```

packages/dht/src/connection/websocket/AutoCertifierClientFacade.ts
```diff
@@ -77,13 +77,18 @@ export class AutoCertifierClientFacade {
...
-        await Promise.all([
-            waitForEvent3(this.autoCertifierClient as any, 'updatedCertificate', START_TIMEOUT),
-            this.autoCertifierClient.start()
-        ])
+
+        // waiting for 'updatedCertificate' event is not needed here because
+        // all paths of this.autoCertifierClient.start() will block until 'updatedCertificate' 
+        // event is emitted or an exception is thrown
+
+        await withTimeout(this.autoCertifierClient.start(), START_TIMEOUT,
```
teogeb, line 85: Maybe we should separate the starting of the client and obtaining the initial certificate. It doesn't take long time to start the client, but in practice we wait for the certificate?

Also could rename `AutoCertifierClientFacade` to `AutoCertifierFacade` as we don't seem to use any of the client's functionality methods (we just call the `start` and `stop` methods).

-> maybe in a separate PR

packages/dht/src/connection/websocket/WebsocketConnector.ts
```diff
@@ -334,10 +334,12 @@ export class WebsocketConnector {
 
     public async destroy(): Promise<void> {
         this.abortController.abort()
-        this.rpcCommunicator.destroy()
 
-        const requests = Array.from(this.ongoingConnectRequests.values())
-        await Promise.allSettled(requests.map((conn) => conn.close(false)))
```
teogeb, line 340: Why lines `339`-`340` were removed?

teogeb (changes requested):

PR #2184, opened 2023-12-08
feat(client, broker): [NET-1221] Use `ethers` overriders in operator plugin

teogeb:
## Summary

Operator plugin use higher `gasPrice` value in Polygon environment.

- Added new public `getEthersOverrides()` method to `StreamrClient`. Broker uses the method and gives the overrides as a parameter when it submits transactions in `ContractFacade`. As client's default value for `highGasPriceStrategy` flag is `true` in Polygon environment, the client returns a higher-than-normal `gasPrice` value in the `Overrides` object.
- Changed `highGasPrice` strategy to use +30% increment instead of increasing the gas price by `10000000000`.

## Small changes

Also changed `getEthersOverrides` to NOT to use `highGasPrice` strategy if explicit `gasPrice` is defined in client's config overrides.

## Fixes

In the client the high gas price strategy should have been the enabled by default in the `polygon` environment, but it was not enabled because we used case-sensitive chain name comparison. The name changed from `polygon` to `Polygon` in https://github.com/streamr-dev/network/pull/2130.
-  https://github.com/streamr-dev/network/blob/e88eacad4010d26b76583b0339d2a3dbb8f9e80e/packages/client/src/Ethereum.ts#L36

## Client config refactoring (breaking change)

Restructred client's `contracts.ethereumNetworks` config block because client no longer uses separate chains (`mainchain` and `sidechain`). This restructing allowed us also to add schema validation for the config block.

The option name and structure was changed from
```ts
ethereumNetworks?: Record<string, EthereumNetworkConfig>
export interface EthereumNetworkConfig {
    chainId: number
    overrides?: Overrides
    highGasPriceStrategy?: boolean
}
```
to
```ts
ethereumNetwork?: EthereumNetworkConfig
export interface EthereumNetworkConfig {
    overrides?: Overrides
    highGasPriceStrategy?: boolean
}
```

## Future improvements

Should add tests more automated tests for this feature. For the automated tests we'd need some abstraction to mock the contracts e.g. in `ContractFacade.ts`/`Ethereum.ts`.


packages/client/src/Ethereum.ts
```diff
@@ -31,24 +31,22 @@ const getRpcProviders = (connectionInfo: ChainConnectionInfo, pollInterval?: num
...
-const getOverrides = (chainName: string, provider: Provider, config: Pick<StrictStreamrClientConfig, 'contracts'>): Overrides => {
-    const chainConfig = config.contracts.ethereumNetworks[chainName]
-    if (chainConfig === undefined) { return {} }
+export const getEthersOverrides = (config: Pick<StrictStreamrClientConfig, 'contracts'>): Overrides => {
+    const chainConfig = config.contracts.ethereumNetwork
     const overrides = chainConfig.overrides ?? {}
-    if (chainConfig.highGasPriceStrategy) {
-        const gasPriceStrategy = (estimatedGasPrice: BigNumber) => estimatedGasPrice.add('10000000000') 
+    if ((chainConfig.highGasPriceStrategy) && (chainConfig.overrides?.gasPrice === undefined)) {
+        const primaryProvider = getStreamRegistryChainProviders(config)[0]
```
jtakalai, line 42: this feels a bit indirect. If we use highGasPriceStrategy, then we presumably use it in a network, and we should probably use the provider of that network, not necessarily "stream registry chain provider". I'd pass the provider as argument to this function the use it directly.

Typical use would look like so:
```
        const ethersOverrides = this.getEthersOverrides(this.operatorContract.provider)
        ethersOverrides.gasLimit = '200000'
        await (await this.operatorContract.voteOnFlag(
            sponsorship,
            targetOperator,
            voteData,
            ethersOverrides
        )).wait()
```
or if we have just one provider in use in that instance, use that.

jtakalai (approved):

packages/client/src/Ethereum.ts
jtakalai: of course if we know we only work in one network/chain, then sure, let's keep it that way.

PR #2185, opened 2023-12-08
refactor(dht): Refactor `StoreManager`

teogeb:
Changes to `StoreManager`:
- Fixed boolean usage of `selfIsPrimaryStorer` in `replicateDataToNeighbors`. The name and the content of the variable didn't match. **Check: it was maybe already used correctly at line 211?**
- Simplified `StoreManagerConfig#getClosestNeighborsTo` return type
- Simplified and optimized the logic of `selfIsOneOfClosestPeers` method
- Removed `doNotConnect` parameter which was always `false` (using explicit `false` in `StoreRpcRemote#replicateData` instead)

Added unit test. Added `StoreManagerConfig#createRpcRemote` dependency field so that we can mock that in the tests.

## Other changes

- Replaced some usages of `SortedContactList#getAllContacts` with `getClosestContactId`/`getContactIds` (we didn't need the `Contact` objects in those cases, just `NodeID`s)

### DIFF
--- a/packages/dht/src/dht/store/StoreRpcRemote.ts
+++ b/packages/dht/src/dht/store/StoreRpcRemote.ts
@@ -20,10 +20,10 @@ export class StoreRpcRemote extends RpcRemote<IStoreRpcClient> {
         }
     }
 
-    async replicateData(request: ReplicateDataRequest, doNotConnect: boolean = false): Promise<void> {
+    async replicateData(request: ReplicateDataRequest): Promise<void> {
         const options = this.formDhtRpcOptions({
             timeout: EXISTING_CONNECTION_TIMEOUT,
-            doNotConnect
+            doNotConnect: false
         })
         return this.getClient().replicateData(request, options)
     }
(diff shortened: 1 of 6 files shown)


packages/dht/src/dht/store/StoreRpcRemote.ts
```diff
@@ -20,10 +20,10 @@ export class StoreRpcRemote extends RpcRemote<IStoreRpcClient> {
         }
     }
 
-    async replicateData(request: ReplicateDataRequest, doNotConnect: boolean = false): Promise<void> {
+    async replicateData(request: ReplicateDataRequest): Promise<void> {
         const options = this.formDhtRpcOptions({
             timeout: EXISTING_CONNECTION_TIMEOUT,
-            doNotConnect
+            doNotConnect: false
```
juslesan, line 26: `doNotConnect` should be set true atleast for when calling `this.replicateDataToClosestNodes()` from the destroy function. This way the node will not open any new soon to be disconnected connections during tear down.

Setting `doNotConnect: true` makes sense in general as all nodes in the kBucket structure should be connected to.

The only exception would be the replication that is done on new contacts, as the node might not be connected to yet.

juslesan (approved):

PR #2186, opened 2023-12-08
fix: Fix all hanging tests, reduce flakyness of other tests

ptesavol:
Fix all hanging tests. Includes a protobuf change.

* Bugfix: change gracefulDiconnect into notification (it was previously partly treated as a notification)
* Bugfix: do not open new connections when ConnectionManager is in Stopping state 

* Make timeouts unique by single milliseconds to make it easier to recognize in debug tools which timeout is which
* Force SimulatorConnection to send 'disconnect' event in close() the same way as websocket connections
* Add constructor option 'stopGivenTransport' DhtNode to indicate that the given transport should not be stopped when the DhtNode is stopped.
* Use the 'stopGivenTransport' in trackerless tests (the transports were not previously stopped at all)
* Call stop() on simulator in tests if it was started

* remove --force-exit from test scripts
* remove memory leaks from client by adding handling of destroy signal to many places and by removing closures
* fix mqtt bridge test to properly wait for arriving messages
* fix websocket connection establishment in broker websocket plugin test (previously 'error' waitForEvent would time out and cause unhandled promise exception on slower machines)
* await for sub1.unsubscribe() in memory leak test to get the subscription to non-random state before testing for memory leaks


packages/trackerless-network/test/integration/Inspect.test.ts
```diff
@@ -68,6 +69,10 @@ describe('inspect', () => {
         Simulator.useFakeTimers(false)
     })
 
+    afterAll(() => {
```
juslesan, line 70: move to afterEach

packages/dht/src/connection/simulator/SimulatorConnector.ts
```diff
@@ -32,6 +32,9 @@ export class SimulatorConnector {
     }
 
     public connect(targetPeerDescriptor: PeerDescriptor): ManagedConnection {
+        if (this.stopped) {
+            console.error('connect() called on stopped SimulatorConnector')
```
juslesan, line 35: remove console.error

juslesan (changes requested):

packages/trackerless-network/test/integration/Inspect.test.ts
```diff
@@ -30,7 +30,8 @@ describe('inspect', () => {
             layer0: {
                 entryPoints: [publisherDescriptor],
                 peerDescriptor,
-                transport
+                transport,
+                stopGivenTransport: true
```
teogeb, line 34: Maybe we could use the same pattern as in `dht` test utils:
https://github.com/streamr-dev/network/blob/e88eacad4010d26b76583b0339d2a3dbb8f9e80e/packages/dht/test/utils/utils.ts#L75-L76

Or we could add e.g. `stop` event which the creator of the instance could listen.

ptesavol: How about passing a transport factory instead the transport itself?

.github/workflows/reusable/cached-build/action.yml
```diff
@@ -35,5 +35,7 @@ runs:
       shell: bash
     - name: build
       if: steps.cache-build.outputs.cache-hit != 'true'
-      run: npm run build
+      run: npm run build && npm ci --no-audit
```
juslesan, line 39: Add comment as to why this is necessary?

.github/workflows/validate.yml
```diff
@@ -115,7 +115,7 @@ jobs:
       package: broker
       docker-services: cassandra init-keyspace dev-chain-fast deploy-network-subgraphs-fastchain
       run-entry-point: true
-      command: npm run test-integration
+      command: npm run test-sequential && npm run test-integration
```
juslesan, line 125: run sequential after integration

packages/broker/test/integration/plugins/mqtt/MqttPlugin.test.ts
```diff
@@ -24,7 +24,7 @@ createMessagingPluginTest('mqtt',
             await client.publish(streamId, JSON.stringify(msg))
         },
         subscribe: async (messageQueue: Queue<Message>, streamId: string, client: AsyncMqttClient): Promise<void> => {
-            client.once('message', (topic: string, message: Buffer) => {
+            client.on('message', (topic: string, message: Buffer) => {
```
juslesan, line 27: Is this an unnecessary change?

packages/broker/test/integration/plugins/operator/OperatorPlugin.test.ts
```diff
@@ -56,7 +56,7 @@ describe('OperatorPlugin', () => {
         })
         // wait for a while so that MaintainTopologyService has time to handle addStakedStreams
         // events emitted during Broker start
-        await wait(500)
+        await wait(5000)
```
juslesan: Does this fix test flakyness? Would probably make sense to somehow wait for the broker to have subscribed to the stream

packages/broker/test/integration/plugins/storage/StorageConfig.test.ts
```diff
@@ -64,7 +65,7 @@ describe('StorageConfig', () => {
         await waitForCondition(async () => {
             const result = await cassandraClient.execute('SELECT COUNT(*) FROM stream_data WHERE stream_id = ? ALLOW FILTERING', [stream.id])
             return (result.first().count > 0)
-        })
+        }, 15000)
```
juslesan, line 68: Does this fix test flakyness?

packages/client/src/DestroySignal.ts
```diff
@@ -16,13 +16,17 @@ export class DestroySignal {
...
     constructor() {
-        const controller = new AbortController()
-        this.abortSignal = controller.signal
-        this.onDestroy.listen(() => {
-            controller.abort()
-        })
+        this.handleDestroy = this.handleDestroy.bind(this)
+        this.abortController = new AbortController()
+        this.abortSignal = this.abortController.signal
+        this.onDestroy.listen(this.handleDestroy)
```
juslesan, line 25: Should be possible to keep this as an arrow function?

packages/client/test/end-to-end/MemoryLeaks.test.ts
```diff
@@ -232,20 +232,21 @@ describe('MemoryLeaks', () => {
...
+                        received.push(metadata)
                     })
 
                     leaksDetector.add(instanceId(sub), sub)
 
                     await publishTestMessages(MAX_MESSAGES)
-                    await wait(1000)
+                    await waitForCondition(() => received.length >= MAX_MESSAGES)
+                    await sub.unsubscribe()
+                    await client.destroy()
```
juslesan, line 247: Is the client supposed to be destroyed in this test?

packages/client/test/end-to-end/MemoryLeaks.test.ts
```diff
@@ -294,8 +303,8 @@ describe('MemoryLeaks', () => {
                     expect(received1).toHaveLength(SOME_MESSAGES)
                     expect(received2).toHaveLength(MAX_MESSAGES)
                     await sub2.unsubscribe()
-                    await wait(1000)
-                }, TIMEOUT)
+                    await client.destroy()
```
juslesan, line 306: Is the client supposed to be destroyed here?

packages/client/test/end-to-end/publish-subscribe.test.ts
```diff
@@ -92,12 +90,26 @@ describe('publish-subscribe', () => {
...
+
+            beforeEach(async () => {
+                networkNode = await startNetworkNode(stream.id)
+            }, TIMEOUT)
+
+            afterEach(async () => {
+                await networkNode.stop()
+            }, TIMEOUT)
+
+            it('messages are published encrypted', (done) => {
```
juslesan, line 105: Could avoid using the done pattern here? Should be possible to implement with async/await

packages/client/test/test-utils/LeaksDetector.ts
```diff
@@ -28,9 +28,11 @@ export class LeaksDetector {
 
         this.resetGC()
```
juslesan, line 29: Revert changes in this file

teogeb:
There are many great improvements 👍 Would it make sense to divide this quite big PR so that some of the improvement are in separate PRs? (E.g. the bug fixes in one PR, test teardown fixes in one etc.)

That way it would be easy to review and manage the changes and we'd get these merged sooner.  (We can't maybe land a PR with 100 changed files e.g. just a day before the release). Also we'll have less git conflicts when we can land some of these changes sooner.

juslesan (changes requested):
Revert changes to the timeouts in the DHT and trackerless-network packages to their original values

.github/workflows/reusable/cached-build/action.yml
ptesavol: This is a known bug in npm, another npm ci after build is needed to install the binaries generated by submodules

teogeb: @ptesavol Which binaries weren't installed in practice? Could you provide a link to that bug?

ptesavol: streamr-keyserver was not installed. https://github.com/npm/cli/issues/4591

PR #2189, opened 2023-12-11
refactor(dht): Refactor `RecursiveOperationSession`

teogeb:
Refactored `RecursiveOperationSession` and added unit test for it.

## Changes

- `RecursiveOperationSession` creates its own id
- Added `RecursiveOperationSession#start()` (previously the sending of the request was done in  `RecursiveOperationManager`)
- Removed type check from `doRouteRequest` as the function implementation assumes non-null (by using `!` assertions for the `recursiveOperationRequest` variable)
- Moved `isPeerCloserToIdThanSelf` from `DhtNode` to `RecursiveOperationManager` as it doesn't any `DhtNode` fields
- Simplify `data` param handling in `RecursiveOperationManager#sendResponse`
  - e.g. removed unnecessary `DataEntry.create.bind(DataEntry)` transformation
- Renamed `doSendResponse` -> `onResponseReceived` as the method doesn't send a response, but is a callback of RPC local's method which receives the method (which is sent by the remote by calling the `sendResponse` RPC call)
- Remove the payload of `completed` event as it was no used by any component
- Convert `RecursiveOperationSessiongetResults` to normal method

## Test utils

- Added `FakeTransport` class. When we create multiple instances of these transports via `FakeEnvironment#createTransport` the instances can automatically communicate with each other.

## Other changes

- Use arrow function in `ListeningRpcCommunicator` constructor so that `transport.send` works without binding

## Future improvements

- The `RecursiveOperationSession` test could assert peer descriptors (instead of the length of the array), and we could modify the test setup so that the test completes by receiving all data it wants (instead of completing after 4s timeout)
- Would it make sense to give `excludedPeer` as one of the fields `RecursiveOperationSession`? 
  - If it  needs to exclude the peer only in `start`, but not for when it receives data, the current setup is maybe ok
- Could simplify the logic in `RecursiveOperationManager#sendResponse`, which doesn't always send but instead calls  `onResponseReceived`. Should we e.g. move the `isOwnNode` check to the caller, or rename the method?

### DIFF
--- a/packages/dht/test/unit/RecursiveOperationSession.test.ts
+++ b/packages/dht/test/unit/RecursiveOperationSession.test.ts
@@ -1,0 +1,63 @@
+import { toProtoRpcClient } from '@streamr/proto-rpc'
+import { waitForCondition } from '@streamr/utils'
+import { range } from 'lodash'
+import { RecursiveOperationSession } from '../../src/dht/recursive-operation/RecursiveOperationSession'
+import { RecursiveOperationSessionRpcRemote } from '../../src/dht/recursive-operation/RecursiveOperationSessionRpcRemote'
+import { ServiceID } from '../../src/exports'
+import { createRandomNodeId } from '../../src/helpers/nodeId'
+import { Message, PeerDescriptor, RecursiveOperation } from '../../src/proto/packages/dht/protos/DhtRpc'
+import { RecursiveOperationSessionRpcClient } from '../../src/proto/packages/dht/protos/DhtRpc.client'
+import { RoutingRpcCommunicator } from '../../src/transport/RoutingRpcCommunicator'
+import { FakeEnvironment } from '../utils/FakeTransport'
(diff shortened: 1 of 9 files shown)


packages/dht/test/unit/RecursiveOperationSession.test.ts
```diff
@@ -0,0 +1,68 @@
...
+        localPeerDescriptor = createMockPeerDescriptor()
+    })
+
+    it('happy path', async () => {
+        const doRouteRequest = jest.fn()
+        const session = new RecursiveOperationSession({
+            transport: environment.createTransport(),
+            targetId: createRandomNodeId(),
+            localPeerDescriptor,
+            waitedRoutingPathCompletions: 3,
```
juslesan, line 41: This is 2 by default since we use two routing paths

juslesan (approved):

PR #2193, opened 2023-12-11
test(dht): Remove `Simulator#useFakeTimers`

teogeb:
Virtual time is no longer used (https://github.com/streamr-dev/network/pull/2181). Removed the virtual time functionality from the simulator.

As `sinon` is no longer used as prod dependency after this change, this PR resolves the packaging issue mentioned in NET-1137:

> Older version of sinonjs/fake-timers library causes issues with typescript when es6 modules are expected. This issue has already been fixed for the sinon/js library since version https://github.com/sinonjs/fake-timers/commits/v11.2.0 , but jest libraries are still using an older version of sinon.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6372,29 +6372,6 @@
                 "@sinonjs/commons": "^3.0.0"
             }
         },
-        "node_modules/@sinonjs/samsam": {
-            "version": "8.0.0",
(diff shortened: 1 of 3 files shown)


juslesan (approved):
We might want to use simulated time in the future for benchmarking certain scalability elements. On such occasions the simulated time could be cherry-picked from here to a testing branch.

PR #2194, opened 2023-12-11
refactor(dht): Options for `ITransport#send`

teogeb:
Change `ITransport#send` to use options object instead of positional parameters.

## Open questions

- Would it make sense to rename the booleans to be non-negated:
  - `doNotMindStopped` -> `ignoreIfStopped`
  - `doNotConnect` -> `ignoreIfNotConnected` (or just `connect`, and the default would be `true`)


juslesan (approved):
> doNotMindStopped -> ignoreIfStopped
The names would not match here. If the `doNotMindStopped` is set `true` the message will be sent even if the CM is stopped. Could be `ignoreStopped`

> doNotConnect -> ignoreIfNotConnected (or just connect, and the default would be true)
connect could be a good name

PR #2199, opened 2023-12-12
NET-1133: Update config wizard to create v1.0 configs

mondoreale:
## Summary

New Streamr config wizard! 🧙🏻

## Changes

The way the wizard is built is different. I still use the `inquirer` library buy flavoured slightly different. It's more about individual prompt methods (`confirm`, `select`, etc.).

I also test it different. `@inquirer/testing` helped me make the tests more "integrational". I pretend to be the user and validate the output of each scenario.

## How to run locally for a quick test?

```
npm run clean
npm run bootstrap
cd packages/broker
npx esbuild
npx esbuild bin/config-wizard.ts --platform=node --target=node18 --bundle --packages=external > dist/wizard.js && node dist/wizard.js
```

## Example output
```
✓ Congratulations, you've set up your Streamr Network node!
  Your node address is 0x394c8d686B6B083b9bd91DB9f6238140D961e922
  Your node's generated name is Deer Good Foam

  You can start your Streamr node now with
  streamr-broker /Users/mariusz/.streamr/config/default.json

  For environment specific run instructions, see
  https://docs.streamr.network/guides/how-to-run-streamr-node
```

example

### DIFF
--- a/packages/broker/bin/config-wizard.ts
+++ b/packages/broker/bin/config-wizard.ts
@@ -8,7 +8,7 @@ program
     .name('broker-config-wizard')
     .description('Run the configuration wizard for the broker')
 
-void (async () => {
+;(async () => {
     try {
         await start()
     } catch (e) {
--- a/packages/broker/bin/config-wizard.ts
+++ b/packages/broker/bin/config-wizard.ts
@@ -8,4 +8,10 @@ program
     .name('broker-config-wizard')
     .description('Run the configuration wizard for the broker')
 
-start()
+void (async () => {
+    try {
+        await start()
+    } catch (e) {
+        console.error('Streamr Node Config Wizard encountered an error:\n', e)
+    }
+})()
--- a/packages/broker/src/config/ConfigWizard.ts
+++ b/packages/broker/src/config/ConfigWizard.ts
@@ -18,9 +18,9 @@ import * as WebsocketConfigSchema from '../plugins/websocket/config.schema.json'
 import * as BrokerConfigSchema from './config.schema.json'
 import { ConfigFile, getDefaultFile } from './config'
 
-const MinBalance = utils.parseEther('0.1')
+const MIN_BALANCE = utils.parseEther('0.1')
 
-export const start = async (): Promise<void> => {
+export async function start(): Promise<void> {
     log(`
         >
         > ***Welcome to the Streamr Network!***
@@ -56,9 +56,9 @@ export const start = async (): Promise<void> => {
 
         const storagePath = await getStoragePath()
 
-        const httpServer = http?.port ? { port: http.port } : void 0
+        const httpServer = http?.port ? { port: http.port } : undefined
 
-        const config: ConfigFile = {
+        let config: ConfigFile = {
             $schema: formSchemaUrl(CURRENT_CONFIGURATION_VERSION),
             client: {
                 auth: {
@@ -72,9 +72,13 @@ export const start = async (): Promise<void> => {
             httpServer,
         }
 
+        if (network === 'mumbai') {
+            config = getMumbaiConfig(config)
+        }
+
         persistConfig(
             storagePath,
-            network === 'polygon' ? config : getMumbaiConfig(config)
+            config,
         )
 
         log(`
@@ -98,7 +102,7 @@ export const start = async (): Promise<void> => {
 
                 resume()
 
-                if (balance.lt(MinBalance)) {
+                if (balance.lt(MIN_BALANCE)) {
                     log(`
                         > ! ${content}. You'll need to fund it with a small amount of MATIC tokens.
                     `)
@@ -174,7 +178,7 @@ export const start = async (): Promise<void> => {
 /**
  * Generates a mnemonic for a given private key.
  */
-export const getNodeMnemonic = (privateKey: string): string => {
+export function getNodeMnemonic(privateKey: string): string {
     return generateMnemonicFromAddress(
         toEthereumAddress(new Wallet(privateKey).address)
     )
@@ -183,7 +187,7 @@ export const getNodeMnemonic = (privateKey: string): string => {
 /**
  * Lets the user generate a private key or import an existing private key.
  */
-async function getPrivateKey() {
+async function getPrivateKey(): Promise<string> {
     const privateKeySource = await select<'Generate' | 'Import'>({
         message:
             'Do you want to generate a new Ethereum private key or import an existing one?',
@@ -225,8 +229,8 @@ async function getPrivateKey() {
 /**
  * Lets the user decide the desired network for their node.
  */
-async function getNetwork() {
-    return select<'polygon' | 'mumbai'>({
+async function getNetwork(): Promise<'polygon' | 'mumbai'> {
+    return select<Awaited<ReturnType<typeof getNetwork>>>({
         message:
             'Which network do you want to configure your node to connect to?',
         choices: [
@@ -242,8 +246,10 @@ async function getNetwork() {
 
 /**
  * Lets the user gather and configure desired operator plugins.
+ * @returns A valid Ethereum address if the user decide to participate
+ * in earning rewards, and `undefined` otherwise.
  */
-async function getOperatorAddress() {
+async function getOperatorAddress(): Promise<string | undefined> {
     const setupOperator = await confirm({
         message:
             'Do you wish to participate in earning rewards by staking on stream Sponsorships?',
@@ -270,7 +276,7 @@ interface PubsubPlugin {
 
 type PubsubPluginKey = 'websocket' | 'mqtt' | 'http'
 
-const DefaultPort: Record<PubsubPluginKey, number> = {
+const DEFAULT_PORTS: Record<PubsubPluginKey, number> = {
     websocket: WebsocketConfigSchema.properties.port.default,
     mqtt: MqttConfigSchema.properties.port.default,
     http: BrokerConfigSchema.properties.httpServer.properties.port.default,
@@ -279,7 +285,7 @@ const DefaultPort: Record<PubsubPluginKey, number> = {
 /**
  * Lets the user select and configure desired pub/sub plugins.
  */
-async function getPubsubPlugins() {
+async function getPubsubPlugins(): Promise<Partial<Record<PubsubPluginKey, PubsubPlugin>>> {
     const setupPubsub = await confirm({
         message:
             'Do you wish to use your node for data publishing/subscribing?',
@@ -299,10 +305,10 @@ async function getPubsubPlugins() {
         ],
     })
 
-    const pubsubPlugins: Partial<Record<PubsubPluginKey, PubsubPlugin>> = {}
+    const pubsubPlugins: Awaited<ReturnType<typeof getPubsubPlugins>> = {}
 
     for (const key of keys) {
-        const defaultPort = DefaultPort[key]
+        const defaultPort = DEFAULT_PORTS[key]
 
         const port = Number(
             await input({
@@ -324,7 +330,7 @@ async function getPubsubPlugins() {
                                         ([pluginKey, plugin]) =>
                                             value ===
                                             (plugin.port ||
-                                                DefaultPort[
+                                                DEFAULT_PORTS[
                                                     pluginKey as PubsubPluginKey
                                                 ])
                                     ) || []
@@ -355,7 +361,7 @@ async function getPubsubPlugins() {
 /**
  * Lets the user decide where to write the config file.
  */
-async function getStoragePath() {
+async function getStoragePath(): Promise<string> {
     // eslint-disable-next-line no-constant-condition
     while (true) {
         const path = await input({
@@ -377,9 +383,9 @@ async function getStoragePath() {
 }
 
 /**
- * Writes given config structure into a file.
+ * Writes the config into a file.
  */
-function persistConfig(storagePath: string, config: ConfigFile) {
+function persistConfig(storagePath: string, config: ConfigFile): void {
     const dirPath = path.dirname(storagePath)
 
     if (!existsSync(dirPath)) {
@@ -394,7 +400,7 @@ function persistConfig(storagePath: string, config: ConfigFile) {
 }
 
 /**
- * Adjusts the given config for the Mumbai test environment.
+ * Adjusts `config` for the Mumbai test environment.
  */
 function getMumbaiConfig(config: ConfigFile): ConfigFile {
     return produce(config, (draft) => {
@@ -437,6 +443,10 @@ function getMumbaiConfig(config: ConfigFile): ConfigFile {
     })
 }
 
+/**
+ * Gets a wallet balance of the network-native token for the given
+ * wallet address.
+ */
 async function getNativeBalance(
     network: 'polygon' | 'mumbai',
     address: string
@@ -450,6 +460,10 @@ async function getNativeBalance(
     return new providers.JsonRpcProvider(url).getBalance(address)
 }
 
+/**
+ * Gets an array of node addresses associated with the given operator
+ * contract address on the given network.
+ */
 async function getOperatorNodeAddresses(
     network: 'polygon' | 'mumbai',
     operatorAddress: string
@@ -477,12 +491,22 @@ async function getOperatorNodeAddresses(
     return data.operator?.nodes || null
 }
 
+/**
+ * Prints out an animated busyness indicator and does not move on
+ * to the next line until torn down.
+ * @returns A teardown callback that cleans up the line and brings
+ * the cursor to BOL.
+ */
 function progress(fn: (frame: string) => string): () => void {
     const frames = '◢◣◤◥'
 
     let frameNo = 0
 
     function tick() {
+        /**
+         * `isTTY` is false in CI which also means `clearLine` and
+         * `cursorTo` are not functions.
+         */
         if (process.stdout.isTTY) {
             process.stdout.clearLine(0)
 
@@ -501,6 +525,10 @@ function progress(fn: (frame: string) => string): () => void {
     return () => {
         clearInterval(intervalId)
 
+        /**
+         * `isTTY` is false in CI which also means `clearLine` and
+         * `cursorTo` are not functions.
+         */
         if (process.stdout.isTTY) {
             process.stdout.clearLine(0)
 
@@ -509,7 +537,18 @@ function progress(fn: (frame: string) => string): () => void {
     }
 }
 
-function style(message: string) {
+/**
+ * Formats the given message with colors and styles with the use
+ * of a markdown-ish syntax, like:
+ * - \*\*bold\*\*,
+ * - \_dim\_,
+ * - \*bright white\*,
+ * - \> indent normally,
+ * - \> ! indent as a warning,
+ * - \> x indent as an error, and
+ * - \> ~ indent as a confirmation or success.
+ */
+function style(message: string): string {
     let result = message
 
     const filters: [RegExp, chalk.Chalk][] = [
@@ -539,6 +578,6 @@ function style(message: string) {
         .trim()
 }
 
-function log(message = '') {
+function log(message = ''): void {
     console.info(style(message))
 }
--- a/packages/broker/src/config/ConfigWizard.ts
+++ b/packages/broker/src/config/ConfigWizard.ts
@@ -17,28 +17,15 @@ import { ConfigFile, getDefaultFile } from './config'
 const MinBalance = utils.parseEther('0.1')
 
 export const start = async (): Promise<void> => {
-    function notify(...args: unknown[]) {
-        console.info(chalk.bgGrey(' '), ...args)
-    }
-
-    console.info()
-
-    notify()
-
-    notify(' ', chalk.whiteBright.bold('Welcome to the Streamr Network!'))
-
-    notify(' ', 'This Config Wizard will help you setup your node.')
-
-    notify(' ', 'The steps are documented here:')
-
-    notify(
-        ' ',
-        'https://docs.streamr.network/guides/how-to-run-streamr-node#config-wizard'
-    )
-
-    notify()
-
-    console.info()
+    log(`
+        >
+        > ***Welcome to the Streamr Network!***
+        > This Config Wizard will help you setup your node.
+        >
+        > The steps are documented here:
+        > *https://docs.streamr.network/guides/how-to-run-streamr-node#config-wizard*
+        >
+    `)
 
     try {
         const privateKey = await getPrivateKey()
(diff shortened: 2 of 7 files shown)


fonty1:
Generated config file:
Screenshot 2023-12-12 at 12 52 47
It shouldn't have this line:
❌    "$schema": "https://schema.streamr.network/config-v2.schema.json",
I.e. it should match this template:
https://docs.streamr.network/guides/become-an-operator#testnet-node-config

mondoreale:
Looks like `npx streamr-broker-init` does it too (copied it over). Gone in [e82b4da](https://github.com/streamr-dev/network/pull/2199/commits/e82b4dab8a16e55a79764d463fbda919ae96761f).

fonty1 (approved):
LGTM. Probably best that @teogeb casts an eye before merging.

mondoreale:
Current state.

https://github.com/streamr-dev/network/assets/320066/0d4d4af8-fb65-42a2-a17b-46e33eb4a8f2

Open for suggestions.

mondoreale:
Pre-approved version (0):

example

teogeb:
> It shouldn't have this line: ❌ "$schema": "https://schema.streamr.network/config-v2.schema.json", I.e. it should match this template: https://docs.streamr.network/guides/become-an-operator#testnet-node-config

It is good that the schema definition is there as it tells the version number of the config file and therefore allows Broker's auto-migration to be used.

Brokers work ok without the definition. For testnets the version number is not relevant as we don't plan to implement auto-migration between testnets.

mondoreale:
@teogeb but doesn't it prevent the client from getting up? Having `$schema` in it constitutes an invalid structure from the schema validator's point of view, no?

mondoreale:
Also,

```
curl https://schema.streamr.network/config-v2.schema.json
```

gives

```xml
<Error>
<script/>
<Code>AccessDenied</Code>
<Message>Access Denied</Message>
<RequestId>A60029N99F3JGEY2</RequestId>
<HostId>BupW5G6spzJcilmeTJU1tZBzsP2cynaCIlOHMq8oWXkFFLSFINn8F7zUrQp5iR7wxqAyYQ6DExY=</HostId>
</Error>
```

packages/broker/src/config/migration.ts
```diff
@@ -144,7 +145,7 @@ export const readConfigAndMigrateIfNeeded = (fileName: string | undefined): Conf
     let explicitTargetFile = undefined
     if (fileName === undefined) {
         const defaultTargetFile = getDefaultFile()
-        const legacyTargetFile = getLegacyDefaultFile()
+        const legacyTargetFile = path.join(os.homedir(), '.streamr/broker-config.json')
```
teogeb: Why `getLegacyDefaultFile()` was inlined? Or did the functionality change?

packages/broker/src/config/config.ts
```diff
@@ -30,20 +30,11 @@ export interface ConfigFile extends Config {
     $schema?: string
 }
 
-export interface ConfigFile extends Config {
-    $schema?: string
```
teogeb, line 34: Revert this as `$schema` should be used (so that auto-migration works).

mondoreale:
> @teogeb but doesn't it prevent the client from getting up? Having `$schema` in it constitutes an invalid structure from the schema validator's point of view, no?

Oh, it doesn't fail, eh? I thought @fonty1's note came up from a schema validation failure.

Sounds like it's a good idea to bring it back. @fonty1?

packages/broker/bin/config-wizard.ts
```diff
@@ -8,4 +8,10 @@ program
     .name('broker-config-wizard')
     .description('Run the configuration wizard for the broker')
 
-start()
+void (async () => {
```
teogeb:
```suggestion
;(async () => {
```

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -1,254 +1,539 @@
...
+    port?: number
 }
 
-const PLUGIN_NAMES: Record<string, string> = {
-    WS: 'websocket',
-    MQTT: 'mqtt',
-    HTTP: 'http'
+type PubsubPluginKey = 'websocket' | 'mqtt' | 'http'
+
+const DefaultPort: Record<PubsubPluginKey, number> = {
```
teogeb:
```suggestion
const DEFAULT_PORTS: Record<PubsubPluginKey, number> = {
```

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -1,254 +1,539 @@
...
+    mqtt: MqttConfigSchema.properties.port.default,
+    http: BrokerConfigSchema.properties.httpServer.properties.port.default,
 }
 
-const PRIVATE_KEY_SOURCE_GENERATE = 'Generate'
-const PRIVATE_KEY_SOURCE_IMPORT = 'Import'
+/**
+ * Lets the user select and configure desired pub/sub plugins.
+ */
+async function getPubsubPlugins() {
```
teogeb: Could add return type for all functions in this file

fonty1:
> > @teogeb but doesn't it prevent the client from getting up? Having `$schema` in it constitutes an invalid structure from the schema validator's point of view, no?
> 
> Oh, it doesn't fail, eh? I thought @fonty1's note came up from a schema validation failure.
> 
> Sounds like it's a good idea to bring it back. @fonty1?

I assumed it was dropped in 1.0 since that what we've been doing to date :)

> Brokers work ok without the definition. For testnets the version number is not relevant as we don't plan to implement auto-migration between testnets.

New information to me, yeah, follow @teogeb's lead and I'll make a note of this in our Docs as well. Thanks!

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -1,254 +1,539 @@
...
+
+        if (http) {
+            Object.assign(pubsubPlugins, {
+                http: {},
+            })
+        }
+
+        const storagePath = await getStoragePath()
+
+        const httpServer = http?.port ? { port: http.port } : void 0
```
teogeb:
```suggestion
        const httpServer = http?.port ? { port: http.port } : undefined
```

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -1,254 +1,539 @@
...
+            },
+        } : {}
+
+        const { http, ...pubsubPlugins } = await getPubsubPlugins()
+
+        if (http) {
+            Object.assign(pubsubPlugins, {
+                http: {},
+            })
+        }
```
teogeb: Why is this needed?

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -1,254 +1,539 @@
...
-                return answers.enabledApiPlugins.includes(name)
+    writeFileSync(storagePath, JSON.stringify(config, null, 4))
+
+    chmodSync(storagePath, '600')
+}
+
+/**
+ * Adjusts the given config for the Mumbai test environment.
+ */
+function getMumbaiConfig(config: ConfigFile): ConfigFile {
```
teogeb: We can use the new environment preset, i.e. just defined
```
    environment: 'mumbai'
```

https://github.com/streamr-dev/network/pull/2139/files

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -1,254 +1,539 @@
...
+            plugins: {
+                ...operatorPlugins,
+                ...pubsubPlugins,
+            },
+            httpServer,
+        }
+
+        persistConfig(
+            storagePath,
+            network === 'polygon' ? config : getMumbaiConfig(config)
```
teogeb: Could introduce a string literal enum for networks, and also convert this ternary something like `if (network === 'mumbai') { config = getMumbaiConfig(config) }` so that possible third network will be handled correctly (if we add that at some point)

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -1,254 +1,539 @@
...
+                    log(`> ${content}`)
+                }
+            } catch (e) {
+                resume()
+
+                log('> x Failed to fetch node\'s balance')
+            }
+        }
+
+        if (operator) {
```
teogeb, line 133: Coudl use explicit `undefined` checks here and in other if statements

mondoreale:
I bring `$schema` back in [19d1058](https://github.com/streamr-dev/network/pull/2199/commits/19d1058fddabee3076db89d321342d89be4eb1f1).

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -1,254 +1,544 @@
...
+            > Your node address is ${chalk.greenBright(nodeAddress)}
+            > Your node's generated name is ${chalk.greenBright(getNodeMnemonic(privateKey))}
+        `)
+
+        if (operator) {
+            const resume = progress((f) =>
+                style(`> Your node address has *${f} MATIC* _– checking balance…_`)
+            )
+
+            try {
```
teogeb, line 107: Why is this additional formating needed? Is it for removing `ETH` word?

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -1,254 +1,544 @@
...
+                log('> x Failed to fetch operator nodes')
+            }
+        }
+
+        log(`
+            >
+            > You can start your Streamr node now with
+            > *streamr-broker ${storagePath}*
+            >
+            > For environment specific run instructions, see
```
teogeb, line 182: Which error case this handles ? E.g. ctrl+c or something else?

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -1,254 +1,544 @@
...
+    const { data } = z
+        .object({
+            data: z.object({
+                operator: z.union([
+                    z.null(),
+                    z.object({ nodes: z.array(z.string()) }),
+                ]),
+            }),
+        })
+        .parse(await resp.json())
```
teogeb, line 470: Does this fix some issue with `inquirer/prompts` library?

teogeb (commented):
First round of comments

teogeb:
> @teogeb but doesn't it prevent the client from getting up? Having `$schema` in it constitutes an invalid structure from the schema validator's point of view, no?

This is a Broker config, not a client config. In Broker there is `$schema` property at top level of the config: https://github.com/streamr-dev/network/blob/c09c373a1585c7d3df01aecf7f566802fc8f1e00/packages/broker/src/config/config.schema.json#L8

teogeb:
> schema validation failure.

Yes https://linear.app/streamr/issue/NET-1029/v2-broker-config-schema-is-not-accessible-online

packages/broker/bin/config-wizard.ts
mondoreale: `void` is underrated. #nosemis

packages/broker/src/config/ConfigWizard.ts
mondoreale: I'mma port the comments from the OG implementation…

packages/broker/src/config/ConfigWizard.ts
mondoreale: I'll reformat, but a union takes us just as far, currently. I'll keep it simple.

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -1,254 +1,544 @@
...
+        default: true,
+    })
 
-export const CONFIG_TEMPLATE: any = {
-    $schema: formSchemaUrl(CURRENT_CONFIGURATION_VERSION),
-    client: {
-        auth: {}
-    },
-    plugins: {},
-    apiAuthentication: {
```
teogeb, line 69: Keep the `apiAuthentication` block. It should contain an auto-generated key if any API plugins (i.e. pubsub plugins) have been enabled.

teogeb (changes requested):

packages/broker/src/config/ConfigWizard.ts
mondoreale: `progress` prints out a busyness indicator while we're checking the wallet balance. This entire line gets erased and replaced with `Your node address has *12.34 MATIC* (…)` once we know the value.

Both texts are formatted the same way and emphasise the important part.

packages/broker/src/config/ConfigWizard.ts
mondoreale: `operator` is either `undefined` or a valid ethereum address.

packages/broker/src/config/ConfigWizard.ts
fonty1: IIRC and IIUC :P including a plugin with default settings is done by adding an empty object like `http: {}` to the plugins section. I can't recall the exact reasoning behind it all though, perhaps we can look at documenting it together when Eric gets back?

packages/broker/src/config/ConfigWizard.ts
mondoreale: Ctrl+c, yes.

packages/broker/src/config/ConfigWizard.ts
mondoreale: Very handy! Let's make proper adjustments once the env feature is live.

packages/broker/src/config/ConfigWizard.ts
mondoreale: What issues?

packages/broker/src/config/ConfigWizard.ts
mondoreale: Good to have you, Sir. I'll revert.

packages/broker/src/config/migration.ts
mondoreale: It used to be used once, here, and assigned to `legacyTargetFile`. Does not seem to _need_ a utility function per se, no?

packages/broker/src/config/config.ts
mondoreale: Now I see why am I getting über-confused by this, haha. `ConfigFile` got de-dupped here.

Screenshot 2023-12-13 at 12 44 06 PM

packages/broker/src/config/ConfigWizard.ts
mondoreale: This is the only comment that's out there. Doesn't really explain anything.
https://github.com/streamr-dev/network/blob/streamr-1.0/packages/broker/src/config/ConfigWizard.ts#L187

On Slack I described the outcome of this behaviour.

packages/broker/src/config/ConfigWizard.ts
teogeb: Yes, therefore `if (operator !== undefined)` would communicate that check clearly

packages/broker/src/config/ConfigWizard.ts
teogeb: It will be live in the same release as this config wizard

packages/broker/src/config/ConfigWizard.ts
teogeb: Refered to `.replace(/\.(\d+)/, (f) => f.substring(0, 3))` snippet (maybe that line was the line `92` at the time of comment, or something)

packages/broker/src/config/ConfigWizard.ts
teogeb: Ah time line numbers have changed. Maybe referred the `progress` function.

packages/broker/src/config/migration.ts
teogeb: Nothing _needs_ a utility function... But we put that into a utility function so that config-related facts are in `config.ts`. If a function which gives us the default file is there, then a function that does almost the same could be also in the same place.

Maybe the actual solution would be to drop the support for legacy file location. The legacy location was used in back in Brubeck, so shouldn't be relevant in Tatum era 👍

packages/broker/bin/config-wizard.ts
teogeb: What is the benefit of `void` here? Should we start to use this patten in `network` monorepo codebase?

The `void` is so underrated, that maybe most JS developers (me included) don't even know what this does without guessing / reading the documentation. That's one aspect why this adds unnecessary complexity.

The documentation says:
The `void` operator evaluates the given expression and then returns `undefined` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void)

Here the intention is not to evaluate any expression. Of course calling a function technically "evaluates" it as every function returns a value. But here the point is just to execute the function.

packages/broker/src/config/ConfigWizard.ts
teogeb: The comment is very bad, agree. The summary of Slack discussion:
- for each plugin there should be a block inside plugins block, i.e. if `httpPlugin` enabled, then there is a `http` block
- typically the non-default port is defined inside the plugin's block
  - if a default is used, no port information need to be written as we'll use defaults from the plugin schema, e.g. https://github.com/streamr-dev/network/blob/e00d704440bc68d4455c8a614442c310db9684b4/packages/broker/src/plugins/mqtt/config.schema.json#L11
- the only exception is `httpPlugin`: in that case the port is defined in the `httpServer` root level block
 - if we use default port, then no need to define the root level `httpServer` block (as empty is the default value for it in the Broker schema, and that way it uses the default port automatically) https://github.com/streamr-dev/network/blob/e00d704440bc68d4455c8a614442c310db9684b4/packages/broker/src/config/config.schema.json#L59-L60

packages/broker/src/config/ConfigWizard.ts
teogeb: Does the throw error /force closed/ come from the inquired library? Could add a comment there

(Referring to line `172` here, line numbers corrupted again, maybe because of the force push or something)

packages/broker/src/config/ConfigWizard.ts
teogeb: Maybe we wouldn’t need any exceptional cases at lines `50`-`56`, i.e. we’d have just
```ts
const pubsubPlugins = await getPubsubPlugins()
```
and then when we write pubsubPlugins in plugins, we’d omit the `httpPlugin`’s port:
```ts
plugins: {
    ...operatorPlugins.map((p) => p === ‘http’ ? omit(p, ’port) : p),
    ...pubsubPlugins,
},
```
That way it would a bit clearer that this whole thing is about the port, and if getPubsubPlugins start to give other options in the future, we don’t accidentally omit all httpPlugin options at line 54 (where we currently

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -1,254 +1,591 @@
...
+    const privateKey = await (async () => {
+        if (privateKeySource === 'Generate') {
+            return Wallet.createRandom().privateKey
         }
+
+        return password({
+            message: 'Please provide the private key to import',
+            validate(value) {
+                try {
+                    return !!new Wallet(value)
```
teogeb: Why `!!` is used?

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -1,254 +1,591 @@
...
-    plugins: {},
-    apiAuthentication: {
-        keys: [generateApiKey()]
+        const port = Number(
+            await input({
+                message: `Provide a port for the ${key} plugin`,
+                default: defaultPort.toString(),
+                validate(value) {
+                    try {
+                        return !!z.coerce
```
teogeb: Why `!!` is used?

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -1,254 +1,591 @@
...
+    return privateKey
 }
 
-const generateApiKey = (): string => {
-    const hex = uuid().split('-').join('')
-    return Buffer.from(hex).toString('base64').replace(/[^0-9a-z]/gi, '')
+/**
+ * Lets the user decide the desired network for their node.
+ */
+async function getNetwork(): Promise<'polygon' | 'mumbai'> {
```
teogeb: Could introduce an enum for the networks

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -1,254 +1,591 @@
...
+                    } catch (e: unknown) {
+                        return (e as z.ZodError).issues
+                            .map(({ message }) => message)
+                            .join(', ')
+                    }
+                },
+            })
+        )
+
+        pubsubPlugins[key] = port !== defaultPort ? { port } : {}
```
teogeb, line 380: We check if the port has already been taken at line 349. To support that, maybe it would be good to do the omitting of default ports after all the ports have been asked? 

(And it would maybe make sense to omit just the `port` from the object instead of assuming that the port is the only config option we ask. Not too relevant here, but if we add other config options, it is easier if the port omitting feature doesn't give us a surprise by omitting all options.)

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -1,254 +1,591 @@
...
+ * Gets a wallet balance of the network-native token for the given
+ * wallet address.
+ */
+async function getNativeBalance(
+    network: 'polygon' | 'mumbai',
+    address: string
+): Promise<BigNumber> {
+    const url = config[network].rpcEndpoints[0]?.url || ''
+
+    if (!/^https?:/i.test(url)) {
```
teogeb:
```suggestion
    const url = config[network].rpcEndpoints[0]?.url

    if ((url === undefined) || !/^https?:/i.test(url)) {
```

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -1,254 +1,591 @@
...
+        })
+        .parse(await resp.json())
 
-const selectStoragePath = async (): Promise<StorageAnswers> => {
-    let answers
-    do {
-        answers = await inquirer.prompt(storagePathPrompts)
-    } while (answers.overwrite === false)
-    return answers as any
+    return data.operator?.nodes || null
```
teogeb:
```suggestion
    return data.operator?.nodes ?? []
```

the return type of this function can be just `string[]` and the check at line `146` be` if (nodes.length > 0)`

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -1,254 +1,591 @@
...
+                    if (!nodes.includes(nodeAddress.toLowerCase())) {
+                        log(
+                            '> ! You will need to pair your node with your Operator:'
+                        )
+                    } else {
+                        log('> Your node has been paired with your Operator:')
+                    }
+                } else {
+                    log(`
+                        > x Your Operator could not be found on the **${network.replace(/\w/, (l) => l.toUpperCase())}** network, see
```
teogeb: What the `network.replace(/\w/, (l) => l.toUpperCase())` does? E.g. does it capitalize the string (if that is what it does, could use https://lodash.com/docs/4.17.15#capitalize)

packages/broker/test/unit/ConfigWizard.test.ts
```diff
@@ -1,257 +1,1689 @@
...
+    const { answers, logs }: Scenario = {
+        answers: [],
+        logs: [],
+    }
+
+    jest.spyOn(console, 'info').mockImplementation((...args: unknown[]) => {
+        const log = args
+            .join('')
+            // eslint-disable-next-line no-control-regex
+            .replace(/\x1B\[\d+m/g, '')
```
teogeb: What this does? (Removes colors, maybe?)

packages/broker/bin/config-wizard.ts
mondoreale: Now it's an awkward argument. Let's have it your way. Doesn't make any difference here anyhway.

packages/broker/src/config/ConfigWizard.ts
mondoreale: A comment here (there) is a decent idea.

packages/broker/src/config/ConfigWizard.ts
mondoreale: I'll give it a union type so I don't have to typeof it later. Enum brings no benefit.

packages/broker/src/config/ConfigWizard.ts
mondoreale: It already covers the case and is sufficiently obvious. I'll leave it.

packages/broker/src/config/ConfigWizard.ts
mondoreale: That `map` is ugly af, lol. Plus, `pubsubPlugins` is an object. I got your point about the other fields we potentially lose here though. I'll get that fixed.

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -1,254 +1,591 @@
...
+            const resume = progress((f) =>
+                style(`> Your node address has *${f} MATIC* _– checking balance…_`)
+            )
+
+            try {
+                const balance = await getNativeBalance(network, nodeAddress)
+
+                const content = `Your node address has *${utils
+                    .formatEther(balance)
+                    .replace(/\.(\d+)/, (f) => f.substring(0, 3))} MATIC*`
```
mondoreale: Oh, this? Say you have a balance of `19.7465234786528937`. This makes it `19.74`.

packages/broker/src/config/ConfigWizard.ts
mondoreale: Reformatted.

packages/broker/src/config/ConfigWizard.ts
mondoreale: Reformatted.

packages/broker/src/config/ConfigWizard.ts
mondoreale: What if the `operator` is null? (= meaning not found)

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -1,254 +1,607 @@
...
+            return Wallet.createRandom().privateKey
         }
+
+        return password({
+            message: 'Please provide the private key to import',
+            validate(value) {
+                try {
+                    new Wallet(value)
+
+                    return true
```
mondoreale, line 229: `!!` gone.

packages/broker/src/config/ConfigWizard.ts
```diff
@@ -1,254 +1,607 @@
...
-    plugins: {},
-    apiAuthentication: {
-        keys: [generateApiKey()]
+        const port = Number(
+            await input({
+                message: `Provide a port for the ${key} plugin`,
+                default: defaultPort.toString(),
+                validate(value) {
+                    try {
+                        z.coerce
```
mondoreale, line 342: And here, `!!` gone.

packages/broker/src/config/ConfigWizard.ts
mondoreale: > We check if the port has already been taken at line 349. To support that, maybe it would be good to do the omitting of default ports after all the ports have been asked?

Not sure if I understand you right, but I already prevent conflicts on both custom and default ports. It's also tested.

> (And it would maybe make sense to omit just the `port` from the object instead of assuming that the port is the only config option we ask. Not too relevant here, but if we add other config options, it is easier if the port omitting feature doesn't give us a surprise by omitting all options.)

Let's be today.

packages/broker/src/config/migration.ts
mondoreale: The util is back.

mondoreale:
@teogeb the environment field is in. The PR turned a bit chaotic with all the things flying in, and the comments spread out unevenly. I don't remember gh being so… inconsistent about rebasing and comments ¯\\_(ツ)_/¯.

mondoreale:
This is now better. I dropped the unexpected `@inquirer/testing` from the root `package.json` and kept the `@streamr/config@5.0.5` update from `streamr-1.0`.

teogeb (approved):

PR #2202, opened 2023-12-13
refactor(dht): Simplify `SendOptions`

teogeb:
Negate `doNoConnect` -> `connect`, rename `doNotMindStopped` -> `sendIfStopped`. The parameters are required. Also rename related parameter of `DhtCallContect` and `DhtRpcOptions`.

## Future improvements

- Maybe `sendOptions` could be a separate block inside `callContext`
- Maybe `sendIfStopped` is no longer needed?

### DIFF
--- a/packages/dht/src/transport/RoutingRpcCommunicator.ts
+++ b/packages/dht/src/transport/RoutingRpcCommunicator.ts
@@ -42,7 +42,7 @@ export class RoutingRpcCommunicator extends RpcCommunicator {
             // TODO maybe sendOptions could be a separate block inside callContext
             const sendOpts = (msg.header.response !== undefined)
                 ? {
-                    // typically we already have a connection, but if it has disconnected for some reasonm
+                    // typically we already have a connection, but if it has disconnected for some reason
                     // maybe the receiver has gone offline (or it is no longer a neighbord) and therefore there
                     // is no point for trying form a new connection
                     connect: false,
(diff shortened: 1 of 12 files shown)


packages/dht/src/transport/RoutingRpcCommunicator.ts
```diff
@@ -39,15 +39,20 @@ export class RoutingRpcCommunicator extends RpcCommunicator {
...
-            } else if (msg.header.response || callContext && callContext.doNotConnect) {
-                return this.sendFn(message, { doNotConnect: true })
-            } else {
-                return this.sendFn(message)
-            }
-
+            // TODO maybe sendOptions could be a separate block inside callContext
+            const sendOpts = (msg.header.response !== undefined)
+                ? {
+                    // typically we already have a connection, but if it has disconnected for some reasonm
```
juslesan:
```suggestion
                    // typically we already have a connection, but if it has disconnected for some reason
```

packages/dht/src/transport/RoutingRpcCommunicator.ts
```diff
@@ -39,15 +39,20 @@ export class RoutingRpcCommunicator extends RpcCommunicator {
...
+                    // maybe the receiver has gone offline (or it is no longer a neighbord) and therefore there
```
juslesan:
```suggestion
                    // the receiver could have gone offline (or it is no longer a neighbor) and therefore there
```

packages/dht/src/transport/RoutingRpcCommunicator.ts
```diff
@@ -39,15 +39,20 @@ export class RoutingRpcCommunicator extends RpcCommunicator {
...
+                    // is no point for trying form a new connection
```
juslesan:
```suggestion
                    // is no point in trying form a new connection
```

juslesan (approved):

PR #2204, opened 2023-12-14
Fix(DHT): Critical WebSocket connection request bug

juslesan:
## Summary

No longer attempt to open a connection to Websocket server when receiving ConnectionRequests if a connection already exists. Opening duplicate client connections caused significant problems as duplicate connection handling is not meant to handle duplicates originating from the same side.

In the worst case, there appeared to be an infinite loop of duplicate connection replacement.

Fixes many flakyness issues on the client and broker tests.

Seems to fix client `end-to-end/publish-subscribe.test.ts` in a full CI setup with an entry point and 3 brokers. However, the test can still fail when running an entry-point only. Investigating this further currently.

## Future improvements
- Why does the simultaneous connections test not catch this issue?
- Add testing to ensure this does not happen again


ptesavol (approved):
All looks good, please merge asap.

PR #2213, opened 2023-12-15
fix(DHT): Fix staleness corner case in small networks

juslesan:
## Summary

Fixed a corner case where staleness of data was set incorrectly in smaller networks. Stale is no longer set true if the number of neighbors a node has in total is less than `redundancyFactor`


ptesavol (approved):
LGTM

PR #2217, opened 2023-12-15
test: use `127.0.0.1` instead of `localhost`

teogeb:
Use `127.0.0.1` in tests because our WebSocket implementation doesn't support IPv6. By using `127.0.0.1` instead of `localhost` we force the incoming connection attempts to use IPv4, and force the WebSocket server to resolve addresses only to IPv4.

### DIFF
--- a/packages/dht/src/helpers/Connectivity.ts
+++ b/packages/dht/src/helpers/Connectivity.ts
@@ -3,7 +3,7 @@ import { ConnectivityMethod, NodeType, PeerDescriptor } from '../proto/packages/
 import { isPrivateIPv4 } from './AddressTools'
 
 export const canOpenConnectionFromBrowser = (websocketServer: ConnectivityMethod): boolean => {
-    const hasPrivateAddress = ((websocketServer.host === 'localhost') || isPrivateIPv4(websocketServer.host))
+    const hasPrivateAddress = ((websocketServer.host === '127.0.0.1') || isPrivateIPv4(websocketServer.host))
     return websocketServer.tls || hasPrivateAddress
 }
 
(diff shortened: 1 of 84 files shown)


packages/dht/src/helpers/Connectivity.ts
```diff
@@ -3,7 +3,7 @@ import { ConnectivityMethod, NodeType, PeerDescriptor } from '../proto/packages/
 import { isPrivateIPv4 } from './AddressTools'
 
 export const canOpenConnectionFromBrowser = (websocketServer: ConnectivityMethod): boolean => {
-    const hasPrivateAddress = ((websocketServer.host === 'localhost') || isPrivateIPv4(websocketServer.host))
+    const hasPrivateAddress = ((websocketServer.host === '127.0.0.1') || isPrivateIPv4(websocketServer.host))
```
juslesan: Keep this as localhost, it is there intentionally. isPrivateIPv4() will include 127.0.0.1

juslesan (changes requested):

packages/dht/src/helpers/Connectivity.ts
teogeb: Do you mean `expect(isPrivateIPv4('127.0.0.1')).toBe(true)` should pass? It doesn't.

(added this unit test case in commit 482e4bd90)

teogeb: Reverted this change. The `isPrivateIPv4` changed in a separate PR: https://github.com/streamr-dev/network/pull/2239

juslesan (approved):

PR #2219, opened 2023-12-16
feat(dht): Check version compatibility on handshakes

juslesan:
## Summary

Check version compatibility on handshakes.

This ensures that nodes running incompatible older versions of the software cannot join networks that assume the use of newer versions.

## Future Improvements

- The version field might no longer be needed in the handshake response in the future. Once all nodes add the version number to the handshake request we can safely assume that the requests version is enough to decide compatibility.
- The version numbering could be separated from the monorepo version to a protocol version. The Protocol version could be sent instead of the monorepo version. This would ensure that backwards breaking changes to the client or broker would not require network splitting on the lower layers.
- The fallback value `BEFORE_TESTNET_TWO_VERSION` will likely not be needed in the future and can be removed.


packages/dht/src/helpers/versionCompatibility.ts
```diff
@@ -0,0 +1,13 @@
...
+    const minorSourceVersion = excludePatchVersion(version1)
+    const minorOwnVersion = excludePatchVersion(version2)
+    return minorSourceVersion === minorOwnVersion
+}
+
+export const excludePatchVersion = (version: string): string => {
+    const versionParts = version.split('.')
+    versionParts.pop()
+    return versionParts.join('.')
+}
```
harbu, line 13: Maybe a unit test case for empty string ensuring it doesn't throw.

packages/dht/src/helpers/versionCompatibility.ts
```diff
@@ -0,0 +1,13 @@
+// Is able to compare versions such as 1.2.3 and 1.2.4
+// can also compare versions such as 100.0.0-pretestnet.0 and 100.0.0-pretestnet.40
+export const isCompatibleVersion = (version1: string, version2: string): boolean => {
+    const minorSourceVersion = excludePatchVersion(version1)
+    const minorOwnVersion = excludePatchVersion(version2)
+    return minorSourceVersion === minorOwnVersion
```
teogeb:
```suggestion
    const minorVersion1 = excludePatchVersion(version1)
    const minorVersion2 = excludePatchVersion(version2)
    return minorVersion1 === minorVersion2
```

packages/dht/src/connection/websocket/WebsocketConnector.ts
```diff
@@ -308,7 +315,9 @@ export class WebsocketConnector {
                 targetPeerDescriptor
             )
             managedConnection.setRemotePeerDescriptor(sourcePeerDescriptor)
-            if (targetPeerDescriptor && !areEqualPeerDescriptors(this.localPeerDescriptor!, targetPeerDescriptor)) {
+            if (!isCompatibleVersion(sourceVersion, version)) {
+                managedConnection.rejectHandshake(HandshakeError.UNSUPPORTED_VERSION)
```
harbu, line 319: Duplication of 
```ts
if (!isCompatibleVersion(sourceVersion, version)) {
    managedConnection.rejectHandshake(HandshakeError.UNSUPPORTED_VERSION)
}
```
Move before (outside) if-else block?

harbu (changes requested):

teogeb (commented):
We can remove some fallbacks in the future when all nodes send version info? Could add TODO comments (and link to Linear issue) in those places.

At that point, do we still need both request and response side handling?

teogeb (approved):

packages/dht/src/connection/websocket/WebsocketConnector.ts
juslesan: Yeah would be good to not have to do duplication here but it is necessary in this case

PR #2230, opened 2023-12-18
fix(DHT): More redundant requests per join

juslesan:
## Summary

Double the number of redundant `getClosestPeers` requests per DiscoverySession to ensure that smaller Layer0 networks do not split


juslesan:
Does not appear to be needed in the end.

PR #2235, opened 2023-12-19
feat(DHT): Router exclude parallel path root

juslesan:
## Summary

The root of the primary routing path is excluded when routing on the secondary path in recursive operations. This improves the creation of the secondary path and fixes an problematic corner case in DHTs of 3 nodes in total. 

The feature is only enabled for recursive operations as they are rarely targeted towards exact nodeIds. Even in cases where they are, the exact nodeId can be found without completing multiple routing paths. In the remaining cases (forwarding, regular routing), the sent message is targeted to a peer. In these cases it makes more sense to explore the second path completely.

This feature fixes some of the flakyness seen in client and broker end-to-end tests.

This is a breaking change headed towards testnet3


packages/dht/protos/DhtRpc.proto
```diff
@@ -164,6 +164,7 @@ message RouteMessageWrapper {
   Message message = 4;
   repeated PeerDescriptor reachableThrough = 5;
   repeated PeerDescriptor routingPath = 6;
+  repeated string parallelRoots = 7;
```
harbu: The name could more clearly convey that these are nodes, e.g. `parallelRootNodes` or `parallelRootNodeIds` ?

harbu (approved):

packages/dht/protos/DhtRpc.proto
```diff
@@ -164,6 +164,7 @@ message RouteMessageWrapper {
   Message message = 4;
   repeated PeerDescriptor reachableThrough = 5;
   repeated PeerDescriptor routingPath = 6;
+  repeated string parallelRootNodeIds = 7;
```
teogeb, line 167: nodeIds are sent as bytes in protocol level. Can this be e.g. array of bytes?

juslesan: Due to added complexity from binary comparison issues I'd rather just keep this as a list of strings

harbu (approved):

PR #2236, opened 2023-12-19
fix(dht): Recursive operation second path close condition

juslesan:
## Summary

In cases where the second routing path for recursive operations was no longer finding closer peers than the first path the RecursiveOperationSession could easily enter timeout state. this is because routing to the closest peer is not possible twice as it rejects the second message as a duplicate. To counter this if the second path finds the same closest node as the first path the operation will resolve.

This change also speeds up closing RecursiveFindSession with the FETCH_DATA mode when no data is behind the key


packages/dht/src/dht/recursive-operation/RecursiveOperationSession.ts
```diff
@@ -45,6 +46,7 @@ export class RecursiveOperationSession extends EventEmitter<RecursiveOperationSe
     private timeoutTask?: NodeJS.Timeout 
     private completionEventEmitted = false
     private noCloserNodesReceivedCounter = 0
+    private noCloserNodesReceivedFrom: Set<NodeID> = new Set()
```
harbu: This set is never read from (only written to)?

packages/dht/test/integration/StoreOnDhtWithThreeNodes.test.ts
```diff
@@ -0,0 +1,56 @@
...
+        await node1.stop()
+        await node2.stop()
+        simulator?.stop()
+    })
+
+    it('Node can store on three peer DHT', async () => {
+        const storedData1 = createMockDataEntry()
+        const storedData2 = createMockDataEntry()
+        await node1.getDataFromDht(storedData1.key)
+        await node2.getDataFromDht(storedData1.key)
```
harbu: Why do we `getDataFromDht` before `storeDataToDht`? And why is the same not done to `entryPoint`?

harbu (changes requested):

packages/dht/test/integration/StoreOnDhtWithThreeNodes.test.ts
juslesan: Before this PR fetching null data from the DHT would take a lot longer in smaller networks. Without the changes the test would timeout on these first calls

packages/dht/src/dht/recursive-operation/RecursiveOperationSession.ts
harbu: What I meant with this comment is that the set is only `add`:ed to but never read or checked anywhere in the code?

packages/dht/test/integration/StoreOnDhtWithThreeNodes.test.ts
```diff
@@ -0,0 +1,57 @@
...
+        await node1.stop()
+        await node2.stop()
+        simulator?.stop()
+    })
+
+    it('Node can store on three peer DHT', async () => {
+        const storedData1 = createMockDataEntry()
+        const storedData2 = createMockDataEntry()
+        await node1.getDataFromDht(getDhtAddressFromRaw(storedData1.key))
+        await node2.getDataFromDht(getDhtAddressFromRaw(storedData1.key))
```
harbu, line 45:
```suggestion
        // Here we effectively test that fetching "null" data doesn't take too long. A test timeout here indicates an issue. 
        await node1.getDataFromDht(getDhtAddressFromRaw(storedData1.key))
        await node2.getDataFromDht(getDhtAddressFromRaw(storedData1.key))
        
```

harbu (commented):
I would add a comment here for clarity.

harbu (approved):

PR #2237, opened 2023-12-19
refactor: `DhtAddress` type

teogeb:
New `DhtAddress` type which replaces the existing `NodeID` type. This type can be used also for data keys.

The type is defined in the `dht` package, and it is used in all packages: `dht`, `trackerless-network`, `client` and `broker`.

There is also another helper type `DhtAddressRaw`, which is an alias to `Uint8Array`.

This PR doesn't contain functionality changes. Also there are no protocol level changes, and therefore this is not a breaking change.

## Future improvements

- Improve naming of some fields/variables/etc., e.g.
  - `targetId` -> `targetAddress`
  - `referenceId` -> `referenceAddress`
  - `nodeIdOrDataKeyRaw` -> `addressRaw`
  - `ConnectionLockerHandler` and maybe others: `id` -> `nodeId`


ptesavol (approved):
please merge

juslesan (approved):

PR #2245, opened 2023-12-22
test(dht): Fix `Find` test assertions

teogeb:
The node ids weren't asserted (the expressions were just `expect(true)`). Also refactored the check to use `DhtAddress` instead of `PeerID`.


harbu:
FYI: changed base from `streamr-1.0` to `main`

juslesan (approved):

PR #2246, opened 2023-12-22
test(dht): Simplify `RouteMessage` test

teogeb:
Removed `receiveMatrix` which wasn't written by not read. 

Also removed `getPeerId(node)` checks used `console.error` instead of `jest` assertions. (At least the first part of the check tested the just node id generation which we do at line `45`).

## Open questions

- Should we keep `receiveMatrix` and assert something about it?
- Was the intention of `(parseInt(getPeerId(node).toString()) === 0)` check to assert that the message was not emitted by the entry point node? If that is the case, should we have a `jest` assertion which checks that fact?
  - the id of the entry point was `0` before PR #2187


harbu:
FYI: changed base from `streamr-1.0` to `main`

juslesan (approved):
The receive matrix was originally added for debugging purposes and could be used for stricter test asserions.

PR #1910, opened 2023-10-02
feat(protocol, client, network): NET-1009 Publish binary content

juslesan:
## Summary

Added support for publishing binary content in the StreamrClient's `publish()` method

## Changes

- Streamr protocol has a new ContentType: `BINARY`
- ContentType is now a required field when creating a new StreamMessage-object
- The client's MessageFactory#createMessage detects if a new message is of binary type.

## Limitations and future improvements

- Lazy loading of content in the protocol's `StreamMessage#getContent`
- rename `serializedContent` in StreamMessage to `content`
- Support binary payloads in broker plugins
- Support binary payloads in cli-tools

PR #2262, opened 2024-01-04
feat NET-1036[!]: binary storage and resend support

harbu:
## Summary

**breaking change**

- StreamMessages are stored in Cassandra in protobuf binary format instead of JSON array serialization format.
- Resend HTTP streaming is now done in length-prefixed binary frames instead of JSON array serialization format.

In the previously landed pull request #1910, binary payload support was added to `streamr-client` package. This PR follows up that work by optimizing away some uses of binary text encoding / decoding in the storage node and resends. Instead data is stored and transmitted in binary and represented with `Buffer` and/or `Uint8Array` when appropriate.

- We can now probably get rid of the JSON array serialization format for good?
- Rename `serializedContent` to `content` in StreamMessage.ts. Renamed `parsedContent` to `cachedSomethingSomething`.
- Maximum payload size in resends is what can be represented using 4 unsigned bytes. This should be more than reasonable given WebRTC limitations and how payloads are assembled presently.
- Messages will get stored in Cassandra using a specific version of the protobuf definition. Any conversions between versions would be the responsibility of the receiver (not the storage node). 


packages/utils/src/lengthPrefixedFrameUtils.ts
```diff
@@ -0,0 +1,44 @@
+import { Transform, TransformCallback } from 'stream'
+
+/**
+ * Utilities to deal with length-prefixed frames, i.e. | length [4 bytes] | payload [variable bytes] |
+ */
+
+export const toLengthPrefixedFrame = (payload: Uint8Array): Buffer => {
+    const length = Buffer.alloc(4)
```
teogeb: Introduce constant

packages/utils/test/lengthPrefixFrameUtils.test.ts
```diff
@@ -0,0 +1,129 @@
...
+            decoder = new LengthPrefixedFrameDecoder()
+        })
+
+        it('can decode a single frame', (done) => {
+            const payload = Buffer.from('Hello, world!', 'utf-8')
+            const frame = toLengthPrefixedFrame(payload)
+
+            decoder.write(frame)
+
+            decoder.on('data', (data: Buffer) => {
```
teogeb, line 33: Should we assign the listener before doing the action? (`decoder.on` before `decoder.write`)

(also in other test cases)

packages/client/test/end-to-end/resend.test.ts
```diff
@@ -14,10 +17,14 @@ const TIMEOUT = 30 * 1000
...
     let publisherClient: StreamrClient
     let resendClient: StreamrClient
+    let payloads: Array<Uint8Array | { idx: number }>
 
     beforeEach(async () => {
         publisherClient = createTestClient(await fetchPrivateKeyWithGas(), 43232)
         resendClient = createTestClient(fastPrivateKey(), 43233)
+        const binaryPayloads = range(NUM_OF_MESSAGES / 2).map(() => randomBytes(random(0, 256)))
+        const jsonPayloads = range(NUM_OF_MESSAGES / 2).map((idx) => ({ idx }))
+        payloads = shuffle([... binaryPayloads, ...jsonPayloads])
```
teogeb:
```suggestion
        payloads = shuffle([...binaryPayloads, ...jsonPayloads])
```

packages/broker/test/unit/plugins/storage/dataQueryEndpoint.test.ts
```diff
@@ -8,24 +8,28 @@ import {
...
+import { MetricsContext, toEthereumAddress, hexToBinary, utf8ToBinary, toLengthPrefixedFrame } from '@streamr/utils'
+import { convertStreamMessageToBytes } from '@streamr/trackerless-network'
 
 const createEmptyStream = () => {
     const stream = new PassThrough()
     stream.push(null)
     return stream
 }
 
+const intoBinaryStream = (msg: StreamMessage[]): Readable => {
```
teogeb:
```suggestion
const createOutputStream = (msg: StreamMessage[]): Readable => {
```

Also could remove `createEmptyStream` and use `createOutputStream([])` instead of it?

packages/broker/test/integration/plugins/storage/BatchManager.test.ts
```diff
@@ -13,28 +13,23 @@ const keyspace = 'streamr_dev_v2'
...
-        signature: hexToBinary('0x1234'),
-        contentType: ContentType.JSON
-    })
+): InsertRecord {
+    return {
+        streamId,
+        partition,
+        timestamp,
+        sequenceNo,
+        publisherId,
```
teogeb: The `publisherId` is always `publisherOne`. Could remove `publisherId` and `msgChainId` parameters from this function and remove `defaultPublisherId` variable.

packages/broker/src/plugins/storage/Storage.ts
```diff
@@ -229,9 +240,9 @@ export class Storage extends EventEmitter {
             writeBytesPerSecond: new RateMetric()
         }
         metricsContext.addMetrics('broker.plugin.storage', metrics)
-        this.on('read', (streamMessage: StreamMessage) => {
+        this.on('read', (bytes: Uint8Array) => {
```
teogeb:
```suggestion
        this.on('read', (streamMessage: Uint8Array) => {
```

packages/broker/src/plugins/storage/DataQueryFormat.ts
```diff
@@ -41,22 +40,14 @@ export const toObject = (msg: StreamMessage): any => {
         contentType: msg.contentType,
         encryptionType: msg.encryptionType,
         groupKeyId: msg.groupKeyId,
-        content: (msg.encryptionType === EncryptionType.NONE ? msg.getParsedContent() : msg.getSerializedContent()),
+        content: parsedContent instanceof Uint8Array ? binaryToHex(parsedContent) : parsedContent,
```
teogeb, line 43: Could we analyze `ContentType.BINARY` instead of using `instanceof`?

PR #2264, opened 2024-01-04
refactor(dht): Rename `DhtNodeOptions#peerId`

teogeb:
Renamed `DhtNodeOptions#peerId` to `nodeId` as that config option controls the id of the node (could be just `id` if there we had only few config options in `DhtNodeOptions`).

Also renamed `PeerDescritor#peerId` -> `nodeId` in `TestProtos` so that it is consistent with the `dht` package.

### DIFF
--- a/packages/dht/src/dht/DhtNode.ts
+++ b/packages/dht/src/dht/DhtNode.ts
@@ -75,7 +75,7 @@ export interface DhtNodeOptions {
     websocketHost?: string
     websocketPortRange?: PortRange
     websocketServerEnableTls?: boolean
-    id?: DhtAddress
+    nodeId?: DhtAddress
 
     rpcRequestTimeout?: number
     iceServers?: IceServer[]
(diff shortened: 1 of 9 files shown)


packages/dht/src/dht/DhtNode.ts
```diff
@@ -75,7 +75,7 @@ export interface DhtNodeOptions {
     websocketHost?: string
     websocketPortRange?: PortRange
     websocketServerEnableTls?: boolean
-    peerId?: DhtAddress
+    id?: DhtAddress
```
harbu: I would name this as `nodeId`. I think the `DhtNode` class is so large and has so many responsibilities that it would not be apparent that `id` would refer to the node's id.

teogeb: Changed

harbu (approved):

PR #2267, opened 2024-01-05
ci: run eslint & versions & check separately from build phase

harbu:
## Summary

Increase parallelization of CI validation run by spinning off linting (and `npm run versions`) into its own job and running `npm run check` on each individual package in each package's own test run.

Currently this saves around 30 seconds to 1 minute in total wait time. If we implement eslint rules that require type-information in the future (very likely) the eslint runs will become much slower warranting the need for this change even more. 

- `npm run check` gets run multiple time for packages that have multiple test jobs defined. This is a _minor_ issue imo since the command's run time is negligible and I feel like making an optional argument for excluding the run would introduce more complexity than is worth. 

### DIFF
--- a/.github/workflows/test-setup.yml
+++ b/.github/workflows/test-setup.yml
@@ -30,6 +30,7 @@ jobs:
       - name: build (cached)
         uses: ./.github/workflows/reusable/cached-build
       - run: npm run check
+        working-directory: 'packages/${{ inputs.package }}'
       - name: start docker services
         if: ${{ inputs.docker-services != '' }}
         uses: streamr-dev/streamr-docker-dev-action@v1.0.1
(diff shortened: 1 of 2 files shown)


.github/workflows/test-setup.yml
```diff
@@ -29,6 +29,8 @@ jobs:
       - uses: actions/checkout@v4
       - name: build (cached)
         uses: ./.github/workflows/reusable/cached-build
+      - run: npm run check
+        working-directory: 'packages/${{ inputs.package }}'
```
harbu, line 33: Notice that `npm run check` is placed deliberately before spinning up docker services. No point in starting up services if tests don't compile :shrug:

teogeb (approved):

PR #2272, opened 2024-01-09
fix(DHT): fix flaky `integration/StoreAndDelete.test.ts`

juslesan:
## Summary

Wait 500ms for delete operations to propagate in the test. Test was flaky in 1/20 runs due to the delete operation not propagating fully before trying to find the deleted data from the DHT.


harbu (approved):

packages/dht/test/integration/StoreAndDelete.test.ts
```diff
@@ -42,6 +43,8 @@ describe('Storing data in DHT', () => {
         const successfulStorers = await storingNode.storeDataToDht(getDhtAddressFromRaw(entry.key), entry.data!)
         expect(successfulStorers.length).toBeGreaterThan(4)
         await storingNode.deleteDataFromDht(getDhtAddressFromRaw(entry.key), true)
+        // Wait for the delete operation to propagate
+        await wait(500)
```
teogeb, line 47: Would it make sense to use `waitForCondition`?

juslesan: Could do that, but it is not really necessary to do polling. Just need to wait a while for the delete operations to propagate

PR #2281, opened 2024-01-10
refactor: Unify names of functions, variables representing counts

teogeb:
Use naming convention `getFooCount` for counter variables/fields/etc. Previously we have used names like `getNumberOfFoo` and `getNumOfFoo`.


packages/dht/src/dht/routing/DuplicateDetector.ts
```diff
@@ -2,18 +2,18 @@ export class DuplicateDetector {
 
     private values: Set<string> = new Set()
     private queue: Array<string> = []
-    private maxNumberOfValues: number
+    private maxValueCount: number
```
harbu:
```suggestion
    private maxItemCount: number
```

harbu (changes requested):

harbu (approved):

PR #2284, opened 2024-01-10
fix(DHT): use distant id instead of random id in additional join

juslesan:
## Summary

Instead of joining on a random id, flip the binary and join on the opposite side of the kademlia structure. This significantly improves the network structure as nodes discover more peers on average which makes routing paths more robust.


packages/dht/src/dht/discovery/PeerDiscovery.ts
```diff
@@ -36,14 +36,14 @@ export class PeerDiscovery {
 
     async joinDht(
         entryPoints: PeerDescriptor[],
-        doAdditionalRandomPeerDiscovery = true,
+        doAdditionalFlippedPeerDiscovery = true,
```
teogeb:
```suggestion
        doAdditionalDistantPeerDiscovery = true,
```

packages/dht/src/identifiers.ts
```diff
@@ -16,6 +16,12 @@ export const getRawFromDhtAddress = (address: DhtAddress): DhtAddressRaw => {
     return hexToBinary(address) as unknown as DhtAddressRaw
 }
 
+export const getFlippedDhtAddress = (address: DhtAddress): DhtAddress => {
```
teogeb:
```suggestion
export const createDistantDhtAddress = (address: DhtAddress): DhtAddress => {
```

and move to `PeerDiscovery.ts`

teogeb (approved):

PR #2291, opened 2024-01-12
refactor: "Immutable" `StreamMessage`

harbu:
## Summary

- Make all fields of `StreamMessage` readonly (hinting towards but not necessarily enforcing immutability)
- Remove `StreamMessage#clone` prefer copy constructor pattern

- There is a small risk of something breaking here but based on analysis of code this shouldn't occur. Basically after decryption we used to put the decrypted key in `newGroupKey` (against the type definition), now we fixed the code so that this doesn't happen which affects behavior but is more in line with the type system.

### DIFF
--- a/packages/client/src/encryption/EncryptionUtil.ts
+++ b/packages/client/src/encryption/EncryptionUtil.ts
@@ -1,5 +1,5 @@
 import crypto, { CipherKey } from 'crypto'
-import { EncryptionType, StreamMessage, StreamMessageError } from '@streamr/protocol'
+import { EncryptionType, StreamMessage, StreamMessageAESEncrypted, StreamMessageError } from '@streamr/protocol'
 import { GroupKey } from './GroupKey'
 
 export class DecryptError extends StreamMessageError {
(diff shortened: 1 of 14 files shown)


packages/client/src/encryption/EncryptionUtil.ts
```diff
@@ -48,32 +48,32 @@ export class EncryptionUtil {
...
 
+        let newGroupKey: GroupKey | undefined = undefined
         try {
-            const { newGroupKey } = streamMessage
-            if (newGroupKey) {
-                // newGroupKey should be EncryptedGroupKey | GroupKey, but GroupKey is not defined in protocol
-                // @ts-expect-error expecting EncryptedGroupKey
-                streamMessage.newGroupKey = groupKey.decryptNextGroupKey(newGroupKey)
+            if (streamMessage.newGroupKey) {
+                newGroupKey = groupKey.decryptNextGroupKey(streamMessage.newGroupKey)
```
harbu, line 62: Any idea why the wrong type was being used in the previous implementation?

teogeb (approved):

PR #2292, opened 2024-01-12
feat(DHT): use a separate contacted set for additional distant joins

juslesan:
## Summary

In many cases distant joins were not even properly started due to the contacted set being shared with the contacted sets used for regular joining. By ensuring that the sets are unique both operations are fully completed.

### DIFF
--- a/packages/dht/src/dht/discovery/PeerDiscovery.ts
+++ b/packages/dht/src/dht/discovery/PeerDiscovery.ts
@@ -47,7 +47,7 @@ export class PeerDiscovery {
     ): Promise<void> {
         const contactedPeers = new Set<DhtAddress>()
         const distantJoinConfig = doAdditionalDistantPeerDiscovery 
-            ? { enabled: true, contactedPeers: new Set<DhtAddress>() } : { enabled: false } as { enabled: false }
+            ? { enabled: true, contactedPeers: new Set<DhtAddress>() } : { enabled: false } as const
         await Promise.all(entryPoints.map((entryPoint) => this.joinThroughEntryPoint(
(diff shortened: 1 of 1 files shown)


packages/dht/src/dht/discovery/PeerDiscovery.ts
```diff
@@ -58,7 +59,7 @@ export class PeerDiscovery {
         entryPointDescriptor: PeerDescriptor,
         // Note that this set is mutated by DiscoverySession
         contactedPeers: Set<DhtAddress>,
-        doAdditionalDistantPeerDiscovery = true,
+        additionalDistantJoinConfig?: { contactedPeers: Set<DhtAddress> },
```
teogeb:
```suggestion
        additionalDistantJoin: { enabled: true, contactedPeers: Set<DhtAddress> } | { enabled: false } |,
```

packages/dht/src/dht/discovery/PeerDiscovery.ts
```diff
@@ -46,10 +46,12 @@ export class PeerDiscovery {
         retry = true
     ): Promise<void> {
         const contactedPeers = new Set<DhtAddress>()
+        const distantJoinConfig = doAdditionalDistantPeerDiscovery 
+            ? { enabled: true, contactedPeers: new Set<DhtAddress>() } : { enabled: false } as { enabled: false }
```
harbu: To narrow the type:
```suggestion
            ? { enabled: true, contactedPeers: new Set<DhtAddress>() } : { enabled: false } as const
```

harbu (approved):

PR #2293, opened 2024-01-12
fix(proto-rpc): toProtoRpcClient() in case of calls with responses with all-optional response parameters

ptesavol:
Fix toProtoRpcClient() not to turn calls with responses with all-optional parameters into notifications


juslesan (approved):
Might make sense to consider removing the ProtoRpcClient type in the future and default back to using the ts-protobuf default behaviour for simplicity

PR #2299, opened 2024-01-15
refactor(protocol): rm runtime type validation from classes

harbu:
## Summary

Now that we don't de-serialize directly into `StreamMessage` but go thru the protobuf equivalent first, there is (imo) no need to do runtime type validation in the protocol class anymore. Semantic validation is still reasonable to keep around (e.g. `msgId` is strictly ahead of `prevMsgRef`, when using `AES` check that `groupKeyId` is present).

## Changes

- Remove `EncryptionType.RSA` (was only used internally, protobuf equivalent has only NONE / AES)
- Remove Enum runtime validation in StreamMessage (ContentType, MessageType, EncryptionType, SignatureType)
- Remove some runtime validation from `StreamMessage`, `MessageID`, and `EncrypedGroupKey`.

- We could add validation that if `LEGACY_SECP256K1` is used then `BINARY` content type is disallowed.
- It is worth thinking thru where semantic validation should take place. It is kind of late in the process to validate it only here in `StreamMessage`, it could be argued that this should occur much earlier in the process of de-serializing messages.

### DIFF
--- a/packages/protocol/src/protocol/message_layer/StreamMessage.ts
+++ b/packages/protocol/src/protocol/message_layer/StreamMessage.ts
@@ -22,7 +22,6 @@ export enum ContentType {
 
 export enum EncryptionType {
     NONE = 0,
-    RSA = 1,
     AES = 2
 }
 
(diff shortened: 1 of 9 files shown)


packages/protocol/src/protocol/message_layer/StreamMessage.ts
```diff
@@ -81,29 +75,22 @@ export default class StreamMessage implements StreamMessageOptions {
         groupKeyId,
         newGroupKey,
     }: StreamMessageOptions) {
-        validateIsType('messageId', messageId, 'MessageID', MessageID)
+        validateSequence(messageId, prevMsgRef)
+        validateIsNotEmptyByteArray('content', content)
```
teogeb, line 103: We could allow empty array

PR #2302, opened 2024-01-15
Refactor peer descriptor v2

ptesavol:
* Add fields ipAddress, salt and signature to PeerDescriptor
* Add a add-hoc serialization function for serializing the PeerDescriptor fileds that are to be signed


packages/dht/src/identifiers.ts
```diff
@@ -28,3 +28,16 @@ export const createRandomDhtAddress = (): DhtAddress => {
     return getDhtAddressFromRaw(crypto.randomBytes(KADEMLIA_ID_LENGTH_IN_BYTES))
 }
 
+export const serializePeerDescriptorForSigning = (peerDescriptor: PeerDescriptor): Uint8Array => {
+    const undefinedBuffer = Buffer.from('undefined')
```
teogeb: If a field is `undefined`, we could use `new Uint8Array(0)`, similarly to the usage in the `protocol` package:
https://github.com/streamr-dev/network/blob/a7c2de649b918b33c8f755cb43131d779f4bd560/packages/protocol/src/protocol/message_layer/signature.ts#L28-L29

packages/dht/src/identifiers.ts
```diff
@@ -28,3 +28,16 @@ export const createRandomDhtAddress = (): DhtAddress => {
...
+    const buffers = [ 
+        peerDescriptor.type !== undefined ? Buffer.from([peerDescriptor.type]) : undefinedBuffer,
+        peerDescriptor.udp !== undefined ? ConnectivityMethod.toBinary(peerDescriptor.udp) : undefinedBuffer,
+        peerDescriptor.tcp !== undefined ? ConnectivityMethod.toBinary(peerDescriptor.tcp) : undefinedBuffer,
+        peerDescriptor.websocket !== undefined ? ConnectivityMethod.toBinary(peerDescriptor.websocket) : undefinedBuffer,
+        peerDescriptor.region !== undefined ? Buffer.from([peerDescriptor.region]) : undefinedBuffer,
```
teogeb: Can we be sure that region is always one byte value (i.e. less than `256`)

packages/dht/src/identifiers.ts
ptesavol: Then there is risk of confusion, the following will have the same serialization:
{
1: 121,
2: undefined
3: 345
}

{
1: 121,
2: 345
3: undefined
}

packages/dht/src/identifiers.ts
ptesavol: no

packages/dht/src/identifiers.ts
ptesavol: this is exactly the kind of source of error I wanted to avoid by placing the signed part in a separate field and using standard protobuf serialization

teogeb: One alternative would be to include separators.

packages/dht/src/identifiers.ts
teogeb: Then we should some other way to convert region to a `Buffer`. Currently it just gets the last byte: e.g.  `0x1E240: Buffer.from([123456]) -> 0x40`

Same applies also to `NodeType`, but maybe there we can be sure it is just one byte value?

packages/dht/src/identifiers.ts
harbu: Either approach is fine IMO

packages/dht/src/identifiers.ts
harbu: Not sure I understood this point? Let's discuss on call for example.

harbu (commented):
Discussed with Teo, something like the below would make sense, thoughts @ptesavol ?

```ts
function convertUnsignedIntegerToBuffer(number: number): Buffer {
    const buffer = Buffer.alloc(4)
    buffer.writeUInt32BE(number)
    return buffer
}

export const createPeerDescriptorSignaturePayload = (peerDescriptor: PeerDescriptor): Uint8Array => {
    const separator = Buffer.from(',')
    const buffers = [
        peerDescriptor.type !== undefined ? convertUnsignedIntegerToBuffer(peerDescriptor.type) : new Uint8Array(0),
        separator,
        peerDescriptor.udp !== undefined ? ConnectivityMethod.toBinary(peerDescriptor.udp) : new Uint8Array(0),
        separator,
        peerDescriptor.tcp !== undefined ? ConnectivityMethod.toBinary(peerDescriptor.tcp) : new Uint8Array(0),
        separator,
        peerDescriptor.websocket !== undefined ? ConnectivityMethod.toBinary(peerDescriptor.websocket) : new Uint8Array(0),
        separator,
        peerDescriptor.region !== undefined ? convertUnsignedIntegerToBuffer(peerDescriptor.region) : new Uint8Array(0),
        separator,
        peerDescriptor.ipAddress !== undefined ? convertUnsignedIntegerToBuffer(peerDescriptor.ipAddress) : new Uint8Array(0),
        separator,
        peerDescriptor.salt !== undefined ? Buffer.from(peerDescriptor.salt) : new Uint8Array(0)
    ]
    return Buffer.concat(buffers)
}
```

And move to its own file from identifiers.ts.

harbu (approved):

teogeb (approved):

PR #2304, opened 2024-01-15
refactor(client): Use `urls` storage node metadata field

teogeb:
Use `urls` field from storage node metadata to read the URL of the storage node. The field contains a `string` array. Previously the URL was stored in the `http` field and it contained only one URL instead of an array of URLs.

With this change we can configure different URLs for 1.0 clients and Brubeck era clients in the same contract.

From the url list we pick a random item. We don't currently support any retry logic within one storage node. (We already have a retry logic if there are multiple storage nodes.)

## Future improvements

- If there is no `urls` field in the metadata, the resend `Promise` rejects with an error `StreamrClientError`: Only absolute URLs are supported, code=`STORAGE_NODE_ERROR`. Maybe the error message could describe the root cause.
- We could maybe retry the resend by using another URL if a resend fails a randomly picked URL.

### DIFF
--- a/packages/broker/test/utils.ts
+++ b/packages/broker/test/utils.ts
@@ -138,7 +138,7 @@ export async function startStorageNode(
     })
     try {
         await client.setStorageNodeMetadata({
-            http: `http://127.0.0.1:${httpPort}`
+            urls: [`http://127.0.0.1:${httpPort}`]
         })
         await createAssignmentStream(client)
     } finally {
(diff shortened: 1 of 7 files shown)


harbu (approved):
> We don't validate the storage node metadata when we read the url: if there is no urls field in the metadata, the return promise may reject?

Not a big issue imo. If you do implement validation, I would recommend using `zod`.

PR #2308, opened 2024-01-16
feat(network): disconnect excess stream connections when detected

juslesan:
## Summary

If the neighbor update interval detects that the local and remote node boht have more then N number of connections they will disconnect the shared connection. This way the target number of connections is kept near the target number N

## Future improvements

- Rename `neighborCount` to eg `neighborTargetCount`?


packages/trackerless-network/src/logic/neighbor-discovery/NeighborUpdateManager.ts
```diff
@@ -16,6 +16,7 @@ interface NeighborUpdateManagerConfig {
     streamPartId: StreamPartID
     rpcCommunicator: ListeningRpcCommunicator
     neighborUpdateInterval: number
+    neighborCount: number
```
teogeb, line 19: Is this the preferred neighbor count? I.e. the number of neigbors we'd like to have, but not necessarily currently have?

If that's the case, could rename e.g. to `preferredNeighborCount`

packages/trackerless-network/src/logic/neighbor-discovery/NeighborUpdateRpcLocal.ts
```diff
@@ -25,11 +26,24 @@ export class NeighborUpdateRpcLocal implements INeighborUpdateRpc {
...
+    }
+
     // INeighborUpdateRpc server method
     async neighborUpdate(message: NeighborUpdate, context: ServerCallContext): Promise<NeighborUpdate> {
         const senderPeerDescriptor = (context as DhtCallContext).incomingSourceDescriptor!
         const senderId = getNodeIdFromPeerDescriptor(senderPeerDescriptor)
-        if (this.config.neighbors.has(senderId)) {
+        if (this.config.neighbors.has(senderId) 
+            && this.config.neighbors.size() > this.config.neighborCount
+            && message.neighborDescriptors.length > this.config.neighborCount
```
teogeb: Here we compare the how many neighbors the remote has, and how many we'd like to have? But could it be that the remote has been configured differently, and its "preferered neighbor count" is something else than what is configured here in the local node?

teogeb (commented):
First round of comments

packages/trackerless-network/src/logic/neighbor-discovery/NeighborUpdateManager.ts
harbu: Could be misleading as well, e.g. "what is a preferred neighbor and how is that different from normal neighbor?"

I would've suggested `neighborTargetCount`?

packages/trackerless-network/src/logic/neighbor-discovery/NeighborUpdateRpcLocal.ts
```diff
@@ -44,19 +58,9 @@ export class NeighborUpdateRpcLocal implements INeighborUpdateRpc {
...
-            return response
+            return this.createResponse(false)
         } else {
-            const response: NeighborUpdate = {
-                streamPartId: this.config.streamPartId,
-                neighborDescriptors: this.config.neighbors.getAll().map((neighbor) => neighbor.getPeerDescriptor()),
-                removeMe: true
-            }
-            return response
+            return this.createResponse(true)
```
teogeb: Could simplify the if so that there are only two `if` branches, one which returns `this.createResponse(true)` and another which returns `this.createResponse(false)`

packages/trackerless-network/test/unit/NeighborUpdateRpcLocal.test.ts
```diff
@@ -0,0 +1,116 @@
...
+            rpcCommunicator,
+            neighborCount
+        })
+    })
+
+    afterEach(() => {
+        rpcCommunicator.destroy()
+    })
+
+    it('Returns neighbors', async () => {
```
harbu:
```suggestion
    it('response contains neighbor list of expected size', async () => {
```

packages/trackerless-network/test/unit/NeighborUpdateRpcLocal.test.ts
```diff
@@ -0,0 +1,116 @@
...
+        }
+        const res = await rpcLocal.neighborUpdate({
+            streamPartId,
+            neighborDescriptors: [localPeerDescriptor],
+            removeMe: false
+        }, { incomingSourceDescriptor: createMockPeerDescriptor() } as any)
+        expect(res.neighborDescriptors.length).toEqual(neighborCount)
+    })
+
+    it('Caller is a neighbor', async () => {
```
harbu:
```suggestion
    it('does not ask to be removed if caller is a neighbor', async () => {
```

packages/trackerless-network/test/unit/NeighborUpdateRpcLocal.test.ts
```diff
@@ -0,0 +1,116 @@
...
+        neighbors.add(neighbor)
+        const res = await rpcLocal.neighborUpdate({
+            streamPartId,
+            neighborDescriptors: [localPeerDescriptor],
+            removeMe: false
+        }, { incomingSourceDescriptor: caller } as any)
+        expect(res.removeMe).toEqual(false)
+    })
+
+    it('Caller is not a neighbor', async () => {
```
harbu:
```suggestion
    it('asks to be removed if caller is not a neighbor', async () => {
```

packages/trackerless-network/test/unit/NeighborUpdateRpcLocal.test.ts
```diff
@@ -0,0 +1,116 @@
...
+        const caller = createMockPeerDescriptor()
+        const res = await rpcLocal.neighborUpdate({
+            streamPartId,
+            neighborDescriptors: [localPeerDescriptor],
+            removeMe: false
+        }, { incomingSourceDescriptor: caller } as any)
+        expect(res.removeMe).toEqual(true)
+    })
+
+    it('Caller is a neighbor and both have too many neighbors', async () => {
```
harbu:
```suggestion
    it('asks to be removed if caller is a neighbor and both have too many neighbors', async () => {
```

harbu (changes requested):

packages/trackerless-network/src/logic/neighbor-discovery/NeighborUpdateRpcLocal.ts
juslesan: Does not matter. Then the remote node has to find someone with a free neighbor slot

packages/trackerless-network/src/logic/neighbor-discovery/NeighborUpdateManager.ts
juslesan: Should probably rename the variable across all uses then

harbu: I see what you mean.

The `neighborCount` here is passed all the way down from the `StreamrNodeConfig#streamPartitionNeighborCount`. We would have to rename all the places along the call chain (as well as a few other uses) and also change the packages/client/src/config.schema.json file.

To keep the PR compact and to the point, I would propose we keep the name as-is and add to future improvements section the name change and land that as a separate refactor PR later. Thoughts?

packages/trackerless-network/src/logic/neighbor-discovery/NeighborUpdateRpcLocal.ts
harbu: Played around with copilot "simplify" feature and it suggested the following, looks kinda good? What did you have in mind?

```ts
if (!this.config.neighbors.has(senderId)) {
    return this.createResponse(true)
} else {
    const isOverNeighborCount = this.config.neighbors.size() > this.config.neighborCount
      && message.neighborDescriptors.length > this.config.neighborCount

    if (!isOverNeighborCount) {
     // do all the updating stuff here
        this.config.neighborFinder.start()
    }

    return this.createResponse(isOverNeighborCount)
}
```

packages/trackerless-network/src/logic/neighbor-discovery/NeighborUpdateRpcLocal.ts
harbu: My intuition:

The local node always has a mandate to cut connections as necessary to arrive at its `neighborCount`config setting. This condition is represented by `this.config.neighbors.size() > this.config.neighborCount`.

The second condition `message.neighborDescriptors.length > this.config.neighborCount` is there to do it "politely". We only ask the other side to cut the connection if they are above a reasonable neighbor count (e.g.  the setting we are using).

Think of it like this: if our neighbor _A_ has only 2 neighbors of its own while another neighbor of ours, _B_, has 8 neighbors of its own, and we have too many neighbors right now,  which neighbor would you ask to cut connection with us?

packages/trackerless-network/src/logic/neighbor-discovery/NeighborUpdateManager.ts
juslesan: agree

packages/trackerless-network/src/logic/neighbor-discovery/NeighborUpdateRpcLocal.ts
teogeb: Maybe we could add a comment for line `43`. Something like this: we use `this.config.neighborCount` as an approximation (as we don't know the config value of used by the remote), and we kind of assume that if the remote has more neighbors than that, it has "enough neighbors" and therefore we can remove the connection.

packages/trackerless-network/src/logic/neighbor-discovery/NeighborUpdateRpcLocal.ts
teogeb: Now we always call this.updateContacts. Is this intentional? Before commit 92d01df75df83aa4ead54a10c67fe974db3d3792 this logic was only in `else if (this.config.neighbors.has(senderId))` branch.

juslesan: Yes this is intentional. The contacts can be updated regardless of weather or not the neighbor is asked to disconnect

packages/trackerless-network/src/logic/neighbor-discovery/NeighborUpdateManager.ts
harbu: Created ticket NET-1257 on this.

packages/trackerless-network/src/logic/neighbor-discovery/NeighborUpdateRpcLocal.ts
harbu:
```
// We don't know the remote's neighborCount setting here. We only ask to cut connections if the remote has a "sufficient" number of neighbors, where "sufficient" means our neighborCount setting.
```

teogeb (approved):

harbu (approved):

PR #2310, opened 2024-01-16
chore: Enable `@typescript-eslint/no-unnecessary-type-assertion` eslint rule

teogeb:
Enable the rule and remove unnecessary type assertions.


harbu (approved):
👍

PR #2313, opened 2024-01-17
feat(dht): Generate nodeIds in new format when not explicitly set

ptesavol:
Generate `nodeIs` in new format (new signing keys are generated every time `DhtNode` starts)

## Future improvements
- Client's `convertPeerDescriptorToNetworkPeerDescriptor` doesn't include all `PeerDescriptor` fields, only the field of `NetworkPeerDescriptor`. Maybe should update `NetworkPeerDescriptor` to match `PeerDescriptor`? https://linear.app/streamr/issue/NET-1255/clients-networkpeerdescriptor-should-include-allmost-fields-of-dhts
- Support for passing optional explicit private key via config
- Disable explicit `nodeId`. Do not allow setting an explicit `nodeId` instead rely on hardcoded things for generating consistent id.


packages/dht/src/dht/DhtNode.ts
```diff
@@ -76,6 +79,8 @@ export interface DhtNodeOptions {
     websocketPortRange?: PortRange
     websocketServerEnableTls?: boolean
     nodeId?: DhtAddress
+    signingModule?: ISigningModule
+    nodeName?: string   // used as salt in nodeId generation if given, otherwise random salt is used
```
harbu:
```suggestion
    identityKey?: string   // used as salt in nodeId generation if given, otherwise random salt is used
```

packages/dht/protos/DhtRpc.proto
```diff
@@ -197,6 +197,7 @@ message ConnectivityResponse {
   string host = 1;
   string natType = 2;
   ConnectivityMethod websocket = 3;
+  string probedIp = 4;
```
harbu: Two questions:
- type `uint32` vs ` string`
- name `ipAddress` vs. `probedIpAddress`

packages/dht/src/connection/websocket/ServerWebsocket.ts
```diff
@@ -128,4 +128,13 @@ export class ServerWebsocket extends EventEmitter<ConnectionEvents> implements I
             return ''
         }
     }
+
+    public getRemoteIp(): string {
+        if (!this.stopped && this.socket) {
+            return this.socket.socket.remoteAddress!
+        } else {
+            logger.error('Tried to get the remoteAddress of a stopped connection')
+            return ''
```
harbu: Throw in this part (unexpected situation that should never occur)

packages/dht/src/connection/websocket/ServerWebsocket.ts
```diff
@@ -128,4 +128,13 @@ export class ServerWebsocket extends EventEmitter<ConnectionEvents> implements I
             return ''
         }
     }
+
+    public getRemoteIp(): string {
+        if (!this.stopped && this.socket) {
```
harbu, line 137: Where to do the closing?

packages/dht/src/dht/DhtNode.ts
```diff
@@ -112,24 +117,62 @@ const logger = new Logger(module)
...
+    const signature = signingModule.sign(idPartToSign)
+
+    const nodeIdRaw = Buffer.concat([ipHash.slice(ipHash.length - 13, ipHash.length),
+        signature.slice(signature.length - 7, signature.length)])
+
+    return nodeIdRaw
+}
+
+export const createPeerDescriptor = (signingModule: ISigningModule, 
+    msg?: ConnectivityResponse, nodeId?: DhtAddress, nodeName?: string): PeerDescriptor => {
```
harbu:
```suggestion
export const createPeerDescriptor = (signingModule: ISigningModule, msg?: ConnectivityResponse, nodeId?: DhtAddress, nodeName?: string): PeerDescriptor => {
```

teogeb: We won't support that users could give explicit nodeId, i.e this parameter and all usages to this can be removed.

packages/dht/src/dht/DhtNode.ts
```diff
@@ -112,24 +117,62 @@ const logger = new Logger(module)
 
 export type Events = TransportEvents & DhtNodeEvents
 
-export const createPeerDescriptor = (msg?: ConnectivityResponse, nodeId?: DhtAddress): PeerDescriptor => {
+const calculateNodeIdRaw = (signingModule: ISigningModule, ipAddress: number,
```
teogeb: We could instantiate new `EthereumSigningModule` here and not give this as a parameter. Maybe could even inline the methods `EthereumSigningModule`, and just call methods of `@streamr/utils`.

The `ISigningModule` is not needed as we support only signing type (when we'll add the support for multiple signing types, we'll most likely add a new signatureType to `PeerDescriptor`)

packages/dht/test/integration/ConnectionManager.test.ts
```diff
@@ -40,7 +40,8 @@ describe('ConnectionManager', () => {
     }
 
     beforeEach(() => {
-        createLocalPeerDescriptor = jest.fn().mockImplementation((response) => createPeerDescriptor(response))
+        const signingModule = new EthereumSigningModule()
+        createLocalPeerDescriptor = jest.fn().mockImplementation((response) => createPeerDescriptor(signingModule, response))
```
harbu: Could we use ` createMockPeerDescriptor`

packages/dht/test/utils/utils.ts
```diff
@@ -21,21 +22,26 @@ import {
...
+    const signingModule = new EthereumSigningModule()
+    const connectivityResponse: ConnectivityResponse = {
+        host: '127.0.0.1',
+        natType: NatType.OPEN_INTERNET,
+        probedIp: '127.0.0.1'
+    } 
+
+    const ret = createPeerDescriptor(signingModule, connectivityResponse)
+    Object.assign(ret, opts)
+    return ret
```
harbu: First create PeerDescriptor as a whole then sign?

Let's see the implementation how it will pan out.

packages/dht/src/dht/DhtNode.ts
```diff
@@ -112,24 +114,47 @@ const logger = new Logger(module)
...
+    const nodeIdRaw = Buffer.concat([ipHash.slice(ipHash.length - 13, ipHash.length),
+        signature.slice(signature.length - 7, signature.length)])
+
+    return nodeIdRaw
+}
+
+export const createPeerDescriptor = (msg: ConnectivityResponse, nodeId?: DhtAddress): PeerDescriptor => {
+
+    const privateKey = crypto.randomBytes(32)
+    const publicKey = crypto.randomBytes(20)  // TODO calculate publicKey from privateKey
```
harbu: Do we still need publicKey, see Petri's point. If we do, we need to see if we can recover it here without `ethers`.

harbu: Move function

packages/dht/src/dht/DhtNode.ts
```diff
@@ -112,24 +114,47 @@ const logger = new Logger(module)
 
 export type Events = TransportEvents & DhtNodeEvents
 
-export const createPeerDescriptor = (msg?: ConnectivityResponse, nodeId?: DhtAddress): PeerDescriptor => {
+const calculateNodeIdRaw = (ipAddress: number, privateKey: Uint8Array): DhtAddressRaw => {
```
harbu: Move function

packages/dht/src/dht/DhtNode.ts
```diff
@@ -112,24 +114,47 @@ const logger = new Logger(module)
...
-    if ((nodeId === undefined) && (msg !== undefined)) {
-        nodeIdRaw = new Uint8Array(20)
-        const ipNum = msg.host.split('.').map((octet, index, array) => {
-            return parseInt(octet) * Math.pow(256, (array.length - index - 1))
-        }).reduce((prev, curr) => prev + curr)
-        const view = new DataView(nodeIdRaw.buffer)
-        view.setInt32(0, ipNum)
-        nodeIdRaw.set(crypto.randomBytes(20 - 4), 4)
+    
+    // ToDo: add checking that the nodeId is valid
```
harbu: remove

packages/dht/src/dht/DhtNode.ts
```diff
@@ -112,24 +114,47 @@ const logger = new Logger(module)
...
     } else {
-        nodeIdRaw = getRawFromDhtAddress(nodeId!)
+        nodeIdRaw = calculateNodeIdRaw(msg.ipAddress, privateKey)
     }
+
     const nodeType = isBrowserEnvironment() ? NodeType.BROWSER : NodeType.NODEJS
-    const ret: PeerDescriptor = { nodeId: nodeIdRaw, type: nodeType }
-    if (msg && msg.websocket) {
+
+    const ret: PeerDescriptor = { nodeId: nodeIdRaw, type: nodeType, ipAddress: msg.ipAddress, publicKey }
```
harbu: nitipick: Make multline

packages/dht/src/dht/DhtNode.ts
harbu: Support for now.

packages/dht/src/dht/DhtNode.ts
harbu: Discussed. We do need the publicKey.

PR #2318, opened 2024-01-19
fix(network): propagation during inspection, passing inspection

juslesan:
## Summary

Fix propagation during inspection. If a requested temporary connection is open to a node it should always backpropagate.

No longer filter nodes from propagation if they are being currently inspected.

Pass inspection if data has only been received from the inspected node. (There is no "truth" to compare to)

## Future improvements / TODO

There is a remaining problematic case that occurs during churn, that should be worked out: 
A node receives some data from the remaining stream when inspection starts but never from the inspected node. After this, data is always received from the inspected node first.

- Could rework temporary connections to work so that they no longer use the sendData() rpc call. This way normal propagation and temporary connections would be completely separate. This makes inspection even more detectable than before


juslesan:
@harbu this is the propagation fix for inspection

packages/trackerless-network/src/logic/RandomGraphNode.ts
```diff
@@ -308,7 +308,8 @@ export class RandomGraphNode extends EventEmitter<Events> {
             markAndCheckDuplicate(this.duplicateDetectors, msg.messageId!, msg.previousMessageRef)
         }
         this.emit('message', msg)
-        this.config.propagation.feedUnseenMessage(msg, this.getPropagationTargets(msg), previousNode ?? null)
+        const shouldAvoidBackPropagation = previousNode !== undefined && !this.config.temporaryConnectionRpcLocal.hasNode(previousNode)
+        this.config.propagation.feedUnseenMessage(msg, this.getPropagationTargets(msg), shouldAvoidBackPropagation ? previousNode : null)
```
harbu:
```suggestion
        const skipBackPropagation = previousNode !== undefined && !this.config.temporaryConnectionRpcLocal.hasNode(previousNode)
        this.config.propagation.feedUnseenMessage(msg, this.getPropagationTargets(msg), skipBackPropagation ? previousNode : null)
```

harbu (approved):

teogeb (approved):

PR #2320, opened 2024-01-22
feat(network): Connectivity check versioning

juslesan:
## Summary

Add a version field to the ConnectivityResponse. The server adds its version to the response and the client will decide weather or not to discard the response based on the received value.

## Other changes

Pumped up the number of failed connectivity requests before the stopping connectivity checking from 5 to 10

- When running connectivity checks on entry points, the entry points that could not be reached should be tracked and excluded.
- Could also check for version mismatches on the server side. For now older clients making connectivity requests to newer servers should be compatible. 
- The version field could be added to the PeerDescriptor. This way incompatible connections can be avoided before ever opening a connection.


ptesavol (approved):
Looking good

packages/dht/src/connection/websocket/WebsocketConnector.ts
```diff
@@ -193,10 +195,11 @@ export class WebsocketConnector {
                             port: this.selectedPort!,
                             host: this.host,
                             tls: this.config.serverEnableTls,
-                            selfSigned
+                            selfSigned,
+                            version
                         }
                         if (!this.abortController.signal.aborted) {
-                            return await sendConnectivityRequest(connectivityRequest, entryPoint)
+                            return await sendConnectivityRequest(connectivityRequest, entryPoint, version)
```
teogeb: Could maybe rename `version` as `localVersion`  when importing so that it is clear when we use it?

packages/dht/src/connection/connectivityChecker.ts
```diff
@@ -78,9 +80,15 @@ export const sendConnectivityRequest = async (
                     if (message.body.oneofKind === 'connectivityResponse') {
                         logger.debug('ConnectivityResponse received: ' + JSON.stringify(Message.toJson(message)))
                         const connectivityResponseMessage = message.body.connectivityResponse
+                        const remoteVersion = connectivityResponseMessage.version
                         outgoingConnection!.off('data', listener)
                         clearTimeout(timeoutId)
-                        resolve(connectivityResponseMessage)
+                        // remoteVersion can be undefined if the remote peer is running an older version
+                        if (remoteVersion !== undefined && isCompatibleVersion(localVersion, remoteVersion)) {
```
harbu: Are you sure it is `undefined` and not empty string `''` when remote peer is running an older version? Asking because the `ConnectivityResponse#ipAddress` field apparently gets default value 0 when not present? So is there a same kind of "assign default value if not present" logic here w.r.t. protobuf behavior?

packages/dht/src/connection/websocket/WebsocketConnector.ts
```diff
@@ -193,10 +195,11 @@ export class WebsocketConnector {
                             port: this.selectedPort!,
                             host: this.host,
                             tls: this.config.serverEnableTls,
-                            selfSigned
+                            selfSigned,
+                            version
```
harbu: There is no `version` field in the protobuf declaration for `ConnectivityRequest`. So remove `version` field here or add it to the protobuf?

harbu (changes requested):

packages/dht/src/connection/connectivityChecker.ts
juslesan: Good point, here it is indeed an empty string instead of undefined as well. So the undefined check is likely completely unnecessary since the isCompatibleVersion correctly handles empty strings in this case

harbu: Hey I think if we set the protobuf type to `optional` we can get `undefined`?

harbu (approved):

PR #2325, opened 2024-01-23
quickfix: only accept regular incoming websocket connections if local…

ptesavol:
* No not accept incoming websocket connections if localPeerDescriptor is undefined, but close the connection instead

ToDo: find out whether closing the incoming connection is the right thing to do if localPeerDescriptor is undefined


juslesan:
Moved to #2328

PR #2327, opened 2024-01-24
fix(DHT): Only emit newContact if contact was added to SortedContactList

juslesan:
## Summary

the `newContact` event is now only emitted if the contact was actually added to the contact list. Previously if a peer was outside the max length of the structure `newContact` was emitted each time it was seen. This caused performance problems in some nodes.

## Future Improvements

- Rename event to contactAdded


packages/dht/src/dht/contact/SortedContactList.ts
```diff
@@ -70,15 +77,13 @@ export class SortedContactList<C extends { getNodeId: () => DhtAddress }> extend
                         removedContact,
                         this.getClosestContacts()
                     )
+                    this.emit(
+                        'newContact',
+                        contact,
+                        this.getClosestContacts()
```
teogeb: Could calculate `this.getClosestContacts()` only once?

teogeb (approved):

harbu (approved):
LGTM; Future improvement: add typing to events

PR #2328, opened 2024-01-24
fix(DHT): Reject WS server connections during start

juslesan:
## Summary

Immediately close close on WS server if local PeerDescriptor has not been set. This reduces the likelihood of hanging connections in the entry points.

### DIFF
--- a/packages/dht/src/connection/websocket/WebsocketConnector.ts
+++ b/packages/dht/src/connection/websocket/WebsocketConnector.ts
@@ -153,6 +153,9 @@ export class WebsocketConnector {
                 } else if (action === 'connectivityProbe') {
                     // no-op
                 } else {
+                    // The localPeerDescriptor can be undefined here as the WS server is used for connectivity checks
+                    // before the localPeerDescriptor is set during start.
+                    // Handshaked connections should be rejected before the localPeerDescriptor is set.
                     if (this.localPeerDescriptor !== undefined) {
                         this.attachHandshaker(connection)
                     } else {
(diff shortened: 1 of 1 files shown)


harbu (approved):

harbu:
It's not necessarily obvious why the websocket server is started before the peerDescriptor is set. I assume this is done so because in order to do the connectivity check, we need the websocket server on? And only after doing the connecitvity check can we generate the peerDescriptor?

juslesan:
> It's not necessarily obvious why the websocket server is started before the peerDescriptor is set. I assume this is done so because in order to do the connectivity check, we need the websocket server on? And only after doing the connecitvity check can we generate the peerDescriptor?

Yes the server is on for connectivity checking. The node has to accept connections to the WS server during start to verify that the starting node is reachable on the server.

packages/dht/src/connection/websocket/WebsocketConnector.ts
```diff
@@ -153,7 +153,12 @@ export class WebsocketConnector {
                 } else if (action === 'connectivityProbe') {
                     // no-op
                 } else {
-                    this.attachHandshaker(connection)
+                    if (this.localPeerDescriptor !== undefined) {
```
teogeb, line 159: Worth adding a comment why the `localPeerDescription` can be `undefined` (e.g. similar info which was written here: https://github.com/streamr-dev/network/pull/2328#issuecomment-1908062807)

teogeb (approved):

PR #2329, opened 2024-01-24
test(broker): convert test for `MaintainTopologyHelper` into unit tests

harbu:
## Summary
Replace the integration tests for `MaintainTopologyHelper` with unit tests. This shaves off ~40 sec of integration test run time and fixes some flakyness.

It should be noted that the class is effectively end-to-end tested via _test/integration/MaintainTopologyService.test.ts_. In that sense the code is still under end-to-end testing coverage.


packages/broker/src/plugins/operator/OperatorPlugin.ts
```diff
@@ -80,7 +80,7 @@ export class OperatorPlugin extends Plugin<OperatorPluginConfig> {
         logger.info('Fetched my redundancy factor', { redundancyFactor })
 
         const contractFacade = ContractFacade.createInstance(serviceConfig)
-        const maintainTopologyHelper = new MaintainTopologyHelper(serviceConfig)
+        const maintainTopologyHelper = new MaintainTopologyHelper(ContractFacade.createInstance(serviceConfig))
```
teogeb:
```suggestion
        const maintainTopologyHelper = new MaintainTopologyHelper(contractFacade)
```

packages/broker/test/unit/plugins/operator/MaintainTopologyHelper.test.ts
```diff
@@ -0,0 +1,202 @@
...
+const SPONSORSHIP_TWO = randomEthereumAddress()
+
+describe(MaintainTopologyHelper, () => {
+    let smartContractEventEmitter: EventEmitter3
+    let contractFacade: MockProxy<ContractFacade>
+    let helper: MaintainTopologyHelper
+
+    async function emitSmartContractEvent(event: 'Staked' | 'Unstaked', sponsorship: string): Promise<void> {
+        smartContractEventEmitter.emit(event, sponsorship)
+        await wait(0)
```
teogeb, line 28: Why `wait` is needed?

packages/broker/test/unit/plugins/operator/MaintainTopologyHelper.test.ts
```diff
@@ -0,0 +1,202 @@
...
+        })
+
+        it('event "addStakedStreams" is not emitted thereafter if staking to an already staked stream', async () => {
+            const events = eventsWithArgsToArray(helper as any, ['addStakedStreams'])
+            await helper.start()
+            events.length = 0
+
+            await emitSmartContractEvent('Staked', SPONSORSHIP_ONE)
+
+            expect(events).toEqual([])
```
teogeb:
```suggestion
            const eventCountBefore = events.length

            await emitSmartContractEvent('Staked', SPONSORSHIP_ONE)

            expect(events.length).toEqual(eventCountBefore)
```

packages/broker/test/unit/plugins/operator/MaintainTopologyHelper.test.ts
harbu: The event handler for event `Staked` is async in `MaintainTopologyHelper`. A pause occurs when the handler gets the stream id for the sponsorship from the contractFacade. Therefore we need to pause here to allow for the event handler to go thru.

We don't technically need `wait(0)` when the event is `Unstaked` but I don't think it is worth adding an if-clause for this.

packages/broker/test/unit/plugins/operator/MaintainTopologyHelper.test.ts
harbu: Fixed in another way, see 30075a3f5

packages/broker/test/unit/plugins/operator/MaintainTopologyHelper.test.ts
teogeb: Could add a comment about the wait (or rename the function e.g. to `triggerEventHandler`)

teogeb (approved):

PR #2330, opened 2024-01-25
feat(DHT): validate explicit peer descriptor nodeId length

juslesan:
## Summary

Validate length of the nodeId if an explicit PeerDescriptor is given to the DhtNode.

### DIFF
--- a/packages/dht/src/dht/DhtNode.ts
+++ b/packages/dht/src/dht/DhtNode.ts
@@ -157,6 +157,11 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
                 throw new Error(`Invalid nodeId, the length of the nodeId should be ${expectedNodeIdLength}`)
             }
         }
+        if (this.config.peerDescriptor !== undefined) {
+            if (this.config.peerDescriptor.nodeId.length !== KADEMLIA_ID_LENGTH_IN_BYTES) {
+                throw new Error(`Invalid peerDescriptor, the length of the nodeId should be ${KADEMLIA_ID_LENGTH_IN_BYTES} bytes`)
+            }
+        }
     }
 
     public async start(): Promise<void> {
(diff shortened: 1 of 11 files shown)


packages/trackerless-network/test/utils/utils.ts
```diff
@@ -92,7 +92,7 @@ export const createStreamMessage = (
 }
 
 export const createRandomNodeId = (): DhtAddress => {
-    return getDhtAddressFromRaw(randomBytes(10))
+    return getDhtAddressFromRaw(randomBytes(20))
```
harbu: Is this same as Kademlia ID, so should we do?
```suggestion
    return getDhtAddressFromRaw(randomBytes(KADEMLIA_ID_LENGTH_IN_BYTES))
```

packages/dht/src/dht/DhtNode.ts
```diff
@@ -157,6 +157,11 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
                 throw new Error(`Invalid nodeId, the length of the nodeId should be ${expectedNodeIdLength}`)
             }
         }
+        if (this.config.peerDescriptor !== undefined) {
+            if (this.config.peerDescriptor.nodeId.length !== KADEMLIA_ID_LENGTH_IN_BYTES) {
```
harbu, line 161: peerDescriptor.nodeId should be `KADEMLIA_ID_LENGTH_IN_BYTES` but above in the code this.config.nodeId should be `KADEMLIA_ID_LENGTH_IN_BYTES * 2`. Not sure I follow, should these be the same?

juslesan: The type is different. If a nodeId is given explicitly is given in string as hex. In the PeerDescriptor it is a Uint8Array.

harbu: Got ya

teogeb (approved):

harbu (approved):

PR #2332, opened 2024-01-25
fix(DHT): Performance: no longer add the source of a routed message as a contact

juslesan:
## Summary

No longer add the source of a routed message as a contact. In cases where the contact never fits the routing node's contact lists the CPU cost for continuous attempting can be significant


harbu (approved):

harbu:
Is there a downside to removing this? Seems like there has been some reason this has originally been thought of as a good idea?

juslesan:
The "downside" is that nodes will know less peers in the network. This is not necessarily a downside since the main scalability design is that nodes only know a small view of the network. The performance win is more than worth it especially in under high throughput routing in larger networks. The functionality under `src/dht/discovery` should take care of finding the appropriate peers in the DHT

PR #2333, opened 2024-01-25
fix(DHT): minor performance improvement by limiting number of contacts in StoreManager

juslesan:
## Summary

Limit the number of contacts used in `storeManager#replicateAndUpdateStaleState`


packages/dht/src/dht/store/StoreManager.ts
```diff
@@ -63,10 +63,10 @@ export class StoreManager {
         const newNodeId = getNodeIdFromPeerDescriptor(newNode)
         const key = getDhtAddressFromRaw(dataEntry.key)
         // TODO use config option or named constant?
-        const closestToData = this.config.getClosestNeighborsTo(key, 10)
+        const closestToData = this.config.getClosestNeighborsTo(key, this.config.redundancyFactor)
```
teogeb, line 63: Remove TODO comment

packages/dht/src/dht/store/StoreManager.ts
```diff
@@ -63,10 +63,10 @@ export class StoreManager {
...
         const sortedList = new SortedContactList<Contact>({
             referenceId: key, 
-            maxSize: 20,  // TODO use config option or named constant?
+            maxSize: this.config.redundancyFactor,  // TODO use config option or named constant?
```
teogeb: Remove TODO comment

packages/dht/src/dht/store/StoreManager.ts
```diff
@@ -80,11 +79,11 @@ export class StoreManager {
         if (selfIsPrimaryStorer) {
             sortedList.addContact(new Contact(newNode))
             const sorted = sortedList.getContactIds()
-            // findIndex should never return -1 here because we just added the new node to the list
+            // findIndex returns -1 if the node did not fit the redundancyFactor closest nodes
             const index = findIndex(sorted, (nodeId) => (nodeId === newNodeId))
             // if new node is within the storageRedundancyFactor closest nodes to the data
             // do replicate data to it
-            if (index < this.config.redundancyFactor) {
+            if (index !== -1) {
```
teogeb: This could be just `sortedList.has(newNodeId)`, i.e. we don't need to calculate `index`?  (and maybe rename `sortedList` to e.g. `redundancyFactorNodes`)

There is no `SortedContactList#has` yet, but it trivial to add the helper function:
```typescript
    has(id: DhtAddress): boolean {
        return this.contactsById.has(id)
    }
```

harbu (approved):

PR #2334, opened 2024-01-25
fix(DHT): Check if data should be replicated per key instead of per entry

juslesan:
## Summary

Run the necessary checks to know if data should be replicated to a new contact only once per key, instead of per data entry. 

Major CPU improvement for the operation.

## Future Improvements

- Transfer the replicated messages in a per key batch instead of sending each entry individually

### DIFF
--- a/packages/dht/src/dht/store/StoreManager.ts
+++ b/packages/dht/src/dht/store/StoreManager.ts
@@ -54,14 +54,13 @@ export class StoreManager {
     }
 
     onNewContact(peerDescriptor: PeerDescriptor): void {
-        for (const dataEntry of this.config.localDataStore.values()) {
-            this.replicateAndUpdateStaleState(dataEntry, peerDescriptor)
+        for (const key of this.config.localDataStore.keys()) {
+            this.replicateAndUpdateStaleState(key, peerDescriptor)
         }
     }
 
-    private replicateAndUpdateStaleState(dataEntry: DataEntry, newNode: PeerDescriptor): void {
+    private replicateAndUpdateStaleState(key: DhtAddress, newNode: PeerDescriptor): void {
         const newNodeId = getNodeIdFromPeerDescriptor(newNode)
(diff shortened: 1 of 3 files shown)


packages/dht/src/dht/store/StoreManager.ts
```diff
@@ -86,11 +85,12 @@ export class StoreManager {
...
             if (index < this.config.redundancyFactor) {
                 setImmediate(async () => {
-                    await this.replicateDataToContact(dataEntry, newNode)
+                    const dataEntries = Array.from(this.config.localDataStore.values(key))
+                    await Promise.all(dataEntries.map(async (dataEntry) => this.replicateDataToContact(dataEntry, newNode)))
                 })
             }
         } else if (!this.selfIsWithinRedundancyFactor(key)) {
-            this.config.localDataStore.setStale(key, getDhtAddressFromRaw(dataEntry.creator), true)
+            this.config.localDataStore.setAllEntriesAsStale(key)
```
teogeb, line 93: Remove `setStale` method from the `LocalDataStore` class (not used by any component)

packages/dht/test/unit/StoreManager.test.ts
```diff
@@ -39,7 +40,12 @@ describe('StoreManager', () => {
                 } as any,
                 recursiveOperationManager: undefined as any,
                 localPeerDescriptor: { nodeId: getRawFromDhtAddress(localNodeId), type: NodeType.NODEJS },
-                localDataStore: { values: () => [DATA_ENTRY], setStale } as any,
+                localDataStore: { 
+                    keys: () => [getDhtAddressFromRaw(DATA_ENTRY.key)],
+                    values: () => [DATA_ENTRY],
+                    setStale,
```
teogeb: Remove all occurrences of `setStale` and use `setAllEntriesAsStale` instead

harbu (approved):

teogeb (approved):

PR #2336, opened 2024-01-26
fix(DHT) limit the contactList size in RoutingSession

juslesan:
## Summary

The RoutingSession's contact list had a maxSize of 10000. As the message is only attempted to be routed to the closest peers to the targetId there is no need to know the order of this many connections. This will significantly reduce memory use, and narrowly improve CPU use.


packages/dht/src/dht/routing/RoutingSession.ts
```diff
@@ -90,7 +90,7 @@ export class RoutingSession extends EventEmitter<RoutingSessionEvents> {
         const previousId = previousPeer ? getNodeIdFromPeerDescriptor(previousPeer) : undefined
         this.contactList = new SortedContactList({
             referenceId: getDhtAddressFromRaw(config.routedMessage.target),
-            maxSize: 10000,  // TODO use config option or named constant?
+            maxSize: 10,  // TODO use config option or named constant?
```
teogeb: Do we have a some config option for `DhtNode` which correlates to this value? E.g. would it make sense if this be would always have slightly greater value than `numberOfNodesPerKBucket`, or something like that? (If that is the case, maybe we derive the value from the option it correlates with.)

juslesan: yeah, a name would be good. There's no need to have this be configurable. It's just more of a sanity limit for the structure. There's no need to keep a sorted list of 10000 connections when we only want to know the closest few. The limit could be even lower than 10 because of this...

perhaps something like `ROUTING_CANDIDATE_LIMIT`

harbu (approved):

teogeb (approved):

PR #2339, opened 2024-01-26
fix(network): unlock connections with removeMe in NeighborUpdates

juslesan:
## Summary

unlock connections after receiving removeMe or after sending one and removing the neighbor. Fixes an issue with inflating numbers of connection locks on streams

## Future improvements

Should have better mechanisms to enforce releasing locks.

Could start using weakLocks for stream connections. Stream connections should be bidirectional so they would never be disconnected unless there's a bug somewhere

Ensure that proxy clients unlock connections when closed

### DIFF
--- a/packages/trackerless-network/src/logic/neighbor-discovery/NeighborUpdateManager.ts
+++ b/packages/trackerless-network/src/logic/neighbor-discovery/NeighborUpdateManager.ts
@@ -1,5 +1,5 @@
 import { NeighborUpdate } from '../../proto/packages/trackerless-network/protos/NetworkRpc'
-import { ListeningRpcCommunicator, PeerDescriptor, getNodeIdFromPeerDescriptor } from '@streamr/dht'
+import { ConnectionLocker, ListeningRpcCommunicator, PeerDescriptor, getNodeIdFromPeerDescriptor } from '@streamr/dht'
 import { NeighborUpdateRpcClient } from '../../proto/packages/trackerless-network/protos/NetworkRpc.client'
 import { Logger, scheduleAtInterval } from '@streamr/utils'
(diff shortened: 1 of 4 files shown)


harbu (approved):

packages/trackerless-network/src/logic/neighbor-discovery/NeighborUpdateManager.ts
```diff
@@ -51,6 +52,7 @@ export class NeighborUpdateManager {
             if (res.removeMe) {
                 const nodeId = getNodeIdFromPeerDescriptor(neighbor.getPeerDescriptor())
                 this.config.neighbors.remove(nodeId)
+                this.config.connectionLocker.unlockConnection(neighbor.getPeerDescriptor(), this.config.streamPartId)
```
teogeb, line 55: Should we unlock connection in all cases where a node is removed the neighbor list? I.e. some of these:
- `NeighborUpdateManager:53`
- `NeighborUpdateRpcLocal:69`
- `ProxyClient:225`

teogeb (commented):
First round of comments

packages/trackerless-network/src/logic/neighbor-discovery/NeighborUpdateManager.ts
juslesan: Unlocking was added to the first two in the list but the proxy needs a bigger refactor I'd rather include in another PR

teogeb: Maybe worth adding that proxy refactoring to "Future improvements" and/or create a Linear ticket?

teogeb (approved):

PR #2347, opened 2024-01-31
fix(dht): Criticial forwarding fix to requestIds (duplicate detection)

juslesan:
## Summary

Once a forwarded message was routed to the actual destination the forwarded messages request id was used for the routed message. This caused the actual routed message to be processed as a duplicate back on the rout e to the destination


ptesavol (approved):
This is an important bux fix, merge ASAP

PR #2348, opened 2024-01-31
feat(DHT): connectivity check even if websocket server was not started

juslesan:
## Summary

Run connectivity check to discover own IP address even when WS server was not started. Fixes a clustering around NodeIds starting with 85 for all nodes without a WS server as they use ip `127.0.0.1`.

Configuring entry points is now necessary when running a DhtNode without a WS server, unless the node is configured with the proper values to generate a PeerDescriptor without a ConnectivityCheck.

(Is backwards compatible)

## Future improvements

- Could refactor ConnectivityRequestHandler further
- Change ConnectivityRequest port to an optional field
- Perhaps ConnectivityCheck could be renamed to reflect that any kind of checking is no longer done. The point is to find out the node's connectivity information
- Could rename the entry point config in the connection layer? Entry point still makes since since ConnectivityChecks are an operation that all entry points should support.

### DIFF
--- a/packages/dht/src/connection/connectivityRequestHandler.ts
+++ b/packages/dht/src/connection/connectivityRequestHandler.ts
@@ -36,7 +36,19 @@ export const attachConnectivityRequestHandler = (connectionToListenTo: ServerWeb
 const handleIncomingConnectivityRequest = async (connection: ServerWebsocket, connectivityRequest: ConnectivityRequest): Promise<void> => {
     const host = connectivityRequest.host ?? connection.getRemoteAddress()
     const ipAddress = connection.getRemoteIp()
-    const connectivityResponse = await connectivityProbe(connectivityRequest, ipAddress, host)
+    let connectivityResponse: ConnectivityResponse
+    if (connectivityRequest.port === 0) {
+        logger.trace('ConnectivityRequest port is 0, replying without connectivityProbe')
+        connectivityResponse = {
+            host,
+            natType: NatType.UNKNOWN,
+            ipAddress: ipv4ToNumber(ipAddress),
+            version: localVersion
+        }
+    } else {
+        connectivityResponse = await connectivityProbe(connectivityRequest, ipAddress, host)
+    }
+
     const msg: Message = {
         serviceId: CONNECTIVITY_CHECKER_SERVICE_ID,
         messageType: MessageType.CONNECTIVITY_RESPONSE,
--- a/packages/dht/src/connection/connectivityRequestHandler.ts
+++ b/packages/dht/src/connection/connectivityRequestHandler.ts
@@ -37,7 +37,9 @@ const handleIncomingConnectivityRequest = async (connection: ServerWebsocket, co
     const host = connectivityRequest.host ?? connection.getRemoteAddress()
     const ipAddress = connection.getRemoteIp()
     let connectivityResponse: ConnectivityResponse
-    if (connectivityRequest.port === 0) {
+    if (connectivityRequest.port !== 0) {
+        connectivityResponse = await connectivityProbe(connectivityRequest, ipAddress, host)
+    } else {
         logger.trace('ConnectivityRequest port is 0, replying without connectivityProbe')
         connectivityResponse = {
             host,
@@ -45,10 +47,7 @@ const handleIncomingConnectivityRequest = async (connection: ServerWebsocket, co
             ipAddress: ipv4ToNumber(ipAddress),
             version: localVersion
         }
-    } else {
-        connectivityResponse = await connectivityProbe(connectivityRequest, ipAddress, host)
     }
-
     const msg: Message = {
         serviceId: CONNECTIVITY_CHECKER_SERVICE_ID,
         messageType: MessageType.CONNECTIVITY_RESPONSE,
--- a/packages/dht/src/connection/connectivityRequestHandler.ts
+++ b/packages/dht/src/connection/connectivityRequestHandler.ts
@@ -11,6 +11,8 @@ import { ServerWebsocket } from './websocket/ServerWebsocket'
 import { connectivityMethodToWebsocketUrl } from './websocket/WebsocketConnector'
 import { version as localVersion } from '../../package.json'
 
+export const DISABLE_CONNECTIVITY_PROBE = 0
+
 const logger = new Logger(module)
 
 export const attachConnectivityRequestHandler = (connectionToListenTo: ServerWebsocket): void => {
@@ -37,7 +39,7 @@ const handleIncomingConnectivityRequest = async (connection: ServerWebsocket, co
     const host = connectivityRequest.host ?? connection.getRemoteAddress()
     const ipAddress = connection.getRemoteIp()
     let connectivityResponse: ConnectivityResponse
-    if (connectivityRequest.port !== 0) {
+    if (connectivityRequest.port !== DISABLE_CONNECTIVITY_PROBE) {
         connectivityResponse = await connectivityProbe(connectivityRequest, ipAddress, host)
     } else {
         logger.trace('ConnectivityRequest port is 0, replying without connectivityProbe')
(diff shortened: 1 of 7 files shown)


harbu (approved):

packages/dht/src/connection/websocket/WebsocketConnector.ts
```diff
@@ -183,33 +183,29 @@ export class WebsocketConnector {
...
-                            natType: NatType.OPEN_INTERNET,
-                            websocket: { host: this.host!, port: this.selectedPort!, tls: this.config.tlsCertificate !== undefined },
-                            // TODO: maybe do a DNS lookup here?
-                            ipAddress: ipv4ToNumber('127.0.0.1'),
-                            version: localVersion
-                        }
-                        return preconfiguredConnectivityResponse
+                    // Do real connectivity checking
+                    const connectivityRequest = {
+                        port: this.selectedPort ?? 0,
```
teogeb: Would it make sense to annotate `port` as optional field?

packages/dht/src/connection/connectivityRequestHandler.ts
```diff
@@ -36,7 +36,19 @@ export const attachConnectivityRequestHandler = (connectionToListenTo: ServerWeb
...
+        logger.trace('ConnectivityRequest port is 0, replying without connectivityProbe')
+        connectivityResponse = {
+            host,
+            natType: NatType.UNKNOWN,
+            ipAddress: ipv4ToNumber(ipAddress),
+            version: localVersion
+        }
+    } else {
+        connectivityResponse = await connectivityProbe(connectivityRequest, ipAddress, host)
+    }
```
teogeb: Could swap the order of `if` statement branches as the probing is significantly more typical use case:

```suggestion
    if (connectivityRequest.port !== 0) {
        connectivityResponse = await connectivityProbe(connectivityRequest, ipAddress, host)
    } else {
        logger.trace('ConnectivityRequest port is 0, replying without connectivityProbe')
        connectivityResponse = {
            host,
            natType: NatType.UNKNOWN,
            ipAddress: ipv4ToNumber(ipAddress),
            version: localVersion
        }
    }
```

packages/dht/test/end-to-end/Layer0Webrtc-Layer1.test.ts
```diff
@@ -28,22 +28,26 @@ describe('Layer 1 on Layer 0 with mocked connections', () => {
 
         const layer0Node1Id = createRandomDhtAddress()
         layer0Node1 = new DhtNode({
-            nodeId: layer0Node1Id
+            nodeId: layer0Node1Id,
+            entryPoints: [entrypointDescriptor]
```
teogeb, line 32: Are the explicit entry points needed because of the changes in this PR?

packages/dht/src/connection/websocket/WebsocketConnector.ts
juslesan: I would prefer to avoid using optional fields (perhaps completely remove them in the future) since they are not standard protobuf. We use a custom flag of the `ts-protobuf` library to use them.

packages/dht/test/end-to-end/Layer0Webrtc-Layer1.test.ts
juslesan: Yes, previously connectivity checking was disabled in cases where a websocket server was configured. so the entry point list was not necessary in such cases

packages/dht/src/connection/websocket/WebsocketConnector.ts
teogeb: The `optional` annotation is a good way to communicate whether we consider some field to be always included in the payload of valid messages or not. We can then build the entity validation based on that high level logic of field presence, and have correct TypeScript there too.

Do we currently have an accessor which we can use to check if an object contains a value for a field? That would very straightforward way to check if port was given or not.

Without optionality we need this kind of magic values. It is too easy to forget to handle the magic value case when we process the message.

packages/dht/test/end-to-end/Layer0Webrtc-Layer1.test.ts
teogeb: This is a new requirement? Worth documenting in the PR description which kind of configurations are no longer supported.

packages/dht/src/connection/websocket/WebsocketConnector.ts
juslesan: This should be left for validation not protobuf. There are 0 guarantees that the optional fields would ever work if someone wants to use our protobuf definitions to communicate with the network using another language.

juslesan: Seems that optional is supported since protobuf 3.15... Could change this to optional in the future.

packages/dht/src/connection/websocket/WebsocketConnector.ts
```diff
@@ -183,33 +183,29 @@ export class WebsocketConnector {
         for (const reattempt of range(ENTRY_POINT_CONNECTION_ATTEMPTS)) {
```
teogeb, line 196: Could rename the `noServerConnectivityResponse` variable at line `174` as it is not accurate anymore. Or maybe the use case `(this.abortController.signal.aborted)`, could just throw an Error as we are not supposed to run connectivity checks when the application has been stopped (we do that only while we start the application, i.e. before it is allowed to call `stop()`).

packages/dht/src/connection/websocket/WebsocketConnector.ts
teogeb: If we don't do that in this PR, could just introduce a constant for the 0 value. E.g. const `DISABLE_CONNECTIVITY_PROBE = 0`

packages/dht/src/connection/websocket/WebsocketConnector.ts
```diff
@@ -183,33 +183,29 @@ export class WebsocketConnector {
...
             const entryPoint = sample(this.config.entrypoints)!
             try {
-                if (!this.websocketServer) {
-                    return noServerConnectivityResponse
+                if (!this.config.entrypoints || this.config.entrypoints.length === 0) {
```
teogeb: As `config.entrypoints` doesn't change during the for loop iteration, this block could be moved out of the for loop? I.e. would be checked once before entering the for loop.

(Could also rename `config.entrypoints` e.g. to `config.connectivityRequestTargets` in a separate PR as discussed yesterday. Maybe in a separate PR.)

packages/dht/test/end-to-end/Layer0Webrtc-Layer1.test.ts
teogeb: Based on yesterday's discussion: these settings were changed to modify the test behavior so that nodeId is generated from the connectivity response. The `entryPoints` list is needed for making a connectivity a request as we use entrypoints as connectivity request targests. 

-> could remove redundant explicit nodeId as we get the id from connectivity response

teogeb:
> Configuring entry points is now necessary when running a DhtNode without a WS server, unless the node is configured with the proper values to generate a PeerDescriptor without a ConnectivityCheck.

Which are these proper values? Either `NodeId` and `websocketHost` or explicit `peerDescriptor`?

packages/dht/test/end-to-end/Layer0Webrtc-Layer1.test.ts
juslesan: No need to remove here since explicit nodeId is still supported

juslesan:
> > Configuring entry points is now necessary when running a DhtNode without a WS server, unless the node is configured with the proper values to generate a PeerDescriptor without a ConnectivityCheck.
> 
> Which are these proper values? Either `NodeId` and `websocketHost` or explicit `peerDescriptor`?

Each value that is required to create a valid PeerDescriptor. Alternatively pass an explicit peerDescriptor for now

PR #2351, opened 2024-02-01
docs(network): update network README

juslesan:
## Summary

Added a general guide on how to run a node and publish/subscribe.

### DIFF
--- a/packages/trackerless-network/README.md
+++ b/packages/trackerless-network/README.md
@@ -1,6 +1,63 @@
 # Trackerless Network
 
+A P2P topic publish-subscribe network.
+
 The Trackerless Network package is a reimplementation of the old (Corea-Brubeck) network package. 
 The package is reimplemented to use the network and transport stacks of the proto-rpc and DHT packages.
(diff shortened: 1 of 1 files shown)


packages/trackerless-network/README.md
```diff
@@ -1,6 +1,63 @@
...
+
+Running a network node requires a DhtNode from the `@streamr/dht` to use as control layer. For more details on how to configure the DHT node go [here](packages/dht/README.md) The control layer node is configured as follows:
+
+```js
+const networkNode = new NetworkNode({
+    // Give all control layer DHT configs here.
+    layer0: {
+
+    },
+    // Content layer specific configurations:
```
teogeb, line 20: Content layer is a good name 👍 Does it refer to layer1 + layer2, i.e. the stream control layer + stream delivery layer?

teogeb (approved):

packages/trackerless-network/README.md
juslesan: My thought here is that it would refer to both.

PR #2358, opened 2024-02-06
feat(DHT): NET-1256 protocol versioning

juslesan:
## Summary

Start using protocol versioning instead of application versioning when handshaking connections. 

Current idea:

All peers include their own version and the list of versions they support in the handshake request. The responding peer will see if the version is compatible in either direction. If it is then the handshake will be accepted, else it will be rejected.

This is a draft PR for now to gather ideas on how to further implement the feature from here.

## TODO

- ConnectivityRequest versioning?
- Add protocol version to PeerDescriptor, could directly reject too old versions?


teogeb:
Implemented in https://github.com/streamr-dev/network/pull/2404

PR #2361, opened 2024-02-07
fix(utils): Add `secp256k1` dependency

teogeb:
Move `secp256k1` dependency from `client` package to `utils` package. The functionality was already moved in this PR: https://github.com/streamr-dev/network/pull/2300


harbu (approved):
good catch

PR #2365, opened 2024-02-12
fix(network): fix infinite loop corner case in parallel handshakes

juslesan:
## Summary

There was an unfortunate corner case where in some cases when trying to find random peers to connect to a node could enter an infinite loop. It is not clear if nodes ever entered it in production or testing but this change ensures that the case is no longer possible.

### DIFF
--- a/packages/trackerless-network/src/logic/neighbor-discovery/Handshaker.ts
+++ b/packages/trackerless-network/src/logic/neighbor-discovery/Handshaker.ts
@@ -32,10 +32,6 @@ const logger = new Logger(module)
 
 const PARALLEL_HANDSHAKE_COUNT = 2
 
-const getNodeIdsFromDeliveryRpcRemotes = (DeliveryRpcRemote: DeliveryRpcRemote[]): DhtAddress[] => {
-    return DeliveryRpcRemote.map((rpc) => getNodeIdFromPeerDescriptor(rpc.getPeerDescriptor()))
-}
-
 export class Handshaker {
 
     private readonly ongoingHandshakes: Set<DhtAddress> = new Set()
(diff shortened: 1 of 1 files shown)


packages/trackerless-network/src/logic/neighbor-discovery/Handshaker.ts
```diff
@@ -78,11 +78,11 @@ export class Handshaker {
 
     private selectParallelTargets(excludedIds: DhtAddress[]): HandshakeRpcRemote[] {
         const neighbors = this.config.nearbyNodeView.getFirstAndLast(excludedIds)
-        while (neighbors.length < PARALLEL_HANDSHAKE_COUNT && this.config.randomNodeView.size(excludedIds) > 0) {
-            const random = this.config.randomNodeView.getRandom(excludedIds)
-            if (random) {
-                neighbors.push(random)
-            }
+        const excludeFromRandomView = [...excludedIds, ...neighbors.map((neighbor) => getNodeIdFromPeerDescriptor(neighbor.getPeerDescriptor()))]
```
teogeb: Could simplify the state: this can always be calculated from `excludedIds` and `neighbors`.

harbu (approved):

packages/trackerless-network/src/logic/neighbor-discovery/Handshaker.ts
harbu: Not sure I follow @teogeb ?

teogeb: @juslesan The point is that we don't need the `excludeFromRandomView` variable as it can be calculated on-demand:

`const getExcludeFromRandomView = () => [...excludedIds, ...neighbors.map((neighbor) => getNodeIdFromPeerDescriptor(neighbor.getPeerDescriptor()))]`

It can be a local helper function inside the `selectParallelTargets` method.

teogeb (approved):

PR #2367, opened 2024-02-12
refactor(dht, trackerless-network): Remove unnecessary type assertions

teogeb:
Removed some unnecessary type assertions. These assertions would be `@typescript-eslint/no-unnecessary-type-assertion` errors  when we land  https://github.com/streamr-dev/network/pull/2364

## Open questions

Why e.g. this line in `LocalDataStore.test.ts` no longer needs type assertion? We used to have exclamation mark to assert that the `notDeletedData[0]` is not `undefined` if the array would not contain any items:
```ts
const notDeletedData = Array.from(localDataStore.values(getDhtAddressFromRaw(storedEntry.key)))
expect(notDeletedData[0].deleted).toBeFalse()
```

### DIFF
--- a/packages/dht/test/integration/RouteMessage.test.ts
+++ b/packages/dht/test/integration/RouteMessage.test.ts
@@ -176,8 +176,8 @@ describe('Route Message With Mock Connections', () => {
                 oneofKind: 'rpcMessage',
                 rpcMessage: closestPeersRequest
             },
-            sourceDescriptor: sourceNode.getLocalPeerDescriptor()!,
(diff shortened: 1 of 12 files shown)


harbu (approved):

harbu:
> Why e.g. this line in LocalDataStore.test.ts no longer needs type assertion? We used to have exclamation mark to assert that the notDeletedData[0] is not undefined if the array would not contain any items:

I'm guessing this is a choice of TypeScript to make the language more "ergonomic" and not have to use exclamation mark every time you access elements of on array.

PR #2369, opened 2024-02-13
fix(DHT): avoid unnecessary updateAndGetRoutablePeers calls

juslesan:
## Summary

Low hanging fruit CPU optimisation in RoutingSession: 
No longer run `updateAndGetRoutablePeers()` if the session is closing in `onRequestSucceeded()` or `onRequestFailed()`

### DIFF
--- a/packages/dht/src/dht/routing/RoutingSession.ts
+++ b/packages/dht/src/dht/routing/RoutingSession.ts
@@ -108,6 +108,12 @@ export class RoutingSession extends EventEmitter<RoutingSessionEvents> {
             this.ongoingRequests.delete(nodeId)
         }
         this.deleteParallelRootIfSource(nodeId)
+        this.failedHopCounter += 1
+        if (this.failedHopCounter >= MAX_FAILED_HOPS) {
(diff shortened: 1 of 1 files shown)


harbu (approved):

packages/dht/src/dht/routing/RoutingSession.ts
```diff
@@ -108,14 +108,19 @@ export class RoutingSession extends EventEmitter<RoutingSessionEvents> {
             this.ongoingRequests.delete(nodeId)
         }
         this.deleteParallelRootIfSource(nodeId)
+        this.failedHopCounter += 1
+        if (this.failedHopCounter >= MAX_FAILED_HOPS) {
+            logger.trace(`Stopping routing after ${MAX_FAILED_HOPS} failed attempts for sessionId: ${this.sessionId}`)
```
teogeb, line 113: If we are stopping the routing, should we set `this.stopped = true`?

juslesan: https://github.com/streamr-dev/network/pull/2370

PR #2371, opened 2024-02-14
fix(DHT) off handshaker onData after handshake is completed.

juslesan:
## Summary

The handshaker would remain listening to data from managed connections, parsing data, until the connection would be closed. Now the handshaker will stop listening to data once connection is handshaked.

### DIFF
--- a/packages/dht/src/connection/Handshaker.ts
+++ b/packages/dht/src/connection/Handshaker.ts
@@ -57,6 +57,7 @@ export class Handshaker extends EventEmitter<HandshakerEvents> {
                 } else {
                     this.emit('handshakeCompleted', handshake.sourcePeerDescriptor!)
                 }
+                this.connection.off('data', (data: Uint8Array) => this.onData(data))
             }
         } catch (err) {
             logger.error('error while parsing handshake message', err)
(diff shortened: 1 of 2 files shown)


packages/dht/src/connection/Handshaker.ts
```diff
@@ -101,4 +101,8 @@ export class Handshaker extends EventEmitter<HandshakerEvents> {
         this.connection.send(Message.toBinary(msg))
         logger.trace('handshake response sent')
     }
+
+    public stop(): void {
+        this.connection.off('data', (data: Uint8Array) => this.onData(data))
```
harbu: Shouldn't the 2nd argument here be a reference to the same function that was assigned to listen to the event originally. E.g.
  ```suggestion
        this.connection.off('data', this.onDataListener)
```

And in constructor:
```ts
constructor(...) {
    // ...
    this.onDataListener = (data: Uint8Array) => this.onData(data)
    this.connection.on('data', this.onDataListener)
    // ...
}
```

juslesan: right... somehow I keep forgeting

harbu (approved):

teogeb (approved):

PR #2372, opened 2024-02-14
feat(DHT): LRU routing table cache

juslesan:
## Summary

Cache routing tables based on targetId and previousId using `lru-cache`. This is a significant performance improvement.

Calculating the routing table for each individual routed message is expensive can be expensive in high throughput scenarios. By default, routing tables are cached for 15 seconds and up to 1000 tables. The cached routing tables will be updated on new connections and disconnections to keep them up-to-date. Adding to a table is O(log n) deleting is O(1). Complete reconstruction is O(n log n). Thus it makes more sense to cache and maintain frequently used tables instead of  reconstructing them on demand.

## Future improvements

- Could make cache options configurable. It could be useful in some cases to store more tables or keep them stored for longer.
- Remove the `contacted` field from `SortedContactList` values as it is no longer needed anywhere.

### DIFF
--- a/packages/dht/src/dht/routing/Router.ts
+++ b/packages/dht/src/dht/routing/Router.ts
@@ -8,7 +8,7 @@ import { DhtNodeRpcRemote } from '../DhtNodeRpcRemote'
 import { v4 } from 'uuid'
 import { RouterRpcLocal, createRouteMessageAck } from './RouterRpcLocal'
 import { DhtAddress, areEqualPeerDescriptors, getDhtAddressFromRaw, getNodeIdFromPeerDescriptor } from '../../identifiers'
-import { RoutingTablesCache } from './RoutingTableCache'
+import { RoutingTablesCache } from './RoutingTablesCache'
 
 export interface RouterConfig {
     rpcCommunicator: RoutingRpcCommunicator
--- a/packages/dht/src/dht/routing/Router.ts
+++ b/packages/dht/src/dht/routing/Router.ts
@@ -8,6 +8,7 @@ import { DhtNodeRpcRemote } from '../DhtNodeRpcRemote'
 import { v4 } from 'uuid'
 import { RouterRpcLocal, createRouteMessageAck } from './RouterRpcLocal'
 import { DhtAddress, areEqualPeerDescriptors, getDhtAddressFromRaw, getNodeIdFromPeerDescriptor } from '../../identifiers'
+import { RoutingTableCache } from './RoutingTableCache'
 
 export interface RouterConfig {
     rpcCommunicator: RoutingRpcCommunicator
@@ -27,6 +28,7 @@ const logger = new Logger(module)
 export class Router {
 
     private readonly forwardingTable: Map<DhtAddress, ForwardingTableEntry> = new Map()
+    private readonly routingTableCache = new RoutingTableCache()
     private ongoingRoutingSessions: Map<string, RoutingSession> = new Map()
     // TODO use config option or named constant?
     private readonly duplicateRequestDetector: DuplicateDetector = new DuplicateDetector(100000)
@@ -168,7 +170,8 @@ export class Router {
             // TODO use config option or named constant?
             parallelism: areEqualPeerDescriptors(this.config.localPeerDescriptor, routedMessage.sourcePeer!) ? 2 : 1,
             mode,
-            excludedNodeIds
+            excludedNodeIds,
+            routingTableCache: this.routingTableCache
         })
     }
 
@@ -186,6 +189,10 @@ export class Router {
 
     public removeRoutingSession(sessionId: string): void {
         this.ongoingRoutingSessions.delete(sessionId)
+    }
+
+    public resetCache(): void {
+        this.routingTableCache.reset()
     }
 
     public stop(): void {
--- a/packages/dht/src/dht/routing/Router.ts
+++ b/packages/dht/src/dht/routing/Router.ts
@@ -194,10 +194,6 @@ export class Router {
     onNodeConnected(peerDescriptor: PeerDescriptor): void {
         const remote = new RoutingRemoteContact(peerDescriptor, this.config.localPeerDescriptor, this.config.rpcCommunicator)
         this.routingTableCache.onNodeConnected(remote)
-        if (this.ongoingRoutingSessions.size > 1000) {
-            logger.warn('Too many ongoing routing sessions, clearing cache')
-            this.resetCache()
-        }
     }
 
     onNodeDisconnected(peerDescriptor: PeerDescriptor): void {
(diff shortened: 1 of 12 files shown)


packages/dht/src/dht/routing/RoutingSession.ts
```diff
@@ -167,21 +159,37 @@ export class RoutingSession extends EventEmitter<RoutingSessionEvents> {
...
+                .map((peer) => new RoutingRemoteContact(
+                    peer.getPeerDescriptor(),
+                    this.config.localPeerDescriptor,
+                    this.config.rpcCommunicator
+                ))
+            routingTable.addContacts(contacts)
+            this.config.routingTableCache.set(targetId, routingTable, previousId)
+        }
+        return routingTable.getAllContacts()
+            .filter((contact) => !this.contactedPeers.has(contact.getNodeId()) && !this.config.excludedNodeIds.has(contact.getNodeId()))
```
teogeb, line 188: Do we need `this.config.excludedNodeIds.has(contact.getNodeId())` as we set `excludedNodeIds: this.config.excludedNodeIds` at line `176`?

packages/dht/src/dht/routing/Router.ts
```diff
@@ -188,6 +191,19 @@ export class Router {
         this.ongoingRoutingSessions.delete(sessionId)
     }
 
+    onNodeConnected(peerDescriptor: PeerDescriptor): void {
```
ptesavol, line 194: Should this be called handleNodeConnected() as this is not an event listener but a public function called from outside (don't know if this convention is still in effect?)

packages/dht/src/dht/routing/Router.ts
```diff
@@ -188,6 +191,19 @@ export class Router {
...
+        const remote = new RoutingRemoteContact(peerDescriptor, this.config.localPeerDescriptor, this.config.rpcCommunicator)
+        this.routingTableCache.onNodeConnected(remote)
+    }
+
+    onNodeDisconnected(peerDescriptor: PeerDescriptor): void {
```
ptesavol, line 199: same comment here

packages/dht/src/dht/routing/RoutingSession.ts
```diff
@@ -13,29 +13,30 @@ import { EXISTING_CONNECTION_TIMEOUT } from '../contact/RpcRemote'
 import { getPreviousPeer } from './getPreviousPeer'
 import { DhtAddress, areEqualPeerDescriptors, getDhtAddressFromRaw, getNodeIdFromPeerDescriptor } from '../../identifiers'
 import { pull } from 'lodash'
+import { RoutingTableCache } from './RoutingTableCache'
 
 const logger = new Logger(module)
 
 const MAX_FAILED_HOPS = 2
-const CONTACT_LIST_MAX_SIZE = 10
+const ROUTING_TABLE_MAX_SIZE = 20
```
teogeb, line 21: Maybe could be more explicit:

```suggestion
const ROUTING_TABLE_MAX_CONTACT_COUNT = 20
```

packages/dht/src/dht/routing/RoutingTableCache.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    set(targetId: DhtAddress, table: SortedContactList<RoutingRemoteContact>, previousId?: DhtAddress): void {
+        this.tables.set(createRoutingTableId(targetId, previousId), table)
+    }
+
+    has(targetId: DhtAddress, previousId?: DhtAddress): boolean {
+        return this.tables.has(createRoutingTableId(targetId, previousId))
+    }
+
+    onNodeDisconnected(nodeId: DhtAddress): void {
```
ptesavol: handle..

packages/dht/src/dht/routing/RoutingTableCache.ts
```diff
@@ -0,0 +1,57 @@
...
+        this.tables.forEach((table) => table.removeContact(nodeId))
+    }
+
+    onNodeConnected(remote: RoutingRemoteContact): void {
```
ptesavol: handle..

ptesavol (approved):
Looks good otherwise, but check the comments on onSomething() vs handleSomething() (however, I don't know if this convention is out-of-date already)

packages/dht/src/dht/routing/RoutingTableCache.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Storing the previousId is important as it is used as a minimum distance for the contacts in the table.
+ * Calculating a RoutingTable from scratch is an O(n log n) operation (n = number of connections of a node)
+ * However,
+ * - Adding a contact to a RoutingTable is an O(log n) operation.
+ * - Deleting a contact from a RoutingTable is an O(1) operation.
+ * Thus, holding the most frequently used routing tables in memory to be updated on 
+ * connections and disconnections is hugely beneficial in terms of performance.
+*/
+
+export class RoutingTableCache {
```
teogeb: Could be plural to communicate explicitly that the caches contains multiple tables:

```suggestion
export class RoutingTablesCache {
```

packages/dht/src/dht/routing/RoutingTableCache.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    private readonly tables: LRUCache<RoutingTableID, SortedContactList<RoutingRemoteContact>> = new LRUCache(DEFAULT_LRU_OPTIONS)
```
teogeb: Could maybe introduce a helper type:
`type RoutingTable = Pick<SortedContactList<RoutingRemoteContact>, 'addContacts' | 'getAllContacts'>`

teogeb (commented):
First round of comments

packages/dht/src/dht/routing/RoutingSession.ts
juslesan: Yeah actually line 176 should not be there as it makes the tables potentially unusable between sessions... Fixing this

packages/dht/src/dht/routing/RoutingSession.ts
juslesan: I prefer `ROUTING_TABLE_MAX_SIZE `

packages/dht/src/dht/routing/Router.ts
teogeb: We currently apply this `onFooBar` naming style in many components, e.g. in `PeerManager#onContactConnected,` `RoutingSession#onRequestSucceeded`, `DiscoverySession#onClosestPeersRequestFailed.` 

The point is that the method is called when some event/action happens, so from the component's point of view it is just an event listener. In many cases the listener is triggered by external component (as that component knows that the event/action happened), but that doesn't make big difference.

harbu: Maybe let's try to keep the `onXXX` convention (or alternatively use a method name that describe what is done)?

PR #2380, opened 2024-02-19
refactor(trackerless-network): `StreamMessage` structure

teogeb:
## Summary

Divided `StreamMessage` entity to three sub entities: `ContentMessage`, `GroupKeyRequest` and `GroupKeyResponse`:

```proto
message StreamMessage {
  MessageID messageId = 1;
  optional MessageRef previousMessageRef = 2;
  bytes signature = 3;
  SignatureType signatureType = 4;
  oneof body {
    ContentMessage contentMessage = 5;
    GroupKeyRequest groupKeyRequest = 6;
    GroupKeyResponse groupKeyResponse = 7;
  }
}
```

The similar structure was already used in the `DhtRpc.proto`'s `Message` entity.

## Other changes

Small change in broadcast metrics: we now include only stream messages, not group key messages. This makes sense, as we already reported just the message content length, i.e. we didn't include e.g. the message headers or other kind of control data. The reason for this change is that there is no longer `content` field in group key messages from which we could calculate the content length.

## Open questions

Should the `previousMessageRef` field be in the`StreamMessage` entity or in the `ContentMessage`? Only content messages have references to previous messages, but maybe some messages in the future could also have that kind of sequencing. If we move it, it would make sense to move the `messageId#messageChainId` too as the previous message reference is always `messageChain`-specific. Therefore it would maybe be most consistent if we keep it in the `StreamMessage` (i.e. as it is now in this PR).

## Future improvements

The `DhtRpc.proto`'s `Message` entity has a separate `messageType` field. That is maybe not needed as we can infer the type from the `oneof` object (e.g. `if (msg.body.oneofKind === 'contentMessage') ...`).

### DIFF
--- a/packages/trackerless-network/src/logic/StreamrNode.ts
+++ b/packages/trackerless-network/src/logic/StreamrNode.ts
@@ -113,8 +113,10 @@ export class StreamrNode extends EventEmitter<Events> {
         logger.debug(`Broadcasting to stream part ${streamPartId}`)
         this.joinStreamPart(streamPartId)
         this.streamParts.get(streamPartId)!.broadcast(msg)
-        this.metrics.broadcastMessagesPerSecond.record(1)
-        this.metrics.broadcastBytesPerSecond.record(msg.content.length)
+        if (msg.body.oneofKind === 'contentMessage') {
+            this.metrics.broadcastMessagesPerSecond.record(1)
+            this.metrics.broadcastBytesPerSecond.record(msg.body.contentMessage.content.length)
+        }
     }
 
     async leaveStreamPart(streamPartId: StreamPartID): Promise<void> {
--- a/packages/trackerless-network/protos/NetworkRpc.proto
+++ b/packages/trackerless-network/protos/NetworkRpc.proto
@@ -69,17 +69,24 @@ enum SignatureType {
 }
 
 message StreamMessage {
-  StreamMessageType messageType = 1;
+  // this is a required field but in generated NetworkRpc.ts it is incorrectly annotated as optional (NET-1082)
+  MessageID messageId = 1;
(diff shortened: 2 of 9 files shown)


packages/trackerless-network/src/logic/StreamrNode.ts
```diff
@@ -113,8 +113,10 @@ export class StreamrNode extends EventEmitter<Events> {
         logger.debug(`Broadcasting to stream part ${streamPartId}`)
         this.joinStreamPart(streamPartId)
         this.streamParts.get(streamPartId)!.broadcast(msg)
-        this.metrics.broadcastMessagesPerSecond.record(1)
-        this.metrics.broadcastBytesPerSecond.record(msg.content.length)
+        if (msg.body.oneofKind === 'contentMessage') {
+            this.metrics.broadcastMessagesPerSecond.record(1)
+            this.metrics.broadcastBytesPerSecond.record(msg.body.contentMessage.content.length)
+        }
```
harbu, line 119: Behavioral change?

harbu (approved):

packages/trackerless-network/src/logic/StreamrNode.ts
harbu: NVM answered in description.

harbu:
> Should the previousMessageRef field be in theStreamMessage entity and or in the ContentMessage? Only content messages have references to previous messages, but maybe some messages in the future could also have that kind of sequencing.

Yeah good point, we could move it. OTOH as you said if we need sequencing for other message types in future it could be handy to be at the `StreamMessage`. I would maybe keep it there for now?

juslesan (approved):

PR #2383, opened 2024-02-20
chore(DHT): use WS library instead of websocket in the server side

juslesan:
## Summary

Start using the ws library instead of websocket in the server side. This is slight performance improvement when the number of received incoming and outgoing messages to a node is high.

## Other changes

- Renamed ServerWebsocket -> WebsocketServerConnection.
- Removed `getRemoteAddress` and `getRemoteIP` from WebsocketServerConnection. Instead the class has a new public field `remoteIpAddress`. This is always the public facing IP address.
- Removed `NodeJsWsServer` variable from WebsocketServer as it is no longer required for electron tests.

## Future improvements
- Disconnect WebsocketServerConnections if they receive string data


packages/dht/src/connection/websocket/WebsocketServerConnection.ts
```diff
@@ -44,20 +41,21 @@ export class ServerWebsocket extends EventEmitter<ConnectionEvents> implements I
...
+            logger.trace('Received string Message')
+        } else if (message instanceof Buffer) {
+            logger.trace('Received Buffer Message')
+            this.emit('data', new Uint8Array(message))
+        // If in an Karma / Electron test, use the NodeJS implementation
+        } else if (typeof NodeJsBuffer !== 'undefined' && message instanceof NodeJsBuffer) {
+            logger.trace('Received NodeJsBuffer Message')
+            this.emit('data', new Uint8Array(message))
         }
     }
```
harbu: Is there a real need for the last else if block (else than logging), just collapse with the `message instanceof Buffer` one?

harbu (approved):

packages/dht/src/connection/websocket/WebsocketConnector.ts
```diff
@@ -143,10 +143,10 @@ export class WebsocketConnector {
     public async start(): Promise<void> {
         if (!this.abortController.signal.aborted && this.websocketServer) {
             this.websocketServer.on('connected', (connection: IConnection) => {
-                const serverSocket = connection as unknown as ServerWebsocket
-                const query = serverSocket.resourceURL.query as unknown as (ParsedUrlQuery | null)
+                const serverSocket = connection as unknown as WebsocketServerConnection
+                const query = queryString.parse(serverSocket.resourceURL.query as string)
```
teogeb: If there is on query (the `?foo=bar` part), will this `value` be empty or `null` or `undefined`? If can be `null` or `undefined`, could adjust the type. If not, could remove the question mark at line `148`.

packages/dht/src/connection/websocket/WebsocketServer.ts
```diff
@@ -139,7 +131,10 @@ export class WebsocketServer extends EventEmitter<ConnectionSourceEvents> {
         this.abortController.abort()
         this.removeAllListeners()
         return new Promise((resolve, _reject) => {
-            this.wsServer?.shutDown()
+            this.wsServer!.close()
```
teogeb, line 134: It would maybe make sense to close the connections (i.e. clients) first and then the server? If the behavior of closed server is not well defined.

packages/dht/src/connection/websocket/WebsocketServerConnection.ts
```diff
@@ -44,20 +41,21 @@ export class ServerWebsocket extends EventEmitter<ConnectionEvents> implements I
...
-        if (message.type === MessageType.UTF8) {
-            logger.debug('Received string Message: ' + message.utf8Data)
-        } else if (message.type === MessageType.BINARY) {
-            logger.trace('Received Binary Message of ' + message.binaryData.length + ' bytes')
-            this.emit('data',
-                new Uint8Array(message.binaryData.buffer, message.binaryData.byteOffset,
-                    message.binaryData.byteLength / Uint8Array.BYTES_PER_ELEMENT))
+    private onMessage(message: WebSocket.RawData, isBinary: boolean): void {
+        if (!isBinary) {
+            logger.trace('Received string Message')
```
teogeb, line 40: Don't we emit any data from string messages? Is sending strings messages a supported use case or not?

teogeb (changes requested):

packages/dht/src/connection/websocket/WebsocketServerConnection.ts
juslesan: no string messages are not supported

teogeb: Would it be possible to close the WebSocket connection with an error message? And maybe write a log message which says explicitly that this is not supported.

packages/dht/src/connection/websocket/WebsocketServer.ts
juslesan: The WS server should always be closed first. This ensures that new connections are no longer accepted

packages/dht/src/connection/websocket/WebsocketServerConnection.ts
juslesan: removing the NodeJsBuffer completely as it should not be necessary.

packages/dht/src/connection/websocket/WebsocketServerConnection.ts
juslesan: It could be possible and probably would make sense as it violates the protocol. However, I would leave it out of this PR.

packages/dht/src/connection/websocket/WebsocketConnector.ts
teogeb: The `serverSocket.resourceURL.query` is be `null` if there is no query string. E.g. if we do `websocat 'ws://10.200.10.1:40500/'`.

teogeb: The `serverSocket.resourceURL.query` is be `null` if there is no query string. E.g. if we do `websocat 'ws://10.200.10.1:40500/'`.

packages/dht/src/connection/websocket/WebsocketServerConnection.ts
teogeb: Could add that improvement to the "Future improvements" section.

packages/dht/src/connection/websocket/WebsocketConnector.ts
juslesan: const query = {} at line 147 in this case

teogeb: This would maybe work, and be also type-correct:

```suggestion
                const query = queryString.parse(serverSocket.resourceURL.query ?? '')
```

teogeb (approved):

ptesavol (approved):

PR #2390, opened 2024-02-21
refactor(network): NET-1271 lock/unlock connections based on neighbor events in RandomGraphNode

juslesan:
## Summary

Lock and unlock connections in Layer2 based on nodeAdded and nodeRemoved events in RandomGraphNode instead of doing the operation saparetely each time nodes are removed or added to the neighbor struct

## Future improvements

Add a parameter for enabling emitting nodeRemoved events in the replaceNodes method.

### DIFF
--- a/packages/trackerless-network/src/logic/NodeList.ts
+++ b/packages/trackerless-network/src/logic/NodeList.ts
@@ -4,7 +4,8 @@ import { DeliveryRpcRemote } from './DeliveryRpcRemote'
 import { EventEmitter } from 'eventemitter3'
 
 export interface Events {
-    nodeAdded: (id: DhtAddress, remote: DeliveryRpcRemote) => any
+    nodeAdded: (id: DhtAddress, remote: DeliveryRpcRemote) => void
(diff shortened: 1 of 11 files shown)


packages/trackerless-network/src/logic/NodeList.ts
```diff
@@ -40,13 +41,18 @@ export class NodeList extends EventEmitter<Events> {
...
+            this.nodes.delete(nodeId)
+            this.emit('nodeRemoved', nodeId, remote)
+        }   
     }
 
     has(nodeId: DhtAddress): boolean {
         return this.nodes.has(nodeId)
     }
 
+    // Replace nodes does not emit nodeRemoved events, use with caution
```
teogeb, line 55: Maybe it could emit the events? Or could add a parameter which controls the event emitting.

juslesan: Yeah a parameter could make sense. The method is never called for the neighbor structure, only for contacts. Contact structures don't need the event for anything at the moment

teogeb (approved):

PR #2414, opened 2024-03-04
fix(dht): Provide all parameters to `NodeWebrtcConnection`

teogeb:
## Summary

If there a WebRTC answerer behind a NAT, it doesn't connect to STUN server.

When `WebrtcConnectorLocalRpc` created connection instances, it ignored most of the constructor parameter which we should pass to `NodeWebrtcConnection`. It passed just the remote peer, but e.g. not ICE servers:
```ts
new NodeWebrtcConnection({ remotePeerDescriptor: remotePeer })
```

## Changes

Moved the connection construction out of `WebrtcConnectorLocalRpc` by introducing `config.createConnection()` function.

## Open questions

- Some tests are failing, maybe as a consequence of this change: https://github.com/streamr-dev/network/actions/runs/8140879404/job/22246994771?pr=2414

## Future improvements

- Improve test coverage so that we test also situations which use STUN servers. 
- Should mark `iceServers` config option and maybe other options as required so that we can avoid this kind of bug in the future. For arrays we can use explicit empty array.

### DIFF
--- a/packages/dht/src/connection/webrtc/NodeWebrtcConnection.ts
+++ b/packages/dht/src/connection/webrtc/NodeWebrtcConnection.ts
@@ -26,7 +26,7 @@ export interface Params {
     bufferThresholdLow?: number
     connectingTimeout?: number
     maxMessageSize?: number
-    iceServers?: IceServer[]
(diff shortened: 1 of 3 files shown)


ptesavol (approved):
Please merge ASAP

juslesan (approved):
The failing tests are known flaky tests and should not be affected by the changes

PR #2416, opened 2024-03-04
feat(DHT): Exclude already attempted entry points during connectivity checks

juslesan:
## Summary

Exclude already attempted entry points when attempting connectivity checks.

Improves start up reliability in production. In dev environment connecting to starting a entry point can be more difficult as it will be excluded after the first attempt.

## Other changes

Added abort signal to loop condition and `wait` in checkConnectivity


packages/dht/src/connection/websocket/WebsocketConnector.ts
```diff
@@ -192,8 +192,16 @@ export class WebsocketConnector {
...
+        const entrypoints = new Map<DhtAddress, PeerDescriptor>()
+        for (const entrypoint of this.config.entrypoints) {
+            entrypoints.set(getNodeIdFromPeerDescriptor(entrypoint), entrypoint)
+        }
+        const exclude: DhtAddress[] = []
+        const getExcludedIds = () => Array.from(entrypoints.keys()).filter((key) => !exclude.includes(key))
+        while (getExcludedIds().length > 0) {
+            const excludedIds = getExcludedIds()
+            const entryPointId = sample(excludedIds)!
+            const entryPoint = entrypoints.get(entryPointId)!
```
harbu: Could use lodash/shuffle here to simplify?
```suggestion
        for (const entryPoint of shuffle(this.config.entrypoints)) {
```

packages/dht/src/connection/websocket/WebsocketConnector.ts
```diff
@@ -208,10 +216,11 @@ export class WebsocketConnector {
                     throw new Err.ConnectionFailed('ConnectivityChecker is destroyed')
```
harbu, line 209: Not necessarily in this PR but looks like we do re-tries even when `this.abortController.signal.aborted` is true? Should this be checked in the catch-block to avoid unnecessary waiting and reattempts?

packages/dht/src/connection/websocket/WebsocketConnector.ts
juslesan: I'd prefer to shuffle first, always pick the first index and then remove it in a while loop to make excluding possible

harbu (approved):

PR #2425, opened 2024-03-11
feat(client): Config option to disable WebSocket server

teogeb:
Client config option `network.controlLayer.websocketPortRange`  can be `null`. In that case no WebSocket server is started.


packages/client/src/Config.ts
```diff
@@ -109,7 +109,7 @@ export interface ControlLayerConfig {
      * The server is used by the network layer to accept incoming connections
      * over the public internet to improve the network node's connectivity.
      */
-    websocketPortRange?: PortRange
+    websocketPortRange?: PortRange | null
```
harbu, line 112: Should we update the comment above as follows?

> If not specified, a server will not be started.

into

> If *set to null*, a server will not be started.

juslesan (approved):

packages/client/src/Config.ts
teogeb: Updated

harbu (approved):

PR #2430, opened 2024-03-14
refactor(trackerless-network): Rename `DeliveryRpc`

teogeb:
Rename `DeliveryRpc` -> `ContentDeliveryRpc` so that the naming is consistent e.g. with the new 
`ContentDeliveryManager` class (https://github.com/streamr-dev/network/pull/2140).

### DIFF
--- a/packages/trackerless-network/protos/NetworkRpc.proto
+++ b/packages/trackerless-network/protos/NetworkRpc.proto
@@ -168,6 +168,7 @@ enum ProxyDirection {
 message StreamPartitionInfo {
   string id = 1;
   repeated dht.PeerDescriptor controlLayerNeighbors = 2;
+  // could be "contentDelivery" instead of "deliveryLayer", but that is a breaking change
   repeated dht.PeerDescriptor deliveryLayerNeighbors = 3;
 }
 
--- a/packages/trackerless-network/protos/NetworkRpc.proto
+++ b/packages/trackerless-network/protos/NetworkRpc.proto
@@ -6,7 +6,7 @@ import "google/protobuf/empty.proto";
 import "google/protobuf/any.proto";
 import "packages/dht/protos/DhtRpc.proto";
 
-service DeliveryRpc {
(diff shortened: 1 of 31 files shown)


packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -168,6 +168,7 @@ enum ProxyDirection {
 message StreamPartitionInfo {
   string id = 1;
   repeated dht.PeerDescriptor controlLayerNeighbors = 2;
+  // could be "contentDelivery" instead of "deliveryLayer", but that is a breaking change
```
juslesan: Is only a breaking change for the API and would it would be caught in the build process. In the protocol layer renaming fields is not a breaking change.

For example the crawler running an older version would run without any need for changes. After updating the crawler to a version with this change, any code that touches this field needs to adapt to the renaming change.

packages/trackerless-network/src/logic/formStreamPartDeliveryServiceId.ts
```diff
@@ -1,6 +1,7 @@
 import { ServiceID } from '@streamr/dht'
 import { StreamPartID } from '@streamr/protocol'
 
-export const formStreamPartDeliveryServiceId = (streamPartId: StreamPartID): ServiceID => {
+export const formStreamPartContentDeliveryServiceId = (streamPartId: StreamPartID): ServiceID => {
+    // could be "content-delivery" instead of "delivery", but that is a breaking change
```
juslesan, line 5: Now this would a breaking change in the protocol layer, but not in the application layer.

juslesan (approved):

packages/trackerless-network/protos/NetworkRpc.proto
teogeb: Renamed the field

harbu (approved):

juslesan (approved):

PR #2432, opened 2024-03-14
fix(DHT) do distant peer discovery on rejoin

juslesan:
## Summary

Do distant peer discovery on rejoins to ensure that the node is connects to nodes further away in the kademlia id space


packages/dht/src/dht/discovery/PeerDiscovery.ts
```diff
@@ -155,14 +155,18 @@ export class PeerDiscovery {
         }
     }
 
-    public async rejoinDht(entryPoint: PeerDescriptor): Promise<void> {
+    public async rejoinDht(
+        entryPoint: PeerDescriptor,
+        contactedPeers: Set<DhtAddress> = new Set(),
```
teogeb, line 160: These parameter are not used by any caller?

teogeb (changes requested):

teogeb (approved):

PR #2437, opened 2024-03-17
refactor(dht): Simplify `PeerManager#addContact`

teogeb:
- now takes only one contact as an argument, not an array of contacts
- removed obsolete `setActive` parameter 
  - that parameter was previously used in `RouterRpcLocal`, but the functionality was removed in PR #2332)
  - removed also the obsolete router config options related to this change


juslesan (approved):

juslesan:
We might need something like this to mark which contacts are known to be online. This way the chance of propagation of offline contacts in the network is less likely. Although it would be better to just do more work to remove contacts that are offline from the contact structures

PR #2447, opened 2024-03-20
refactor(dht): Remove `ContactState#contacted`

teogeb:
Removed the `contacted` field of `ContactState` and the relevant accessors as those were not used anywhere.


juslesan (approved):

packages/dht/test/benchmark/kademlia-simulation/SimulationNode.ts
```diff
@@ -71,7 +71,6 @@ export class SimulationNode {
 
     private findMoreContacts(contactList: Contact[], shortlist: SortedContactList<Contact>) {
         contactList.forEach((contact) => {
-            shortlist.setContacted(contact.getNodeId())
```
juslesan: Here we set contactedn and in further lines we `getUncontactedContacts`. In the changed logic we use `getActiveContacts` so the logic has reversed.

PR #2455, opened 2024-03-27
refactor(dht): Use `SortedContactList#getClosestContacts`

teogeb:
Use the `getClosestContacts` and `getFurthestContacts` methods instead of the `getAllContacts` method when querying the closes/furthest contacts from a contact list.

The `getFurthestContacts` is a new method and it is just a reverse of `getClosestContacts`.

Added also a check for the `limit` parameter: if it is negative, no items are returned.

### DIFF
--- a/packages/dht/src/dht/contact/SortedContactList.ts
+++ b/packages/dht/src/dht/contact/SortedContactList.ts
@@ -116,7 +116,7 @@ export class SortedContactList<C extends { getNodeId: () => DhtAddress }> extend
      * Furthest first then others in descending distance order
      */
     getFurthestContacts(limit?: number): C[] {
-        const ret = this.getClosestContacts().reverse()
+        const ret = this.getClosestContacts().toReversed()
         return (limit === undefined) 
             ? ret 
             : ret.slice(0, Math.max(limit, 0))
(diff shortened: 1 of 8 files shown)


packages/dht/src/dht/contact/SortedContactList.ts
```diff
@@ -102,11 +102,24 @@ export class SortedContactList<C extends { getNodeId: () => DhtAddress }> extend
...
             ? ret 
-            : ret.slice(0, limit)
+            : ret.slice(0, Math.max(limit, 0))
+    }
+
+    /*
+     * Furthest first then others in descending distance order
+     */
+    getFurthestContacts(limit?: number): C[] {
+        const ret = this.getClosestContacts().reverse()
```
juslesan: Not urgently important here but:

`reverse()` mutates the original array and returns the mutated array. `toReversed()` returns the reversed array without mutating the original array.

juslesan (approved):

packages/dht/src/dht/contact/SortedContactList.ts
teogeb: Fixed

PR #2456, opened 2024-03-27
refactor(dht): Manage active states in `PeerManager`

teogeb:
Moved the state management of active contacts from `SortedContactList` to `PeerManager`.  The state is not relevant in other use cases of `SortedContactList`. This change allows us to remove the `ContactState` wrapper and therefore optimize some `SortedContactList` methods significantly.

## Future improvements

- Remove `ContactState` wrapper
- Rename `PeerManagerEvents` arguments


juslesan (approved):
Future consideration:

The active contacts are quite poorly maintained. They are only set when the node initially joins the DHT. After this nodes are no longer set as active. 

For long running nodes this can lead to poor ability to find new neighbors if pings fail. In the worst cases there will be 0 active contacts remaining in the set. Could make sense to have better upkeep for active contacts. For example if another node calls the `getClosestPeers` in DhtNodeRpcLocal the contact could be set as active.

PR #2464, opened 2024-04-02
feat(client) [NET-1281]: Include nodeId in metrics payload

teogeb:
In Brubeck era we used the `publisherId` of the message to identify the publisher node. Now it makes sense to add a separate field as there is no trivial mapping between `publisherId` and `nodeId`.


mondoreale (approved):
👍🏻

harbu (approved):

PR #2465, opened 2024-04-02
GeoIp location

ptesavol:
* Add geoip-location package that can map an ip address to latitude and longitude using MaxMind's GeoLite City database

* If "geoIpDatabaseFolder" config variable is given to DhtNode, use geoip-location package server-side in DhtNode's ConnectivityRequest handler to add latitude and longitude to each ConnectivityReply

* Add function to cdn-location package for mapping coordinates to a local region number
 
* In DHTNode client-side, calculate local region number based on the latitude and longitude contained in ConnectivityReply. If ConnectivityReply does not contain latitude and longitude, fallback to using cdn-location.

* Take geoip-location server into use in Broker


packages/dht/src/dht/DhtNode.ts
```diff
@@ -422,11 +417,22 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
...
-            this.localPeerDescriptor = createPeerDescriptor(connectivityResponse, this.region!, this.config.nodeId)
+            let region: number | undefined = undefined
+            if (this.config.region !== undefined) {
+                region = this.config.region
+            } else if (connectivityResponse.latitude !== undefined && connectivityResponse.longitude !== undefined) {
+                region = getLocalRegionByCoordinates(connectivityResponse.latitude, connectivityResponse.longitude)
+            } else {
+                // as a fallback get the region from the CDN
+                // and if it's not available, use a random region
+                region = await getLocalRegion()
```
juslesan: This function does not need to be async if the cdn-region is queried during start "just in case". As the `generatePeerDescriptorCallBack` is vsllrf multiple times if autocertification is enabled it could be reasonable to do so

packages/geoip-location/tsconfig.jest.json
```diff
@@ -0,0 +1,17 @@
...
+        "noImplicitOverride": false
+    },
+    "include": [
+        "src/**/*",
+        "test/**/*"
+    ],
+    "references": [
+        { "path": "../utils/tsconfig.node.json" },
+        { "path": "../test-utils/tsconfig.node.json" },
+        { "path": "../proto-rpc/tsconfig.node.json" },
```
juslesan: remove unnecessary references

packages/geoip-location/src/downloadGeoIpDatabase.ts
```diff
@@ -0,0 +1,156 @@
+import crypto from 'crypto'
+import fs from 'fs'
+import { Readable } from 'stream'
+import { finished } from 'stream/promises'
+import tar from 'tar'
+import NodePath from 'path'     // use NodePath to avoid conflict with other 'path' symbols
+import { CityResponse, Reader } from 'mmdb-lib'
+
+const GEOIP_MIRROR_URL = 'https://raw.githubusercontent.com/GitSquared/node-geolite2-redist/master/redist/'
```
juslesan, line 8: Would it make sense to have a more static version being downloaded than the master branch? Just in case...

packages/geoip-location/src/downloadGeoIpDatabase.ts
```diff
@@ -0,0 +1,156 @@
...
+            fs.rmSync(dbPath + '.download', { recursive: true })
+        } catch (e2) {
+            // ignore error when removing the temporary folder
+        }
+        throw new Error('Downloaded database hash does not match the expected hash')
+    }
+
+    try {
+        // move the extracted file to the correct location
+        fs.renameSync(dbPath + '.download/' + DB_NAME + DB_SUFFIX, dbPath + DB_NAME + DB_SUFFIX)
```
juslesan: Make the correct path into a variable

packages/geoip-location/src/downloadGeoIpDatabase.ts
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    const remoteHash = await downloadRemoteHash(abortSignal)
+    
+    const isDbValid = await isCurrentDbValid(dbFolder, remoteHash)
+    
+    if (isDbValid === false) {
+        await downloadNewDb(GEOIP_MIRROR_URL + DB_NAME + TAR_SUFFFIX, dbFolder, remoteHash, abortSignal)
+        // return new reader if db was downloaded
+        return new Reader<CityResponse>(fs.readFileSync(dbFolder + DB_NAME + DB_SUFFIX))
```
juslesan: make file path into a variable

packages/dht/src/dht/DhtNode.ts
juslesan: Or just cache the result

packages/dht/src/connection/websocket/WebsocketServerConnection.ts
```diff
@@ -26,14 +26,19 @@ export class WebsocketServerConnection extends EventEmitter<ConnectionEvents> im
...
+        this.remoteAddress = remoteAddress
 
         socket.on('message', this.onMessage)
         socket.on('close', this.onClose)
         socket.on('error', this.onError)
 
         this.socket = socket
     }
+    
+    // use a getter to make it possible to mock the value in tests
```
juslesan, line 38: The value is not mocked in any tests so could just revert the change

packages/cdn-location/data-generation/final-data/airportCodeToRegionNumber.ts
```diff
@@ -1,345 +1,345 @@
...
-    LCK: 33500,
-    CLT: 33600,
-    RDU: 33700,
-    ORF: 33800,
-    RIC: 33900,
-    IAD: 34000,
-    DCA: 34100,
-    PHL: 34200,
-    NYC: 34300,
+export const airportCodeToRegionNumber: Record<string, [regionNumber: number, latitude: number, longitude: number]> = {
```
juslesan: Could rename the function since it now also returns coordinates.

juslesan: Could also return the value in a map instead of an array to make the code that uses the function more understandable.

packages/geoip-location/src/downloadGeoIpDatabase.ts
```diff
@@ -0,0 +1,156 @@
...
+import NodePath from 'path'     // use NodePath to avoid conflict with other 'path' symbols
+import { CityResponse, Reader } from 'mmdb-lib'
+
+const GEOIP_MIRROR_URL = 'https://raw.githubusercontent.com/GitSquared/node-geolite2-redist/master/redist/'
+const DB_NAME = 'GeoLite2-City'
+const TAR_SUFFFIX = '.tar.gz'
+const DB_SUFFIX = '.mmdb'
+const HASH_SUFFIX = '.mmdb.sha384'
+
+const downloadNewDb = async (url: string, dbPath: string, remoteHash: string,
```
juslesan: Wondering what happens if permissions do not allow for writing or reading in the directory

juslesan (commented):
Some comments

packages/dht/src/dht/DhtNode.ts
ptesavol: Actually this function not being async has been very problematic also in the past (had to invent all kinds of workarounds), I think it is good that we finally got rid of this problem and need no workarounds ever again.

packages/geoip-location/src/downloadGeoIpDatabase.ts
ptesavol: This package seems to keep the db up-to-date and EULA-compliant by committing a new db to the master branch twice a week, so downloading from the master means that we get the same database as the node-geolite-redist package users. If the db in the master branch is sometimes broken, our own geoip package keeps using the old db it alredy has.

packages/dht/src/connection/websocket/WebsocketServerConnection.ts
ptesavol: Yes it is mocked, for example here: https://github.com/streamr-dev/network/blob/7284f0195dd7380fe8095c0b8026d0e3a03f2ae7/packages/dht/test/end-to-end/GeoIpLayer0.test.ts#L34

packages/geoip-location/src/downloadGeoIpDatabase.ts
ptesavol: It throws, tested here: https://github.com/streamr-dev/network/blob/7284f0195dd7380fe8095c0b8026d0e3a03f2ae7/packages/geoip-location/test/unit/downloadGeoIpDatabase.test.ts#L19

packages/dht/src/dht/DhtNode.ts
juslesan: Probably still a good idea to cache the results of the cdn-location query.

packages/geoip-location/src/downloadGeoIpDatabase.ts
ptesavol: In order to alleviate security concerns, forced the x chmod flag to be off on all downloaded files.

packages/geoip-location/tsconfig.jest.json
ptesavol: done

packages/cdn-location/data-generation/final-data/airportCodeToRegionNumber.ts
ptesavol: This is actually generated data (not a function), and there are a lot of code that depends on the structure of this data (the data gets sorted etc). The file is large already, and converting it into objects would almost double its size. I think it is better not to touch this.

packages/dht/src/dht/DhtNode.ts
ptesavol: added getLocalRegionWithCache() function to cdn-location

packages/cdn-location/data-generation/final-data/airportCodeToRegionNumber.ts
ptesavol: renamed airportCodeToRegionNumber to airportCodeToRegion

juslesan:
Should add the geoip-location package to the `release.sh` file

packages/dht/src/connection/websocket/WebsocketConnector.ts
```diff
@@ -77,6 +80,7 @@ export class WebsocketConnector {
             maxMessageSize: config.maxMessageSize,
             enableTls: config.serverEnableTls
         }) : undefined
+
```
juslesan:
```suggestion
```

packages/dht/src/connection/websocket/WebsocketConnector.ts
```diff
@@ -159,12 +163,22 @@ export class WebsocketConnector {
                         this.attachHandshaker(connection)
                     } else {
                         logger.trace('incoming Websocket connection before localPeerDescriptor was set, closing connection')
-                        connection.close(false).catch(() => {})
+                        connection.close(false).catch(() => { })
```
juslesan:
```suggestion
                        connection.close(false).catch(() => {})
```

packages/dht/src/connection/websocket/WebsocketConnector.ts
```diff
@@ -159,12 +163,22 @@ export class WebsocketConnector {
...
                     }
                 }
             })
             const port = await this.websocketServer.start()
             this.selectedPort = port
+
+            if (this.config.geoIpDatabaseFolder) {
```
juslesan, line 163: Could encapsulate the geoip-location related logic to a facade module.

juslesan: Perhaps it would make sense to start the geoip-location before setting the server listener?

packages/dht/src/connection/websocket/WebsocketServerConnection.ts
```diff
@@ -26,14 +26,19 @@ export class WebsocketServerConnection extends EventEmitter<ConnectionEvents> im
...
+    public get remoteIpAddress(): string {
```
juslesan:
```suggestion
    public getRemoteIpAddress(): string {
```

packages/geoip-location/src/GeoIpLocator.ts
```diff
@@ -0,0 +1,113 @@
...
+    private readonly dbCheckInterval: number
+    private readonly dbCheckErrorInterval: number
+    private readonly mirrorUrl?: string
+    private reader?: Reader<CityResponse>
+    private dbCheckTimeout?: LongTimeout.Timeout
+
+    // By default, check the database every 30 days
+    // If the check fails, retry after 24 hours
+
+    constructor(geoIpDatabaseFolder: string, dbCheckInterval: number = 30 * 24 * 60 * 60 * 1000,
```
juslesan: Make this default checkIntervals to contants

packages/geoip-location/test/unit/GeoIpLocator-intervals.test.ts
```diff
@@ -0,0 +1,107 @@
...
+    let testServer: TestServer | undefined
+
+    const getDbDir = () => {
+        dirCounter++
+        return dbPath + '/geolitelocator2-intervals' + dirCounter
+    }
+
+    beforeAll(async () => {
+        testServer = new TestServer()
+        await testServer.start(31991)
```
juslesan:
```suggestion
        await testServer.start(serverPort)
```

packages/geoip-location/test/unit/tarHelper.test.ts
```diff
@@ -0,0 +1,102 @@
...
+import { TestServer } from '../helpers/TestServer'
+
+let sPort = 3197
+
+const getServerPort = () => {
+    const port = sPort
+    sPort++
+    return port
+}
+describe('tarHelper', () => {
```
juslesan, line 5: Start and stop server is beforeEach and afterEach

packages/geoip-location/test/unit/GeoIpLocator-no-network-at-monthly.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    let locator: GeoIpLocator | undefined
+
+    const getDbDir = () => {
+        dirCounter++
+        return dbPath + '/geolite2-no-nw-monthly' + dirCounter
+    }
+
+    beforeAll(async () => {
+        testServer = new TestServer()
+        await testServer.start(31990)
```
juslesan: add serverPort constant, use it on line 23 as well

packages/geoip-location/test/unit/GeoIpLocator.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    const dbPath = '/tmp'
+
+    const getDbDir = () => {
+        dirCounter++
+        return dbPath + '/geolite2-' + dirCounter
+    }
+
+    beforeAll(async () => {
+        testServer = new TestServer()
+        await testServer.start(31992)
```
juslesan: add serverPort constant, use it on line 31 as well

juslesan (changes requested):
Should we remove the geoip-location packages from the browser bundle in the client?

packages/dht/src/connection/websocket/WebsocketServerConnection.ts
ptesavol: The use of an actual js getter is intentional, the user of WebSocketServerConnection does not need to know this field has been changed into a getter.

packages/geoip-location/test/unit/tarHelper.test.ts
ptesavol: Putting server start and stop in tests themselves is intentional. The servers are unique, they run on different ports and some of them get aborted in the middle. I tried using beforeEach() and afterEach(), which resulted in the servers of wrong test tests cases getting closed.

packages/dht/src/connection/websocket/WebsocketConnector.ts
ptesavol: done

packages/dht/src/connection/websocket/WebsocketConnector.ts
ptesavol: Changed the starting order, I don't think there is a need for a facade for such a simple component as this.

packages/dht/src/dht/DhtNode.ts
ptesavol: make cacheAge a optional parameter with default constant in cdnlocation

packages/dht/src/connection/websocket/WebsocketServerConnection.ts
ptesavol: change to java-style getter function

packages/geoip-location/test/unit/tarHelper.test.ts
ptesavol: Try beforeEach/afterEach with same port number for all test cases

packages/geoip-location/src/GeoIpLocator.ts
```diff
@@ -0,0 +1,109 @@
...
+        if (this.reader === undefined) {
+            logger.warn('GeoIpLocator: lookup called before database is ready (maybe start() was not called?')
+            return undefined
+        }
+
+        // If ip is undefined of null, the library will crash
+        // this might happen despite the ts typings because the ip address 
+        // comes from the ws server socket and is not under our control
+
+        if (!(ip as unknown)) {
```
juslesan: Is there any case where this would happen in the real world?

ptesavol: It did happen while testing, and that's why I added the check. 

In all, this is a timer-run system that is run very rarely, so we should make it as difficult to crash as possible, as crashes have a high probability of staying unnoticed ("It has been running fine for 20 days, so probably it will keep that way")

juslesan (approved):

PR #2468, opened 2024-04-02
feat(client): Use node ID as partition key in metrics publishing

teogeb:
The partition is now determined by `nodeId` instead of the `userId`. This way e.g. the Network Explorer can calculate a partition number in order to receive some node-specific metrics messages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Changes before Tatum release are not documented in this file.
 
 #### Changed
 
-- The partition of the metrics streams is calculated from the node ID
+- Change the way in which the partition of the metrics stream is calculated (based on node ID)
 
 #### Deprecated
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -12,8 +12,12 @@ Changes before Tatum release are not documented in this file.
 
 #### Added
 
+- Metrics stream messages contain the node ID
```
```suggestion
- Add node ID to metrics messages
```

CHANGELOG.md
```diff
@@ -12,8 +12,12 @@ Changes before Tatum release are not documented in this file.
...
+
 #### Changed
 
+- The partition of the metrics streams is calculated from the node ID
```
```suggestion
- Change the way in which the partition of the metrics stream is calculated (based on node ID)
```

harbu (approved):

PR #2470, opened 2024-04-02
refactor(trackerless-network): Rename `RandomGraphNode`

teogeb:
Rename `RandomGraphNode` as it no longer does random connections. Instead it nowadays uses closest nodes from Kademlia and the ring DHT.

Also removed obsolete `RandomGraphSimulation.ts` .

## Future improvements

- Rename `Layer1Node` to `StreamPartControlLayerNode` and `Layer0Node` to `ControlLayerNode`/`GlobalControlLayerNode`


packages/trackerless-network/src/logic/ContentDeliveryManager.ts
```diff
@@ -22,8 +22,8 @@ import { ProxyDirection, StreamMessage, StreamPartitionInfo } from '../proto/pac
 import { EntryPointDiscovery, NETWORK_SPLIT_AVOIDANCE_LIMIT } from './EntryPointDiscovery'
 import { Layer0Node } from './Layer0Node'
 import { Layer1Node } from './Layer1Node'
-import { RandomGraphNode } from './RandomGraphNode'
-import { createRandomGraphNode } from './createRandomGraphNode'
+import { ContentDeliveryLayerNode } from './ContentDeliveryLayerNode'
+import { createContentDeliveryLayerNode } from './createContentDeliveryLayerNode'
 import { ProxyClient } from './proxy/ProxyClient'
 
 export type StreamPartDelivery = {
```
juslesan, line 29: Probably need to think about the name of the type `StreamPartDelivery` at some point.

juslesan (approved):

PR #2471, opened 2024-04-02
docs: Changelog for v100.2.1

teogeb:
Created a section for v100.2.1 and move the items from unreleased section into that section. Also added two missing items.

Also:
- changed packages of some previous releases
- fixed a PR link (the changelog was added in one PR, but the feature was implemented in two)


harbu (approved):
:+1:

PR #2473, opened 2024-04-03
Use latest node-datachannel with help of EsmLoader

ptesavol:
* Add EsmLoader for loading esm modules into our CommonJS project
* Use EsmLoader to load the latest node-datachannel library
* Remove externals definitions from browser tests to make them work with the node-datachannel esm library
* Add --experimental-vm-modules env variable to jest invocations to make them work with esm modules


packages/dht/src/connection/ConnectorFacade.ts
```diff
@@ -152,8 +151,8 @@ export class DefaultConnectorFacade implements ConnectorFacade {
     }
 
     async stop(): Promise<void> {
-        await this.websocketConnector!.destroy()
-        await this.webrtcConnector!.stop()
+        await this.websocketConnector?.destroy()
```
juslesan: revert changes in the stop method.

packages/browser-test-runner/webpack.config.js
```diff
@@ -67,8 +67,10 @@ module.exports = function({ entry, libraryName, alias = {} }) {
                 umdNamedDefine: true,
             },
             externals: {
+                /*
```
juslesan: remove the commented lines completely

packages/dht/src/helpers/EsmLoader.ts
```diff
@@ -0,0 +1,37 @@
+import { isBrowserEnvironment } from './browser/isBrowserEnvironment'
+import { waitForCondition } from '@streamr/utils'
+
```
juslesan, line 3: Could add some doc comments here to explain why this is needed.

juslesan (changes requested):

packages/browser-test-runner/webpack.config.js
```diff
@@ -10,7 +10,7 @@ module.exports = function({ entry, libraryName, alias = {} }) {
             },
             mode: 'development',
             entry,
-            devtool: 'source-map',
+            devtool: 'eval-source-map',
```
harbu, line 13: What's the background of this change?

packages/dht/src/helpers/EsmLoader.ts
```diff
@@ -0,0 +1,37 @@
...
+declare global {
+    // eslint-disable-next-line no-var, no-underscore-dangle
+    var _streamr_electron_test: boolean
+}
+
+export type NodeDataChannel = typeof import('node-datachannel', { with: { "resolution-mode": "import" } })
+
+// eslint-disable-next-line @typescript-eslint/no-extraneous-class
+export class EsmLoader {
```
harbu, line 43: Is this something we can get rid of once (or if) we move to ESM packages ourselves? Maybe add a comment about this and a reference to ticket NET-963 related to moving the project into ESM.

harbu:
Does this change affect the end-users of say `@streamr/sdk`? Do they need to use the `--experimental-vm-modules` flag when running their code?

packages/browser-test-runner/webpack.config.js
ptesavol: source-map devtool does not seem to work at all (you can just see a lot of 'source map not found' warnings in beginning). eval-source-map seems to work, it finds the maps and prints out references to .ts linenumbers

ptesavol:
@harbu The --experimental-vm-modules flag is only needed by jest. jest uses some exotic virtual machine api to load esm:s  https://stackoverflow.com/questions/74069138/node-js-experimental-vm-modules-command-line-option-vs-type-module-in-pac

packages/browser-test-runner/webpack.config.js
harbu: got ya!

packages/dht/src/helpers/EsmLoader.ts
ptesavol: Yes, this EsmLoader is just a hack to make ESM modules work in our CommonJS codebase. Nothing like this will be needed if we move the whole project to ESM.

packages/dht/src/connection/ConnectorFacade.ts
ptesavol: done

harbu (approved):

juslesan (approved):
probably a good idea to do an experimental release of this change just in case something goes wrong with the npm packaging @harbu

harbu:
> probably a good idea to do an experimental release of this change just in case something goes wrong with the npm packaging @harbu

Good idea. Let's discuss on Monday call how we want to roll this out. (e.g. Keep branch or merge to main)

juslesan:
Can close now after node-datachannel v0.7.0? 
https://github.com/streamr-dev/network/pull/2505

juslesan:
No longer needed as CJS is supported bu node-datachannel since v0.8.0

PR #2476, opened 2024-04-04
feat(dht): `DiscoverySession` waits for all branches

teogeb:
When `DiscoverySession` sees a node, which doesn't have any `uncontacted` contacts, it is very likely that there are there parallel operations ongoing. Before this PR the whole session was stopped when any of the parallel branches saw that kind of situation.

Now the session stops only if there are no ongoing requests.

## Open questions

How about the other check at the same line: if we reach the `noProgressCounter` but still have ongoing requests, should we wait for those request?

## Future improvements

- Maybe the could be simplified more: e.g. maybe we could check that the `this.noProgressCounter >= this.config.noProgressLimit` check near the place where we increment the value?
- Could apply this change also to `RingDiscoverySession` (or combine these two classes).


juslesan (approved):
Adding a check for ongoingRequests to complete and the noProgressLimit would probably require more logic to stop new requests from being opened after noProgressLimit is reached. Otherwise there's a chance that the operation does not complete until the entire network is crawled.

PR #2477, opened 2024-04-04
refactor(dht): Simplify `DiscoverySession`

teogeb:
Simplify the state management of `DiscoverySession` by using a `Gate` object instead of internal `EventEmitter`. Also using `AbortController` instead of `stop()` method as there is already an `AbortController` instance in PeerDiscovery which it can just pass to the session.

## Future improvements

- Could apply this change also to `RingDiscoverySession` (or combine these two classes).


packages/dht/src/dht/discovery/DiscoverySession.ts
```diff
@@ -112,17 +105,13 @@ export class DiscoverySession {
...
-    }
 
-    public stop(): void {
-        this.stopped = true
-        this.emitter.emit('discoveryCompleted')
-        this.emitter.removeAllListeners()
+        setImmediate(() => {
+            this.findMoreContacts()
+        })
+        await Promise.race([
```
juslesan: Would something this work here?

```suggestion
        await withTimeout(this.doneGate.waitUntilOpen(), timeout, 'discovery session timed out', this.config.abortSignal)
```

teogeb: Good improvement 👍 Fixed

juslesan (approved):

PR #2478, opened 2024-04-04
feat: Topology visualization tool

teogeb:
Add new developer tool for topology visualization. It reads a network topology definition from a JSON and creates a DOT graph. The JSON may also contain labels for some of the nodes and a color-coded route information.

## Future improvements

Now the tool is in new `developer-tools` directory to monorepo root. But maybe we could implement this as a `cli-tools` command.


juslesan (approved):
I think having the developer-tools package is a good idea.

In the future it would be a good idea to create additional scripts that query the crawler and draw graphs for streams / DHTs

PR #2483, opened 2024-04-08
feat(client): Use StaticJsonProvider

teogeb:
Use `StaticJsonProvider` instead of normal `JsonProvider` so that we don't do unnecessary chainId requests. Before this PR we typically did a chainId request before each read.

## Future improvements

- We could use same provider instance for multiple contracts so that the caching is more efficient

### DIFF
--- a/packages/client/src/Ethereum.ts
+++ b/packages/client/src/Ethereum.ts
@@ -2,7 +2,7 @@
  * Config and utilities for interating with identity & Ethereum chain.
  */
 import { Wallet } from '@ethersproject/wallet'
-import { JsonRpcProvider } from '@ethersproject/providers'
+import { StaticJsonRpcProvider } from '@ethersproject/providers'
 import type { Provider } from '@ethersproject/providers'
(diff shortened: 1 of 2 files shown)


packages/client/src/Ethereum.ts
```diff
@@ -23,7 +23,7 @@ export const getStreamRegistryChainProviders = (config: Pick<StrictStreamrClient
 
 const getRpcProviders = (connectionInfo: ChainConnectionInfo, pollInterval?: number): Provider[] => {
     return connectionInfo.rpcs.map((c: ConnectionInfo) => {
-        const provider = new JsonRpcProvider(c)
+        const provider = new StaticJsonRpcProvider(c)
```
harbu, line 26: Should we add a timeout while we are at it?
```suggestion
        const provider = new StaticJsonRpcProvider({
            ...c,
            timeout: 30 * 1000
        })
```

harbu: In follow-up PR

harbu (approved):

PR #2489, opened 2024-04-09
chore: Enable `no-lonely-if` eslint rule

teogeb:
https://eslint.org/docs/latest/rules/no-lonely-if

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -72,6 +72,7 @@ module.exports = {
         'default-case': ['error'],
         'no-useless-return': 'error',
         'promise/always-return': ['error', { ignoreLastCallback: true }],
+        'no-lonely-if': 'error',
         // TODO: in follow up PRs, select which rules we should enable and fix the code. When all recommended rules
         //  have been enabled, consider enabling the "strict" preset.
         '@typescript-eslint/require-await': 'off',
--- a/packages/broker/src/plugins/storage/dataQueryEndpoint.ts
+++ b/packages/broker/src/plugins/storage/dataQueryEndpoint.ts
@@ -32,10 +32,8 @@ class ResponseTransform extends Transform {
             if (this.format.header !== undefined) {
                 this.push(this.format.header)
             }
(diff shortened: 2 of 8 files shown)


harbu:
Do you think there is any style value here? Can it sometimes be more easy to understand something in this format, eg.

```
if (notLoggedIn) {
   doSomething()
   forwardToLoginPage()
} else {
  if (loggedInViaWallet) {
      doSomethingElse()
  } else if (loggedInViaHttp) {
      doSomethingEvenElse()
  } else {
      doSomethingEvenDifferent()
  }
}
```

teogeb:
> Do you think there is any style value here? Can it sometimes be more easy to understand something in this format, eg.
> 
> ```
> if (notLoggedIn) {
>    doSomething()
>    forwardToLoginPage()
> } else {
>   if (loggedInViaWallet) {
>       doSomethingElse()
>   } else if (loggedInViaHttp) {
>       doSomethingEvenElse()
>   } else {
>       doSomethingEvenDifferent()
>   }
> }
> ```

Yes, sometimes that is more readable. In that case could add `eslint-ignore` to the code. How about the changes I made in this PR, would it make sense to add `eslint-ignore` to some of those?

harbu:
> > Do you think there is any style value here? Can it sometimes be more easy to understand something in this format, eg.
> > ```
> > if (notLoggedIn) {
> >    doSomething()
> >    forwardToLoginPage()
> > } else {
> >   if (loggedInViaWallet) {
> >       doSomethingElse()
> >   } else if (loggedInViaHttp) {
> >       doSomethingEvenElse()
> >   } else {
> >       doSomethingEvenDifferent()
> >   }
> > }
> > ```
> 
> Yes, sometimes that is more readable. In that case could add `eslint-ignore` to the code. How about the changes I made in this PR, would it make sense to add `eslint-ignore` to some of those?

Yeah good point. I think the changes here made sense. Approved.

harbu (approved):

PR #2491, opened 2024-04-09
refactor(dht): Fix ring peer naming

teogeb:
We should avoid using `peer` term, as it is used to refer both `neighbor` and `contact`.

In this PR fixed term `ringPeer` -> `ringContact` as the `neighbor` term refers to Kademlia neighbors, and therefore the ring contacts are always `contacts`, not `neighbors`.


packages/dht/protos/DhtRpc.proto
```diff
@@ -102,11 +103,13 @@ message ClosestPeersResponse {
   string requestId = 2;
 }
 
+// TODO rename to ClosestRingContactsRequest
 message ClosestRingPeersRequest {
```
juslesan, line 107: These message names and the names of their fields could be renamed without it being a breaking breaking change. But could keep them as is for consistency

juslesan (approved):

PR #2496, opened 2024-04-10
feat(client): add configurable timeout for JSON RPC calls

harbu:
## Summary

Add a configuration option for JSON RPC call timeouts. It is now set to 30 seconds by default. It used to be 120 seconds (Ethers.js default).

Also _removed_ ad-hoc solution where if `dev2` was chosen as environment, we applied 30 second timeout.

### Open questions
- Is 30 seconds too short?


packages/client/src/config.schema.json
```diff
@@ -454,6 +454,10 @@
                         }
                     },
                     "default": {}
+                },
+                "jsonRpcTimeout": {
+                    "type": "number",
+                    "default": 30000
```
teogeb, line 460: Is this enough for writing transactions in production?

PR #2498, opened 2024-04-11
refactor(dht): Fix closest peer naming

teogeb:
We should avoid using `peer` term, as it is used to refer both `neighbor` and `contact`.

In this PR fixed the term in use cases where closest peers are handled.

There is one field renaming in proto file: that should be a backwards compatible change.

## Future improvements

- Rename `closestContacts` -> `nearbyContacts`

### DIFF
--- a/packages/dht/src/dht/discovery/PeerDiscovery.ts
+++ b/packages/dht/src/dht/discovery/PeerDiscovery.ts
@@ -183,11 +183,12 @@ export class PeerDiscovery {
         if (!this.recoveryIntervalStarted) {
             this.recoveryIntervalStarted = true
             // TODO use config option or named constant?
-            await scheduleAtInterval(() => this.fetchClosestPeersFromBucket(), 60000, true, this.abortController.signal)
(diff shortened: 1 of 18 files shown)


packages/dht/src/dht/discovery/PeerDiscovery.ts
```diff
@@ -183,11 +183,12 @@ export class PeerDiscovery {
         if (!this.recoveryIntervalStarted) {
             this.recoveryIntervalStarted = true
             // TODO use config option or named constant?
-            await scheduleAtInterval(() => this.fetchClosestPeersFromBucket(), 60000, true, this.abortController.signal)
+            await scheduleAtInterval(() => this.fetchClosestNeighborsFromBucket(), 60000, true, this.abortController.signal)
         }
     }
 
-    private async fetchClosestPeersFromBucket(): Promise<void> {
+    // TODO rename the method (the bucket is not relevant abstraction for this class)
```
juslesan: fetchClosestNeighbors

juslesan (approved):

PR #2499, opened 2024-04-11
feat:(DHT): NET-1292 option to periodically ping least recently seen contacts

juslesan:
## Summary

Ping least recently seen contacts periodically if the option is set


juslesan:
Closing as the solution adds too much traffic

PR #2501, opened 2024-04-16
feat(DHT): Ping neighbors and ring contacts in layer1

juslesan:
## Summary

Added configuration to periodically ping neighbors and ring contacts. This ensures that the peers in the nodes' structures are periodically refreshed. This ensures that ungracefully leaving peers are eventually detected.

## Changes

- Added utility function to ping contacts.
- Added two new configurations to the DhtNode to enable periodic pinging.
- Query random neighbors in the PeerDiscovery's Recovery interval

## Limitations

There needs to be new mechanisms implemented to ensure that nodes that lose connectivity to the internet can rejoin the layer1 DHT. Previously the contacts and nodes in the buckets would remain there if internet connectivity was lost. Nodes will no longer be able to rejoin the streams properly. Publishing this feature before such mechanisms are in place could be detrimental to the system in the long run.


packages/dht/src/dht/DhtNode.ts
```diff
@@ -299,6 +302,25 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
...
         })
         this.bindRpcLocalMethods()
+
+        if (this.config.periodicallyPingNeighbors === true) {
+            await scheduleAtInterval(async () => {
+                const offlineNeighbors = await this.peerManager!.pingAllNeighbors()
+                offlineNeighbors.forEach((offlineNeighbor) => {
+                    logger.trace('Removing offline neighbor', { node: getNodeIdFromPeerDescriptor(offlineNeighbor) })
+                    this.peerManager!.removeContact(getNodeIdFromPeerDescriptor(offlineNeighbor))
+                })
```
harbu: (nitpick)

I would consider moving this logic into a method into `peerManager`. E.g. `this.peerManager!.pruneOfflineNeighbors()`.

Essentially we are taking the return value of a method of `peerManager` only to remove contacts from `peerManager`. Why not just do this entirely in `peerManager`?

packages/dht/src/dht/DhtNode.ts
```diff
@@ -299,6 +302,25 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
...
+            }, 60000, true, this.abortController.signal)
+        }
+        if (this.config.periodicallyPingRingContacts === true) {
+            await scheduleAtInterval(async () => {
+                const offlineContacts = await this.peerManager!.pingAllRingContacts()
+                offlineContacts.forEach((contact) => {
+                    logger.trace('Removing offline ring contact', { node: getNodeIdFromPeerDescriptor(contact) })
+                    this.peerManager!.removeContact(getNodeIdFromPeerDescriptor(contact))
+                })
```
harbu: Same suggestion here as above.

packages/dht/src/dht/discovery/PeerDiscovery.ts
```diff
@@ -197,14 +204,21 @@ export class PeerDiscovery {
...
                 const contacts = await node.getClosestPeers(localNodeId)
                 for (const contact of contacts) {
                     this.config.peerManager.addContact(contact)
                 }
+            }),
+            ...randomNodes.map(async (node: DhtNodeRpcRemote) => {
+                const contacts = await node.getClosestPeers(createRandomDhtAddress())
+                for (const contact of contacts) {
+                    this.config.peerManager.addContact(contact)
+                }
```
harbu, line 219: This change is not related to pinging, right? Seems to be fetching more contacts in recovery?

harbu (approved):

packages/dht/src/dht/discovery/PeerDiscovery.ts
juslesan: Yeah forgot to include this to the descritption. The point is that in the long run nodes will also find nodes that are further away in the ID space. Currently there are issues where some nodes only know nearby nodes which tends to make them hard to contact.

PR #2502, opened 2024-04-16
feat(network): Manual rejoin trigger after lost connectivity in layer1

juslesan:
## Summary

Manually trigger a rejoin to the layer 1 DHTs if connectivity is lost. This is a required feature after #2501 as the nodes in layer1 will lose contacts when disconnected from the internet. This is also an improvement for rejoining streams in general after lost connectivity.

This is achieved by reating to a new event the DhtNode that is emitted if the number of neighbors drops to 0 and there are no configured entrypoints. When a layer1 DhtNode emits the event,  the network node will attempt to query entry points to a stream from the Layer0 DHT. It will then call `joinDht` on the layer1 node with the queried entry points. This operation is ran in an interval until the node finds at least one neighbor.


ptesavol (approved):
lgtm

PR #2507, opened 2024-04-22
refactor(dht): Rename `PeerManager#closestContacts`

teogeb:
Renamed `PeerManager#closestContacts` -> `nearbyContacts`. This way we can separate the concepts of:
- nearby contacts (all currently known nodes having short Kademlia distance to this node)
- closest contacts/neighbors/etc. (a subset of some contact list, e.g. 5 nearest neighbors)

Also renamed these to use term neighbor instead of bucket:
- `addClosestContactToBucket`
- `getNearbyActiveContactNotInBucket`
- test description `"Adds Nodes from layer1 KBucket to nearbyNodeView if its size is below nodeViewSize"`

Also small refactoring in `ContentDeliveryLayerNode#start`.

## Open questions

Why `PeerManager#getClosestRingContactsTo` returns also nearby contacts?


juslesan:
> Why PeerManager#getClosestRingContactsTo returns also nearby contacts?

Apparently it adds all nearbyContacts to the sorted list before returning. This should not be necessary since the new contacts should be added to the structures at the same time. So this logic could be removed.


Whenever there is a reference to kbucket / bucket rename to neighbors

juslesan (approved):

PR #2510, opened 2024-04-22
refactor(DHT): refactor ping interval logic

juslesan:
## Summary

Refactored ping related logic by removing duplicate logic in pruneOfflineNeighbors and pruneOfflineRingContacts.

### DIFF
--- a/packages/dht/src/dht/DhtNode.ts
+++ b/packages/dht/src/dht/DhtNode.ts
@@ -121,6 +121,8 @@ type StrictDhtNodeOptions = MarkRequired<DhtNodeOptions,
 
 const logger = new Logger(module)
 
+const PERIODICAL_PING_INTERVAL = 60 * 1000
+
 export type Events = TransportEvents & DhtNodeEvents
 
 export class DhtNode extends EventEmitter<Events> implements ITransport {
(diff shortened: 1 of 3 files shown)


packages/dht/src/dht/DhtNode.ts
```diff
@@ -305,10 +307,19 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
         this.bindRpcLocalMethods()
 
         if (this.config.periodicallyPingNeighbors === true) {
```
teogeb, line 310: Could avoid duplicating the scheduling logic, e.g. like this:
```ts
        const pruneTargets = []
        if (this.config.periodicallyPingNeighbors === true) {
            pruneTargets.push(() => this.peerManager!.getNeighbors().map((node) => this.createDhtNodeRpcRemote(node)))
        }
        if (this.config.periodicallyPingRingContacts === true) {
            pruneTargets.push(() => this.peerManager!.getRingContacts().getAllContacts())
        }
        for (const pruneTarget of pruneTargets) {
            await scheduleAtInterval(
                async () => {
                    const nodes = pruneTarget()
                    await this.peerManager!.pruneOfflineNodes(nodes)
                }, PERIODICAL_PING_INTERVAL, false, this.abortController.signal
            )
        }
```

teogeb (changes requested):

teogeb (approved):

PR #2515, opened 2024-04-23
Allow partition id as a parameter for publish command. 

yaruno:
Should not allow both partition key and partition id at the same time.

## Summary

Added partition id as possible parameter for publish command.

## Changes

We'll check if partition key and id are given as parameters and give suitable error msg. It's one or the other but not both as parameter options.

Provide a bullet list or description of known omissions or limitations of the
solution and/or any ideas for future improvement. Leave section ouf if
not applicable.


packages/cli-tools/bin/streamr-stream-publish.ts
```diff
@@ -63,4 +73,5 @@ createClientCommand(async (client: StreamrClient, streamId: string, options: Opt
     .arguments('<streamId>')
     .description('publish to a stream by reading JSON messages from stdin line-by-line')
     .option('-k, --partition-key-field <string>', 'field name in each message to use for assigning the message to a stream partition')
+    .option('-p, --partition [partition]', 'partition')
```
harbu, line 76: Move the use of `createFnParseInt`here. See `streamr-stream-subscribe.ts` for example.

harbu (changes requested):
still want to get this merged?

check eslint; see comment about `createFnParseInt`

PR #2517, opened 2024-04-23
fix(network, DHT): stricter typing for event listeners in interfaces

juslesan:
## Summary

Stricter listener typing for events in the `ITransport` and `Layer1Node` interfaces.

It is no longer possible to set a listener with an inproper listener function for example:
`layer1Node.on('manualRejoinRequired', (peerDescriptor: PeerDescriptor) => {})`
will now be caught by the compiler.

https://linear.app/streamr/issue/NET-1449/fix-event-typing-in-discoverylayernode-and-itransport

## Changes 

- DhtNode no longer export an Events type. It now only exports a DhtNodeEvents type that includes DhtNode and ITransport specific events.


packages/utils/src/EmitterOf.ts
```diff
@@ -0,0 +1,35 @@
+// Utility type EmitterOf<T>. By deriving an interface from this type, 
+// you declare that the interface has the standard EventEmitter3  
+// listener setter functions 'on', 'off', 'once' for each event type
+// defined in T.
+
+// Convert EventEmitter3 event types to corresponding listener
+// setter function types (eg. types of 'on', 'off', 'once').
+// For example, 
+// message: (message: Message) => void
+// -> message: (event: 'message', listener: (message: Message) => void) => void
```
teogeb: Could use different names for function and parameter names here, so that the example is easier to understand?

packages/utils/src/EmitterOf.ts
```diff
@@ -0,0 +1,35 @@
...
+
+type Intersection<T> = {
+    [K in keyof T]: (x: T[K]) => void
+  }[keyof T] extends
+    (x: infer I) => void ? I : never
+
+type OverloadedListenerSetter<T> = Intersection<ListenerSetterTypes<T>> 
+
+// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
+export type EmitterOf<T> = {
```
teogeb: Rename e.g, to `EventEmitterType`?

teogeb: This give bit strange error messages:

e.g.
```
const emitter: Layer1Node = {} as any
emitter.on('manualRejoinRequired', (foo: any) => {})
```

```
No overload matches this call.
  The last overload gave the following error.
    Argument of type '"manualRejoinRequired"' is not assignable to parameter of type '"ringContactRemoved"'.ts(2769)
```

Maybe good to add a bullet point about that into the "Future improvements" section

teogeb (commented):
> Why do the `on`, `off` and `once` types need to be defined multiple times per function type.

This is no longer an issue, could remove it from the "Future improvements" list

packages/utils/src/EmitterOf.ts
juslesan: This seems weird but I believe this is just overloading errors are reported in this case, as the assigned listener has any type in the first parameter.

ptesavol (approved):
I think this can be merged to main.

juslesan:
Eslint broken here for some reason

PR #2519, opened 2024-04-23
fix(DHT): fix hanging connection issue with connectivity requests

juslesan:
## Summary

When ws connections are opened using connection requests, the server checks if the requested connection's presumed peer descriptor is correct. This fixes a problem with hanging connections detected in nodes that do run a WS server.

## Future improvements

This problem went undetected due to the case not being tested. The connection layer needs thorough refactoring to make testing possible.

### DIFF
--- a/packages/dht/src/connection/ConnectionManager.ts
+++ b/packages/dht/src/connection/ConnectionManager.ts
@@ -159,14 +159,7 @@ export class ConnectionManager extends EventEmitter<TransportEvents> implements 
             allowToContainReferenceId: false
         })
         this.connections.forEach((connection, key) => {
-            // TODO: Investigate why multiple invalid WS client connections to the same
-            // server with a different nodeId can remain in the this.connections map.
-            // Seems to only happen if the ConnectionManager acting as client is not running a WS server itself.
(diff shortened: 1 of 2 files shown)


ptesavol:
Could you clarify how this fixes a problem with hanging connections detected in nodes that do run a WS server? What was the problem and how does this PR fix the problem?

juslesan:
> Could you clarify how this fixes a problem with hanging connections detected in nodes that do run a WS server? What was the problem and how does this PR fix the problem? 

@ptesavol 

It should not be possible to open connections to a node without knowing its NodeID first. This fixes a whole array of problems related to nodes being restarted. The issues manifested by nodes having offline nodeIds as neighbors which leads to lots of bugs in the long run. Additionally offline nodes may appear in the connections structure indefinetely. This causes critical problems in the protocol.

The remaining leaking connection are a bit more sinister. If a WS server node would request a serverless node for a connection and restart after routing the request, the client could be able to open a connection to a server and do a handshake. However, the nodeId of the server node would have changed in between. This behaviour caused some connections to hang forever as the client node would keep the connection behind an offline NodeID

ptesavol (approved):
This is OK to be merged.

PR #2523, opened 2024-04-25
ETH-768: try to avoid sending transactions that could revert

jtakalai:
## Summary

reverting tx just waste the nodes' MATIC

### DIFF
--- a/packages/broker/src/plugins/operator/ContractFacade.ts
+++ b/packages/broker/src/plugins/operator/ContractFacade.ts
@@ -463,7 +463,14 @@ export class ContractFacade {
 
     async voteOnFlag(sponsorship: string, targetOperator: string, kick: boolean): Promise<void> {
         const voteData = kick ? VOTE_KICK : VOTE_NO_KICK
-        // typical gas cost 99336, but this has shown insufficient sometimes
+
+        // This will throw if voteOnFlag would revert e.g. due to flag already being closed.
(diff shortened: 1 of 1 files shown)


packages/broker/src/plugins/operator/ContractFacade.ts
```diff
@@ -463,7 +463,14 @@ export class ContractFacade {
 
     async voteOnFlag(sponsorship: string, targetOperator: string, kick: boolean): Promise<void> {
         const voteData = kick ? VOTE_KICK : VOTE_NO_KICK
-        // typical gas cost 99336, but this has shown insufficient sometimes
+
+        // This will throw if voteOnFlag would revert e.g. due to flag already being closed.
+        // Since we bypass gas estimation in the actual transaction (by setting gasLimit),
+        //   we should do this step manually to avoid wasting the nodes' MATIC on reverting transactions.
+        await this.operatorContract.estimateGas.voteOnFlag(sponsorship, targetOperator, voteData)
```
teogeb, line 470: Why exactly this line would throw? The ethers docs says that this should just return a gas price for the transaction, but doesn't mention any rejection situation: https://docs.ethers.org/v5/api/contract/contract/#contract-estimateGas

jtakalai:
This seems not to be practical issue anymore (probably this was due to bad RPCs that were lagging a lot behind)

PR #2532, opened 2024-04-29
refactor(trackerless-network): EntryPointDiscovery refactor

juslesan:
Extracted `StreamPartSplitAvoidance` class from `EntryPointDiscovery`.

Also:
- Simplified stream joining logic. Now there are two separate branches for joining: one where there are known entry point and one where we need discover the entry points first.
- Simplified entry point discovery logic. The `discoverEntryPointsFromDht` gets `excludeSet` instead of `knownEntryPointCount`, and it returns list a simple list instead of `FindEntryPointsResult`.

## Future improvements

- Further simplify internal logic of the `EntryPointDiscovery` and `StreamPartSplitAvoidance` classes
- Maybe `excludeSet` of `discoverEntryPointsFromDht` is not needed: the caller of the method could just filter unwanted entry points from the result set


packages/trackerless-network/src/logic/EntryPointDiscovery.ts
```diff
@@ -20,95 +19,48 @@ const parseEntryPointData = (dataEntries: DataEntry[]): PeerDescriptor[] => {
...
             return filtered
+        // If all found entry points are filtered return the unfiltered list 
         } else {
             return discoveredEntryPoints
         }
     }
 
+    private async discoverEntryPoints(): Promise<PeerDescriptor[]> {
+        const dataKey = streamPartIdToDataKey(this.config.streamPartId)
+        return this.queryEntrypoints(dataKey)
```
teogeb: Could inline `queryEntrypoints` into this method. (Is the logging correct at line `65`?)

packages/trackerless-network/src/logic/EntryPointDiscovery.ts
```diff
@@ -20,95 +19,48 @@ const parseEntryPointData = (dataEntries: DataEntry[]): PeerDescriptor[] => {
...
-    private async discoverEntryPoints(): Promise<PeerDescriptor[]> {
-        const dataKey = streamPartIdToDataKey(this.config.streamPartId)
-        const discoveredEntryPoints = await this.queryEntrypoints(dataKey)
-        const filtered = discoveredEntryPoints.filter((node) => 
-            !this.networkSplitAvoidedNodes.has(getNodeIdFromPeerDescriptor(node)))
-        // If all discovered entry points have previously been detected as offline, try again
+        const filtered = discoveredEntryPoints.filter((node) => !excludeSet.has(getNodeIdFromPeerDescriptor(node)))
         if (filtered.length > 0) {
             return filtered
+        // If all found entry points are filtered return the unfiltered list
```
teogeb, line 53: Better to avoid this kind of surprise. If this logic is needed in caller, let's move it there (i.e. implement one of the "Future improvements")

packages/trackerless-network/src/logic/EntryPointDiscovery.ts
```diff
@@ -119,20 +71,13 @@ export class EntryPointDiscovery {
...
-            this.isLocalNodeStoredAsEntryPoint = true
-            await this.storeSelfAsEntryPoint()
-            await this.keepSelfAsEntryPoint()
-        }
-        if (possibleNetworkSplitDetected) {
-            // TODO should we catch possible promise rejection?
-            setImmediate(() => this.avoidNetworkSplit())
-        }
+        this.isLocalNodeStoredAsEntryPoint = true
+        await this.storeSelfAsEntryPoint()
```
teogeb:
```suggestion
        await this.storeLocalNodeAsEntryPoint()
```

packages/trackerless-network/src/logic/EntryPointDiscovery.ts
```diff
@@ -119,20 +71,13 @@ export class EntryPointDiscovery {
         }
     }
 
-    async storeSelfAsEntryPointIfNecessary(currentEntrypointCount: number): Promise<void> {
+    async storeAndKeepSelfAsEntryPoint(): Promise<void> {
```
teogeb:
```suggestion
    async storeAndKeepLocalNodeAsEntryPoint(): Promise<void> {
```

packages/trackerless-network/src/logic/EntryPointDiscovery.ts
```diff
@@ -119,20 +71,13 @@ export class EntryPointDiscovery {
...
-        if ((currentEntrypointCount < ENTRYPOINT_STORE_LIMIT) || possibleNetworkSplitDetected) {
-            this.isLocalNodeStoredAsEntryPoint = true
-            await this.storeSelfAsEntryPoint()
-            await this.keepSelfAsEntryPoint()
-        }
-        if (possibleNetworkSplitDetected) {
-            // TODO should we catch possible promise rejection?
-            setImmediate(() => this.avoidNetworkSplit())
-        }
+        this.isLocalNodeStoredAsEntryPoint = true
```
teogeb, line 74: Could we move the setting of this field into `storeSelfAsEntryPoint`?

packages/trackerless-network/test/unit/StreamPartSplitAvoidance.test.ts
```diff
@@ -0,0 +1,31 @@
...
+    })
+
+    afterEach(() => {
+        layer1Node.stop()
+        streamPartSplitAvoidance.destroy()
+    })
+
+    it('runs avoidance until number of neighbors is above SPLIT_AVOIDANCE_LIMIT', async () => {
+        await streamPartSplitAvoidance.avoidNetworkSplit()
+        expect(layer1Node.getNeighborCount()).toBeGreaterThan(4)
```
teogeb: Use constant here

packages/trackerless-network/src/logic/StreamPartSplitAvoidance.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export const MIN_NEIGHBOR_COUNT = 4
+
+export interface StreamPartSplitAvoidanceConfig {
+    layer1Node: Layer1Node
+    discoverEntryPoints: (excludedNodes?: Set<DhtAddress>) => Promise<PeerDescriptor[]>
+    exponentialRunOfBaseDelay?: number
+}
+
+export class StreamPartSplitAvoidance {
```
teogeb: Would it make sense to have a word `Network` in the name? Or maybe the functionality is generic, and this could be `NetworkSplitAvoidance`?

packages/trackerless-network/src/logic/EntryPointDiscovery.ts
juslesan: The point is to keep it here as the node will be stored as entry point until it leaves the stream. `keepSelfAsEntryPoint` would call `storeSelfAsEntryPoint `, reassigning the value which would be pointless.

packages/trackerless-network/src/logic/StreamPartSplitAvoidance.ts
teogeb: Rename also fields/variables accordingly

juslesan: which variables would that be?

teogeb: All which have phrase `splitAvoidance` -> `networkSplitAvoidance`. E.g. `streamPartSplitAvoidance` -> `streamPartNetworkSplitAvoidance`.

teogeb (approved):

PR #2535, opened 2024-05-02
fix(DHT): Require connectionsView if transport is given in config

juslesan:
## Summary

Require connectionsView if transport is given in config in the DhtNode

### DIFF
--- a/packages/dht/src/dht/DhtNode.ts
+++ b/packages/dht/src/dht/DhtNode.ts
@@ -610,7 +610,7 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
  
     public async waitForNetworkConnectivity(): Promise<void> {
         await waitForCondition(() => this.connectionsView!.getConnectionCount() > 0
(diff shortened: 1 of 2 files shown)


packages/dht/src/dht/DhtNode.ts
```diff
@@ -604,15 +607,10 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
...
     public async waitForNetworkConnectivity(): Promise<void> {
-        await waitForCondition(() => {
-            if (!this.peerManager) {
-                return false
-            } else {
-                return (this.connectionsView!.getConnectionCount() > 0)
-            }
-        }, this.config.networkConnectivityTimeout, 100, this.abortController.signal)
+        await waitForCondition(() => this.connectionsView!.getConnectionCount() > 0
+            , this.config.networkConnectivityTimeout, 100, this.abortController.signal)
```
teogeb: Fix indentation

teogeb (approved):

PR #2543, opened 2024-05-03
refactor(DHT): StoreRpcLocal unit tests

juslesan:
## Summary

Added unit tests for the StoreRpcLocal class

## Changes

- Moved `replicateDataToNeighbors` from StoreManager to StoreRpcLocal

### DIFF
--- a/packages/dht/src/dht/store/StoreRpcLocal.ts
+++ b/packages/dht/src/dht/store/StoreRpcLocal.ts
@@ -1,5 +1,5 @@
 import { ServerCallContext } from '@protobuf-ts/runtime-rpc'
-import { Logger } from '@streamr/utils'
+import { Logger, executeSafePromise } from '@streamr/utils'
 import { Empty } from '../../proto/google/protobuf/empty'
 import { Timestamp } from '../../proto/google/protobuf/timestamp'
 import {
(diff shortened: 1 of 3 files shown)


packages/dht/src/dht/store/StoreRpcLocal.ts
```diff
@@ -29,10 +30,14 @@ export class StoreRpcLocal implements IStoreRpc {
         this.config = config
     }
 
+    private isLocalNodeStorer(dataKey: DhtAddress): boolean {
+        return this.config.getStorers(dataKey).some((p) => areEqualPeerDescriptors(p, this.config.localPeerDescriptor))    
+    }
+
```
teogeb: Could move this to to the end of the file as this is a utility function used by other methods

packages/dht/src/dht/store/StoreRpcLocal.ts
```diff
@@ -54,13 +59,30 @@ export class StoreRpcLocal implements IStoreRpc {
...
             this.config.localDataStore.setAllEntriesAsStale(key)
         }
         logger.trace('server-side replicateData() at end')
         return {}
     }
+
+    private replicateDataToNeighbors(requestor: PeerDescriptor, dataEntry: DataEntry): void {
+        const dataKey = getDhtAddressFromRaw(dataEntry.key)
+        const storers = this.config.getStorers(dataKey)
+        const selfIsPrimaryStorer = areEqualPeerDescriptors(storers[0], this.config.localPeerDescriptor)
```
teogeb: If we nowadays use term `localNode` instead of `self`?

```suggestion
        const isLocalNodePrimaryStorer = areEqualPeerDescriptors(storers[0], this.config.localPeerDescriptor)
```

packages/dht/test/unit/StoreRpcLocal.test.ts
```diff
@@ -0,0 +1,167 @@
...
+
+        it('replicateData', async () => {
+            const request = {
+                entry: DATA_ENTRY
+            }
+            await storeRpcLocal.replicateData(request as any, { incomingSourceDescriptor: createMockPeerDescriptor() } as any)
+            expect(setAllEntriesAsStale).not.toHaveBeenCalled()
+            // Wait for setImmediate
+            await wait(50)
+            expect(replicateDataToContact).toHaveBeenCalledTimes(4)
```
teogeb: Could calculate 4 from `NODE_COUNT` (if we don't want explicitly assert the payload of the `replicateDataToContact` calls)

```suggestion
            // All other nodes except local node
            expect(replicateDataToContact).toHaveBeenCalledTimes(NODE_COUNT - 1)
```

teogeb (approved):

PR #2548, opened 2024-05-07
feat(DHT): Queried stream entrypoint filtering

juslesan:
## Summary

Removed `discoverEntryPointsFromDht`, made `discoverEntryPoints` a public function.
The excluding logic of `discoverEntryPointsFromDht` is now a concern of the caller.
No longer reset the filtered entry points to the original queried set. This is no longer necessary as nodes remove themselves as entry points when leaving streams gracefully.


teogeb (approved):
> No longer reset the filtered entry points to the original queried set. This is no longer necessary as nodes remove themselves as entry points when leaving streams gracefully.

Looks good 👍 How the functionality changes in practice? Should we change the title prefix from `refactor` to `feat`?

juslesan:
> Looks good 👍 How the functionality changes in practice? Should we change the title prefix from refactor to feat?

@teogeb Previously the filtering reset handled a case where nodes had trouble joining layer1 DHTs when initially starting. This is no longer a common issue but it does remove the fallback if it resurfaces.

PR #2549, opened 2024-05-10
refactor(DHT): remove input buffer from ManagedConnection

juslesan:
## Summary

Removed inputBuffer related logic from the ManagedConnection class. There should be 0 cases where the managedConnection listener is not set before a handShake has been successfully completed. The size of the buffer is always 0.

## Future improvements

- Extract outputBuffer from the ManagedConnection.
- Remove handshaker related logic from the ManagedConnection. Instead refactor the Handshaker to wrap an IConnection until it has been successfully handshaked. Once the connection is ready emit an event with the IConnection.


ptesavol (approved):
LGTM

PR #2551, opened 2024-05-15
refactor(DHT): Extract output buffer from ManagedConnection

juslesan:
## Summary

Extracted outputBuffer from ManagedConnection, now handled in ConnectionManager

## Changes

- Added Endpoint type in ConnectionManager and refactored the connections object to be endpoints. Endpoint includes the OutputBuffer and the ManagedConnection. 
- Removed closing property from MAnagedConnection.
- ManagedConnection send will now throw if it is called before a successful handshake

## TODO

- Should Endpoint encapsulate logic related to buffer handling?
- Add unit tests


ptesavol (approved):
Approved, but please add a todo comment about encapsulating the buffer-handling logic (and everything else that the user does not need to know about) inside the Endpoint.

PR #1361, opened 2023-05-11
Docs fix

yaruno:
Fix links, fix positioning of API documentation, adjusted naming of categories.

PR #2552, opened 2024-05-15
deps(docs): [DREL-247] Update `typedoc` and `typescript`

teogeb:
Updated TypeDoc, TypeScript and two related plugins. 

With these changes Docusaurus seems to generate API documentation, but the output is not optimal. We should tweak it to look more like the output of the previous version (see the output generated in `main` branch, or in https://docs.streamr.network/usage/sdk/api/)

## Changes / TODO

- The `out` directory definition has been updated. Maybe the `docs/` prefix is needed now for some reason?
- Added some exclude definitions (maybe not optimal set of files, should maybe tweak the list)
- Added "readme: none" definition as we don't seem to need that (see `refine-docs/index.js#110`)
- Removed obsolete `refine-api-reference.ts` file (the functionality was moved to `refine-docs/index.js` in https://github.com/streamr-dev/network/pull/1361)
- Removed sidebar config as it doesn't seem to work (doesn't remove version number from the title). Note that there is also `sidebars.js` config file, maybe it conflicts with that? (not clear if either of these configs is obsolete).
- The `refine-docs` plugin is temporarily disabled. We should update some parts of the implementation (`refine-docs/index.js`) so that it is compatible with the current version.


teogeb:
Replaced by PR https://github.com/streamr-dev/network/pull/2917

PR #2558, opened 2024-05-17
test(broker): Increase timeout for `MaintainTopologyService` test

teogeb:
It is not a timeout issues, as it works ok when we only run only that test... Need to investigate.


teogeb:
Solved in https://github.com/streamr-dev/network/pull/2560.

PR #2559, opened 2024-05-20
refactor(DHT): Extract handshaker from managed connection

juslesan:
## Summary

Extracted handshaker from the ManagedConnection class. These made the incoming and outgoing connection setting on the constructor obsolete.

## Changes

- Removed ManagedWebrtcConnection
- Added new functions `createOutgoingHandshaker` and `createIncomingHandshaker`.
- Appropriate refactors for Connector classes to start using new handshaker factories.
- New parameter `doNotRequestConnection` for WebrtcConnector#connect
- Refactored WebrtcConnectorLocal to use WebRtcConnector connect to always create connections.
- Removed connecting timeout from NodeWebrtcConnection. ManagedConnection has timeout.

## Future improvements
- Unit tests
- Use createIncomingHandshaker in WebsocketConnector `attachHandshaker`
- Make separate classes for incoming and outgoing handshakers
- remove onHandshakeCompleted from ManagedConnection. Use event from Connectors instead.
- Explore complete removal of the ManagedConnection class
- Event based connection creation


ptesavol (approved):
Approved. This is going to the right direction. I think the goal should be that Connectors should emit IConnections and not know anything about ManagedConnection (the whole class could be removed). Also the the connectors should not know about handshaking, but instead the handshaking should be initiated in the ConnectionManager

PR #2562, opened 2024-05-23
chore(client) [NET-1287]: use `FallbackProvider`

harbu:
## Summary

Use `FallbackProvider` to provide redundancy when a RPC is down. 

Also use custom event listener implementation (the `ChainEventPoller` class) as the `FallbackProvider` would poll too many RPCs.

## Refactoring

- Renamed `RpcProviderFactory` to `RpcProviderSource`
- Moved `contracts.ts`

## Future improvements

- Maybe the custom event listener implementation is no longer needed when `FallbackProvider` implements `eventQuorum`  and `eventWorkers` functionality. Both of these options are already available, but there is a code comment which states that the functionality is "currently unimplemented": https://github.com/ethers-io/ethers.js/blob/92761872198cf6c9334570da3d110bca2bafa641/src.ts/providers/provider-fallback.ts#L171
- What would be a good value for `pollInterval`. Previously we didn't define an explicit value in the client, and therefore 4s interval was used? https://github.com/ethers-io/ethers.js/blob/92761872198cf6c9334570da3d110bca2bafa641/src.ts/providers/abstract-provider.ts#L428


packages/client/src/RpcProviderFactory.ts
```diff
@@ -1,48 +1,66 @@
-import { ConfigInjectionToken, StrictStreamrClientConfig } from './Config'
 import type { Provider } from 'ethers'
+import { FallbackProvider, FetchRequest, JsonRpcProvider } from 'ethers'
+import { Lifecycle, inject, scoped } from 'tsyringe'
+import { ConfigInjectionToken, StrictStreamrClientConfig } from './Config'
 import { LoggingJsonRpcProvider } from './utils/LoggingJsonRpcProvider'
-import { inject, Lifecycle, scoped } from 'tsyringe'
-import { FetchRequest } from 'ethers'
+
+export const QUORUM = 2
```
harbu: Make configurable

packages/client/src/RpcProviderFactory.ts
```diff
@@ -1,48 +1,66 @@
...
 
 function isDevChain(config: Pick<StrictStreamrClientConfig, 'contracts'>): boolean {
     return config.contracts.streamRegistryChainRPCs?.name === 'dev2'
 }
 
+// TODO: do we even need this class anymore?
```
harbu: Let's keep this class

packages/client/src/RpcProviderFactory.ts
```diff
@@ -1,48 +1,66 @@
...
 @scoped(Lifecycle.ContainerScoped)
 export class RpcProviderFactory {
```
harbu: A better name
```suggestion
export class RpcProviderSource {
```

packages/client/src/RpcProviderFactory.ts
```diff
@@ -1,48 +1,66 @@
...
+    getEventProviders(): Provider[] {
+        return this.config.contracts.streamRegistryChainRPCs.rpcs.map((c) => {
+            const f = new FetchRequest(c.url)
+            f.retryFunc = async () => false  // TODO use this line also for providers in the fallbackprovider?
+            return new JsonRpcProvider(f, {
+                chainId: this.config.contracts.streamRegistryChainRPCs.chainId,
+                name: this.config.contracts.streamRegistryChainRPCs.name
+            }, {
+                staticNetwork: true
+            })
```
harbu: We should aim to use the same providers here that are provided to the `FallbackProvider` above. This means that they need to be cached. Make sure to keep `f.retryFunc = async () => false` around. (Note this will probably break the 429 test which should be a quick fix, we _want_ the 429 not to retry)

packages/client/src/contracts/ChainEventPoller.ts
```diff
@@ -0,0 +1,86 @@
+import { Logger, Multimap, randomString, wait, withTimeout, scheduleAtInterval } from '@streamr/utils'
+import { Contract, EventLog, Provider } from 'ethers'
+import { sample } from 'lodash'
+
+type EventName = string
+type Listener = (...args: any[]) => void
+
+const BLOCK_NUMBER_QUERY_DELAY = 1000
+export const POLL_INTERVAL = 1000  // TODO 5000? create a config option?
+const TIMEOUT = 10 * 1000 // TODO what would be a good value? create a config option?
```
harbu: Use existing config values

`TIMEOUT` => `_timeouts.jsonRpcTimeout`
`BLOCK_NUMBER_QUERY_DELAY` => create a new one config or just keep hard coded?
`POLL_INTERVAL` => `contracts.pollInterval`

packages/client/src/RpcProviderFactory.ts
```diff
@@ -1,48 +1,66 @@
...
+            const providers = this.config.contracts.streamRegistryChainRPCs.rpcs.map((c) => {
                 const fetchRequest = new FetchRequest(c.url)
                 fetchRequest.timeout = timeout
-                const provider = new LoggingJsonRpcProvider(fetchRequest, {
+                return new LoggingJsonRpcProvider(fetchRequest, {
                     chainId: this.config.contracts.streamRegistryChainRPCs.chainId,
                     name: this.config.contracts.streamRegistryChainRPCs.name
                 }, {
                     staticNetwork: true,
+                    pollingInterval: pollInterval,
```
harbu: Can be removed `pollingInterval: pollInterval`

packages/client/src/RpcProviderFactory.ts
```diff
@@ -1,48 +1,66 @@
...
-                    provider.pollingInterval = pollInterval
-                }
-                return provider
+            })
+            this.provider = new FallbackProvider(providers, {
+                chainId: this.config.contracts.streamRegistryChainRPCs.chainId,
+                name: this.config.contracts.streamRegistryChainRPCs.name
+            }, {
+                quorum: Math.min(QUORUM, this.config.contracts.streamRegistryChainRPCs?.rpcs.length),
+                pollingInterval: pollInterval,
```
harbu: Can be removed `pollingInterval: pollInterval`

packages/client/src/contracts/ChainEventPoller.ts
```diff
@@ -0,0 +1,86 @@
...
+        if (started && this.listeners.isEmpty()) {
+            this.abortController!.abort()
+        }
+    }
+
+    private start(): void {
+        const abortController = new AbortController()
+        this.abortController = abortController
+        setImmediate(async () => {
+            const logger = new Logger(module, { traceId: randomString(6) })
```
harbu:
```suggestion
            const logger = new Logger(module, { sessionId: randomString(6) })
```

packages/client/src/contracts/ChainEventPoller.ts
```diff
@@ -0,0 +1,86 @@
...
+                    )) as EventLog[]
+                    logger.info('Polled', { fromBlock, events: events.length })  // TODO debug level
+                    if ((events.length > 0) && (!abortController.signal.aborted)) {
+                        for (const event of events) {
+                            const listeners = this.listeners.get(event.fragment.name)
+                            for (const listener of listeners) {
+                                listener(...event.args, event.blockNumber)
+                            }
+                        }
+                        fromBlock = events[0].blockNumber + 1
```
harbu: We should take max of event blockNumbers here

packages/client/src/contracts/StorageNodeRegistry.ts
```diff
@@ -22,7 +21,7 @@ export interface StorageNodeMetadata {
 export class StorageNodeRegistry {
 
     private nodeRegistryContract?: NodeRegistryContract
-    private readonly nodeRegistryContractsReadonly: NodeRegistryContract[]
+    private readonly nodeRegistryContractsReadonly: NodeRegistryContract
```
harbu: Rename (to singular)
```suggestion
    private readonly nodeRegistryContractReadonly: NodeRegistryContract
```

packages/client/src/contracts/StreamRegistry.ts
```diff
@@ -71,7 +71,7 @@ const CACHE_KEY_SEPARATOR = '|'
 export class StreamRegistry {
 
     private streamRegistryContract?: ObservableContract<StreamRegistryContract>
-    private streamRegistryContractsReadonly: ObservableContract<StreamRegistryContract>[]
+    private streamRegistryContractsReadonly: ObservableContract<StreamRegistryContract>
```
harbu: Rename (to singular)
```suggestion
    private streamRegistryContractReadonly: ObservableContract<StreamRegistryContract>
```

packages/client/src/contracts/StreamRegistry.ts
```diff
@@ -107,23 +107,25 @@ export class StreamRegistry {
...
-            )
-        })
+        this.streamRegistryContractsReadonly = this.contractFactory.createReadContract<StreamRegistryContract>(
+            toEthereumAddress(this.config.contracts.streamRegistryChainAddress),
+            StreamRegistryArtifact,
+            this.rpcProviderFactory.getProvider(),
+            'streamRegistry'
+        )
+        const chainEventPoller = new ChainEventPoller(this.rpcProviderFactory.getEventProviders().map((p) => {
+            // TODO would it make sense to use createDecoratedContract to get logging? or would it do that
```
harbu: Let's keep (non-decorated) plain contracts

packages/client/src/contracts/StreamStorageRegistry.ts
```diff
@@ -35,7 +37,7 @@ interface NodeQueryResult {
 export class StreamStorageRegistry {
 
     private streamStorageRegistryContract?: StreamStorageRegistryContract
-    private readonly streamStorageRegistryContractsReadonly: StreamStorageRegistryContract[]
+    private readonly streamStorageRegistryContractsReadonly: StreamStorageRegistryContract
```
harbu: Rename (to singular)
```suggestion
    private readonly streamStorageRegistryContractReadonly: StreamStorageRegistryContract
```

packages/client/src/ethereumUtils.ts
```diff
@@ -26,7 +26,7 @@ export const getEthersOverrides = async (
     const chainConfig = config.contracts.ethereumNetwork
     const overrides = chainConfig.overrides ?? {}
     if ((chainConfig.highGasPriceStrategy) && (chainConfig.overrides?.gasPrice === undefined)) {
-        const primaryProvider = rpcProviderFactory.getPrimaryProvider()
+        const primaryProvider = rpcProviderFactory.getProvider()
```
harbu: Rename
```suggestion
        const provider = rpcProviderFactory.getProvider()
```

packages/client/test/integration/json-rpc-provider.test.ts
```diff
@@ -0,0 +1,153 @@
...
+            const requests = getNewRequests(now).filter((r) => r.method === 'eth_call')
+            expect(requests).toHaveLength(QUORUM)
+        })
+
+        it('uses another server, if server sends HTTP 503 response', async () => {
+            const requests = await runErrorTest({ httpStatus: 503 } )
+            expect(requests).toHaveLength(QUORUM + 1)
+        })
+
+        it('uses another server, if server sends HTTP 429 response', async () => {
```
harbu, line 84: This should break when `FetchRequest.retryFunc` returns `false`. It if doesn't break this that it is a bit weird...

packages/client/testi.html
```diff
@@ -0,0 +1,40 @@
+<!DOCTYPE html>
```
harbu: Remove before merging

PR #2569, opened 2024-05-28
refactor(DHT): Extract WebsocketServerConnector

juslesan:
## Summary

New classes WebsocketServerConnector and WebsocketClientConnector.

The WebsocketServerConnector is responsible for connectivity checking and creationg of server connections. WebsocketClientConnection only handles  client side connections.


ptesavol (approved):
LGTM

PR #2571, opened 2024-05-29
refactor(DHT): Remove managed connection from connectors

juslesan:
## Summary

Removed ManagedConnection from connectors. ManagedConnections now only exist for handshaked connections in the ConnectionManager.

Connectors now return a `PendingConnection` object. This is passed to the ConnectionManager and Handshakers.

## Changes

- Removed ConnectionType from ManagedConnection.

## Future improvements

- Remove autocertification and websocket server starting related logic from WebsocketServerConnector. Move to WebsocketServer.
- Move PendingConnection handling to ConnectorFacade from ConnectionManager?
- Refactor connection opening logic to be fully event based.
- See if connecting connections can be removed from the Connectors, use ConnectionManagers state only. Holding the state in two places is prone to bugs.


packages/dht/src/connection/ConnectionManager.ts
```diff
@@ -76,11 +77,20 @@ export interface TlsCertificate {
     certFileName: string
 }
 
-interface Endpoint {
-    connection: ManagedConnection
+interface ConnectingEndpoint {
```
ptesavol, line 80: Add todo comment about encapsulating Endpoint into a class/class hierarchy. Currently ConnectionManager needs to know too much about the internals of Endpoints (buffer swithing, type casting etc)

ptesavol (approved):
Approved, this is a good step forwards! Todo comment should be added somewhere about encapsulating Endpoint into a class/set of classes as nowadays ConnectionManager needs to know too much about the internals of an Endpoint.

PR #2506, opened 2024-04-18
Ethers v6 upgrade

harbu:
Update to Ethers v6

## Other changes

- Use explicit version of `@lit-protocol/core` (this doesn't compile (https://github.com/streamr-dev/network/pull/2506/commits/4adafc44ab2040e039c8cd5a6c7d7136bb3a209f)
- Fix eslint warnings in `WebrtcConnectorRpcLocal.ts` (why there are no warnings in `main` for these?)

## Future improvements

- Do not ignore broken links in Docusaurus: use "`throw`" in `onBrokenLinks` in `docusaurus.config.js`.
- Make `@ensdomains/ens-contracts` a dependency of `@streamr/network-contracts-ethers6`.

PR #2580, opened 2024-06-07
fix(client): Require `streamRegistryChainRPCs#name`

teogeb:
Made `name` field of `streamRegistryChainRPCs` required in config schema.

The field is already required in the `Config` class (was changed in https://github.com/streamr-dev/network/pull/2506).

## Open questions

The name of the field is not very clear? We have a separate field called `chainId`, and therefore it would make sense to rename this field to `chainName`/`networkName`/`networkId`? Or introduce a new contract-level block `chain` and move both of the field into that block. (The current config structure made sense more earlier when there were two RPCs blocks, see https://github.com/streamr-dev/network/pull/2131)


teogeb:
This PR is not needed, as the name field will be removed in https://github.com/streamr-dev/network/pull/2581.

PR #2575, opened 2024-06-04
refactor: abstract signature handling

harbu:
## Summary

Refactor signature handling logic to allow for extension of new signature types in the future.

- Extract main classes `MessageSigner` (signing) and `SignatureValidator` (validating)
- Move all signature handling code into `signature` folder
- Rename `Authentication#createMessageSignature` to `Authentication#signWithWallet`

## Future improvements
- Abstract the whole publisherId identity to be any string (follow up PR)

## Checklist before requesting a review

- [x] Is this a breaking change? If it is, be clear in summary.
- [x] Read through code myself one more time.
- [x] Make sure any and all `TODO` comments left behind are meant to be left in.
- [x] Has reasonable passing test coverage?
- [x] Updated changelog if applicable.
- [x] Updated documentation if applicable.

PR #2587, opened 2024-06-11
feat(sdk): abstract message publisher identity

harbu:
## Summary

Based on #2575, land that first.

## Changes

Provide a bullet list of individual changes. Leave this section out if change
set is small and obvious from summary.

- Potentially matching same key with different signature type in LocalGroupKeyStore, problem?
- Removed Legacy group key support. Is this fine?

Provide a bullet list or description of known omissions or limitations of the
solution and/or any ideas for future improvement. Leave section ouf if
not applicable.


packages/client/src/Authentication.ts
```diff
@@ -15,18 +14,13 @@ export interface Authentication {
     // always in lowercase
     getAddress: () => Promise<EthereumAddress>
     signWithWallet: (payload: Uint8Array) => Promise<Uint8Array>
```
teogeb, line 16: Maybe the `withWallet` suffix is not important as there is no other counterpart to distinguish from?  

Also, would it make just expose one `getSigner` function instead of exposing `getStreamRegistryChainSigner` and `signWithWallet`?

```suggestion
    signWithWallet: (payload: Uint8Array) => Promise<Uint8Array>
```

PR #2595, opened 2024-06-14
refactor(trackerless-network): `PeerDescriptorStoreManager`

teogeb:
Renamed and refactored `EntryPointDiscovery` to abstract `PeerDescriptorStoreManager`. 

This makes the module boundaries more explicit: the `PeerDescriptorStoreManager` is an abstract utility class which manages the DHT store operations for `PeerDescriptor` instances.  There is currently only one user of that class, the `ContentDeliveryManager`.

## Future improvements

- Maybe `MAX_NODE_COUNT` could be a configurable option in `PeerDescriptorStoreManager`
- Remove `setStreamPartEntryPoints` method and add the list of known nodes as an optional parameter to join/leave methods.
- Maybe we should use the term _entry point_ only for the global network entry point? Stream part "entry points" are somewhat different concept, i.e. just some known nodes in a given stream. Global entry point nodes have different requirements than stream part "entry points": e.g. the global entry points need to run a WebSocket server, but stream part "entry points" are just some peers.


packages/trackerless-network/src/logic/NodeStoreManager.ts
```diff
@@ -0,0 +1,97 @@
...
+    fetchDataFromDht: (key: DhtAddress) => Promise<DataEntry[]>
+    storeDataToDht: (key: DhtAddress, data: Any) => Promise<PeerDescriptor[]>
+    deleteDataFromDht: (key: DhtAddress, waitForCompletion: boolean) => Promise<void>
+}
+
+/**
+ * For each key there is usually 0-MAX_NODE_COUNT PeerDescriptors stored in the DHT. If there are fewer node,
+ * the peer descriptor of the local node is stored to the DHT.
+ */
+export class NodeStoreManager {
```
juslesan: Could rename to `PeerDescriptorStoreManager`

juslesan (approved):

juslesan:
> Maybe we should use the term entry point only for the global network entry point? Stream part "entry points" are somewhat different concept, i.e. just some known nodes in a given stream. Global entry point nodes have different requirements than stream part "entry points": e.g. the global entry points need to run a WebSocket server, but stream part "entry points" are just some peers.

Stream entry points could be called ie. stream access points

PR #2601, opened 2024-06-18
feat(client, broker): [NET-1310] Use custom event polling logic for Broker features

teogeb:
Moved `OperatorContractFacade` and `operatorContractUtils` from the `broker` package to the `client` package. Renamed it to `Operator`. Added `StreamrClient#getOperator` semi-public method to the client API.

Modified the class to use client's contract utilities, e.g. the `FallbackProvider` and the `ChainEventPoller`.
-  The functions of `operatorContractUtils` don't use `FallbackProvider` and the `ChainEventPoller`. That's maybe ok as those are used in tests only. But if we start to use the  functions in production code, we should start to start to use the  helper classes also in these functions.

## Future improvements
- Replace individual contract artifacts with imports from `@streamr/network-contracts-ethers6`. But need to find a way to do this without bloating webpack bundle size.
- Could use https://typedoc.org/tags/hidden/ in classes/methods which are currently marked as `@deprecated`?
- Do we need to export the `_operatorContractUtils` object? Maybe could add a `StreamrClient#getSponsorship` method? Some methods/functions of `Operator` and `operatorContractUtils` could be moved to the class returned by that method
  - Maybe could remove `OperatorRegistry` class? 
- Lazy load instances of `Contract` in `Operator`?


packages/broker/src/plugins/operator/OperatorPlugin.ts
```diff
@@ -195,49 +194,50 @@ export class OperatorPlugin extends Plugin<OperatorPluginConfig> {
...
-                            },
-                            inspectionIntervalInMs: 8 * 60 * 1000,
-                            maxInspections: 10,
-                            abortSignal: this.abortController.signal
+            addManagedEventListener(
+                operator,
+                'reviewRequested',
+                (event: ReviewRequestEvent): void => {
+                    setImmediate(() => {
+                        executeSafePromise(async () => {
```
harbu: I would just leave out `executeSafePromise` here since it is one big try-catch anyways

packages/broker/src/plugins/operator/inspectRandomNode.ts
```diff
@@ -1,15 +1,14 @@
...
+import { EthereumAddress, Logger, randomString } from '@streamr/utils'
 import { CreateOperatorFleetStateFn } from './OperatorFleetState'
+import { StreamPartAssignments } from './StreamPartAssignments'
 import { inspectOverTime } from './inspectOverTime'
+import { findTarget } from './inspectionUtils'
 
 export async function inspectRandomNode(
     operatorContractAddress: EthereumAddress,
-    contractFacade: ContractFacade,
+    operator: Operator,
```
harbu:
```suggestion
    myOperator: Operator,
```

packages/broker/src/plugins/operator/inspectionUtils.ts
```diff
@@ -41,12 +40,13 @@ function getPartitionsOfStreamAssignedToMe(
 
 export async function findTarget(
     myOperatorContractAddress: EthereumAddress,
-    contractFacade: ContractFacade,
+    operator: Operator,
```
harbu:
```suggestion
    myOperator: Operator,
```

packages/broker/src/plugins/operator/maintainOperatorValue.ts
```diff
@@ -1,25 +1,25 @@
...
-import { ContractFacade } from './ContractFacade'
 
 const logger = new Logger(module)
 
 export const maintainOperatorValue = async (
     withdrawLimitSafetyFraction: number,
     minSponsorshipEarningsInWithdraw: number,
     maxSponsorshipsInWithdraw: number,
-    contractFacade: ContractFacade
+    operator: Operator
```
harbu:
```suggestion
    myOperator: Operator
```

packages/broker/src/plugins/operator/reviewSuspectNode.ts
```diff
@@ -31,7 +30,7 @@ export const reviewSuspectNode = async ({
     sponsorshipAddress,
     targetOperator,
     partition,
-    contractFacade,
+    operator,
```
harbu:
```suggestion
    myOperator,
```

packages/client/src/StreamrClient.ts
```diff
@@ -696,6 +701,24 @@ export class StreamrClient {
         return _getEthersOverrides(this.rpcProviderSource, this.config)
     }
 
+    /**
+     * @deprecated This in an internal method
+     * @hidden
+     */
+    async getOperator(operatorContractAddress: EthereumAddress): Promise<Operator> {
```
harbu: Can be made sync

packages/client/src/contracts/Operator.ts
```diff
@@ -25,6 +33,36 @@ interface EarningsData {
     maxAllowedEarningsDataWei: bigint
 }
 
+/**
+ * @deprecated
+ * @hidden
+ */
+export interface StakeEvent {  // TODO export, do we want to keep the current style where event payloads are always objects?
```
harbu: Yeah I think we should keep this style. Remove TODO?

packages/client/src/contracts/Operator.ts
```diff
@@ -73,67 +111,138 @@ const compareBigInts = (a: bigint, b: bigint) => {
...
-            config
-        )
+    private initEventGateways(
+        contractAddress: EthereumAddress,
+        loggerFactory: LoggerFactory,
+        eventPollInterval: number
+    ): void {
+        const chainEventPoller = new ChainEventPoller(this.rpcProviderSource.getSubProviders().map((p) => {
+            return this.contractFactory.createEventContract(contractAddress, OperatorArtifact, p)
+        // eslint-disable-next-line no-underscore-dangle
```
harbu:
```suggestion
```

packages/client/src/contracts/Operator.ts
```diff
@@ -425,86 +550,88 @@ export class ContractFacade {
...
-                    voteStartTimestampInSecs * 1000,
-                    voteEndTimestampInSecs * 1000
-                )
-            },
-            abortSignal
-        )
-    }
-
+    // TODO could move this method as this is functionality is not specific to one Operator contract instance
     async getStreamId(sponsorshipAddress: string): Promise<StreamID> {
```
harbu:
```suggestion
    async getStreamId(sponsorshipAddress: EthereumAddress): Promise<StreamID> {
```

harbu (changes requested):

harbu (approved):

PR #2609, opened 2024-06-24
feat(cdn-location): Remove `aws.amazon.com` URL

teogeb:
It seems that the `aws.amazon.com` responses no longer include the `X-Amz-Cf-Pop` header used by our `cdn-location` functionality. Therefore removed the URL from the list.


ptesavol (approved):
LGTM

PR #2613, opened 2024-07-01
refactor(client): Change `StreamrClient#getNode` return type

teogeb:
## Summary

Changed `StreamrClient#getNode` to return `NetworkNodeFacade` instead of `NetworkNodeStub`. Now the `NetworkNodeStub` is just an internal implementation detail in the `NetworkNodeFacade` class.

## Changes

- Refactored all client's internal classes to use `NetworkNodeFacade` instead of `NetworkNodeStub`
  - `SubscriptionSession`, `PublisherKeyExchange`, `SubscriberKeyExchange`, `MetricsPublisher`
- Refactored usages in Broker
- Refactored `FakeEnvironment` nodes to be instances of `NetworkNodeFacade` instead of `NetworkNodeStub`
  - As `FakeStorageNode` is no longer a `NetworkNodeStub`, some refactoring was needed. Now `FakeChain` emits events when storage node assignments are written to the chain.  `FakeStorageNode` listens to those events. `FakeStorageNode` and `StreamrClient` teardowns are implemented using the `DestroySignal` functionality.

### DIFF
--- a/packages/client/test/test-utils/fake/FakeChain.ts
+++ b/packages/client/test/test-utils/fake/FakeChain.ts
@@ -3,7 +3,6 @@ import { StreamMetadata } from '../../../src/Stream'
 import { StreamPermission } from '../../../src/permission'
 import { EthereumAddress, Multimap } from '@streamr/utils'
 import { StorageNodeMetadata } from '../../../src/contracts/StorageNodeRegistry'
-import { StorageNodeAssignmentEvent } from '../../../src/contracts/StreamStorageRegistry'
 import { EventEmitter } from 'eventemitter3'
 
 export type PublicPermissionTarget = 'public'
@@ -14,8 +13,13 @@ export interface StreamRegistryItem {
     permissions: Multimap<EthereumAddress | PublicPermissionTarget, StreamPermission>
 }
 
+export interface FakeStorageNodeAssignmentEvent {
+    readonly streamId: StreamID
+    readonly nodeAddress: EthereumAddress
+}
+
 export interface Events {
-    streamAddedToStorageNode: (payload: Omit<StorageNodeAssignmentEvent, 'blockNumber'>) => void
+    streamAddedToStorageNode: (payload: FakeStorageNodeAssignmentEvent) => void
 }
 
 export class FakeChain {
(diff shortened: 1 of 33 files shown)


packages/client/test/test-utils/fake/FakeChain.ts
```diff
@@ -12,9 +14,56 @@ export interface StreamRegistryItem {
...
+        this.streams.set(streamId, registryItem)
+    }
+
+    getStorageAssignments(streamId: StreamID): EthereumAddress[] {
+        return this.storageAssignments.get(streamId)
+    }
+
+    addStorageAssignment(streamId: StreamID, nodeAddress: EthereumAddress): void {
+        this.storageAssignments.add(streamId, nodeAddress)
+        this.eventEmitter.emit('streamAddedToStorageNode', { streamId, nodeAddress })
```
harbu: Do we need to guard against double add?
```suggestion
        const existedBefore = this.storageAssignments.has(streamId, nodeAddress)
        this.storageAssignments.add(streamId, nodeAddress)
        if (!existedBefore) {
            this.eventEmitter.emit('streamAddedToStorageNode', { streamId, nodeAddress })
        }
```

packages/client/test/test-utils/fake/FakeChain.ts
```diff
@@ -12,9 +14,56 @@ export interface StreamRegistryItem {
     permissions: Multimap<EthereumAddress | PublicPermissionTarget, StreamPermission>
 }
 
+export interface Events {
+    streamAddedToStorageNode: (payload: Omit<StorageNodeAssignmentEvent, 'blockNumber'>) => void
```
harbu: Better to just emit the whole thing but always use `blockNumber: 0`?
```suggestion
    streamAddedToStorageNode: (payload: StorageNodeAssignmentEvent) => void
```

harbu:
>Do we need NetworkNodeFacade#publishToNode anymore? Can we just use the NetworkNodeFacade#broadcast method instead? (We can add assertNotDestroyed call e.g. to broadcast if needed)

I would remove

packages/client/test/test-utils/fake/FakeChain.ts
teogeb: There is a guard in the caller (`FakeStreamStorageRegistry#addStreamToStorageNode`), but better move it to here

packages/client/test/test-utils/fake/FakeChain.ts
teogeb: Or maybe a separate interface for the payload as this is not related to the `StreamrClient` event?

harbu (approved):

PR #2614, opened 2024-07-01
refactor(trackerless-network, client): Move protocol translator

teogeb:
Moved `StreamMessageTranslator` functionality from `trackerless-network` to `client`.


packages/broker/src/plugins/storage/DataQueryFormat.ts
```diff
@@ -1,5 +1,5 @@
 import { StreamMessage } from '@streamr/protocol'
-import { convertBytesToStreamMessage } from '@streamr/trackerless-network'
+import { convertBytesToStreamMessage } from '@streamr/sdk'
```
harbu, line 2: Can we get rid of `@streamr/trackerless-network` from broker's package.json?

harbu (approved):

packages/broker/src/plugins/storage/DataQueryFormat.ts
teogeb: Not yet, as `StreamID` functionality is there

teogeb: It wasn't in the `package.json`

PR #2702, opened 2024-08-02
feat(node): NET-1252 OperatorPlugin registers RPC method

juslesan:
## Summary

OperatorPlugin now registers the RPC method that can be used to discover which operators are in which partitions.


packages/node/src/plugins/operator/OperatorPlugin.ts
```diff
@@ -106,6 +114,16 @@ export class OperatorPlugin extends Plugin<OperatorPluginConfig> {
         await fleetState.start()
         await maintainTopologyHelper.start()
 
+        const networkNode = await streamrClient.getNode()
+
+        const rpcServerFunction = async (request: OperatorDiscoveryRequest) => {
+            const streamPartId = StreamPartIDUtils.parse(request.streamPartId)
```
harbu, line 120: Add error handling (try-catch) for the case where string is not a valid stream part id. Do we just return an empty operators list response in that case?

packages/node/test/integration/plugins/operator/OperatorPlugin.test.ts
```diff
@@ -93,4 +93,31 @@ describe('OperatorPlugin', () => {
             await createBroker(config)
         }).rejects.toThrow('Plugin operator doesn\'t support client config value "false" in network.node.acceptProxyConnections')
     })
+    
+    it('Accepts OperatorDiscoveryRequests', async () => {
```
harbu:
```suggestion
    it('accepts OperatorDiscoveryRequests', async () => {
```

packages/node/test/integration/plugins/operator/OperatorPlugin.test.ts
```diff
@@ -93,4 +93,31 @@ describe('OperatorPlugin', () => {
...
+            privateKey: brokerWallet.privateKey,
+            extraPlugins: {
+                operator: {
+                    operatorContractAddress: await operatorContract.getAddress(),
+                }
+            }
+        })
+        await waitForCondition(async () => (await broker.getStreamrClient().getSubscriptions(stream.id)).length > 0)
+        // Ensure that heartbeat has been sent (setting heartbeatUpdateIntervalInMs lower did not help)
+        await wait(10000)
```
harbu: Do we actually need to wait for a heartbeat to be sent here since there is only one broker?

harbu (changes requested):
looking good, see my comments for a few tweaks

packages/node/test/integration/plugins/operator/OperatorPlugin.test.ts
juslesan: At the moment the node does not even know itself before it sees its own heartbeat. I didn't want to change that as I don't know if it's intentional or not

packages/node/src/plugins/operator/OperatorPlugin.ts
juslesan: It is completely ok for a server function to throw. Should probably test what the error message would look like here on the client side

harbu: Alright, so the protobuf package will take care of any error that would be thrown here?

packages/node/test/integration/plugins/operator/OperatorPlugin.test.ts
harbu: To not have to rely on a timer, how about  putting this function in the beginning of the file
```ts
async function waitForHeartbeatMessage(operatorContractAddress: EthereumAddress): Promise<void> {
        const client = createClient(fastPrivateKey())
        const sub = await client.subscribe(formCoordinationStreamId(operatorContractAddress))
        await collect(sub, 1)
        await client?.destroy()
    }
```

and calling it here?

packages/node/src/plugins/operator/OperatorPlugin.ts
```diff
@@ -106,6 +114,16 @@ export class OperatorPlugin extends Plugin<OperatorPluginConfig> {
...
+            const operators = streamPartAssignments.getAssignedNodesForStreamPart(streamPartId)
+            return OperatorDiscoveryResponse.create({ operators: operators.map((operator) => peerDescriptorTranslator(operator)) })
+        }
+
+        networkNode.registerExternalRpcMethod(OperatorDiscoveryRequest, OperatorDiscoveryResponse, 'discoverOperators', rpcServerFunction)
```
harbu: My IDE underlined this. Needs an `await` infront.

packages/node/src/plugins/operator/OperatorPlugin.ts
juslesan: Yeah the errors are caught and reported back to the requestor

harbu (approved):
LGTM; (node-docker-image CI job somehow broken?)

juslesan:
Merging as the docker issue is unrelated

PR #2721, opened 2024-08-16
perf(node): [NET-1113] Optimize operators query in `checkOperatorValueBreach`

teogeb:
The `checkOperatorValueBreach` functionality uses cached list of staked operators instead of querying random operator address on each invocation.

## Changes
- `OperatorPlugin` creates a cache of staked operators, gives that as an to the `checkOperatorValueBreach` call
- added `Cache` utility to `@streamr/utils` package
- renamed `Operator#getOperatorAddresses` -> `getStakedOperators`, it no longer uses `updateRequiredBlockNumber()` functionality


packages/node/src/plugins/operator/OperatorPlugin.ts
```diff
@@ -70,6 +71,8 @@ export interface OperatorServiceConfig {
     getEthersOverrides: () => Promise<Overrides>
 }
 
+const STAKED_OPERATORS_CACHE_MAX_AGE = 7 * 24 * 60 * 60 * 1000
```
harbu: Maybe we can make it a bit shorter to play it safe.
```suggestion
const STAKED_OPERATORS_CACHE_MAX_AGE = 2 * 24 * 60 * 60 * 1000
```

packages/utils/src/Cache.ts
```diff
@@ -0,0 +1,21 @@
...
+    constructor(valueFactory: () => Promise<V>, maxAgeInMilliseconds: number) {
+        this.valueFactory = valueFactory
+        this.maxAgeInMilliseconds = maxAgeInMilliseconds
+    }
+
+    async get(): Promise<V> {
+        const now = Date.now()
+        if ((this.valueTimestamp === undefined) || (now > (this.valueTimestamp + this.maxAgeInMilliseconds))) {
+            this.valueTimestamp = now
+            this.value = await this.valueFactory()
```
harbu: Would it be acceptable for `valueFactory` to reject? If so I would swap the order of the statements here
```suggestion
            this.value = await this.valueFactory()
            this.valueTimestamp = now
```

harbu (changes requested):

harbu (approved):

PR #2725, opened 2024-08-21
fix: NET-1343 don't send a transaction when asked to close a non-existing flag

jtakalai:
## Summary

Do estimateGas before sending a transaction with gasLimit (this circumvents the usual estimateGas that catches reverts)

- [X] Read through code myself one more time.
- [X] Has reasonable passing test coverage?

### DIFF
--- a/packages/node/src/plugins/operator/closeExpiredFlags.ts
+++ b/packages/node/src/plugins/operator/closeExpiredFlags.ts
@@ -21,6 +21,10 @@ export const closeExpiredFlags = async (
         const operatorAddress = flag.targetOperator
         const sponsorship = flag.sponsorship
         logger.info('Close expired flag', { flag })
-        await operator.closeFlag(sponsorship, operatorAddress)
+        try {
+            await operator.closeFlag(sponsorship, operatorAddress)
+        } catch (e) {
+            logger.warn('Failed to close flag (does it exist?)', { sponsorship, operatorAddress, error: e })
+        }
     }
 }
--- a/packages/node/src/plugins/operator/closeExpiredFlags.ts
+++ b/packages/node/src/plugins/operator/closeExpiredFlags.ts
@@ -23,8 +23,8 @@ export const closeExpiredFlags = async (
         logger.info('Close expired flag', { flag })
         try {
             await operator.closeFlag(sponsorship, operatorAddress)
-        } catch (e) {
-            logger.warn('Failed to close flag (does it exist?)', { sponsorship, operatorAddress, error: e })
+        } catch (err) {
+            logger.warn('Failed to close flag (does it exist?)', { sponsorship, operatorAddress, err })
         }
     }
 }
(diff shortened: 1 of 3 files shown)


packages/node/src/plugins/operator/closeExpiredFlags.ts
```diff
@@ -21,6 +21,10 @@ export const closeExpiredFlags = async (
         const operatorAddress = flag.targetOperator
         const sponsorship = flag.sponsorship
         logger.info('Close expired flag', { flag })
-        await operator.closeFlag(sponsorship, operatorAddress)
+        try {
+            await operator.closeFlag(sponsorship, operatorAddress)
+        } catch (e) {
+            logger.warn('Failed to close flag (does it exist?)', { sponsorship, operatorAddress, error: e })
+        }
```
harbu: The logging library we use likes the name `err` (prints full stack trace)
```suggestion
        } catch (err) {
            logger.warn('Failed to close flag (does it exist?)', { sponsorship, operatorAddress, err })
        }
```

packages/sdk/src/contracts/Operator.ts
```diff
@@ -575,6 +575,8 @@ export class Operator {
     async voteOnFlag(sponsorshipAddress: EthereumAddress, targetOperator: EthereumAddress, kick: boolean): Promise<void> {
         const voteData = kick ? VOTE_KICK : VOTE_NO_KICK
         await this.connectToContract()
+        // estimateGas throws if transaction would fail, ignore the return value
```
teogeb: Should the `estimateCall` take to account that we use `gasLimit: '1300000'` in the actual call? Or should we check whether the return value is greater than `1300000` and throw an error in that case?

teogeb: Maybe could also comment why this method uses explicit `estimateGas` call while all other method do not do that?

packages/sdk/test/end-to-end/Operator.test.ts
```diff
@@ -177,6 +177,18 @@ describe('Operator', () => {
...
     }, 60 * 1000)  // TODO why this is slower, takes ~35 seconds?
 
+    it('avoids sending a transaction when trying to close non-existing flags', async () => {
+        const wallet = deployedOperator.nodeWallets[0] as Wallet
+        const operator = await getOperator(wallet, deployedOperator)
+        const nonceBefore = await wallet.getNonce() // nonce = "how many transactions sent so far"
+        await expect(operator.closeFlag(
+            toEthereumAddress(await sponsorship1.getAddress()),
+            toEthereumAddress(await deployedOperator.operatorContract.getAddress())
+        )).rejects.toThrowError()
```
teogeb: Better to wrap the promise creation to a function call so that we don't assume too much about the async implementation of `closeFlag()`. (This style doesn't catch the error if  the method would be a sync function which throws an exception instead of returning a `Promise`.)

```suggestion
        await expect(async () => operator.closeFlag(
            toEthereumAddress(await sponsorship1.getAddress()),
            toEthereumAddress(await deployedOperator.operatorContract.getAddress())
        )).rejects.toThrow()
```

teogeb (changes requested):

packages/sdk/src/contracts/Operator.ts
jtakalai: I suppose we could handle the case where the estimation is higher than 1.3M... since that estimate is conservative (higher than expected) probably we could even not send the transaction. I'll add this logic.

harbu (approved):

harbu:
lgtm; thoughts @teogeb ?

PR #2726, opened 2024-08-22
feat(network): Prioritise ws connections in layer2

juslesan:
## Summary

If a node has 0 neighbors in a stream prioritize picking up one WS server node as a neighbor to improve time to data.

### DIFF
--- a/packages/trackerless-network/src/logic/NodeList.ts
+++ b/packages/trackerless-network/src/logic/NodeList.ts
@@ -8,10 +8,16 @@ export interface Events {
     nodeRemoved: (id: DhtAddress, remote: ContentDeliveryRpcRemote) => void
 }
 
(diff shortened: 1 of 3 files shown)


juslesan:
Still hitting some sad paths with neighbor discovery but the mean and median have gone down (1000 runs):
```
-------------------------------
Time to start network stack
mean: 413.013
median: 311
max: 3873
min: 141
std: 521.7751056068122
-------------------------------
Time to fetch stream entry points
mean: 160.98901098901098
median: 102
max: 10002
min: 11
std: 409.78094935302715
-------------------------------
Time to first neighbor
mean: 1144.8512562814071
median: 579
max: 33513
min: 108
std: 2148.615397557
-------------------------------
Time to first message in network
mean: 1209.987
median: 623.5
max: 33571
min: 164
std: 2192.8854873957735
-------------------------------
Time to validate
mean: 959.806
median: 603.5
max: 13033
min: 443
std: 1238.903815622504
-------------------------------
Time to first message
mean: 2839.313
median: 1932
max: 35263
min: 1206
std: 2581.5059320154583
```

Some of these bad cases are could be caused by older versions passing stale contacts in the network

harbu (approved):

PR #2727, opened 2024-08-23
fix(network): exclude initial ws neighbors for parallel operations

juslesan:
## Summary

As we run parallel handshaking algorithms now it makes sense to push the WS node to the excluded list immediately. This way the same WS server cannot be picked

## Future improvements

Explore more general exclusion in general. Parallel handshaking operations can be done to the same nodes. Naively pushing attempted handshakes to the exclude list breaks some tests.

### DIFF
--- a/packages/trackerless-network/src/logic/neighbor-discovery/Handshaker.ts
+++ b/packages/trackerless-network/src/logic/neighbor-discovery/Handshaker.ts
@@ -83,7 +83,9 @@ export class Handshaker {
                 true
             )
             if (wsNode) {
-                neighbors.set(getNodeIdFromPeerDescriptor(wsNode.getPeerDescriptor()), wsNode)
+                const wsNodeId = getNodeIdFromPeerDescriptor(wsNode.getPeerDescriptor())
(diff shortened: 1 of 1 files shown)


juslesan:
900 runs:

```
-------------------------------
Time to start network stack
mean: 372.1355555555556
median: 292.5
max: 3434
min: 147
std: 461.71627586914144
-------------------------------
Time to fetch stream entry points
mean: 103.43487858719647
median: 80
max: 5140
min: 9
std: 215.12270398084817
-------------------------------
Time to first neighbor
mean: 794.460511679644
median: 429
max: 17402
min: 216
std: 1311.6120511801776
-------------------------------
Time to first message in network
mean: 854.8622222222223
median: 499.5
max: 17422
min: 221
std: 1317.7143464328915
-------------------------------
Time to validate
mean: 977.94
median: 566
max: 13609
min: 437
std: 1415.094531887456
-------------------------------
Time to first message
mean: 2493.6566666666668
median: 1771.5
max: 18558
min: 1142
std: 1945.860233335832
```

harbu (approved):

PR #2736, opened 2024-09-02
fix (DHT): NET-1347 fix hanging sends to connecting connections with `connect: false`

juslesan:
## Summary

No longer allow sending in the connection manager to connecting connections. This fixes a problem with stopping of the nodes taking more than 15 seconds.

This is the intended behaviour as connections should be opened if the connect flag is set false.

## Future improvements

- Fix problem with timeouts not working for notifications in the proto-rpc package.


ptesavol (approved):
LGTM - I also took the change to the C++ version. Btw, does the notification flag really have any effect? Isn't that decided based on the return value of the function?

juslesan:
>  Btw, does the notification flag really have any effect? Isn't that decided based on the return value of the function?

@ptesavol Once entering the RpcCommunicator the notification flag has to be set. This could be set automatically somewhere if the method in question is a notification. If the notification is not set the client will expect an ACK from the server before resolving.

PR #2738, opened 2024-09-04
feat(network): NET-1312 include RTTs to layer2 neighbors in the InfoResponse

juslesan:
## Summary

Include RTTs to neighbors in layer2 in the InfoResponse. The RTT can be initially undefined if the node has not yet pinged the neighbor.


packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -164,6 +164,12 @@ message StreamPartitionInfo {
   string id = 1;
   repeated dht.PeerDescriptor controlLayerNeighbors = 2;
   repeated dht.PeerDescriptor contentDeliveryLayerNeighbors = 3;
+  repeated ContentDeliveryRtt rtts = 4;
```
teogeb:
```suggestion
  repeated ContentDeliveryRtt contentDeliveryLayerRtts = 4;
```

Alternative we could change the structure of `contentDeliveryLayerNeighbors` so that it contains array of something like this:
```
{
   peerDescription: PeerDescription
   rtt: number
}
```

packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -164,6 +164,12 @@ message StreamPartitionInfo {
...
+}
+
+message ContentDeliveryRtt {
+  string nodeId = 1;
```
teogeb:
```suggestion
  bytes nodeId = 1;
```

packages/trackerless-network/src/logic/ContentDeliveryLayerNode.ts
```diff
@@ -390,6 +391,16 @@ export class ContentDeliveryLayerNode extends EventEmitter<Events> {
         return this.options.neighbors.getAll().map((n) => n.getPeerDescriptor())
     }
 
+    getRtts(): ContentDeliveryRtt[] {
+        return this.options.neighbors.getAll().map((n) => {
+            const nodeId = getNodeIdFromPeerDescriptor(n.getPeerDescriptor())
+            return {
+                nodeId,
+                rtt: this.options.neighbors.get(nodeId)?.getRtt()
```
teogeb: Could use exclamation mark as we know that the neighbor exists.

packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -164,6 +164,12 @@ message StreamPartitionInfo {
...
+  optional int32 rtt = 2;
```
teogeb, line 171: We can just leave out the item if it doesn't have value for this field

```suggestion
  int32 rtt = 2;
```

packages/trackerless-network/src/logic/ContentDeliveryLayerNode.ts
teogeb: We should filter out items which have `undefined` rtt?

packages/trackerless-network/src/logic/ContentDeliveryManager.ts
```diff
@@ -260,6 +261,7 @@ export class ContentDeliveryManager extends EventEmitter<Events> {
             neighborTargetCount: this.options.streamPartitionNeighborTargetCount,
             acceptProxyConnections: this.options.acceptProxyConnections,
             rpcRequestTimeout: this.options.rpcRequestTimeout,
+            neighborUpdateInterval: this.options.neighborUpdateInterval,
```
teogeb, line 268: Does this get a correct value: i.e. where we set the default value for `neighborUpdateInterval` config option?

There is handling of default value in `createConfigWithDefaults`, but it is not called here and therefore the default is not applied that way?

teogeb (changes requested):

packages/trackerless-network/protos/NetworkRpc.proto
juslesan: Yeah implemented it like this because of compatibility. Although, as the InfoResponse should contain the version it should not be an issue to have a different schema

packages/trackerless-network/protos/NetworkRpc.proto
juslesan: Can't leave the field empty if it is not marked as optional. Also, the number will be 0 if the RTT is not defined and not optional. Undefined is better here for more accurate numbers.

packages/trackerless-network/src/logic/ContentDeliveryManager.ts
juslesan: `createConfigWithDefaults` is called when calling `createContentDeliveryLayerNode`

juslesan:
Pushed some refactors, will do more once it's decided if we should use a different proto schema in the info response

packages/trackerless-network/src/logic/ContentDeliveryLayerNode.ts
```diff
@@ -390,6 +391,15 @@ export class ContentDeliveryLayerNode extends EventEmitter<Events> {
         return this.options.neighbors.getAll().map((n) => n.getPeerDescriptor())
     }
 
+    getInfo(): ContentDeliveryLayerNeighborInfo[] {
```
teogeb:
```suggestion
    getInfos(): ContentDeliveryLayerNeighborInfo[] {
```

packages/trackerless-network/test/integration/NodeInfoRpc.test.ts
```diff
@@ -89,12 +89,16 @@ describe('NetworkStack NodeInfoRpc', () => {
                 {
                     id: streamPartId1,
                     controlLayerNeighbors: [normalizePeerDescriptor(otherPeerDescriptor)],
-                    contentDeliveryLayerNeighbors: [normalizePeerDescriptor(otherPeerDescriptor)]
+                    contentDeliveryLayerNeighbors: [{
+                        peerDescriptor: normalizePeerDescriptor(otherPeerDescriptor)
+                    }]
```
teogeb, line 94: Could we assert that there is a `rtt` value? (also at line `100`)

packages/trackerless-network/test/unit/ContentDeliveryLayerNode.test.ts
```diff
@@ -97,4 +97,14 @@ describe('ContentDeliveryLayerNode', () => {
...
     }, 25000)
 
+    it('getInfo', () => {
+        const nodeWithRtt = createMockContentDeliveryRpcRemote()
+        neighbors.add(nodeWithRtt)
+        const nodeWithoutRtt = createMockContentDeliveryRpcRemote()
+        neighbors.add(nodeWithoutRtt)
+        nodeWithRtt.setRtt(100)
+        const info = contentDeliveryLayerNode.getInfo()
+        expect(info[0].rtt).toEqual(100)
```
teogeb, line 107: Could assert that the objects have also the `peerDescriptor` field

packages/trackerless-network/src/logic/ContentDeliveryLayerNode.ts
juslesan: This would be kind of grammatically wrong since there is no plural for information.

packages/trackerless-network/test/integration/NodeInfoRpc.test.ts
juslesan: That would require waiting for neighbor updates and assuming that the RTT is always some value in the integration test. This could cause flakiness.

packages/trackerless-network/src/logic/ContentDeliveryLayerNode.ts
teogeb: True. Somehow we need to indicate that we return an array. How about e.g. `getInfoItems`? 

I think that `getInfos` would make the code easier to understand, and that the grammatical correctness doesn't improve readability here.

We've similarly used terms like `metadatas` when referring to arrays of stream metadata objects.

juslesan: Does the typing not indicate the return type?

teogeb: It does. Also the method name should. When we read the code which uses this method, we see the methods name, not the return type
```
const foo = node.getInfos()
``` 

From this snippet we should see that the `foo` is an array. (Of course, we should have a good name for `foo`, preferrably matching the getter.)

juslesan: I don't really see this being necessary since the method is called getInfo in all other modules but whatever

packages/trackerless-network/test/integration/ContentDeliveryManager.test.ts
```diff
@@ -139,4 +139,19 @@ describe('ContentDeliveryManager', () => {
...
+        manager1.joinStreamPart(STREAM_PART_ID)
+        manager2.joinStreamPart(STREAM_PART_ID)
+        await Promise.all([
+            waitForCondition(() => manager1.getNeighbors(STREAM_PART_ID).length === 1),
+            waitForCondition(() => manager2.getNeighbors(STREAM_PART_ID).length === 1)
+        ])
+        // Wait for RTTs to be updated
+        await wait(500)
+        const nodeInfo1 = await manager1.getNodeInfo()
+        const nodeInfo2 = await manager2.getNodeInfo()
```
teogeb:
```suggestion
        const nodeInfo1 = manager1.getNodeInfo()
        const nodeInfo2 = manager2.getNodeInfo()
```

teogeb (approved):

PR #2739, opened 2024-09-05
fix(sdk): Flag expiration timestamp

teogeb:
Before this PR the expiration timestamp is now calculated by adding `maxAge` to `flaggingTimestamp`. Now `voteEndTimestamp` is used instead of `flaggingTimestamp`. It ensures that the flag can't expire before the voting ends.

## Behavior before the fix

If a small value was configured for `maxAge`, the flag expired before the voting ended. In that case the `closeExpiredFlags` service tried to close the "expired" flag during the voting time. That caused the operator to cast a `NO_KICK` vote instead of closing the flag.

This is not an issue in the production environment, as the default value for `maxAge` is one day.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ Changes before Tatum release are not documented in this file.
 
 #### Fixed
 
+- Fixed flag expiration time in `Operator#getExpiredFlags` (https://github.com/streamr-dev/network/pull/2739)
+
 #### Security
 
 ### @streamr/node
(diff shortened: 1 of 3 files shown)


jtakalai (approved):
ok, understood what the idea is: change the meaning of maxAgeInMs so that it's 24h after flag really should've been closed, instead of 24h after it was started. This is a good change, in that any positive number is in principle valid here.

The difference isn't huge for current parameters, since review+voting period is 1h+1.5h << 24h, but at least now it's not possible to (accidentally?) give too low a value here, which might trigger spurious closing.

PR #2756, opened 2024-09-17
feat(network): NET-1356 browser webrtc send buffering

juslesan:
## Summary

Added additional buffering logic to the browser webrtc connections. If the connection buffer fills up messages are dropped. This will alleviate the issue by allowing more messages to be buffered.


packages/dht/src/connection/webrtc/BrowserWebrtcConnection.ts
```diff
@@ -19,6 +19,61 @@ interface Params {
...
+    }
+
+    sendBuffer(): void {
+        if (this.running === true) {
+            return
+        }
+        this.running = true
+        let sendAttempts = 0;
+        (async () => {
+            while (this.queue.length > 0 && this.running) {
```
ptesavol, line 51: Does the bufferedAmout really get smaller in the same tick on browser? I know in node-datachannel it does because there is a worker thread emptying the queue, but is the situation the same in the browser?

packages/dht/src/connection/webrtc/BrowserWebrtcConnection.ts
```diff
@@ -19,6 +19,61 @@ interface Params {
...
+                const messageToSend = this.queue.shift()!
+                try {
+                    this.sendFn(messageToSend)
+                } catch (err) {
+                    this.queue.unshift(messageToSend)
+                }
+                sendAttempts++
+                if (sendAttempts % 200 === 0) {
```
ptesavol, line 59: Isn't 200 attempts quite a lot?

packages/dht/src/connection/webrtc/BrowserWebrtcConnection.ts
```diff
@@ -19,6 +19,61 @@ interface Params {
...
+                    await wait(0) // give up CPU cycle
```
ptesavol, line 60: I think a shorter pause should be used here see https://stackoverflow.com/questions/18826570/settimeout0-vs-window-postmessage-vs-messageport-postmessage on how to implement one. (I have tried this in the past, and the send buffer does get drained also during the shorter pauses)

packages/dht/src/connection/webrtc/BrowserWebrtcConnection.ts
```diff
@@ -166,17 +223,26 @@ export class NodeWebrtcConnection extends EventEmitter<Events> implements IWebrt
...
     public send(data: Uint8Array): void {
+        this.messageQueue.push(data)
+        this.messageQueue.sendBuffer()
+    }
+
+    private doSend(data: Uint8Array): void {
         if (this.lastState === 'connected') {
-            this.dataChannel?.send(data as Buffer)
+            if (this.dataChannel!.bufferedAmount > 1 * 1000 * 1000) {
+                this.messageQueue.pause()
```
ptesavol, line 233: This looks a bit dangerous, doSend() is not an arrow function, so it is not guarantedd that "this" points to the right object. Please make doSend() an arrow function.

ptesavol (changes requested):
I think a browser test should be added to see how the system actually behaves (Logging when different things actually happen, if they do + throughput benchmark).  Also see the comments.

packages/dht/src/connection/webrtc/BrowserWebrtcConnection.ts
juslesan: This isn't really a reattempt or anything, it is simply a counter used for CPU breaking

packages/dht/src/connection/webrtc/BrowserWebrtcConnection.ts
juslesan: In line 96 the doSend function is bound to the message queue using an arrow function. No need to make the class method itself in to an arrow function

packages/dht/src/connection/webrtc/BrowserWebrtcConnection.ts
juslesan: Likely not smaller but it seems that it does get larger. We do have a listener for the onBufferedAmountLow event

PR #2760, opened 2024-09-23
feat(network): React to disconnected event to reject requests

juslesan:
## Summary

Reject requests after disconnected events


ptesavol (approved):
LGTM

PR #2762, opened 2024-09-24
test: Move `global.d.ts` files

teogeb:
Move global test type definition files to `test/types` directories in `dht`, `node` and `utils` packages.  This doesn't change any functionality.


packages/browser-test-runner/webpack.config.js
```diff
@@ -15,7 +15,7 @@ module.exports = function({ entry, libraryName, alias = {} }) {
                 rules: [
                     {
                         test: /\.ts?$/,
-                        exclude: /(node_modules|simulation)/,
+                        exclude: [/(node_modules|simulation)/, /\.d\.ts$/],
```
harbu, line 18: What is this change about?

harbu (approved):

packages/browser-test-runner/webpack.config.js
teogeb: Resolves this error:
```
ERROR in ./test/types/global.d.ts
Module build failed (from ../../node_modules/ts-loader/index.js):
Error: TypeScript emitted no output for ... /network/packages/utils/test/types/global.d.ts.
```

PR #2589, opened 2024-06-13
Generalize userid authentication

harbu:
## Summary

Please provide a summary of the changes and a motivation if applicable.

## Changes

Provide a bullet list of individual changes. Leave this section out if change
set is small and obvious from summary.

## TODO

- [ ] Change 4*`randomUserId` functions to use random length instead of Ethereum-compatible 20 bytes
- [x] Make a separate PR about removing legacy key support, can merge "immediately"
- [x] Move (branded) `type UserID` type from `dht` to `trackerless-network`
- [ ] Grep `address` and `userAddress` and replace with e.g. `userId` in project.
- [x] Test function `randomUserId()` for tests
- [ ] Is the `userAddress` field deprecated in The Graph? Should we use the `userId` field, or both?
- [ ] Could we have a smart `toEthereumAddress` utility so that we wouldn't need `isEthereumAddressByteArray` function?

## Checklist before requesting a review

- [ ] Is this a breaking change? If it is, be clear in summary.
- [ ] Read through code myself one more time.
- [ ] Make sure any and all `TODO` comments left behind are meant to be left in.
- [ ] Has reasonable passing test coverage?
- [ ] Updated changelog if applicable.
- [ ] Updated documentation if applicable.

PR #2763, opened 2024-09-25
refactor: `UserID` annotation

teogeb:
Added new helper type `UserID`, which is currently just an alias to `EthereumAddress`. Replaced the `EthereumAddress`  type with this new type in all places where the address is a user id. 

Also unified some variable/parameter names.

This is a preparation PR for https://github.com/streamr-dev/network/pull/2589, in which the `UserID` type will be changed to `UInt8Array`.

## Other changes

Changed argument type of `StreamrClient#addEncryptionKey` from `EthereumAddress` to `string` as the method is part of our public API.


packages/sdk/src/subscribe/ordering/OrderMessages.ts
```diff
@@ -55,7 +56,7 @@ const createMessageChain = (
  */
 export class OrderMessages {
 
-    private readonly chains: Mapping<[EthereumAddress, string], OrderedMessageChain>
+    private readonly chains: Mapping<[UserID, string], OrderedMessageChain>
```
harbu, line 58: This will be problematic if we go directly to
`type UserID = Uint8Array`

packages/sdk/src/utils/persistence/BrowserPersistence.ts
```diff
@@ -6,7 +6,7 @@ export default class BrowserPersistence implements PersistenceContext {
     private readonly db: IDBPDatabase
 
     static async createInstance(opts: PersistenceContextOptions): Promise<BrowserPersistence> {
-        const db = await openDB(`streamr-sdk::${opts.clientId}`, 1, {
+        const db = await openDB(`streamr-sdk::${opts.ownerId}`, 1, {
```
harbu, line 9: Here we will have to remember to convert to hex?

packages/sdk/src/utils/persistence/ServerPersistence.ts
```diff
@@ -37,13 +37,13 @@ export default class ServerPersistence implements PersistenceContext {
...
         loggerFactory,
-        clientId,
+        ownerId,
         migrationsPath,
         onInit
     }: ServerPersistenceOptions) {
         this.logger = loggerFactory.createLogger(module)
         const paths = envPaths('streamr-sdk')
-        this.dbFilePath = resolve(paths.data, join('./', clientId, `GroupKeys.db`))
+        this.dbFilePath = resolve(paths.data, join('./', ownerId, `GroupKeys.db`))
```
harbu, line 46: Same here, we will have to remember to convert to hex?

harbu (approved):
LGTM;

Few thoughts:
- Should type `UserID` be in utils package instead of trackerless-network? 
- When we eventually make the type `UserID` point to `Uint8Array` we have to be careful to make sure we go thru the places where the type has been used and ensure that it is not dependent on the type being primitive (e.g key of `Map`).

PR #2765, opened 2024-09-26
build: Use `allowJs: false` in `tsconfig.json` 

teogeb:
Do not allow TypeScript to process JavaScript files. We don't typically have JS files in our `src` folders: the only exception is `exports-common.js` in `sdk`.

Changing this setting resolves a VS Code issue. The "problems" section in VS Code is usually full of errors like this:
```
Cannot write file ' ... /network/packages/sdk/dist/types/src/Authentication.d.ts' because it would overwrite input file.ts
```

There may be other ways to solve this issue.


teogeb:
Another approach fixes this issue: https://github.com/streamr-dev/network/pull/2766

PR #2770, opened 2024-09-30
refactor(sdk): Use `signer.getAddress()` in `Authentication`

teogeb:
Simplify `Authentication#getAddress()` by using a method from the `Signer` interface.

This is a cherry-pick from @harbu's https://github.com/streamr-dev/network/pull/2589.

### DIFF
--- a/packages/sdk/src/Authentication.ts
+++ b/packages/sdk/src/Authentication.ts
@@ -41,15 +41,7 @@ export const createAuthentication = (config: Pick<StrictStreamrClientConfig, 'au
         const signer = provider.getSigner()
         return {
             getAddress: pMemoize(async () => {
-                try {
-                    if (!('request' in ethereum && typeof ethereum.request === 'function')) {
-                        throw new Error(`invalid ethereum provider ${ethereum}`)
-                    }
-                    const accounts = await ethereum.request({ method: 'eth_requestAccounts' })
-                    return toEthereumAddress(accounts[0])
-                } catch {
-                    throw new Error('no addresses connected and selected in the custom authentication provider')
-                }
+                return toEthereumAddress(await (await signer).getAddress())
             }),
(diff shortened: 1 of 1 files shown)


harbu (approved):

harbu:
We might want to manually test this in browser before releasing

teogeb:
Tested in Chrome and in Firefox. Works fine in a typical case, where use unlocks Metamask when prompted to do so. If there is a pending unlock (i.e. user has been requested to unlock Metamask earlier, maybe on another page, but he closed the Metamask popup instead of doing the unlock), the promise rejects with this error:

```
Error: could not coalesce error (error={ "code": -32002, "payload": { "method": "eth_requestAccounts", "params": [  ] } }, payload={ "method": "eth_requestAccounts", "params": [  ] }, code=UNKNOWN_ERROR, version=6.13.1)
    at a (errors.js:129:21)
    at s.getRpcError (provider-jsonrpc.js:730:31)
    at s.getRpcError (provider-browser.js:70:22)
    at s.getSigner (provider-browser.js:98:28)
```    

Metamask logs this warning:
```
inpage.js:1 MetaMask - RPC Error: Already processing eth_requestAccounts. Please wait. 
```

The same error message was shown also with the earlier implementation, and therefore there is no regression. A front-end application can easily catch the error in instruct the user to open Metamask.

```
npm run build-browser-production
```

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="utf-8">
    <head>
        <script src="dist/streamr-sdk.web.min.js"></script>
        <script>
            const main = async () => {
                const client = new StreamrClient({
                    auth: {
                        ethereum: window.ethereum
                    }
                })
                try {
                    const address = await client.getAddress()
                    console.log(address)
                } catch (e) {
                    console.log(e)
                }
            }
            main()
        </script>
    </head>
</html>
<body>
</body>
```

PR #2774, opened 2024-10-01
feat(sdk): User id as `HexString`

teogeb:
## Summary

User id items are stored as arbitrary length `string` instead of `EthereumAddress`. 

Renamed `getAddess()` in `StreamrClient` to `getUserId()`. **This is a breaking change.**

This is a preparation PR for https://github.com/streamr-dev/network/pull/2780. This PR doesn't doesn't contain any changes to contract / The Graph interaction and therefore any user ids which are not exactly 20 bytes long don't work yet.  (We should merge this PR to `main` just before that PR, or implement the contract interactions in a PR which has this PR as a base branch.)

## Changes

The `UserID` helper type if now a a separate `BrandedString`. Additionally there is a new helper type `UserIDRaw`, and helper functions `toUserId` and `toUserIdRaw`. There is also the `HexString` helper type which is an alias to `string`.

In the API all user id inputs and outputs are `HexString` values. This is not a breaking change as earlier the inputs and outputs were either `string` or `EthereumAddress` values. (The `EthereumAddress` is a branded string, and it was used only in outputs, i.e. in the API all of these were just `string` values.)

In `Message`:
- field type of `publisherId`

In `StreamrClient`:
- return type of `getUserId()`,
- field type of `UserPermissionQuery#user`
  - it is parameter type in `hasPermission()`
- field type of `UserPermissionAssignment#user`
  - it is parameter/return type in `getPermissions()`, `grantPermissions()`, `revokePermissions()` and `setPermissions()`
- return type of `getStreamPublishers()` and `getStreamSubscribers()`
- parametrer type of `isStreamPublisher()`, `isStreamSubscriber()`
- field type of `SearchStreamsPermissionFilter#user`
  - it is a parameter type in `searchStreams()`
- changed type of `ResendFromOptions#publisherId` and `ResendRangeOptions#publisherId` 
  - these are parameter types in `resend()`
- parameter type of `addEncryptionKey()`

In `Stream`:
- corresponding changes in `hasPermission()`, `getPermissions()`, `grantPermissions()`, `revokePermissions()` and `setPermissions()`

### Internal

In `StreamMessage` the `publisherId` field type is `UserID` (i.e. the branded string) as we don't consider that class to be a part part of the public API. Also the related user id types in `MessageID`/`GroupKeyRequest`/`GroupKeyResponse` are `UserID` strings.

## Future improvements

- Change all `EthereumAddress` outputs to `HexString` outputs (e.g. `getStorageNodes()`) to unify the API 
- Could rename `user` field of `UserPermissionAssignment#user` and `SearchStreamsPermissionFilter#user` to `userId` to unify the API (we have typically the ID suffix in these, e.g. `streamId` and `publisherId`)
- Could refactor some test so that we use `randomUserId` instead of `fastWallet` / `ether`'s `Wallet`


packages/utils/src/UserID.ts
```diff
@@ -1,3 +1,28 @@
...
+import { BrandedString } from './types'
+
+const REGEX = /^0x[a-fA-F0-9]+$/
+
+export type UserID = BrandedString<'UserID'>
+export type UserIDRaw = Uint8Array
+
+export const toUserId = (input: string | UserIDRaw): UserID | never => {
+    if (input instanceof Uint8Array) {
+        return binaryToHex(input, true) as UserID
```
harbu, line 12: Is lower cased result guaranteed?

packages/sdk/src/encryption/PublisherKeyExchange.ts
```diff
@@ -67,10 +75,13 @@ export class PublisherKeyExchange {
         })
         eventEmitter.on('messagePublished', (msg) => {
             if (msg.signatureType === SignatureType.ERC_1271) {
-                const address = msg.getPublisherId()
-                if (!this.erc1271ContractAddresses.has(address)) {
-                    logger.debug('Add ERC-1271 publisher', { address })
-                    this.erc1271ContractAddresses.add(address)
+                const signerUserId = msg.getPublisherId()
```
harbu: This would not be the "recovered" signer but the erc-1271 contract. Maybe the naming here is a bit confusing?

packages/sdk/src/encryption/PublisherKeyExchange.ts
```diff
@@ -67,10 +75,13 @@ export class PublisherKeyExchange {
...
+                if (isEthereumAddressUserId(signerUserId)) {
```
harbu: Maybe worth a warning log if we receive an ERC-2171 signature message but the publisher is not a valid ethereum address?

packages/sdk/src/encryption/LitProtocolFacade.ts
```diff
@@ -59,7 +59,7 @@ const signAuthMessage = async (authentication: Authentication) => {
     const domain = 'dummy.com'
     const uri = 'https://dummy.com'
     const statement = 'dummy'
-    const addressInChecksumCase = ethers.getAddress(await authentication.getAddress())
+    const addressInChecksumCase = ethers.getAddress(toEthereumAddress(toUserId(await authentication.getUserId())))
```
harbu: Isn't `toUserId` unnecessary here?

packages/sdk/src/contracts/searchStreams.ts
```diff
@@ -22,7 +22,6 @@ export interface SearchStreamsOrderBy {
 
 export type SearchStreamsResultItem = {
     id: string
-    userAddress: UserID
```
harbu, line 24: Why removed? Should be replaced with something?

packages/node/jest.config.js
```diff
@@ -2,5 +2,8 @@ const rootConfig = require('../../jest.config')
 module.exports = {
     ...rootConfig,
     globalTeardown: './jest.teardown.js',
+    setupFilesAfterEnv: rootConfig.setupFilesAfterEnv.concat(
```
harbu: Is this here by accident?

packages/sdk/src/contracts/searchStreams.ts
teogeb: It is not needed as we'll eventually just return a collection of `Stream` objects.

Removed also this field from the GraphQL query.

packages/node/jest.config.js
teogeb: No, needed for `toEqualBinary` in `createMessagingPluginTest.test.ts`.

packages/sdk/src/encryption/PublisherKeyExchange.ts
teogeb: This is a message published by this client. We ensure the validity at `MessageFactory:142`.

fonty1:
@teogeb can we keep `uint8Array` at the protocol level and work in hexadecimal at the user/developer interaction level? I'm a little worried about the web interface and command line interactions using `[1,2,3,..]`? Is that a valid concern or am I misreading this update?

teogeb:
> @teogeb can we keep `uint8Array` at the protocol level and work in hexadecimal at the user/developer interaction level? I'm a little worried about the web interface and command line interactions using `[1,2,3,..]`? Is that a valid concern or am I misreading this update?

It is already `UInt8Array` on protocol level (i.e. in `protobuf`), and this PR affects how it is handled in `sdk`. 

Before this PR, user IDs were handled as hex-strings and converted to `EthereumAddress` branded strings. Now, we accept byte arrays as input and convert them to `UserID` branded strings, and then back to byte arrays on output.

Keeping hex-string input/output is still possible. If most users use string-based IDs, it might simplify the API.

teogeb:
I discussed this with @harbu , and we agreed that using a simple string in the public API is the easiest and most understandable approach for JavaScript developers.

We could create a `HexString` helper type as an alias for `string`. Internally, we'd convert `HexString` to a `UserID` branded string to validate user input. When we return the `UserID` branded string, we'd annotate the return type as `HexString`, so users don't have to worry about the internal details.

This likely means the `UserIDRaw` type is no longer necessary.

fonty1:
> I discussed this with @harbu , and we agreed that using a simple string in the public API is the easiest and most understandable approach for JavaScript developers.
> 
> We could create a `HexString` helper type as an alias for `string`. Internally, we'd convert `HexString` to a `UserID` branded string to validate user input. When we return the `UserID` branded string, we'd annotate the return type as `HexString`, so users don't have to worry about the internal details.
> 
> This likely means the `UserIDRaw` type is no longer necessary.

SGTM

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ Changes before Tatum release are not documented in this file.
 
 #### Changed
 
+- **BREAKING CHANGE:** Rename `getAddress()` to `getUserId()` (https://github.com/streamr-dev/network/pull/2774)
```
```suggestion
- Rename `getAddress()` to `getUserId()` (https://github.com/streamr-dev/network/pull/2774)
```

packages/sdk/src/StreamIDBuilder.ts
```diff
@@ -48,7 +49,7 @@ export class StreamIDBuilder {
     async toStreamID(streamIdOrPath: string): Promise<StreamID> {
         let address: EthereumAddress | undefined
         if (StreamIDUtils.isPathOnlyFormat(streamIdOrPath)) {
-            address = await this.authentication.getAddress()
+            address = toEthereumAddress(await this.authentication.getUserId())
```
harbu, line 52: Are we still going with the idea that interface `Authentication` would have method `getEthereumAddress`? Would it be better to use that here?

packages/sdk/src/StreamrClient.ts
```diff
@@ -462,14 +466,14 @@ export class StreamrClient {
      * Grants permissions on a given stream.
      */
     grantPermissions(streamIdOrPath: string, ...assignments: PermissionAssignment[]): Promise<void> {
-        return this.streamRegistry.grantPermissions(streamIdOrPath, ...assignments)
+        return this.streamRegistry.grantPermissions(streamIdOrPath, ...assignments.map((a) => toInternalPermissionAssignment(a)))
```
harbu: Stylistic choice, you can just pass function directly
```suggestion
        return this.streamRegistry.grantPermissions(streamIdOrPath, ...assignments.map(toInternalPermissionAssignment))
```

packages/sdk/src/encryption/PublisherKeyExchange.ts
```diff
@@ -67,10 +75,13 @@ export class PublisherKeyExchange {
...
         eventEmitter.on('messagePublished', (msg) => {
             if (msg.signatureType === SignatureType.ERC_1271) {
-                const address = msg.getPublisherId()
-                if (!this.erc1271ContractAddresses.has(address)) {
-                    logger.debug('Add ERC-1271 publisher', { address })
-                    this.erc1271ContractAddresses.add(address)
+                const publisherId = msg.getPublisherId()
+                if (isEthereumAddressUserId(publisherId)) {
+                    const address = toEthereumAddress(publisherId)
+                    if (!this.erc1271ContractAddresses.has(address)) {
```
harbu: Could this be problematic from performance viewpoint given that this event is emitted every time a message is published?

packages/sdk/src/StreamIDBuilder.ts
teogeb: I think that it is simpler to have only one method, and convert that to other formats as needed.

If two return types are absolutely needed, then I'd have `getUserId()` and `getUserIdAsEthereumAddress()`. But to me this current approach is more readable. There are only 6 places where this kind of conversion is needed (half of which are using `StreamrClient`, others using `Authentication` directly)

packages/sdk/src/StreamrClient.ts
teogeb: Fixed

packages/sdk/src/encryption/PublisherKeyExchange.ts
teogeb: Simplified the logic to have less conversions in commit 5855736dc.

packages/sdk/src/Authentication.ts
```diff
@@ -40,8 +39,8 @@ export const createAuthentication = (config: Pick<StrictStreamrClientConfig, 'au
         const provider = new BrowserProvider(ethereum)
         const signer = provider.getSigner()
         return {
-            getAddress: pMemoize(async () => {
-                return toEthereumAddress(await (await signer).getAddress())
+            getUserId: pMemoize(async () => {
```
jtakalai, line 42: why not leave the getAddress though? It could return same as getUserId, or `toEthereumAddress` or something, maybe throwing if the user id is NOT an Ethereum address.

packages/sdk/src/StreamIDBuilder.ts
jtakalai: why not leave the old method though? Does "authentication" even mean anything for non-address ids? If you can sign, then your ID is an address, currently.

packages/sdk/src/contracts/ERC1271ContractFacade.ts
```diff
@@ -36,7 +36,7 @@ export class ERC1271ContractFacade {
     }
 
     async isValidSignature(contractAddress: EthereumAddress, payload: Uint8Array, signature: Uint8Array): Promise<boolean> {
-        const recoveredSignerUserId = toEthereumAddress(binaryToHex(recoverSignerUserId(signature, payload), true))
+        const recoveredSignerUserId = toUserId(recoverSignerUserId(signature, payload))
```
jtakalai, line 39: I imagine that according to https://eips.ethereum.org/EIPS/eip-1271 the "recovered id" will be an address, so why change this?

jtakalai (approved):
I'm not going to *demand* changes in this repo, but I don't really see what the point of breaking/removing the old function is, especially in cases where (currently) we only support authentication only with Ethereum keys (that will have a corresponding address). It doesn't feel like too much of a concession to leave the existing getAddress etc. methods that return the ID (if it's an address) or throw (if it's not). This wouldn't require so many changes from users, esp those who expect to be using Ethereum addresses only.

packages/sdk/src/contracts/ERC1271ContractFacade.ts
teogeb: We don't query the recovered address from the contract but from our own `recoverSignerUserId` utility from `@streamr/utils`. It returns  `UserIDRaw`.

The `toUserId` call is used to convert the input from `UserIDRaw` (i.e. `Uint8Array`) to `UserID` (i.e. `string`). The fact that user ids are Ethereum addresses in practice is an implementation detail that this facade doesn't need to care about.

teogeb: Removed the `getAddress()` method so that we don't have two different names for the same thing. 

Typically the method is used like this:
```ts
await stream.grantPermissions({
    permissions: [StreamPermission.PUBLISH],
    userId: await publisher.getUserId()
})
```

-> I.e. when there is a `userId` argument in the API, the user should put something like `client.getUserId()` to it. If the method name would be different, it would not be trivial:
```ts
await stream.grantPermissions({
    permissions: [StreamPermission.PUBLISH],
    userId: await publisher.getAddress()
})
```

teogeb:
> I'm not going to _demand_ changes in this repo, but I don't really see what the point of breaking/removing the old function is, especially in cases where (currently) we only support authentication only with Ethereum keys (that will have a corresponding address). It doesn't feel like too much of a concession to leave the existing getAddress etc. methods that return the ID (if it's an address) or throw (if it's not). This wouldn't require so many changes from users, esp those who expect to be using Ethereum addresses only.

Re-added the `getAddress()` as you suggested. We could maybe remove/deprecate it later if we want to simplify the API (o need to have have two different names for the same thing). But as there are no other authentication methods yet, it makes sense to keep the alias for backwards compatibility.

The `getUserId()` is the preferred method to access user's id so that we'll have this kind of straighforward API:
```ts
await stream.grantPermissions({
    permissions: [StreamPermission.PUBLISH],
    userId: await publisher.getUserId()
})
```

PR #2780, opened 2024-10-02
feat(sdk): [NET-1316] Use `StreamRegistry` v5 permission methods

teogeb:
**NOTE: Do not make releases until we've deployed `StreamRegistry` v5 to production.**

## Summary

Updated the sdk to use `StreamRegistry` v5 instead of v4. In v5 there are `*forUserId`-permission methods which support arbitrary length user ids.

The API is same for all permission types, but longer/shorter IDs is can be used only for publish and subscribe permissions. as there is no contract-level support for other types (edit, grant, delete).

## Changes

Replaced usage:
- `hasPermission`, `hasDirectPermission` -> `getPermissionsForUserId`, `getDirectPermissionsForUserId`
  - i.e. read a permissions struct instead of reading a boolean
  - it still is ok to use `hasPublicPermission` contract call to check public access, no changes in that logic
- `grantPermission` -> `grantPermissionForUserId`
- `revokePermission` -> `revokePermissionForUserId`
- `setPermissionsMultipleStreams` -> `setMultipleStreamPermissionsForUserIds`

Also changed  The Graph queries to use `userId` instead of `userAddress`.

## Refactoring

Refactoring the internal call flow in `StreamRegistry` by introducing a helper method `isStreamPublisherOrSubscriber_nonCached`.

## Tests

Changed `randomUserId` to return random-length user ids.


packages/sdk/src/contracts/StreamRegistry.ts
```diff
@@ -405,14 +407,14 @@ export class StreamRegistry {
             * empty assignments cleanup in The Graph.
             */
             if (permissions.length > 0) {
-                if (permissionResult.userAddress === PUBLIC_PERMISSION_ADDRESS) {
+                if (permissionResult.userId === PUBLIC_PERMISSION_ADDRESS) {
```
jtakalai: should that be called PUBLIC_PERMISSION_USER_ID?

jtakalai (approved):

packages/sdk/src/contracts/StreamRegistry.ts
teogeb: Great catch! Fixed

PR #2789, opened 2024-10-04
refactor(dht): Unify naming of `DhtAddress` converter functions

teogeb:
Renamed:
- `getDhtAddressFromRaw` -> `toDhtAddress`
- `getRawFromDhtAddress` -> `toDhtAddressRaw`
- `getNodeIdFromPeerDescriptor` -> `toNodeId`
- `createRandomDhtAddress` -> `randomDhtAddress`

This way we adapt the same naming style, which is already used in the utils package `toEthereumAddress()`. We'll soon add also more functions with that naming style: `toUserId()` and `toUserIdRaw()` (https://github.com/streamr-dev/network/pull/2774).

This PR doesn't contain functionality changes, only the four renames mentioned above.

### DIFF
--- a/packages/dht/src/identifiers.ts
+++ b/packages/dht/src/identifiers.ts
@@ -8,22 +8,22 @@ export const KADEMLIA_ID_LENGTH_IN_BYTES = 20
 export type DhtAddress = BrandedString<'DhtAddress'>
 export type DhtAddressRaw = Uint8Array
 
-export const getDhtAddressFromRaw = (raw: DhtAddressRaw): DhtAddress => {
+export const toDhtAddress = (raw: DhtAddressRaw): DhtAddress => {
     return binaryToHex(raw) as unknown as DhtAddress
 }
 
-export const getRawFromDhtAddress = (address: DhtAddress): DhtAddressRaw => {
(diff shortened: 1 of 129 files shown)


packages/dht/src/identifiers.ts
```diff
@@ -8,22 +8,22 @@ export const KADEMLIA_ID_LENGTH_IN_BYTES = 20
...
 }
 
-export const getRawFromDhtAddress = (address: DhtAddress): DhtAddressRaw => {
+export const toDhtAddressRaw = (address: DhtAddress): DhtAddressRaw => {
     return hexToBinary(address) as unknown as DhtAddressRaw
 }
 
-export const getNodeIdFromPeerDescriptor = (peerDescriptor: PeerDescriptor): DhtAddress => {
-    return getDhtAddressFromRaw(peerDescriptor.nodeId)
+export const toNodeId = (peerDescriptor: PeerDescriptor): DhtAddress => {
```
juslesan, line 19: feels like the conversion naming convention ("to") doesn't really fit here since the nodeId cannot be converted back to a PeerDescriptor. It feels more like a getter.

juslesan (approved):

PR #2790, opened 2024-10-04
feat(network): Improve `StreamPartitionInfo` compatibility

teogeb:
Improve backwards compatibility for `StreamPartitionInfo`. The format for `contentDeliveryLayerNeighbors` field was changed in https://github.com/streamr-dev/network/pull/2738.

## Open questions

- Should we populate real data to the `deprecatedContentDeliveryLayerNeighbors` field to achieve better compatibility? With the current approach caller checks the version number and decides which field to read.


juslesan (approved):
Could populate the deprecated field aswell, but as far as I know the crawler and a debugging script used to check if a peer is online are the only use cases for the endpoint at the moment.

PR #2801, opened 2024-10-10
feat(network): NET-1346 avoid routing through proxy connections

juslesan:
## Summary

Added new functionality in the connection manager to set the manager in private mode. In private mode the node will inform all peers that the connection is private. This way the peers will know not to use the connections during routing. 

This fixes a known issue with routing where routing is attempted through pure proxy client nodes.

Allowing private connections has to be explicitly allowed. This is required to not allow free riding. Private connections are allowed if the `acceptProxyConnections` flag is set true in the SDK. There is currently no need to have an option to only set the `acceptIncomingPrivateConnections` as true since the feature is only used for proxies.

### DIFF
--- a/packages/dht/src/connection/ConnectionManager.ts
+++ b/packages/dht/src/connection/ConnectionManager.ts
@@ -191,7 +191,8 @@ export class ConnectionManager extends EventEmitter<TransportEvents> implements 
         this.endpoints.forEach((endpoint) => {
             if (endpoint.connected) {
                 const connection = endpoint.connection
-                if (!this.locks.isLocked(connection.getNodeId()) && Date.now() - connection.getLastUsedTimestamp() > maxIdleTime) {
+                const nodeId = connection.getNodeId()
+                if (!this.locks.isLocked(nodeId) && !this.locks.isPrivate(nodeId) && Date.now() - connection.getLastUsedTimestamp() > maxIdleTime) {
                     logger.trace('disconnecting in timeout interval: ' + getNodeIdOrUnknownFromPeerDescriptor(connection.getPeerDescriptor()))
                     disconnectionCandidates.addContact(connection)
                 }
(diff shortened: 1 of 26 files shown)


packages/dht/src/connection/ConnectionManager.ts
```diff
@@ -31,6 +32,7 @@ import { PendingConnection } from './PendingConnection'
 export interface ConnectionManagerOptions {
     maxConnections?: number
     metricsContext: MetricsContext
+    allowPrivateConnections: boolean
```
ptesavol: Maybe this could be renamed as "allowIncomingPrivateConnections" to clarify that this is a server-side setting? This could also be given default value of false in order to avoid always passing this option when creating a ConnectionManager.

packages/dht/src/connection/ConnectionManager.ts
```diff
@@ -127,6 +129,7 @@ export class ConnectionManager extends EventEmitter<TransportEvents> implements
     private rpcCommunicator?: RoutingRpcCommunicator
     private disconnectorIntervalRef?: NodeJS.Timeout
     private state = ConnectionManagerState.IDLE
+    private privateMode = false
```
ptesavol: Maybe this could be renamed to "privateClientMode" to clarify that this is client-side setting? Also all the related methods should be renamed. A comment could be added to explain what this mode actually means.

ptesavol (changes requested):
The distinction between client/server settings side could be improved by naming. Looks good otherwise.

teogeb:
> Private connections are allowed if the acceptProxyConnections flag is set true in the SDK.

Are private connections a proxy-connection specific concept? As this is an important new network level concept, the PR description could maybe elaborate on that topic bit more.

ptesavol (approved):

PR #2802, opened 2024-10-11
feat(DHT): NET-1358: time out webrtc connections early if signalling fails

juslesan:
## Summary

Close webrtc connections if remote description does not arrive within 5 seconds. This is an improvement for some worst case time-to-data spikes.

If the description does not arrive it indicates that the other end is not online or routing has failed and a new connection should be eventually attempted.


ptesavol (approved):
LGTM

PR #2804, opened 2024-10-14
test: Random user ids

teogeb:
Use `randomUserId()` instead of random / hardcoded Ethereum addresses.

Also use relative path in `createTestStream()` test utility.


harbu (approved):
lgtm

PR #2809, opened 2024-10-14
refactor(node): Use `UserID` in `Storage`

teogeb:
Use `UserID` in `Storage`-related functionality. Also added query parameter validation for `dataQueryEnpoint` queries.


packages/utils/src/UserID.ts
```diff
@@ -2,6 +2,12 @@ import { EthereumAddress, toEthereumAddress } from './EthereumAddress'
 
 export type UserID = EthereumAddress
 
-export const toUserId = (value: string): UserID => {
-    return toEthereumAddress(value)
+const REGEX = /^0x[a-fA-F0-9]+$/
+
+export const toUserId = (input: string): UserID => {
+    return toEthereumAddress(input)
```
harbu, line 8: toEthereumAddress intentional? Then userId has to be specific length?

teogeb: This is pre-https://github.com/streamr-dev/network/pull/2774 era functionality, that's why we use `EthereumAddress`.

harbu (approved):

PR #2812, opened 2024-10-15
feat(utils): `merge()` does deep merge

teogeb:
Changed the `merge()` utility function to perform a deep merge instead of a flat merge. In most cases, it doesn't matter whether we use flat or deep merging, so the flat behavior may not have been explicitly required. When merging objects, we could implicitly assume that the merge is deep. So maybe this is better this way?

This is part of a preparation for NET-1363, where a deep merge is needed.

Also simplified `applyEnvironmentDefaults()`.

## Usage in Stream#update

The `Streamr#update` is one of the callers of `merge()`. The  API documentation says it "merges metadata with the existing metadata.". It is not clear whether that means a deep or a flat merge?

We'll harmonize the `Stream#update` method soonish so that it matches the functionality of `StreamrClient#updateStream` (NET-1364). That's why we don't change the behavior in this PR. If we'd keep the merging functionality, it should most likely do deep merge.

### DIFF
--- a/packages/utils/src/merge.ts
+++ b/packages/utils/src/merge.ts
@@ -2,7 +2,7 @@ import { isArray, mergeWith } from 'lodash'
 
 /*
  * Does deep merge. This is similar to `lodash` merge, but handles arrays differently:
- * `lodash` merges elements of arrays by their indices, this overrites the existing
+ * `lodash` merges elements of arrays by their indices, this overwrites the existing
  * value with the array
  */
 export const merge = <TTarget>(...sources: (Partial<TTarget> | undefined)[]): TTarget => {
--- a/packages/utils/src/merge.ts
+++ b/packages/utils/src/merge.ts
@@ -9,7 +9,7 @@ export const merge = <TTarget>(...sources: (Partial<TTarget> | undefined)[]): TT
     const result: Record<string, unknown> = {}
     mergeWith(result, ...sources, (_: any, srcValue: any) => {
         if (isArray(srcValue)) {
-            return srcValue
+            return [...srcValue]
         } else {
             return undefined  // no customization: does the default merging for this field
         }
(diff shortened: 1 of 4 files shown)


packages/utils/src/merge.ts
```diff
@@ -1,13 +1,18 @@
+import { isArray, mergeWith } from 'lodash'
+
+/*
+ * Does deep merge. This is similar to `lodash` merge, but handles arrays differently:
+ * `lodash` merges elements of arrays by their indices, this overrites the existing
```
harbu:
```suggestion
 * `lodash` merges elements of arrays by their indices, this overwrites the existing
```

packages/utils/test/merge.test.ts
```diff
@@ -101,4 +115,29 @@ describe('merge', () => {
...
+        // eslint-disable-next-line @typescript-eslint/no-extraneous-class
+        class Foo {}
+        const o1 = {
+            foo: 1
+        }
+        const foo = new Foo()
+        const o2 = {
+            foo
+        }
+        expect(merge(o1, o2).foo).toBe(foo)
```
harbu: I would also check that the values within Foo haven't been affected.
```suggestion
        class Foo {
            values: Record<string, unknown>
        }
        const foo1 = new Foo()
        foo1.values = {
            x: 5,
            y: 6
        }
        const foo2 = new Foo()
        foo2.values = {
            y: 7
            z: 8
        }
        const o1 = {
            foo: foo1
        }
        const o2 = {
            foo: foo2
        }
        const result = merge(o1, o2)
        expect(result.foo).toBe(foo1)
        expect(result.foo.values).toEqual({
            x: 5,
            y: 6
        })
```

teogeb: Good idea. The merging goes other way around: `foo2` should be the value, and values therefore `{ 7, 8 }`

harbu (approved):

PR #2817, opened 2024-10-15
feat(cli-tools)!: [NET-1363] Add support for arbitrary `--env` 

teogeb:
**This is a breaking change**

Added new `--env <id>` flag. It defines the client environment by applying `{ environment: <id> }` setting in the client config. Therefore all current client environments are now supported by `cli-tools`: `polygon`, `polygonAmoy` and `dev2`. 

Removed support for the `--dev` flag. It is same as `--env dev2`.

## Example

```
streamr storage-node list --env dev2
```


packages/sdk/src/Config.ts
```diff
@@ -228,7 +228,9 @@ export interface EthereumNetworkConfig {
 //   in Ethereum network
 export type EnvironmentId = 'polygon' | 'polygonAmoy' | 'dev2'
 
-export const DEFAULT_ENVIRONMENT: EnvironmentId = 'polygon'
+export const ENVIRONMENT_IDS: EnvironmentId[] = ['polygon', 'polygonAmoy', 'dev2']
+
+export const DEFAULT_ENVIRONMENT_ID: EnvironmentId = 'polygon'
```
harbu, line 233: Is it so that any value added here needs to be added ton `config.schema.json` and vice versa?

harbu (approved):

packages/sdk/src/Config.ts
teogeb: Yes

PR #2818, opened 2024-10-16
feat(network): add proxied nodes to info endpoint

juslesan:
## Summary

Added proxied nodes to the to info endpoint


packages/trackerless-network/src/proto/packages/dht/protos/DhtRpc.client.ts
```diff
@@ -336,6 +337,10 @@ export interface IConnectionLockRpcClient {
      * @generated from protobuf rpc: gracefulDisconnect(dht.DisconnectNotice) returns (google.protobuf.Empty);
      */
     gracefulDisconnect(input: DisconnectNotice, options?: RpcOptions): UnaryCall<DisconnectNotice, Empty>;
+    /**
```
teogeb: These changes aren't part of this PRs? Changed in https://github.com/streamr-dev/network/pull/2801 and https://github.com/streamr-dev/network/pull/2815

packages/dht/src/connection/simulator/SimulatorTransport.ts
```diff
@@ -5,11 +5,11 @@ import { Simulator } from './Simulator'
 import { SimulatorConnectorFacade } from '../ConnectorFacade'
 
 export class SimulatorTransport extends ConnectionManager {
-    constructor(localPeerDescriptor: PeerDescriptor, simulator: Simulator) {
+    constructor(localPeerDescriptor: PeerDescriptor, simulator: Simulator, allowIncomingPrivateConnections = false) {
         super({
             createConnectorFacade: () => new SimulatorConnectorFacade(localPeerDescriptor, simulator),
             metricsContext: new MetricsContext(),
-            allowIncomingPrivateConnections: false
+            allowIncomingPrivateConnections
```
teogeb, line 12: Are private connections and proxy connections the same thing? (Do we plan to use private connections to something else, too?)

E.g. we have this comment in `ConnectionManager`:
```
    // Used to filter proxy connections from the connections view
    private remotePrivateConnections: Set<DhtAddress> = new Set()
```

teogeb (changes requested):

packages/dht/src/connection/simulator/SimulatorTransport.ts
juslesan: Proxy connections is a layer2 concept so having a wider conecpt in the lower layer makes sense

packages/trackerless-network/src/proto/packages/dht/protos/DhtRpc.client.ts
juslesan: These are unused generated code from the DHT in the trackerless-network

juslesan: So the it's not specific to this PR

packages/dht/src/connection/simulator/SimulatorTransport.ts
juslesan: Also Not all proxy-connections are private connections. Nodes only use privateClientMode when they only have proxy client connections. If there are any streams where the node participates in normally private mode is disabled

teogeb: > Proxy connections is a layer2 concept so having a wider conecpt in the lower layer makes sense

Ah ok, wasn't aware of this fact 👍

We could modify/remove this "Used to filter proxy connections from the connections view" comment as it is bit misleading.

teogeb: Also, not about this PR, but why the `remotePrivateConnections` has the `remote`-prefix? Is the whole connection private, i.e. both the local and remote see treat it specially? (If that's the case, we could remove the remote-prefix. If it is private only on the remote side, we could rename the accessor to match the field name.)

juslesan: There isn't a real need to maintain the state of the locally private connections as the entire connection manager is in private mode if they exist at all. The remote-prefix is kept to make it clear that they are set by someone else

packages/trackerless-network/src/proto/packages/dht/protos/DhtRpc.client.ts
teogeb: In that case we could revert these changes in this PR.

packages/dht/src/connection/simulator/SimulatorTransport.ts
juslesan: Could remove the comment yeah

PR #2820, opened 2024-10-17
refactor(DHT: NET-1366 move PeerDescriptor out of DhtRpc, no longer import DhtRpc to higher layers

juslesan:
## Summary

Move PeerDescriptor out of DhtRpc to reduce the amount of generated code in higher level packages. DhtRpc is no longer imported in the higher layer packages. Instead only the PeerDescriptor is imported. This means that any imports pointing to the DhtRpc had to be changed

PR is large due the number of generated files being deleted. See commits to see what changes in each package.


packages/trackerless-network/generated/packages/dht/protos/PeerDescriptor.ts
```diff
@@ -0,0 +1,111 @@
+// @generated by protobuf-ts 2.9.4 with parameter server_generic,generate_dependencies,long_type_number
+// @generated from protobuf file "packages/dht/protos/PeerDescriptor.proto" (package "peerDescriptor", syntax proto3)
+// tslint:disable
+import { MessageType } from "@protobuf-ts/runtime";
+/**
+ * @generated from protobuf message peerDescriptor.PeerDescriptor
+ */
+export interface PeerDescriptor {
```
teogeb, line 8: We generate the same file to three places:
- `packages/dht/generated/packages/dht/protos/PeerDescriptor.ts`
- `packages/sdk/src/generated/packages/dht/protos/PeerDescriptor.ts`
- `packages/trackerless-network/generated/packages/dht/protos/PeerDescriptor.ts`

Would it be possible to generate the file only to the dht package?  The class is in `dht`'s `exports.ts` and others packages should use it that way?

packages/trackerless-network/test/unit/ExternalNetworkRpc.test.ts
```diff
@@ -23,7 +23,7 @@ describe('ExternalNetworkRpc', () => {
     })
 
     it('registers method', async () => {
-        rpc.registerRpcMethod(HandshakeRequest, HandshakeResponse, 'handshake', () => fn())
+        rpc.registerRpcMethod(StreamPartHandshakeRequest, StreamPartHandshakeResponse, 'handshake', () => fn())
```
teogeb, line 26: Is this change intentional? Maybe it is, if the point here is that we shouldn't register `dht's` methods in `trackerless-network` package?

Could mention this change in the PR description (or extract this change to a separate PR).

packages/trackerless-network/generated/packages/dht/protos/PeerDescriptor.ts
juslesan: The problem is that we need to do define the PeerDescriptor to the protobuf messages in the .proto files. This has the side effect of generating the files.

packages/trackerless-network/test/unit/ExternalNetworkRpc.test.ts
juslesan: The DhtRpc is no longer imported to the .proto file in trackerless. Therefore, the test needed new foobar messages to use

packages/trackerless-network/generated/packages/dht/protos/PeerDescriptor.ts
teogeb: Maybe easiest to generate, but then delete file:
```sh
sed -i '' 's|import { PeerDescriptor } from "../../dht/protos/PeerDescriptor";|import { PeerDescriptor } from "@streamr/dht";|' generated/packages/trackerless-network/protos/NetworkRpc.ts
rm generated/packages/dht/protos/PeerDescriptor.ts
```

Adding these two lines to `proto.sh` solves the issue.

packages/trackerless-network/test/unit/ExternalNetworkRpc.test.ts
teogeb: Good to update to PR description. This big PR is just about moving `PeerDescriptor`, and it is not trivial to know all prerequisites/consequences related to that.

packages/trackerless-network/generated/packages/dht/protos/PeerDescriptor.ts
juslesan: Can't build the generated NetworkRpc files if PeerDescriptor is removed from the proto build

teogeb: What error message you get?

For me e.g. this runs all ok:
```
cd packages/trackerless-network
./proto.sh
npm run build
```

(did `npm run clean && npm run bootstrap` before this, so should the workspace should be clean)

Could that error be caused by other changes? When I DIDN'T include this two line snippet, and deleted the `generated` directory in `trackerless-network,` I got some errors like this when doing `npm run build`:
```
Cannot find module '../generated/packages/trackerless-network/protos/NetworkRpc' or its corresponding type declarations.
8 import { ProxyDirection, StreamMessage } from '../generated/packages/trackerless-network/protos/NetworkRpc'
```

juslesan: I mean I suppose we could just stab the imports but it seems too hacky to do so. Especially since we persist the generated files only to get around a known issue in the ts-protobuf library (https://github.com/timostamm/protobuf-ts/issues/270).

juslesan: I'll try to fix this completely in https://linear.app/streamr/issue/NET-1371/fix-problem-with-building-protobuf-files-in-linux-to-get-rid-of

teogeb (approved):

PR #2822, opened 2024-10-18
chore(trackerless-network): Export `NodeInfo` types

teogeb:
Export these types which are field types in the `NodeInfo` structure:
- `StreamPartitionInfo`
- `ContentDeliveryLayerNeighborInfo`
- `ControlLayerInfo`

These are the type-safe versions of generated proto-rpc interfaces. The type-safety means that some fields are marked as required. These are required fields in practice, but marked as optional because of proto-rpc. 

## Future improvements

- Could do the same thing also to other interfaces which are part of `trackerless-network`'s public API:


packages/trackerless-network/package.json
```diff
@@ -45,7 +46,6 @@
     "@types/uuid": "^10.0.0",
     "@types/yallist": "^4.0.1",
     "expect": "^29.6.2",
-    "ts-essentials": "^10.0.2",
```
juslesan, line 48: Is this change necessary?

teogeb: With the current approach yes, as we use it here:
```
export type ContentDeliveryLayerNeighborInfo = MarkRequired<ContentDeliveryLayerNeighborInfo_, 'peerDescriptor'>
```

In the `sdk` we use it similarly, and there it is a dependency, too.

juslesan (approved):

PR #2825, opened 2024-10-18
feat(sdk)!: `StreamMetadata#partitions` is nullable

teogeb:
**This is a breaking change** as this changes the API

Changed `StreamMetadata` field type: the `partitions` field now nullable.

We can't ensure that the data contains some values (or even that metadata is a JSON). Therefore it is better that the type system doesn't assert anything about the content of the metadata.

Also added `Stream#getPartititionCount()` method. It reads the info from the metadata and defaults to 1 if there is no data available. It also validates that the partition count is valid (between 1 and 100).

Also added `DEFAULT_PARTITITION_COUNT` constant to `@streamr/utils`.

## Future improvements

Maybe the best type for metadata would be just `Record<string, any>`?
- also maybe the `Stream` constructor and `Stream#parseMetadata` shouldn't inject any default value for the `partitions` field

### DIFF
--- a/packages/sdk/src/publish/MessageFactory.ts
+++ b/packages/sdk/src/publish/MessageFactory.ts
@@ -1,4 +1,4 @@
-import { DEFAULT_PARTITION_COUNT, StreamID, UserID, keyToArrayIndex, toEthereumAddress, utf8ToBinary } from '@streamr/utils'
+import { StreamID, UserID, keyToArrayIndex, toEthereumAddress, utf8ToBinary } from '@streamr/utils'
 import random from 'lodash/random'
 import { Authentication } from '../Authentication'
 import { StreamrClientError } from '../StreamrClientError'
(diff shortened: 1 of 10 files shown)


packages/sdk/src/publish/MessageFactory.ts
```diff
@@ -68,7 +68,7 @@ export class MessageFactory {
             throw new StreamrClientError(`You don't have permission to publish to this stream. Using address: ${publisherId}`, 'MISSING_PERMISSION')
         }
 
-        const partitionCount = (await this.streamRegistry.getStream(this.streamId)).getMetadata().partitions
+        const partitionCount = (await this.streamRegistry.getStream(this.streamId)).getMetadata().partitions ?? DEFAULT_PARTITION_COUNT
```
harbu: It feels a bit off that this code needs to know about default partition count. Should we add a getter to `Stream` for handling this? `Stream#getPartitionCount()`

harbu (approved):

PR #2826, opened 2024-10-18
feat(sdk)!: `Stream#update()` doesn't merge

teogeb:
**This is a breaking change**

The `Stream#update()` doesn't merge the metadata with the existing data. It simply overwrites the existing data with the new data.

Now `Stream#update()` and `StreamrClient#updateStream()` have almost the same functionality.

## Future improvements

Maybe the two methods should have exactly the same functions: i.e. we'd clear the cache also in `StreamrClient#updateStream()`.


harbu (approved):

packages/sdk/src/Stream.ts
```diff
@@ -151,19 +137,15 @@ export class Stream {
     }
 
     /**
-     * Updates the metadata of the stream by merging with the existing metadata.
+     * Updates the metadata of a stream.
      */
     async update(metadata: Partial<StreamMetadata>): Promise<void> {
```
harbu, line 156: Maybe add temporary merging behavior to `detectFields` (until we figure out can we remove that method or not)

packages/sdk/src/Stream.ts
```diff
@@ -208,11 +222,12 @@ export class Stream {
         }).filter(Boolean) as Field[] // see https://github.com/microsoft/TypeScript/issues/30621
 
         // Save field config back to the stream
-        await this.update({
+        const merged = flatMerge(this.getMetadata(), this.getMetadata(), {
```
harbu: Two times `this.getMetadata()`?

harbu (approved):

PR #2830, opened 2024-10-21
chore(node)!: Remove deprecated config wizard wrappers

teogeb:
## Summary

There was three ways to start the configuration wizard:
- `bin/config-wizard.js`
- `streamr-broker-init`
- `streamr-node-init`

The `streamr-node-init` command is the currently supported way to do it, and others were deprecated versions of the same task. Removed two other commands/scripts.

## Changes

Modified the Docker file to add symbolic links to `bin` directory so that we can reference the `streamr-node-init` from the command line:
```
docker run -it -v ${pwd}:/home/streamr/.streamr streamr/node bin/streamr-node-init
```

Removed the`bin/config-wizard.js` file from the `bin` directory.  That was primarily there to support the easy command line usage, which was like this:
```
docker run -it -v ${pwd}:/home/streamr/.streamr streamr/node bin/config-wizard
```

(That file not the compiled version of `config-wizard.ts`, but an explicitly added file which started the compiled version of `config-wizard.ts` from the `dist/bin` directory. The `dist/bin/config-wizard.js` file then just started the`streamr-node-init` command internally.)

Removed `streamr-broker-init` bin script command from `package.json`. When Streamr Node is installed via NPM, the `streamr-node-init` command should be used instead of the deprecated `streamr-broker-init`.

## Docs

Updated the Docker usage. The NPM use case was already up-to-date.


fonty1 (approved):
LGTM. Please "Run Workflow" here,
https://github.com/streamr-dev/network/actions/workflows/deploy-docs.yml to update the docs, once this has been merged to Main.

harbu (approved):

CHANGELOG.md
```diff
@@ -57,6 +57,9 @@ Changes before Tatum release are not documented in this file.
 
 #### Removed
 
+- Remove deprecated `bin/config-wizard` script (i.e. the `streamr-broker-init` command)
```
harbu: Breaking change infront?

PR #2848, opened 2024-10-25
chore: Lint browser implementations

teogeb:
## Summary

Do not not exclude browser implementations from `eslint` and `check` tasks. Removed `Browser*` files from excludes list of `tsconfig.jest.json` files.

## Other changes

Converted the `BrowserPersistence` file to a ES module. Before this change we got this error:
```
src/utils/persistence/BrowserPersistence.ts:1:38 - error TS1479: The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("idb")' call instead.
  To convert this file to an ECMAScript module, change its file extension to '.mts', or add the field `"type": "module"` to ...package.json'.
import { openDB, IDBPDatabase } from 'idb'
```

Alternatively could have tweaked the import? It was not clear what the root cause was, as the idb package seems to support CommonJS: https://github.com/jakearchibald/idb/blob/eb2fc14bb3588d09aaa5e86a83bf3519b06e10b3/package.json#L13

Added DOM types for `BrowserWebrtcConnection`.

## Test

There is no regression for the  browser persistence functionality at least with recent Chrome and Firefox versions. We can see from Chrome/Firefox dev tools that this inserts the key to Indexed DB.

```html
<!DOCTYPE html>
<html>
<body>
    <script src="dist/streamr-sdk.web.js"></script>
    <script>
        const client = new StreamrClient()
        client.addEncryptionKey(StreamrClient.EncryptionKey.generate(), '0x1234567890123456789012345678901234567890')
    </script>
</body>
</html>
```


packages/sdk/src/utils/persistence/BrowserPersistence.mts
```diff
@@ -1,5 +1,6 @@
 import { openDB, IDBPDatabase } from 'idb'
-import { PersistenceContext, PersistenceContextOptions } from './PersistenceContext'
+// eslint-disable-next-line import/no-unresolved
+import { PersistenceContext, PersistenceContextOptions } from './PersistenceContext.js'
 
 export default class BrowserPersistence implements PersistenceContext {
```
harbu, line 20: Add description why this is .mts file.

juslesan (approved):

PR #2868, opened 2024-11-11
refactor(utils): Rename `waitForCondition()`

teogeb:
Renamed to `until`.

There is a similar function with the same name already in `sdk`.  We'll remove that function in a follow-up PR.

Also removed obsolete `runAndWaitForCondition` utility function from `test-utils`.


packages/test-utils/src/index.ts
```diff
@@ -102,7 +102,7 @@ export const runAndRaceEvents = async (
  * conditions evaluate to true on a retry attempt within timeout. If timeout
  * is reached with conditionFn never evaluating to true, rejects.
  */
-export const runAndWaitForConditions = async (
+export const runAnduntils = async (
```
harbu:
```suggestion
export const runAndWaitForCondition = async (
```

harbu (changes requested):

juslesan (approved):

PR #2877, opened 2024-11-13
chore(sdk): Use `StreamrClient` cache config in `GapFiller`

teogeb:
## Summary

`GapFiller's` storage node cache use `StreamrClient`'s cache config option.

## Background

There is only one cache configuration option in client. Most of the caches should use the `maxSize`/`maxAge` options configured at client level. In some specific cases it could make sense to use hardcoded `maxSize`/`maxAge` values instead. Before this PR storage node cache used 30 min cache, but there was a TODO comment which stated "`TODO maybe caching should be configurable`".


harbu (approved):
For changelog I would put it under `Changed`

PR #2626, opened 2024-07-04
chore(node): introduce `streamr-node` and `streamr-node-init` commands

harbu:
## Summary

- Add commands `streamr-node` and `streamr-node-init`
- Deprecate `streamr-broker` and `streamr-broker-init` (same behavior but with warning message)
- Update docs to account for new commands (left out Brubeck section, it is based on the old stuff anyways)
- Update Dockerfile entrypoint command 

Example of deprecation warning when running old command
```
The command "streamr-broker" is deprecated and will be removed in the future. Please switch to command "streamr-node" instead.
WARN [2024-07-04T13:38:01.485] (.                        ): The command "streamr-broker" is deprecated and will be removed in the future. Please switch to command "streamr-node" instead.
INFO [2024-07-04T13:38:02.232] (broker                   ): Start Streamr node version 101.0.0-beta.4
INFO [2024-07-04T13:38:02.430] (NetworkStack             ): Starting a Streamr Network Node
INFO [2024-07-04T13:38:03.929] (AutoCertifierClient      ): updateSubdomainIp() called for e12f6842-d716-4379-a1a8-5051ed202d04.streamr-nodes.xyz
INFO [2024-07-04T13:38:03.994] (AutoCertifierClient      ): hasSession() called 1 ongoing sessions
INFO [2024-07-04T13:38:04.468] (AutoCertifierClient      ): 2147483647 milliseconds until certificate update
INFO [2024-07-04T13:38:04.902] (NetworkStack             ): Node id is f8619ca67b65ec5310426a8715c14ad1976e381b
```

- [x] TODO: test linked binaries by hand

## Checklist before requesting a review

- [x] Is this a breaking change? If it is, be clear in summary.
- [x] Read through code myself one more time.
- [x] Make sure any and all `TODO` comments left behind are meant to be left in.
- [x] Has reasonable passing test coverage?
- [x] Updated changelog if applicable.
- [x] Updated documentation if applicable.

PR #2881, opened 2024-11-14
feat(node): Remove `streamr-broker` command

teogeb:
Removed the deprecated command.

Maybe it is ok remove it now as it was deprecated in the previous major release (i.e. v101, see https://github.com/streamr-dev/network/pull/2626)

## Other chages

Fixed changelog item for PR https://github.com/streamr-dev/network/pull/2830


harbu (approved):
Should docs be updated?

teogeb:
> Should docs be updated?

No. Mentioned only in `brubeck-network` docs which refer to previous `sdk` versions.

PR #2885, opened 2024-11-18
refactor(sdk): `StreamRegistry` cache invalidation

teogeb:
Refactor `StreamRegistry` caching:
- added private `populateMetadataCache()` method to populate the cache
- added `invalidateCache()` helper function to simplify cache invalidation
- renamed cache fields to be nouns instead of method verbs

## Small functionality changes

- The metadata is in cache after `setStreamMetadata()` and `createStream()` calls (we can easily populate the cache as the metadata is a method parameter)


packages/sdk/src/contracts/StreamRegistry.ts
```diff
@@ -108,6 +108,11 @@ const formCacheKeyPrefix = (streamId: StreamID): string => {
     return `${streamId}|`
 }
 
+const invalidateCache = (cache: CachingMap<string, any, any>, streamId: StreamID) => {
```
harbu: (nitpick) Missing return type

packages/sdk/src/contracts/StreamRegistry.ts
```diff
@@ -511,44 +517,39 @@ export class StreamRegistry {
...
-    getStreamMetadata(streamId: StreamID, useCache = true): Promise<StreamMetadata> {
-        if (useCache) {
-            return this.getStreamMetadata_cached.get(streamId)
-        } else {
-            return this.getStreamMetadata_nonCached(streamId)
+    async getStreamMetadata(streamId: StreamID, allowCached = true): Promise<StreamMetadata> {
+        if (!allowCached) {
+            invalidateCache(this.metadataCache, streamId)
         }
+        return this.metadataCache.get(streamId)
```
harbu: The way this used to work is that when setting `useCache = false` it would just do a direct call and not use cache. Now it seems to be clearing cache. Is there a problematic scenario where if the underlying call rejects (e.g. JSON RPC is down) we clear a cache but can't provide a replace value?

Comment applies to later usages in this PR as well.

packages/sdk/src/utils/CachingMap.ts
```diff
@@ -18,14 +24,11 @@ export class CachingMap<K extends MapKey, V, P extends any[]> {
 
     private readonly cachedFn: (...args: P) => Promise<V>
     private readonly cache: LRU<K, { data: V, maxAge: number }>
+    private readonly opts: Options<P, K>
```
harbu, line 27: (nitpick) A `Readonly<>` here could provide more clarity to the reader that the options are not meant to change during the life of an instance of `CachingMap`.
```suggestion
    private readonly opts: Readonly<Options<P, K>>
```

packages/sdk/src/contracts/StreamRegistry.ts
```diff
@@ -511,44 +517,39 @@ export class StreamRegistry {
...
     }
 
-    isStreamPublisher(streamId: StreamID, userId: UserID, useCache = true): Promise<boolean> {
-        if (useCache) {
-            return this.isStreamPublisher_cached.get(streamId, userId)
-        } else {
-            return this.isStreamPublisherOrSubscriber_nonCached(streamId, userId, StreamPermission.PUBLISH)
+    isStreamPublisher(streamId: StreamID, userId: UserID, allowCached = true): Promise<boolean> {
+        if (allowCached) {
```
harbu:
```suggestion
        if (!allowCached) {
```

packages/sdk/src/contracts/StreamRegistry.ts
```diff
@@ -511,44 +517,39 @@ export class StreamRegistry {
...
+        return this.publisherCache.get(streamId, userId)
     }
 
-    isStreamSubscriber(streamId: StreamID, userId: UserID, useCache = true): Promise<boolean> {
-        if (useCache) {
-            return this.isStreamSubscriber_cached.get(streamId, userId)
-        } else {
-            return this.isStreamPublisherOrSubscriber_nonCached(streamId, userId, StreamPermission.SUBSCRIBE)
+    isStreamSubscriber(streamId: StreamID, userId: UserID, allowCached = true): Promise<boolean> {
+        if (allowCached) {
```
harbu:
```suggestion
        if (!allowCached) {
```

harbu (changes requested):

packages/sdk/src/contracts/StreamRegistry.ts
teogeb: Yes, that was described in the "small functionality changes" section. Calling the function with `allowCached = false` indicates that the caller needs the value to be up-to-date.  Before this change one component could have seen that stream metadata is `foo`, and another that it is `bar`. The current application wide consistency is most likely better.

I.e. it follows the principle that it is better to fail a task than to allow data corruption to happen.

packages/sdk/src/utils/CachingMap.ts
teogeb: We haven't used this kind of deep annotation in the `sdk` code base before.

Many of our objects are immutable, but adding many `Readonly<>` annotations to all of those would just decrease readability. Using `readonly` in the field type is not the same, but it communicates the intention quite well.

(Also the name and the type of the field should communicate clearly that we are using the Configuration Option Pattern which implies immutability by default.)

packages/sdk/src/contracts/StreamRegistry.ts
harbu: Yes but there can be multiple callers to the same instance of this class. Some of them might be fine with cached values and others are not. Now it could happen that one of those that is fined with cached values no longer receives the cached value because someone called this function requiring a non-cached value but the e.g. fetching fails?

teogeb: Created a separate PR for this issue: https://github.com/streamr-dev/network/pull/2889

harbu (approved):
<del>LGTM</del>

harbu (approved):

PR #2894, opened 2024-11-21
test(sdk): Unify `toThrowStreamrError()`

teogeb:
The output is now similar to `toEqualPeerDescriptor()` (see https://github.com/streamr-dev/network/pull/2892).

The custom matcher implementation now uses stricter types. Also improved output when the error thrown is a primitive value.


packages/sdk/test/unit/customMatchers.test.ts
```diff
@@ -0,0 +1,59 @@
...
+                expect(() => {
+                    // eslint-disable-next-line @typescript-eslint/only-throw-error
+                    expect(() => { throw new TestClass() }).toThrowStreamrError({
+                        message: 'Foobar',
+                        code: 'UNSUPPORTED_OPERATION'
+                    })
+                }).toThrow('Not an instance of StreamrClientError:\nReceived: "TestClass"')
+            })
+
+            it('unexpected privitive', () => {
```
juslesan:
```suggestion
            it('unexpected primitive', () => {
```

juslesan (approved):

PR #2899, opened 2024-11-21
feat(sdk): `Mapping` concurrency

teogeb:
Improved concurrency for `Mapping` class. Now concurrent equal `get()` calls share the same promise, i.e. they evaluate the `valueFactory` only once.

## Example

```ts
const valueFactory = async () => ...
const mapping = new Mapping(valueFactory)
await Promise.all([
    mapping.get('foo'),
    mapping.get('bar'),
    mapping.get('bar'),
    mapping.get('foo'),
    mapping.get('foo')
])
```

Triggers two calls to `valueFactory` (one for `foo` and another for `bar`). Before this PR it triggered 5 calls.

## Small refactoring

- Added `readonly` field annotations
- Updated deprecated `jest` method call

### DIFF
--- a/packages/sdk/src/utils/Mapping.ts
+++ b/packages/sdk/src/utils/Mapping.ts
@@ -14,6 +14,7 @@ interface ValueWrapper<V> {
 export class Mapping<K extends (string | number)[], V> {
 
     private readonly delegate: Map<string, ValueWrapper<V>> = new Map()
+    private readonly pendingPromises: Map<string, Promise<V>> = new Map()
     private readonly valueFactory: (...args: K) => Promise<V>
 
     constructor(valueFactory: (...args: K) => Promise<V>) {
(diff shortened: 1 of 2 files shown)


packages/sdk/src/utils/Mapping.ts
```diff
@@ -13,22 +13,35 @@ interface ValueWrapper<V> {
...
     }
 
     async get(...args: K): Promise<V> {
         const key = formLookupKey(...args)
-        let valueWrapper = this.delegate.get(key)
-        if (valueWrapper === undefined) {
-            const value = await this.valueFactory(...args)
-            valueWrapper = { value }
-            this.delegate.set(key, valueWrapper)
+        const pendingPromises = this.pendingPromises.get(key)
```
harbu:
```suggestion
        const pendingPromise = this.pendingPromises.get(key)
```

harbu (approved):

PR #2903, opened 2024-11-22
refactor(sdk): Use `Mapping` instead of `CachingMap`

teogeb:
Use `Mapping` cache instead of `CachingMap` in `StreamRegistry` and `StreamStorageRegistry`. After these modifications there are no usages for `CacheMap` in `sdk`, and we can remove it (in a follow-up PR).

## Testing

The implementations of `Mapping` and `CachingMap` are very different, but the functionality should be almost/exactly same. Double-check that there are no corner case in which behavior is significantly different. Checked that CachingMap tests pass also with the new implementation (see temporary commit [e5bfdcf](https://github.com/streamr-dev/network/pull/2903/commits/e5bfdcfc94db2af5c0fc46c22261f4022a8473db)). The same functionality was already covered by `Mapping.test.ts`, so there was no need keep any `CachingMap.test.ts` test cases for the `Mapping` class.

## Other changes

- Changed argument type of `Mapping#get()` and `Mapping#valueFactory()`, and return type of `Mapping#values()`

## Future improvements

- Remove `CachingMap`


packages/sdk/src/utils/Mapping.ts
```diff
@@ -37,45 +37,55 @@ export class Mapping<K extends KeyType, V> {
...
-            this.delegate = new Map<string, ValueWrapper<V>>()
+            this.delegate = new Map<string, Item<K, V>>()
         }
         this.opts = opts
     }
 
-    async get(...args: K): Promise<V> {
-        const key = formLookupKey(...args)
-        const pendingPromise = this.pendingPromises.get(key)
+    async get(...key: K): Promise<V> {
```
mondoreale: `key` format is inconsistent. I get values using `get(1,2,3,4)` and set values using `set([1,2,3,4], 'foobar')`. Maybe `get([1,2,3,4])` to keep it the same?

mondoreale (approved):
Looks good to me. I have a couple of loose suggestions:

I'd see the key to be a primitive by default OR something more complex if needed.

For a tuple
```const a = new Mapping<[number, string], string>()```

And for a string
```const a = new Mapping()```

And for a number
```const a = new Mapping<number>()```


So, in short, basic `KeyType` could be anything that satisfies `K extends string | number | symbol`. Then the applied key could be of type `K | K[]`.

```ts
type Key<K> = K | K[]

class Mapping<K extends Key<string | number | symbol> = string, V = string> {}
```

To simplify usage of more complex keys we could rewrite `formLookupKey` to `JSON.stringify(key)`.

↑, maybe worth considering. It'd also allow to bypass 1-element tuple keys.

packages/sdk/src/utils/Mapping.ts
teogeb: Fixed

teogeb:
> I'd see the key to be a primitive by default OR something more complex if needed.

Added this functionality 👍 Didn't set any default types as `string` is just one possibility.

harbu (approved):
had a quick read thru; lgtm

PR #2906, opened 2024-11-25
test(node): Fix config wizard storage path tests

teogeb:
Fixed config wizard storage path assertions to be compatible with Mac OS. In the assertion we compare the real path to a symbolic link in Mac OS and thererore the test fails.

## Change

Now compares normalized paths, i.e. uses `realpathSync` which resolves symbolic links.

## Error

```
    expect(received).toInclude(expected)

    Expected string to include:
      "streamr-node /var/folders/_1/sztmfs_j15zg4t29k2nql0040000gn/T/test-config-wizardhSM4Wz/config.json
    "
    Received:
      "Welcome to the Streamr Network!
        This Config Wizard will help you setup your node.·····
        The steps are documented here:
        https://docs.streamr.network/guides/how-to-run-streamr-node#config-wizard
    ✓ Congratulations, you've setup your Streamr node!
        Your node address is 0x909DC59FF7A3b23126bc6F86ad44dD808fd424Dc
        Your node's generated name is Mountain Until Gun
    You can start your Streamr node now with
        streamr-node /var/folders/1/sztmfsj15zg4t29k2nql0040000gn/T/test-config-wizardhSM4Wz/config.json·····
        For environment specific run instructions, see
        https://docs.streamr.network/guides/how-to-run-streamr-node"

      158 |         expect(summary).toInclude(`generated name is Mountain Until Gun\n`)
      159 |
    > 160 |         expect(summary).toInclude(`streamr-node ${storagePath}\n`)
          |                         ^
      161 |     })
      162 |
      163 |     it('prints out the generated private key onto the screen if told to', async () => {

      at Object.<anonymous> (test/unit/ConfigWizard.test.ts:160:25)
```

### DIFF
--- a/packages/node/test/unit/ConfigWizard.test.ts
+++ b/packages/node/test/unit/ConfigWizard.test.ts
@@ -80,11 +80,7 @@ const OPERATOR_ADDRESS = '0x54d68882d5329397928787ec496da3ba8e45c48c'
 
 const parseStoragePath = (summary: string): string | undefined => {
     const match = summary.match(/streamr-node ([^\s]w+)/)
-    if (match !== null) {
-        return match[1]
-    } else {
-        return undefined
-    }
+    return (match !== null) ? match[1] : undefined
 }
 
 const expectPathsEqual = (actual: string | undefined, expected: string): void => {
--- a/packages/node/test/unit/ConfigWizard.test.ts
+++ b/packages/node/test/unit/ConfigWizard.test.ts
@@ -78,7 +78,7 @@ const IMPORTED_PRIVATE_KEY =
 
 const OPERATOR_ADDRESS = '0x54d68882d5329397928787ec496da3ba8e45c48c'
 
-const parseStoragePath = (summary: string): string | undefined => {
+const extractStoragePath = (summary: string): string | undefined => {
     const match = summary.match(/streamr-node ([^\s]w+)/)
     return (match !== null) ? match[1] : undefined
 }
@@ -172,7 +172,7 @@ describe('Config wizard', () => {
 
         expect(summary).toInclude(`generated name is Mountain Until Gun\n`)
 
-        expectPathsEqual(parseStoragePath(summary), storagePath)
+        expectPathsEqual(extractStoragePath(summary), storagePath)
     })
 
     it('prints out the generated private key onto the screen if told to', async () => {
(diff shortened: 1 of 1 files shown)


packages/node/test/unit/ConfigWizard.test.ts
```diff
@@ -78,10 +78,29 @@ const IMPORTED_PRIVATE_KEY =
 
 const OPERATOR_ADDRESS = '0x54d68882d5329397928787ec496da3ba8e45c48c'
 
+const parseStoragePath = (summary: string): string | undefined => {
+    const match = summary.match(/streamr-node ([^\s]w+)/)
+    if (match !== null) {
+        return match[1]
+    } else {
+        return undefined
+    }
```
mondoreale: `match` can only be a `null` or a "match".

I'd consider either

```ts
// 1
return match?.[1] ?? undefined
```

or 

```ts
// 2
return match ? match[1] : undefined
```

or at least

```ts
// 3
if (match !== null) {
    return match[1]
}

return undefined
```

packages/node/test/unit/ConfigWizard.test.ts
```diff
@@ -78,10 +78,29 @@ const IMPORTED_PRIVATE_KEY =
 
 const OPERATOR_ADDRESS = '0x54d68882d5329397928787ec496da3ba8e45c48c'
 
+const parseStoragePath = (summary: string): string | undefined => {
```
mondoreale: I'd suggest to call this `extractStoragePath` rather than `parseStoragePath`.

My reasoning:
1. One "extracts" when they pull out a substring that matches a pattern. That's it.
2. One "parses" when they analyse to understand and, for instance, splits something into smaller pieces, etc.

In other words, `parseStoragePath` would have to take \*a storage path* as an argument because that's what it parses. What we currently have is `parseSummary`.

mondoreale (commented):
Interesting. 🤔

I like it. 

See my comments.

mondoreale:
Let's get this one out asap.

packages/node/test/unit/ConfigWizard.test.ts
teogeb: Applied the ternary with explicit `null` expression: `return (match !== null) ? match[1] : undefined`

packages/node/test/unit/ConfigWizard.test.ts
teogeb: Very good improvement, thanks 👍

mondoreale (approved):

PR #2907, opened 2024-11-25
test: use Jest instead of NodeJS assertions

teogeb:
Replaced usages of `assert` NodeJS calls with equivalent Jest calls. There is no need to use two separate test utilities in our code base.


juslesan (approved):
👍

PR #2910, opened 2024-11-25
chore: Add `@typescript-eslint/no-confusing-void-expression` rule

teogeb:
Added new `eslint` rule: https://typescript-eslint.io/rules/no-confusing-void-expression/

This rule disallows e.g. using `return foobar()` when `foobar()` method returns `void`. It is better to just call the method and do explicit return separately, if needed. (Quite typically the `return` is not needed at all as the line is the last line of the method.)

Simplified `multiple-publisher-plugins.test.ts` so that it uses `Promise<void>` instead of `Promise<unknown>` and therefore doesn't need any special handling related to this rule.

### DIFF
--- a/packages/dht/test/integration/ConnectionLocking.test.ts
+++ b/packages/dht/test/integration/ConnectionLocking.test.ts
@@ -61,6 +61,7 @@ describe('Connection Locking', () => {
         const nodeId2 = toNodeId(mockPeerDescriptor2)
         await Promise.all([
             until(() => connectionManager2.hasRemoteLockedConnection(nodeId1)),
+            // eslint-disable-next-line @typescript-eslint/no-confusing-void-expression
             connectionManager1.lockConnection(mockPeerDescriptor2, 'testLock')
         ])
         expect(connectionManager1.hasConnection(nodeId2)).toEqual(true)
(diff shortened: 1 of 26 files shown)


packages/dht/test/integration/ConnectionLocking.test.ts
```diff
@@ -61,6 +61,7 @@ describe('Connection Locking', () => {
         const nodeId2 = toNodeId(mockPeerDescriptor2)
         await Promise.all([
             until(() => connectionManager2.hasRemoteLockedConnection(nodeId1)),
+            // eslint-disable-next-line @typescript-eslint/no-confusing-void-expression
```
juslesan, line 64: Why does this line need to be ignored?

packages/node/test/unit/helpers/weightedSample.test.ts
```diff
@@ -3,6 +3,7 @@ import { range, repeat, sum } from 'lodash'
 
 describe(weightedSample, () => {
     it('returns undefined on empty array', () => {
+        // eslint-disable-next-line @typescript-eslint/no-confusing-void-expression
```
juslesan: Why does this line need to be ignored?

packages/trackerless-network/test/end-to-end/content-delivery-layer-node-with-real-connections.test.ts
```diff
@@ -106,10 +106,15 @@ describe('content delivery layer node with real connections', () => {
             dhtNode2.stop(),
             dhtNode3.stop(),
             dhtNode4.stop(),
+            // eslint-disable-next-line @typescript-eslint/no-confusing-void-expression
```
juslesan, line 109: Why does this line need to be ignored?

juslesan (commented):
Why does this rule need to be ignored in Promise.all calls and some other places?

teogeb:
> Why does this rule need to be ignored in Promise.all calls and some other places?

It would be better to refactor these tests. In practice, the arguments sent to `Promise.all()` are evaluated before the `Promise.all() `call (like we do for all function calls: we need the value for parameters before we can pass those as arguments). Therefore it would be clearer if we'd move void calls out from `Promise.all()`, i.e. execute the lines explicitly before the `Promise.all()` is called.

packages/node/test/unit/helpers/weightedSample.test.ts
teogeb: Added type generics. Without it, the generics type `T` evaluated to void.

juslesan (approved):
Could move the void expressions above the Promise.all blocks in all cases to get rid of the eslint-disable comments

PR #2913, opened 2024-11-26
test(dht): Fix `getClosestNodes` unit test

teogeb:
Fixed the test to use two excluded nodes.

The intention in the test is to exclude two nodes, but in practice it used only one excluded node. The `sampleSize` didn't received any count parameter because the parameter `2` was incorrectly passed to the `map` function instead of the `sampleSize` function. (I.e. it was  uses as the `thisArg` parameter of the `map()` function.)


mondoreale (approved):
Ouch!

PR #2918, opened 2024-11-28
feat(node): [NET-1381] Add version to operator heartbeats

teogeb:
Add `applicationVersion` field to the heartbeat message (published by `announceNodeToStream`).

## Testing

Manually tested that the schema validation in v101 accepts the message which has the additional version field. This test snippet passes in a temp branch which was checked out from the v101.1.2 tag:
```ts
it('v102 messages passes validation in v101 schema', () => {
    const msg = {
        "msgType": "heartbeat",
        "peerDescriptor": {
            "nodeId": "nodeId",
            "websocket": {
                "port": 31313,
                "host": "127.0.0.1",
                "tls": false
            }
        },
        "applicationVersion": "102.0.0-beta.1"
    }
    expect(() => HeartbeatMessageSchema.parse(msg)).not.toThrow()
})
```


packages/node/test/unit/plugins/operator/heartbeatUtils.test.ts
```diff
@@ -14,6 +15,21 @@ describe('heartbeatUtils', () => {
...
+            createHeartbeatMessage({
+                nodeId: 'nodeId',
+                websocket: {
+                    port: 31313,
+                    host: '127.0.0.1',
+                    tls: false
+                }
+            }),
+            'version'
+        )
```
mondoreale: No `omit` needed if you do

```ts
const { version: _, ...msg } = createHeartbeatMessage(…)
```

Cheers!

packages/node/test/unit/plugins/operator/heartbeatUtils.test.ts
```diff
@@ -1,5 +1,6 @@
-import { createHeartbeatMessage, HeartbeatMessageSchema } from '../../../../src/plugins/operator/heartbeatUtils'
+import { omit } from 'lodash'
```
mondoreale, line 1: Maybe

```ts
import omit from 'lodash/omit'
```

mondoreale (approved):
Looks good. See comments.

packages/node/test/unit/plugins/operator/heartbeatUtils.test.ts
teogeb: @mondoreale Is this needed anymore when we start to use `rollup`? We have multiple `import { ... } from 'lodash'`  style imports currently in the codebase. If we change this, should we change all other occurences, too?

packages/node/test/unit/plugins/operator/heartbeatUtils.test.ts
teogeb: Using `omit` is simpler and more straightforward as we don't need version stored into a variable.

packages/node/test/unit/plugins/operator/heartbeatUtils.test.ts
harbu: This is `@streamr/node` so no browser target here (but yeah generally with `@streamr/sdk` and below layers we should use the `lodash/*` pattern for now until we have rollup)

harbu (approved):

PR #2919, opened 2024-11-28
refactor(sdk): `operatorContractUtils` type safety

teogeb:
- use `bigint` type when a number is a wei amount
- use `EthereumAddress` type for contract addresses
- use `Signer` instead of `Wallet`
  - this allow us to use `client.getSigner()` as an argument
  
## Other changes

- rename some methods which are used only for test setup:
  - `generateWalletWithGasAndTokens() `-> `createTestWallet()`
  - `setupOperatorContract(`) -> `setupTestOperatorContract()`
  - `deployOperatorContract()` -> `deployTestOperatorContract()`
  - `deploySponsorshipContract()` -> `deployTestSponsorshipContract()`
  - `getProvider()` -> `getTestProvider()`
 - remove some obsolete parameters from these methods and from `getTestAdminWallet()` (not used by any component)
 - do not export `transferTokens()` from `operatorContractUtils.ts` (it is an internal helper method)
 
 ## Future improvements

- Refactor the utilities so that we can use `sponsor`/`stake`/`delegate` etc. methods both for in production environment and in tests
  - e.g. `token` parameter for `transferTokens` should be required or injected from the context (now it is implicitly test token)
- Maybe merge `setupTestOperatorContract()` and `deployTestOperatorContract()`
- Maybe `deployTestOperatorContract()` could return client's `Operator` instance?


packages/sdk/src/contracts/operatorContractUtils.ts
```diff
@@ -188,22 +153,16 @@ export function getTestTokenContract(): DATATokenContract {
...
-
-interface GenerateWalletWithGasAndTokensOpts {
-    chainConfig?: { contracts: { DATA: string } }
+export const getTestAdminWallet = (): Wallet => {
+    return new Wallet(TEST_CHAIN_CONFIG.adminPrivateKey).connect(getTestProvider())
 }
 
-export async function generateWalletWithGasAndTokens(opts?: GenerateWalletWithGasAndTokensOpts): Promise<Wallet & SignerWithProvider> {
-    const provider = getProvider()
+export async function createTestWallet(): Promise<Wallet & SignerWithProvider> {
```
harbu, line 160: We don't always want to spend time waiting for a wallet to be generated that contains tokens, therefore I like the old name since it indicates _what kind of wallet_ we are generating. Sometimes in tests we have wallets that don't need to tokens or gas (just generate random private key), need only gas (use KeyServer) or need both (use this function). I think the name should indicate that we are now generating a certain kind of wallet.

harbu (commented):
In a strict sense, I don't mind the contract addresses being just plain strings here since `EthereumAddress` type comes with the lower casing requirement which is not strictly necessary here.  However I'm not against using it here either so I'm fine with that change.

packages/sdk/src/contracts/operatorContractUtils.ts
teogeb: Maybe it would make sense to modify `dev-chain-fast `so that they KeyServer keys have also some tokens. Then this utility function would be obsolete. https://linear.app/streamr/issue/ETH-800/dev-chain-fast-mint-data-tokens-for-pre-baked-accounts

PR #2921, opened 2024-12-02
refactor(sdk): Remove obsolete `onError` index parameter

teogeb:
Removed the `index` parameter from `onError` signal. It was not used by any component.

## Future improvements

- `MsgChainUtils#onError` has a `StreamMessage` parameter, which is not used. We should remove the parameter or start to use it. (I.e. investigate if we should remove the whole `onError` listener in `createMessagePipeline()` if the `ignoreMessages` handling is no longer needed.)


mondoreale (approved):
Eslint and ts are happy so, lgtm!

PR #2922, opened 2024-12-02
fix(network): Revert backwards incompatible PR #2820

juslesan:
Reverting PR #2820 due to backwards incompatibility with production. The typeUrl of the PeerDescriptors in the Any protobuf message has changes from `dht.PeerDescriptor` to  `peerDescriptor.PeerDescriptor` in this change. This makes it impossible to parse the stream entry points stored stored in the DHT causing stream joining to not work.

The clean up of generated protobuf code will be implemented in #2842


teogeb (approved):
We could add a smoke test which does some simple backwards compatibility checks. (In a follow-up PR, as not strictly related to this PR.)

PR #2924, opened 2024-12-02
refactor(sdk): Replace `quick-lru` with `lru-cache`

mondoreale:
_(Hopefully one of many small PRs.)_

## Summary

`quick-lru` is an ES-only package and in order to build with it for CJS we currently move it (raw ESM) over (to `/vendor`) and transform to CommonJS as part of SDK's build process.

`lru-cache` is already being used by DHT and it organically comes in 2 flavours: CJS and ESM. No magic required, and no magic = less maintenance.

## Changes

- Replace references.
- Address differences in tests, in types, and in logic.
- Fix and annotate tests for `Mapping`.
- Remove "vendor" magic (no longer required).

- [x] ~~Is this a breaking change? If it is, be clear in summary.~~
- [x] ~~Updated changelog if applicable.~~
- [x] ~~Updated documentation if applicable.~~

## Remaining todos
- [x] Fix tests


packages/sdk/test/unit/customMatchers.test.ts
mondoreale, line 1: Let me revert this one. Most likely node v18 vs v20 issue.

mondoreale: Can't get this to pass locally. "Me" problem, it seems. CI sees no issues.

mondoreale:
Oh, this needs a bump. Lotta conflicts.

packages/sdk/src/utils/Mapping.ts
```diff
@@ -19,6 +19,14 @@ interface Item<K, V> {
     value: V
 }
 
+interface LRU<K extends string, V> {
```
teogeb: Could this pick something like `type AbstractMap = Pick<Map, ...>`. In the `else` branch we create an instance, which is not a LRU (just a simple Map) and therefore could rename e.g. to `AbstractMap`.

packages/sdk/test/unit/Mapping.test.ts
```diff
@@ -97,25 +97,35 @@ describe('Mapping', () => {
     })
 
     it('max size', async () => {
-        const SIZE = 3
+        const maxSize = 3
```
teogeb: Good rename. Let's use naming style of constants.

```suggestion
        const MAX_SIZE = 3
```

packages/sdk/src/utils/Mapping.ts
```diff
@@ -45,9 +53,9 @@ export class Mapping<K extends LookupKeyType, V> {
      **/
     constructor(opts: CacheMapOptions<K, V> | LazyMapOptions<K, V>) {
         if ('maxSize' in opts) {
-            this.delegate = new LRU<string, Item<K, V>>({
-                maxSize: opts.maxSize,
-                maxAge: opts.maxAge
+            this.delegate = new LRUCache<string, Item<K, V>>({
+                max: opts.maxSize,
```
teogeb: Would it make sense to use `maxSize` option of LRUCache instead of the `max` option? Both of these options seem to limit the item count, but `max` option allocates the storage pre-emptively. This may be more performant, but also bit wasteful, as very typically our caches are not in anywhere near the maximum limit.

packages/sdk/test/unit/Mapping.test.ts
```diff
@@ -97,25 +97,35 @@ describe('Mapping', () => {
...
+        /**
+         * The current list of keys (most-to-least recently used) goes as follows: foo-1,
+         * foo2, foo1. Going through all ids now will reconstruct the initial collection
+         * causing 3 hits.
+         */
         for (const id of ids) {
             await mapping.get(['foo', id])
         }
-        expect(valueFactory).toHaveBeenCalledTimes(5)
+        expect(valueFactory).toHaveBeenCalledTimes(maxSize + 1 + 3)
```
teogeb: Why this is now 7 instead of 5? Assumed it should be the initial 3 evaluations + evaluation of `foo-1` + re-evaluation of dropped `foo0` 

Also: could use `MAX_SIZE` here in place of 3?

teogeb (changes requested):

packages/sdk/src/utils/Mapping.ts
mondoreale: Types of `Map` and `LRUCache` are incompatible. I'll keep the explicit matching-all type and change its name to `CacheMap`.

packages/sdk/src/utils/Mapping.ts
mondoreale: I guess. That's how it used to be. I'll adjust.

packages/sdk/test/unit/Mapping.test.ts
mondoreale: This tests does the following
| | Operation | Slot 0 | Slot 1 | Slot 2 |
| - | --------- | ------ | ------ | ------ |
| 1 | `get('foo',  0)` | foo0|||
| 2 | `get('foo',  1)` | foo1|foo0|
| 3 | `get('foo',  2)` | foo2|foo1|foo0|
| 4 | `get('foo', -1)` | foo-1|foo2|foo1|
| 5 | `get('foo',  0)` | foo0|foo-1|foo2|
| 6 | `get('foo',  1)` | foo1|foo0|foo-1|
| 7 | `get('foo',  2)` | foo2|foo1|foo0|

Each row is a hit to the value factory because we happen to always hit an entry that's either not yet in the collection or has just been discarded.

teogeb: Thanks, great explanation 👍

teogeb (approved):
Good improvement 👍

mondoreale:
I'll merge it in real quick. Cheers!

mondoreale:
For the record, size of the SDK browser build (streamr-sdk.web.min.js):
- before 5.07MB
- after 5.06MB

As expected, `quick-lru` is gone, and `lru-cache` did not add anything on top of what's already been there.

Just FYI.

PR #2925, opened 2024-12-03
refactor(sdk): Improve `InternalPermissionQuery` type safety

teogeb:
The type of `streamId` field is now `StreamID` instead of `string`.

Also remove obsolete `toInternalPermissionQuery()` call in `Stream#hasPermissions()`. There is no need to convert the query as the we call the client method which does the same conversion.

Also fixed method signatures in `FakeStreamRegistry` so that they matched `StreamRegistry` methods.

### DIFF
--- a/packages/sdk/src/permission.ts
+++ b/packages/sdk/src/permission.ts
@@ -1,5 +1,6 @@
-import { ChangeFieldType, HexString, toUserId, UserID } from '@streamr/utils'
+import { ChangeFieldType, HexString, StreamID, toUserId, UserID } from '@streamr/utils'
 import { MaxUint256 } from 'ethers'
+import { StreamIDBuilder } from './StreamIDBuilder'
 
 export enum StreamPermission {
     EDIT = 'edit',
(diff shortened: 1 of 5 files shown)


packages/sdk/src/permission.ts
```diff
@@ -55,14 +58,18 @@ export interface ChainPermissions {
     canGrant: boolean
 }
 
-export const isPublicPermissionQuery = (query: InternalPermissionQuery): query is PublicPermissionQuery => {
-    return (query as PublicPermissionQuery).public === true
+export const isPublicPermissionQuery = (query: InternalPermissionQuery): query is InternalPublicPermissionQuery => {
+    return (query as InternalPublicPermissionQuery).public === true
```
mondoreale, line 62: I'm guessing `InternalUserPermissionQuery` doesn't have the `public` key that's why the cast (`… as …`). If that's true then the following is sufficient for such a gate:

```ts
return `public` in query
```

No casting!

mondoreale (approved):
Sounds reasonable. Looks good.

packages/sdk/src/permission.ts
teogeb: The current approach is better safer as it errors if `public` isn't a key in the expected interface (InternalPublicPermissionQuery). The `... in query` approach wouldn't error. 

Maybe relevant if we'll rename this field in some future refactoring.

mondoreale: I see. Cheers!

PR #2928, opened 2024-12-04
refactor(sdk): Support multiple contracts in `ChainEventPoller` 

teogeb:
Refactored `ChainEventPoller` to support multiple contracts simultaneously. Now the class is a singleton and we can define contract methods as a listener parameter. This optimizes log fetching by reducing required HTTP calls to the JSON RPC as there needs to be only one poller running even if multiple contract events are listened. 

Also we don't now need to create a `Contract` instance just for adding event a listener. This allows us to simplify e.g. `Operator` class (in a follow-up PR).


harbu:
To address your point

> Added try-catch around the event listener so that the polling doesn't stop if a listener throws an error.

We had a discussion on video meeting (me and Teo ofc), maybe better to leave out and have it as caller's responsibility (especially since it was like this before).

On that front, regarding the `async` case, maybe eslint rule that giving async callback to place where (sync) callback is expected would be an error?

packages/sdk/src/contracts/ChainEventPoller.ts
```diff
@@ -57,25 +69,46 @@ export class ChainEventPoller {
...
                         if (abortController.signal.aborted) {
                             return
                         }
                     }
-
                     logger.debug('Polling', { fromBlock, eventNames })
-                    events = await contract.queryFilter([eventNames], fromBlock) as EventLog[]
+                    const filter = {
+                        address: uniq(this.listeners.map((l) => l.contractAddress)),
+                        topics: [uniq(this.listeners.map((l) => l.contractInterfaceFragment.topicHash))],
```
harbu, line 116: We need to check that this definitely works as we think it does

packages/sdk/test/unit/ChainEventPoller.test.ts
```diff
@@ -1,195 +1,285 @@
...
+            indexed: false,
+            internalType: 'string',
+            name: 'param2',
+            type: 'string'
+        }]
+    }]
+}
+
+const createEventLogItem = (
+    eventName: string, eventArgs: any[], blockNumber: number
```
harbu:
```suggestion
    eventName: string,
    eventArgs: any[],
    blockNumber: number
```

packages/sdk/test/unit/ChainEventPoller.test.ts
```diff
@@ -1,195 +1,285 @@
...
+): Partial<Log> => {
+    const contractInterface = new Interface(createAbi(eventName))
+    return {
+        blockNumber: blockNumber,
```
harbu:
```suggestion
        blockNumber,
```

packages/sdk/src/contracts/ChainEventPoller.ts
```diff
@@ -57,25 +69,46 @@ export class ChainEventPoller {
...
+                        newFromBlock = await provider.getBlockNumber() + 1
                         logger.debug('Fetch next block number explicitly', { newFromBlock } )
                         if (abortController.signal.aborted) {
                             return
                         }
                     }
-
                     logger.debug('Polling', { fromBlock, eventNames })
-                    events = await contract.queryFilter([eventNames], fromBlock) as EventLog[]
+                    const filter = {
```
harbu, line 114: Maybe worth commenting how the filtering logic here works, e.g.

> address == (X OR Y OR Z) AND topic  === (A OR B  OR C)

harbu (approved):

PR #2930, opened 2024-12-05
refactor(sdk)!: Rename storage node event

teogeb:
**This is a breaking change.**

Fixed typo in event name by renaming `streamRemovedFromFromStorageNode` -> `streamRemovedFromStorageNode`.


harbu (approved):
good catch

PR #2943, opened 2024-12-10
deps: Swap `node-fetch` library with a having-it-all native `fetch`

mondoreale:
## Summary

I replace ES-only `node-fetch` with the native [`fetch`](https://nodejs.org/dist/v18.20.5/docs/api/globals.html). ~~a wrapper module that deals with CJS, ESM, and uses `window.fetch` within a browser environment.~~

~~`@streamr/fetch` wraps `node-fetch` v3 (most recent) and some signatures within it have been changed. This PR addresses it, too.~~

## Changes

- Replace `node-fetch` with `fetch`.
- Remove shims, tricks and fixes for `node-fetch` for the browser (mostly SDK).
- `composeAbortSignals` can take an empty list of arguments and arguments can be either instances of `AbortSignal` or `undefined` – makes the utility a bit more flexible, and way more convenient.

- n/a

- [x] ~~Updated changelog if applicable.~~
- [x] ~~Updated documentation if applicable.~~


packages/node/src/helpers/fetchOrThrow.ts
mondoreale, line 1: This isn't being used by anything.

packages/node/package.json
```diff
@@ -63,7 +63,6 @@
     "@types/heap": "^0.2.34",
     "@types/lodash": "^4.17.13",
     "@types/merge2": "^1.4.4",
-    "@types/node-fetch": "^2.6.4",
```
mondoreale, line 66: ~~Types are now exported directly from the @streamr/fetch package.~~ Types come from `@types/node` now!

packages/sdk/src/shim/node-fetch.ts
mondoreale, line 1: Ported this logic to `@streamr/fetch` (in v0.0.2).

packages/sdk/src/shim/http-https.ts
mondoreale, line 1: This doesn't seem to be necessary. Dropping. 🤷🏻

mondoreale (commented):
Good thing about the @streamr/fetch library is that it's highly adaptive. No extra steps to make it work for each of the supported environments.

harbu:
Hey @mondoreale , I think we could maybe get rid of `node-fetch` altogether and use Node.js built-in `fetch`. I think it is available from Node v18 onward? And for the next release we are aiming to support v20 and v22.

In v20 it is still considered "experimental" but probably works good enough https://nodejs.org/dist/v20.11.1/docs/api/globals.html#fetch

In v22 it is already stable https://nodejs.org/dist/v22.12.0/docs/api/globals.html#fetch

mondoreale:
Interesting!

I didn't know you considered native `fetch` ([stable since v21](https://nodejs.org/en/blog/announcements/v21-release-announce)). This will make things much better!

I'll do 2 things from here:
- Repurpose this PR so that it replaces `node-fetch` with the native `fetch`.
- Leave the wrapper package be. Maybe someone finds it helpful.

Cheers!

packages/utils/src/composeAbortSignals.ts
```diff
@@ -10,29 +10,32 @@ export type ComposedAbortSignal = AbortSignal & { destroy: () => void }
  * aforementioned instance of AbortSignal will have more and more listeners added
  * but never cleaned.
  */
-export function composeAbortSignals(...signals: AbortSignal[]): ComposedAbortSignal {
-    if (signals.length === 0) {
-        throw new Error('must provide at least one AbortSignal')
-    }
+export function composeAbortSignals(...signals: (AbortSignal | undefined | null)[]): ComposedAbortSignal {
+    const actualSignals = signals.filter(Boolean) as AbortSignal[]
```
harbu: I have two stylistic points to make here:

Firstly, I would keep the filtering out of this function as it is a separate concern, i.e. the user of this function should filter any undefined or null values. You could make an argument to keep it in if it is very common to need to filter before using the function. I counted two cases where this filtering is needed, I would just move the filtering explicitly to those two cases.

Secondly, I'm not a fan of `.filter(Boolean)`  due to how it handles certain primitive values such as `0` and `''`. It is true that here we aren't dealing with primitives and in general a programmer worth their salt should be paying attention to such things. That being said, it can be rather easy to introduce issues inadvertently through e.g. refactoring. (Maybe a list of objects gets changed to be a list of strings, and somewhere down the call stack, a `.filter(Boolean)` is called on said list, was the intention to really remove empty strings then?)

To avoid the whole confusion about what is being filtered (primitives or non-primitives), I'd just consistently use something that works as expected always, e.g. lodash `compact` or `.filter((v) => v !== undefined)`. (In case you are wondering why I don't check null there, we make it a point in this code base to avoid `null`, see: https://github.com/streamr-dev/network/blob/main/developer-docs/index.md#avoid-null-prefer-undefined, if you want to play it extra safe then check for null as well I suppose)

packages/utils/src/until.ts
```diff
@@ -37,10 +36,10 @@ export const until = async (
         throwError(userAborted, conditionFn, onTimeoutContext)
     }
     abortSignal?.addEventListener('abort', () => { userAborted = true }, { once: true })
-    const timeoutAbortSignal: AbortSignal = (AbortSignal as any).timeout(timeout) // TODO: remove `as any` cast when @types/node has support...
-    const composedSignal = composeAbortSignals(...compact([timeoutAbortSignal, abortSignal]))
+    const timeoutAbortSignal: AbortSignal = AbortSignal.timeout(timeout)
+    const composedSignal = composeAbortSignals(timeoutAbortSignal, abortSignal)
     try {
-        while (true) {
+        for (;;) {
```
harbu: (nitpick) I'd keep this as a `while (true)`  as it is consistent with the rest of the code base. Unless there is some other reason to favor this over `while (true)`?

packages/utils/src/composeAbortSignals.ts
```diff
@@ -10,29 +10,32 @@ export type ComposedAbortSignal = AbortSignal & { destroy: () => void }
...
+
         destroy()
-        abortController.abort()
     }
-    for (const signal of signals) {
-        signal.addEventListener('abort', abort)
+
+    for (const signal of actualSignals) {
+        if (signal.aborted) {
+            onAbort()
```
harbu: I do prefer we check the list for aborted signals before appending any new event listeners. It makes the solution slightly less mutable (no need to add and then remove event listeners in case of an aborted listener) and personally (just my opinion) I find it more readable to have it handled as its own case, i.e.
```ts
// handle pre-aborted signals
for (const signal of actualSignals) {
    if (signal.aborted) {
       return Object.assign(preAbortedSignal, { destroy: () => {} })
    }
}

// at this point, i know all listeners are yet to be aborted, good to add listeners
for (const signal of actualSignals) {
    signal.addEventListener('abort', onAbort, { once: true })
}
```

packages/utils/test/composeAbortSignals.test.ts
```diff
@@ -101,16 +100,14 @@ describe('composeAbortSignals', () => {
         })
     })
 
-    it('throws error if composing empty list', () => {
-        expect(() => {
-            composeAbortSignals()
-        }).toThrow('must provide at least one AbortSignal')
+    it('gives a pending signal for an empty list of signals', () => {
+        expect(composeAbortSignals().aborted).toBeFalse()
```
harbu, line 104: I'm guessing accepting the empty list makes it easier to use:
```ts
const signals = [signal1, signal2, signal3, signal4, signal5].filter((v) => v !== undefined)

// old way:
let composedSignal
if (signals.length !== 0) {
    composedSignal = composeAbortSignals(signals)
}

// new way:
const composedSignal = composeAbortSignals(signals)
```

Originally when I wrote this, I wasn't really sure it made sense to have an abort signal that never aborts so I made it an error. But thinking about it now, I think there are cases where it makes sense such as `fetch`. The request will complete eventually anyway so passing it a signal that never aborts is not an issue.

I guess let's keep this new functionality, maybe add a new entry in PR description under "Changes" such as "composeAbortSignal now allows empty lists".

harbu (commented):
Added a few thoughts mostly on the abort signal changes. Coming it from a "code base convention" angle, so if there's counter-points you feel like you want to bring up, feel free to do so.

packages/utils/src/until.ts
mondoreale: Bad reflex.

I'll revert.

Changed it to avoid constant conditions. It's also shorter. No worries tho.

packages/sdk/src/utils/WebStreamToNodeStream.ts
```diff
@@ -68,6 +72,6 @@ export function WebStreamToNodeStream(
 
     // use PassThrough so we can write to it
     const nodeStream = new PassThrough(nodeStreamOptions)
-    pull(webStream, nodeStream).catch(ignoreAbort)
+    pull(webStream, nodeStream)
```
mondoreale, line 75: The removed piece is now baked into `pull` (see that `finally` above).

packages/node/test/unit/httpServer.test.ts
```diff
@@ -26,9 +24,8 @@ const startTestServer = (...endpoints: Endpoint[]) => {
 }
 
 const createRequest = async (endpoint: string, headers?: Record<string, string>) => {
-    await wait(10) // TODO: remove when fixed https://github.com/node-fetch/node-fetch/issues/1735
```
mondoreale, line 29: Seems to no longer be an issue.

packages/sdk/src/utils/utils.ts
```diff
@@ -210,36 +221,39 @@ export const fetchLengthPrefixedFrameHttpBinaryStream = async function*(
...
+            yield* stream
+        } catch (err) {
+            abortController.abort()
+            throw err
+        } finally {
+            stream?.destroy()
+        }
     } finally {
-        stream?.destroy()
         fetchAbortSignal.destroy()
```
mondoreale, line 257: This is now a stand-alone `try…finally` responsible only for destroying that one composed signal, which AFAIK, in some situations can stay "open" (e.g. above, after each rejected `fetch`).

packages/utils/src/composeAbortSignals.ts
mondoreale: I hear you. Hold my 🍺.

packages/utils/test/composeAbortSignals.test.ts
mondoreale: Yeah, the "new way" could actually be written like so
```ts
const composedSignal = composeAbortSignals(
    signal1,
    signal2,
    signal3,
    signal4,
    signal5
)
```

No pre-filtering needed.

packages/sdk/test/end-to-end/Operator.test.ts
```diff
@@ -33,7 +32,7 @@ const createClient = (privateKey?: string): StreamrClient => {
 const createTheGraphClient = (): TheGraphClient => {
     return new TheGraphClient({
         serverUrl: CHAIN_CONFIG.dev2.theGraphUrl,
-        fetch,
+        fetch: (...params: Parameters<typeof fetch>) => fetch(...params),
```
mondoreale, line 35: Fixes

```
TypeError: Failed to execute 'fetch' on 'Window': Illegal invocation
```

packages/sdk/test/unit/Resends.test.ts
```diff
@@ -48,8 +48,7 @@ describe('Resends', () => {
         }).rejects.toThrowStreamrClientError({
             message: isRunningInElectron()
                 ? 'Failed to fetch'
-                // eslint-disable-next-line max-len
-                : 'request to http://mock.test/streams/stream/data/partitions/0/last?count=1&format=raw failed, reason: getaddrinfo ENOTFOUND mock.test',
+                : 'fetch failed (code=STORAGE_NODE_ERROR)',
```
mondoreale, line 51: `fetch` doesn't provide response for attempts that fail at the get go.

packages/utils/src/composeAbortSignals.ts
```diff
@@ -10,29 +10,52 @@ export type ComposedAbortSignal = AbortSignal & { destroy: () => void }
...
+    const abortController = new AbortController()
+
+    for (const signal of signals) {
+        if (signal?.aborted) {
+            abortController.abort()
 
-    const preAbortedSignal = signals.find((s) => s.aborted)
-    if (preAbortedSignal !== undefined) {
-        return Object.assign(preAbortedSignal, { destroy: () => {} })
+            return Object.assign(abortController.signal, { destroy: () => {} })
```
mondoreale, line 22: @harbu there u have it. I've used the internal signal as \*the* composed signal. No point in assigning `destroy` to one of the given signals (that's a no-no).

packages/sdk/src/utils/WebStreamToNodeStream.ts
```diff
@@ -2,56 +2,60 @@ import { PassThrough, Readable, TransformOptions } from 'stream'
...
+        try {
+            await reader.cancel()
+        } catch (_) {
+            /**
+             * `reader.cancel` can actually throw if called on a response
+             * body reader of a "cancelled" fetch. Do nothing.
+             */
+        }
+
+        reader.releaseLock()
```
mondoreale, line 58: This seems to be a very simple and yet quite powerful memory optimization (always release locks).

packages/utils/src/composeAbortSignals.ts
```diff
@@ -10,29 +10,52 @@ export type ComposedAbortSignal = AbortSignal & { destroy: () => void }
  * aforementioned instance of AbortSignal will have more and more listeners added
  * but never cleaned.
  */
-export function composeAbortSignals(...signals: AbortSignal[]): ComposedAbortSignal {
-    if (signals.length === 0) {
-        throw new Error('must provide at least one AbortSignal')
-    }
+export function composeAbortSignals(
+    ...signals: (AbortSignal | undefined)[]
```
mondoreale, line 14: @harbu, I've change this one to be an array of signals or undefineds, and droppped `null`. It'll match the convention and address one of your feedback comments.

Cheers.

packages/utils/src/composeAbortSignals.ts
```diff
@@ -10,29 +10,42 @@ export type ComposedAbortSignal = AbortSignal & { destroy: () => void }
...
+): ComposedAbortSignal {
+    const abortController = new AbortController()
```
harbu, line 16: (nitpick) This `abortController` could be moved lower in the function body as it is not needed if there is a pre-aborted signal.

harbu (approved):
LGTM; one nitpick but not a blocker

packages/utils/src/composeAbortSignals.ts
mondoreale: Neat! I'll adjust.

packages/utils/src/composeAbortSignals.ts
```diff
@@ -10,29 +10,42 @@ export type ComposedAbortSignal = AbortSignal & { destroy: () => void }
...
+
+    for (const signal of signals) {
+        if (signal?.aborted) {
+            abortController.abort()
```
mondoreale, line 20: The above `abortController` is in the right place. After all, its signal is being used in both "returns".

PR #2944, opened 2024-12-11
refactor(sdk): `NetworkNodeFacade#registerOperator()`

teogeb:
Moved the operator registration functionality from `node` to `sdk`.

## Changes

Added:
- `NetworkNodeFacade#registerOperator()` method

Removed:
- `NetworkNodeFacade#registerExternalRpcMethod()` method
- exports of `OperatorDiscoveryRequest` and `OperatorDiscoveryResponse`

## Other changes

Optimized `OperatorPlugin.test.ts` to use the `warmupPeriodInMs` config option. The `operator discovery` test now takes ~2.5s (i.e. ~80% less time).

## Future improvements

Add unit/integration/end-to-end test to `sdk`? There is currently a test in node (`OperatorPlugin.test.ts`), which covers also `registerOperator()` functionality, but better to have a focused test in `sdk`.

### DIFF
--- a/packages/node/src/plugins/operator/OperatorPlugin.ts
+++ b/packages/node/src/plugins/operator/OperatorPlugin.ts
@@ -1,7 +1,4 @@
 import {
-    OperatorDiscoveryRequest,
-    OperatorDiscoveryResponse,
-    peerDescriptorTranslator,
     ReviewRequestEvent,
     SignerWithProvider,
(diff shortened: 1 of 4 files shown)


harbu (approved):
LGTM; Regarding the open question about whether to include this change in the CHANGELOG, I would say no need as this is a internal affair

PR #2945, opened 2024-12-12
feat(docs): Hide internal methods from API docs

teogeb:
Added `@hidden` TypeDoc annotation for internal methods:
- `getNode()`
- `getConfig()`

This way these methods are not included in the API docs. The internal `StreamrClient#getOperator()` was already annotated like this.

Removed redundant `@internal` annotations from the `NetworkNodeStub` as it is not used by any publicly available method anymore:
- also modified `NetworkNodeFacade#getNode()` to return `Promise<Omit<NetworkNodeStub, 'start' | 'stop'>>` as a consequence of removing the `@internal` annotations
- also removed unnecessary `@internal` annotation from `NetworkNodeFacade#setProxies()`

## Note

It would be possible annotate those method with `@internal`. But then the methods wouldn't be included to the  generated `.d.ts` files (as we use `"stripInternal": true` in `tsconfig.node.json`). The type definitions are needed by `node` and `cli-tools` packages which use these internal methods.

### DIFF
--- a/packages/sdk/src/StreamrClient.ts
+++ b/packages/sdk/src/StreamrClient.ts
@@ -652,6 +652,7 @@ export class StreamrClient {
 
     /**
      * @deprecated This in an internal method
+     * @hidden
      */
     getNode(): NetworkNodeFacade {
         return this.node
@@ -748,6 +749,7 @@ export class StreamrClient {
 
     /**
      * @deprecated This in an internal method
+     * @hidden
      */
     getConfig(): StrictStreamrClientConfig {
         return this.config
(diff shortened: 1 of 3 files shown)


packages/sdk/src/StreamrClient.ts
```diff
@@ -748,6 +749,7 @@ export class StreamrClient {
 
     /**
      * @deprecated This in an internal method
+     * @hidden
      */
     getConfig(): StrictStreamrClientConfig {
```
harbu, line 754: Is there a good reason we've made this internal? I think it could be useful in practice to see the config that is in use?

harbu (approved):

packages/sdk/src/StreamrClient.ts
teogeb: We can make it public in a follow-up pr if needed. Earlier discussion: https://github.com/streamr-dev/network/pull/1859#discussion_r1323122647

PR #2946, opened 2024-12-13
feat(cli-tools): add binary support to mock data generation

harbu:
## Summary

Add new command for generating random binary data `streamr mock-data generate --binary`. Output is newline-delimited hexadecimal strings.


teogeb (commented):
Would it make sense to merge this functionality to the existing `generate` command? 

I.e. we'd add `--binary` option to that command.

The length could be defined e.g. with `--content-length 123-456` option. (Maybe applicable also for non-binary generator, or it could error if used without `--binary`.)

harbu:
> Would it make sense to merge this functionality to the existing `generate` command?
> 
> I.e. we'd add `--binary` option to that command.
> 
> The length could be defined e.g. with `--content-length 123-456` option. (Maybe applicable also for non-binary generator, or it could error if used without `--binary`.)

I thought about the `--binary` option route, however I didn't want to make an exception where `--maxLength` and `--minLength` only work with binary data. (Generating arbitrary hierarchical JSON objects of a certain length is not trivial).

I like the command line parameter values to be simple so opted not for any hyphenated version of giving content length.

packages/cli-tools/bin/streamr-mock-data-generate.ts
```diff
@@ -24,15 +28,29 @@ export const generate = (rate: number): void => {
...
 createCommand()
-    .description('generate and print semi-random JSON data to stdout')
+    .description('generate and print semi-random JSON data or random binary data (hexadecimal) to stdout')
     .option('-r, --rate <n>', 'rate in milliseconds', createFnParseInt('--rate'), 500)
+    .option('--binary', 'generate binary data instead of JSON')
+    .option('--min-length <n>', 'minimum message length in bytes', createFnParseInt('--min-length'), 32)
+    .option('--max-length <n>', 'maximum message length in bytes', createFnParseInt('--max-length'), 64)
     .action((options: Options) => {
-        generate(options.rate)
+        if (options.binary) {
```
teogeb: Move `setInterval` to here so that we don't duplicate it in both `generate*()` functions

packages/cli-tools/bin/streamr-mock-data-generate.ts
```diff
@@ -2,9 +2,13 @@
 import { createFnParseInt } from '../src/common'
 import { createCommand, Options as BaseOptions } from '../src/command'
 import { randomString } from '@streamr/utils'
+import crypto from 'crypto'
 
 interface Options extends BaseOptions {
     rate: number
+    binary: boolean
+    minLength?: number
+    maxLength?: number
```
teogeb: Can these be  `undefined` as there are default values in `.option(...)` definitions?

CHANGELOG.md
```diff
@@ -110,6 +110,8 @@ Changes before Tatum release are not documented in this file.
 
 #### Added
 
+- Add new command `streamr mock-data generate-binary` for generating random binary data (hexadecimal) (https://github.com/streamr-dev/network/pull/2946)
```
teogeb:
```suggestion
- Add binary content support to `streamr mock-data generate` command (https://github.com/streamr-dev/network/pull/2946)
```

packages/cli-tools/bin/streamr-mock-data-generate.ts
harbu: Fixed

packages/cli-tools/bin/streamr-mock-data-generate.ts
harbu: Fixed

teogeb (approved):

PR #2957, opened 2024-12-20
test: Use test token faucet instead of `KeyServer`

teogeb:
## Changes

The `fetchPrivateKeyWithGas()` test utility now uses `generateWalletWithGasAndTokens()` internally to fetch a `Wallet` with gas.

The `KeyServer` was removed as it is no longer needed.

## Background

Before this PR `fetchPrivateKeyWithGas()` used a separate `KeyServer` HTTP server which we started and stopped in test setup/teardown. There are 1000 pre-baked accounts in the `dev-chain-fast` Docker image, and the key server provided one of those accounts to the caller.

After this PR we don't need the pre-baked accounts. 

## Motivation

There was no need to have two separate approaches for providing gas for test tokens. It is better to use the `generateWalletWithGasAndTokens()` instead of pre-baked accounts so that the tests rely less on the pre-defined state of the test environment. Also now we don't need a separate background service. 

As a bonus the test setups are simpler and tests run bit faster.

## Benchmark

This change decreased the average test run time:
- total runtime: ~3210s -> ~3150s
- slowest (i.e. `node-e2e`): ~360s -> ~354s

 There is quite a lot variation in the test runs durations. These results are an average of 10 test runs.

## Future improvements

- We could merge `fetchPrivateKeyWithGas()` and `generateWalletWithGasAndTokens()`. The function could called e.g. `createTestWallet()` and have `opts` `{ gas: boolean, token: boolean }`. 

- Maybe `retry()` logic is not needed in `generateWalletWithGasAndTokens()`?


harbu (approved):

harbu:
> We could merge fetchPrivateKeyWithGas() and generateWalletWithGasAndTokens(). The function could called e.g. createTestWallet() and have opts { gas: boolean, token: boolean }.

Yeah this would be good

> Maybe retry() logic is not needed in generateWalletWithGasAndTokens()?

If I recall correctly this was needed due to concurrency concerns and nonce issues since the same admin wallet is being used

PR #2961, opened 2025-01-07
feat(DHT): NET-1382 add application version to handshake

juslesan:
## Summary

Added `applicationVersion` field to the Handshake request and response. Renamed previous `version` field to `protocolVersion`

## Changes

- NodeInfoEndpoint: the `version` field is now `applicationVersion` 
- Changed naming in convention `WebSocketServerConnector#attachHandshaker` from source/target to remote local


packages/dht/src/connection/Handshaker.ts
```diff
@@ -6,6 +6,7 @@ import { IConnection } from './IConnection'
 import { LOCAL_PROTOCOL_VERSION, isMaybeSupportedVersion } from '../helpers/version'
 import { toNodeId } from '../identifiers'
 import { PendingConnection } from './PendingConnection'
+import { version } from '../../package.json'
```
teogeb:
```suggestion
import { version as applicationVersion } from '../../package.json'
```

teogeb (changes requested):
We could change all occurrences of `version` to `protocolVersion` as it may otherwise be unclear which data is stored to that field/parameter:
- `ConnectivityResponse#version`
- `isMaybeSupportedVersion`
- `HandshakerEvents#handshakeRequest` parameter
- `HandshakeError#UNSUPPORTED_VERSION` (maybe makes sense to have `UNSUPPORTED_PROTOCOL_VERSION` now and a separate `UNSUPPORTED_APPLICATION_VERSION` in the future if needed)

Maybe also `NodeInfoResponse#version` -> `applicationVersion`

juslesan:
> Maybe also `NodeInfoResponse#version` -> `applicationVersion`

Is this necessary since the NodeInfo-endpoint only cares about the application version?

teogeb:
> > Maybe also `NodeInfoResponse#version` -> `applicationVersion`
> 
> Is this necessary since the NodeInfo-endpoint only cares about the application version?

Otherwise it is bit unclear which version this field contains.

packages/dht/src/connection/websocket/WebsocketServerConnector.ts
```diff
@@ -114,23 +114,27 @@ export class WebsocketServerConnector {
     private attachHandshaker(connection: IConnection) {
         // TODO: use createIncomingHandshaker here?
         const handshaker = new Handshaker(this.localPeerDescriptor!, connection)
-        handshaker.once('handshakeRequest', (localPeerDescriptor: PeerDescriptor, sourceVersion: string, remotePeerDescriptor?: PeerDescriptor) => {
-            this.onServerSocketHandshakeRequest(localPeerDescriptor, connection, handshaker, sourceVersion, remotePeerDescriptor)
+        handshaker.once('handshakeRequest', (
+            localPeerDescriptor: PeerDescriptor,
+            remoteProtocolVersion: string,
```
teogeb, line 119: The parameter name was `source...` before this PR, now `remote...`. Is this an intentional change?

(If it is, should we rename `source`->`remote` in some other places, too? In a separate PR maybe.)

teogeb (changes requested):

packages/dht/src/connection/websocket/WebsocketServerConnector.ts
juslesan: Yeah this was an intentional change. We have been using the remote/local naming convention to differentiate between different ends of an exchange

juslesan: Source can be used in cases where there is no clear interaction between two parties. Examples would be routing and recursive operations

teogeb: Are there some usage of `source` naming convention where the interaction is clear and therefore `local`/`remote` would be better?  (e.g. `WebsocketServerConnector`?)

juslesan: https://linear.app/streamr/issue/NET-1392/unify-naming-between-remotelocal-and-source-patterns

juslesan: Yes in the connectors it makes sense to use remote / local, as all individual interactions are between two parties

teogeb (approved):

PR #2967, opened 2025-01-08
chore: Convert Jest configs into typescript

mondoreale:
## Summary

I convert Jest config files into typescript for consistency.

## Changes

- Convert all jest.config.js to jest.config.ts.
- Convert associated setup and teardown files from js to ts, too.
- Changed the `ts-jest` preset to the default, which is best suited for projects written entirely in ts.
- Fix a proto-related warning so that trackerless-network build is clean (unrelated change).

- n/a


juslesan (approved):

jest.config.ts
```diff
@@ -0,0 +1,12 @@
+import type { Config } from '@jest/types'
+
+const config = {
+    transform: {
```
teogeb, line 5: Is it intentional that these files are different? No preset and no explicit tsconfig.

teogeb (changes requested):
If the content changes are intentional, could mention those in the PR description.

jest.config.ts
mondoreale: I've simply gone with defaults across the board, but I hear you, let's be explicit. The things you mention are now back in. Few notes:

- For a ts-only codebase like ours the default preset is optimal. I swapped.
- All projects w/ tests come with tsconfig.json but since they all also come with designated tsconfig.jest.json files I re-added the line.

teogeb: Was the `modulePathIgnorePatterns` redundant? No big difference in the `^.+\\.ts$'` regexp, but the original was slightly better.

packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -3,7 +3,6 @@ syntax = "proto3";
 option optimize_for = CODE_SIZE;
 
 import "google/protobuf/empty.proto";
-import "google/protobuf/any.proto";
```
teogeb, line 6: Good improvement 👍 Maybe worth mentioning in the PR description as this isn't strictly related to this PR.

jest.config.ts
mondoreale: Old regexp is indeed better. I'll revert. Good catch.

As for the ignored `dist`: new preset only processes ts files and I believe there are no logic-carrying ts files in dist, making ignoring it redundant.

(I've also noticed that the regexp itself excludes js. Funny.)

teogeb (approved):

PR #2972, opened 2025-01-13
chore: Formatted with prettier

mondoreale:
## Summary

Format all the codes using Prettier and its refreshed config from `prettier` branch.

@teogeb, [attached is](https://github.com/user-attachments/files/18396471/post-prettier-eslint.log) the result of `npm run eslint` with `indent` enabled.

I've tweaked the branch so that it's the only outstanding rule we have to deal with.

(Note that in this PR the `indent` rule is on.)

Cheers!

## Changes

- Auto-formatted all the files.
- Adjusted selected eslint rules to match Prettier's behaviour:
  - Enabled `avoidEscape` on the `quotes` rule.
  - Changed `@stylistic/member-delimiter-style` to `semi` for singleline types and interfaces (`semi` is typescript's default, too).

- n/a


packages/cli-tools/test/environment.test.ts
```diff
@@ -22,11 +24,15 @@ describe('env command line option', () => {
...
+        'invalid value',
+        async () => {
+            const outputLines = await runCommand('stream show foobar --env invalid-environment', {
+                devEnvironment: false
+            })
+            expect(outputLines).toHaveLength(1)
+            expect(outputLines[0]).toEqual('env must be one of: "polygon", "polygonAmoy", "dev2"')
+        },
+        60 * 1000
+    )
```
mondoreale, line 37: 3rd argument (the timeout) forces Prettier to put the label onto its own line.

mondoreale (commented):
@harbu @teogeb I've reformatted the files using 120 long print width. It definitely makes certain things look much better.

That `it()` formatted differently is a result of passing 3rd argument to it, i.e. `it(label, fn, TIMEOUT)`. It then places the label in its own line. Pinned down one example of it. It'll be applied consistently to all similar scenarios if needed.

packages/utils/src/Metric.ts
```diff
@@ -210,33 +210,37 @@ export class MetricsContext {
...
     ): void {
         const ongoingSamples: Map<string, Sampler> = new Map()
-        scheduleAtFixedRate(async (now: number) => {
-            if (ongoingSamples.size > 0) {
-                const report = {
-                    period: {
-                        start: now - interval,
-                        end: now
+        scheduleAtFixedRate(
+            async (now: number) => {
```
harbu, line 217: I wonder why this was put on a newline (but not the arguments for the callbacks below). Is it because of the `async`  keyword?

packages/utils/src/StreamID.ts
```diff
@@ -20,9 +20,11 @@ export function toStreamID(streamIdOrPath: string, domain?: EthereumAddress | EN
         throw new Error('stream id may not be empty')
     }
     const firstSlashIdx = streamIdOrPath.indexOf('/')
-    if (firstSlashIdx === -1) { // legacy format
+    if (firstSlashIdx === -1) {
+        // legacy format
         return streamIdOrPath as StreamID
-    } else if (firstSlashIdx === 0) { // path-only format
+    } else if (firstSlashIdx === 0) {
+        // path-only format
```
harbu, line 27: Not a fan of this comment enforcement. I think with comments there should be freedom to put them wherever to make it very clear what is reference.

packages/utils/src/waitForEvent3.ts
```diff
@@ -86,19 +85,16 @@ export function raceEvents3<T extends EventEmitter.ValidEventTypes>(
...
-            timeout,
-            'raceEvents3'
-        ).finally(() => {
-            cancelAll()
-        })
+        return withTimeout(Promise.race(promises.map((promise) => promise.task)), timeout, 'raceEvents3').finally(
+            () => {
+                cancelAll()
+            }
+        )
```
harbu, line 92: A lot going on in this statement. I sort of prefer the "before" here due to the complexity being broken down into lines.

packages/utils/test/Logger.test.ts
```diff
@@ -80,26 +88,23 @@ describe('Logger', () => {
 
     describe('name', () => {
         it('short', () => {
-            const expected = typeof _streamr_electron_test === 'undefined'
-                ? 'Logger.test              '
-                : 'Logger.test'
+            const expected = typeof _streamr_electron_test === 'undefined' ? 'Logger.test              ' : 'Logger.test'
```
harbu, line 91: Special case but here previous formatting is superior

packages/utils/test/Logger.test.ts
```diff
@@ -80,26 +88,23 @@ describe('Logger', () => {
...
             expect(Logger.createName(module)).toBe(expected)
         })
         it('application id', () => {
-            const expected = typeof _streamr_electron_test === 'undefined'
-                ? 'APP:Logger.test          '
-                : 'APP:Logger.test'
+            const expected =
+                typeof _streamr_electron_test === 'undefined' ? 'APP:Logger.test          ' : 'APP:Logger.test'
```
harbu, line 96: Special case but here previous formatting is superior

packages/utils/test/Logger.test.ts
```diff
@@ -80,26 +88,23 @@ describe('Logger', () => {
...
             delete process.env.STREAMR_APPLICATION_ID
         })
         it('index', () => {
-            const expected = typeof _streamr_electron_test === 'undefined'
-                ? 'mock                     '
-                : 'mock'
-            expect(Logger.createName({
-                id: ['foo', 'bar', 'mock', 'index'].join(path.sep)
-            } as any)).toBe(expected)
+            const expected = typeof _streamr_electron_test === 'undefined' ? 'mock                     ' : 'mock'
```
harbu, line 102: Special case but here previous formatting is superior

packages/utils/test/retry.test.ts
```diff
@@ -1,36 +1,20 @@
...
         expect(onRetryableFailure).not.toHaveBeenCalled()
     })
 
     it('non-first success', async () => {
         const error = new Error('mock-error')
-        const task = jest.fn()
-            .mockRejectedValueOnce(error)
-            .mockRejectedValueOnce(error)
-            .mockResolvedValue(123)
+        const task = jest.fn().mockRejectedValueOnce(error).mockRejectedValueOnce(error).mockResolvedValue(123)
```
harbu, line 15: Not a fan of this formatting (it's effectively a list we are talking about)

packages/trackerless-network/karma.config.js
```diff
@@ -2,21 +2,25 @@
 const path = require('path')
 const { createKarmaConfig, createWebpackConfig } = require('@streamr/browser-test-runner')
 
-const TEST_PATHS = [
-    './test/unit/**/*.ts',
-    './test/integration/**/*.ts',
-    './test/end-to-end/**/!(webrtc*|websocket*)',
-]
+const TEST_PATHS = ['./test/unit/**/*.ts', './test/integration/**/*.ts', './test/end-to-end/**/!(webrtc*|websocket*)']
```
harbu, line 5: Not a fan of this formatting

packages/trackerless-network/src/logic/DiscoveryLayerNode.ts
```diff
@@ -13,9 +13,15 @@ export interface DiscoveryLayerNodeEvents {
 export interface DiscoveryLayerNode {
     on<T extends keyof DiscoveryLayerNodeEvents>(eventName: T, listener: (peerDescriptor: PeerDescriptor) => void): void
     on<T extends keyof DiscoveryLayerNodeEvents>(eventName: T, listener: () => void): void
-    off<T extends keyof DiscoveryLayerNodeEvents>(eventName: T, listener: (peerDescriptor: PeerDescriptor) => void): void
+    off<T extends keyof DiscoveryLayerNodeEvents>(
```
harbu, line 16: Inconsistency w.r.t. before formatting

packages/trackerless-network/src/logic/DiscoveryLayerNode.ts
```diff
@@ -13,9 +13,15 @@ export interface DiscoveryLayerNodeEvents {
...
+        eventName: T,
+        listener: (peerDescriptor: PeerDescriptor) => void
+    ): void
     off<T extends keyof DiscoveryLayerNodeEvents>(eventName: T, listener: () => void): void
-    once<T extends keyof DiscoveryLayerNodeEvents>(eventName: T, listener: (peerDescriptor: PeerDescriptor) => void): void
+    once<T extends keyof DiscoveryLayerNodeEvents>(
```
harbu, line 21: Inconsistency w.r.t. before formatting

packages/trackerless-network/src/logic/NodeList.ts
```diff
@@ -13,12 +13,10 @@ const getValuesOfIncludedKeys = (
...
     wsOnly = false
 ): ContentDeliveryRpcRemote[] => {
-    const values = wsOnly 
+    const values = wsOnly
         ? Array.from(nodes.entries()).filter(([_, node]) => node.getPeerDescriptor().websocket !== undefined)
         : Array.from(nodes.entries())
-    return values
-        .filter(([id]) => !exclude.includes(id))
-        .map(([_id, node]) => node)
+    return values.filter(([id]) => !exclude.includes(id)).map(([_id, node]) => node)
```
harbu, line 19: Too much going on in this one line, preferred the old formatting

packages/trackerless-network/test/end-to-end/inspect.test.ts
```diff
@@ -98,19 +104,16 @@ describe('inspect', () => {
...
         )
     }, 30000)
 
     afterEach(async () => {
-        await Promise.all([
-            publisherNode.stop(),
-            inspectedNode.stop(),
-            inspectorNode.stop()
-        ])
+        await Promise.all([publisherNode.stop(), inspectedNode.stop(), inspectorNode.stop()])
```
harbu, line 116: I do prefer the earlier formatting (due to the "listness")

packages/trackerless-network/test/end-to-end/proxy-and-full-node.test.ts
```diff
@@ -103,23 +110,53 @@ describe('proxy and full node', () => {
...
         ])
 
-        expect(proxiedNode.stack.getContentDeliveryManager().getStreamPartDelivery(proxiedStreamPart)!.proxied).toBe(true)
-        expect(proxiedNode.stack.getContentDeliveryManager().getStreamPartDelivery(regularStreamPart1)!.proxied).toBe(false)
+        expect(proxiedNode.stack.getContentDeliveryManager().getStreamPartDelivery(proxiedStreamPart)!.proxied).toBe(
+            true
+        )
+        expect(proxiedNode.stack.getContentDeliveryManager().getStreamPartDelivery(regularStreamPart1)!.proxied).toBe(
+            false
+        )
```
harbu, line 118: Weird line break...

packages/trackerless-network/test/end-to-end/proxy-and-full-node.test.ts
```diff
@@ -133,11 +170,20 @@ describe('proxy and full node', () => {
...
+        expect(proxiedNode.stack.getContentDeliveryManager().getStreamPartDelivery(regularStreamPart2)!.proxied).toBe(
+            false
+        )
+        expect(proxiedNode.stack.getContentDeliveryManager().getStreamPartDelivery(regularStreamPart3)!.proxied).toBe(
+            false
+        )
+        expect(proxiedNode.stack.getContentDeliveryManager().getStreamPartDelivery(regularStreamPart4)!.proxied).toBe(
+            false
+        )
```
harbu, line 187: Weird line breaks

packages/trackerless-network/test/end-to-end/proxy-connections.test.ts
```diff
@@ -31,32 +31,37 @@ const MESSAGE: StreamMessage = {
...
     }
-    
+
     const hasConnectionToProxy = (proxyNodeId: DhtAddress, direction: ProxyDirection): boolean => {
-        const client = (proxiedNode.stack.getContentDeliveryManager().getStreamPartDelivery(STREAM_PART_ID) as { client: ProxyClient }).client
+        const client = (
+            proxiedNode.stack.getContentDeliveryManager().getStreamPartDelivery(STREAM_PART_ID) as {
+                client: ProxyClient
+            }
+        ).client
```
harbu, line 64: Kinda weird formatting

packages/trackerless-network/test/integration/ContentDeliveryLayerNode-Layer1Node-Latencies.test.ts
```diff
@@ -20,8 +20,8 @@ describe('ContentDeliveryLayerNode-DhtNode-Latencies', () => {
     beforeEach(async () => {
         const simulator = new Simulator(LatencyType.FIXED, 50)
         const entrypointCm = new SimulatorTransport(entrypointDescriptor, simulator)
-        const cms: SimulatorTransport[] = range(otherNodeCount).map((i) =>
-            new SimulatorTransport(peerDescriptors[i], simulator)
+        const cms: SimulatorTransport[] = range(otherNodeCount).map(
+            (i) => new SimulatorTransport(peerDescriptors[i], simulator)
```
harbu, line 24: Not a fan of this line break

packages/trackerless-network/test/integration/ContentDeliveryLayerNode-Layer1Node-Latencies.test.ts
```diff
@@ -32,20 +32,25 @@ describe('ContentDeliveryLayerNode-DhtNode-Latencies', () => {
...
+        otherDiscoveryLayerNodes = range(otherNodeCount).map(
+            (i) =>
+                new DhtNode({
+                    transport: cms[i],
+                    connectionsView: cms[i],
+                    peerDescriptor: peerDescriptors[i],
+                    serviceId: streamPartId
+                })
+        )
+        otherContentDeliveryLayerNodes = range(otherNodeCount).map((i) =>
```
harbu, line 44: Inconsistency with parameter list line break...

packages/trackerless-network/test/integration/stream-without-default-entrypoints.test.ts
```diff
@@ -85,10 +93,7 @@ describe('stream without default entrypoints', () => {
         nodes[0].addMessageListener((_msg) => {
             receivedMessageCount += 1
         })
-        await Promise.all([
-            until(() => receivedMessageCount === 1, 10000),
-            nodes[1].broadcast(streamMessage)
-        ])
+        await Promise.all([until(() => receivedMessageCount === 1, 10000), nodes[1].broadcast(streamMessage)])
```
harbu, line 96: Imo too much going on in one line

packages/trackerless-network/tsconfig.browser.json
```diff
@@ -5,9 +5,5 @@
     "rootDirs": ["src", "test"],
     "noImplicitOverride": false
   },
-  "include": [
-    "src/**/*",
-    "generated/**/*",
-    "test/**/*"
-  ]
+  "include": ["src/**/*", "generated/**/*", "test/**/*"]
```
harbu, line 8: Kind of like the earlier, ngl

packages/trackerless-network/tsconfig.jest.json
```diff
@@ -1,17 +1,9 @@
...
-        { "path": "../proto-rpc/tsconfig.node.json" },
-        { "path": "../dht/tsconfig.node.json" }
-    ]
+  "extends": "../../tsconfig.jest.json",
+  "compilerOptions": {
+    "noEmit": true,
+    "noImplicitOverride": false
+  },
+  "include": ["src/**/*", "generated/**/*", "test/**/*", "package.json"],
+  "references": [{ "path": "../proto-rpc/tsconfig.node.json" }, { "path": "../dht/tsconfig.node.json" }]
```
harbu, line 8: I don't find this as redable..

packages/trackerless-network/tsconfig.node.json
```diff
@@ -1,17 +1,9 @@
...
-    "references": [
-        { "path": "../dht/tsconfig.node.json" }
-    ]
+  "extends": "../../tsconfig.node.json",
+  "compilerOptions": {
+    "outDir": "dist",
+    "noImplicitOverride": false
+  },
+  "include": ["src/**/*", "generated/**/*", "test/benchmark/first-message.ts", "test/utils/utils.ts", "package.json"],
+  "references": [{ "path": "../dht/tsconfig.node.json" }]
```
harbu, line 8: I don't find this as readable

packages/utils/src/Metric.ts
mondoreale: Because of the presence of the 2nd and the 3rd arguments given to `scheduleAtFixedRate`.

```ts
scheduleAtFixedRate(
    async (now: number) => { /* code */ },
    interval,
    abortSignal
)
```

packages/utils/src/StreamID.ts
mondoreale: It wouldn't touch a comment placed immediately after a statement, e.g.

```ts
if (firstSlashIdx === -1) {
    return streamIdOrPath as StreamID // something here about what just happened
}
```

To me placing a comment immediately after `{` feels a bit strange, generally, so I'm for this behaviour. FYI.

packages/utils/src/waitForEvent3.ts
mondoreale: Those long lines aren't ideal. Maybe smaller print width?

On the other hand... `raceEvents3` could be organically async making it possible to restructure it like so:

```ts
export async function raceEvents3<T extends EventEmitter.ValidEventTypes>(
    emitter: EventEmitter<T>,
    eventNames: (keyof T)[],
    timeout: number | null = 5000
): Promise<RunAndRaceEventsReturnType<T>> {
    // …

    try {
        const taskRace = Promise.race(promises.map((promise) => promise.task))

        if (timeout !== null) {
            return await withTimeout(taskRace, timeout, 'raceEvents3')
        }

        return await taskRace
    } finally {
        promises.forEach((promise) => {
            promise.cancel()
        })
    }
}
```

So it's just possible if it was new code this formatting issue simply wouldn't occur.

packages/utils/test/Logger.test.ts
mondoreale: Agreed.

Maybe if someone wrote it \*now* (accompanied by a formatter) they'd do something like this instead:

```ts
describe('name', () => {
    function pad(str: string) {
        return typeof _streamr_electron_test === 'undefined' ? str.padEnd(25) : str
    }

    it('short', () => {
        expect(Logger.createName(module)).toBe(pad('Logger.test'))
    })

    it('application id', () => {
        process.env.STREAMR_APPLICATION_ID = 'APP'
        expect(Logger.createName(module)).toBe(pad('APP:Logger.test'))
        delete process.env.STREAMR_APPLICATION_ID
    })

    it('index', () => {
        expect(
            Logger.createName({
                id: ['foo', 'bar', 'mock', 'index'].join(path.sep)
            } as any)
        ).toBe(pad('mock'))
    })
})
```

packages/trackerless-network/src/logic/DiscoveryLayerNode.ts
mondoreale: Yeah, that's the most annoying part. Prettier doesn't make sibling nodes consistent with each other.

No way around it.

packages/trackerless-network/src/logic/NodeList.ts
mondoreale: Yep, old one reads better.

On the other hand, chaining tends to get ugly quickly, and very often is also a waste of processing effort. Here for example you don't need multiple passes. Consider this

```ts
const getValuesOfIncludedKeys = (
    nodes: Map<DhtAddress, ContentDeliveryRpcRemote>,
    exclude: DhtAddress[],
    wsOnly = false
): ContentDeliveryRpcRemote[] => {
    const remotes: ContentDeliveryRpcRemote[] = []

    for (const [id, node] of nodes.entries()) {
        if (wsOnly && node.getPeerDescriptor().websocket === undefined) {
            continue
        }

        if (exclude.includes(id)) {
            continue
        }

        remotes.push(node)
    }

    return remotes
}
```

Few benefits
- Single pass
- No extra array allocation (`Array.from`)
- Fairly clean
- Very little room for the formatter to mess up

Plus, to me, it feels much cleaner.

packages/trackerless-network/test/end-to-end/proxy-and-full-node.test.ts
mondoreale: DRY-ing it up a little bit could help but yeah, I'm with you.

packages/trackerless-network/test/end-to-end/proxy-connections.test.ts
mondoreale: I'd throw a rough guess that the type casting doesn't help. 😉

packages/trackerless-network/tsconfig.browser.json
mondoreale: I believe one can simply write

```
["src", "generated", "test"]
```

A matter of preference. I don't mind so I'd go with the flow.

mondoreale (commented):
@harbu thanks for the commentary.

I agree with a lot of it. Having said that, I believe that with a consistent formatter in place most of what's there would've been written differently.

Left a couple of notes on your notes.

Cheers!

harbu:
So I did a bit of a readthru & analysis of the prettier formatting.

- I went thru all the code in packages `utils` and `trackeless-network` .
- I excluded changes related to white space, trailing commas, and interface/type semicolon (those can easily be enforced with any tool)
 - For the remaining changes, I categorized and tallied them into the three categories based on readability
        - Good: change improved readability
        - Neutral: change neither improved or worsened readability
        - Bad: change worsened readability

So for the scores:
**utils**

    Good 24 (23%)
    Neutral 74 (70%)
    Bad 7 (7%)

**trackerless-network**

    Good 51 (19%)
    Neutral 198 (74%)
    Bad 19 (7%)

PR #2976, opened 2025-01-17
NET-1383: Reduce the size of the routing tables to 5 for performance

juslesan:
## Summary

Reduce the size of the routing tables for performance. This reduces used memory in the cache and CPU when constructing the routing tables.


ptesavol (approved):
LGTM

PR #2979, opened 2025-01-20
feat(network): limit pings in layer1 NET-1384

juslesan:
## Summary

Limit the number of neighbors to ping when they are added as neighbors. This fixes a bad looping behaviour during joins to layer1 DHTs

## Side effects

While all nodes during the start of the joins are pinged, after 16 neighbors have been added there is a chance that new neighbors are offline. This will be resolved eventually with the use of periodic pings to all neighbors. However it can lead to issues with layer2 graph generation if multiple nodes leave the network ungracefully in a small time frame.

### DIFF
--- a/.github/workflows/test-setup.yml
+++ b/.github/workflows/test-setup.yml
@@ -25,6 +25,10 @@ on:
       extra-log-prefix:
         required: false
         type: string
+      browser:
+        required: false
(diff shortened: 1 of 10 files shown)


ptesavol (approved):
LGTM

PR #2980, opened 2025-01-20
fix(test): [NET-1400] Electron binary permissions

teogeb:
## Changes

The CI jos changes the properties of the Electron's `chrome-sandbox` binary:
- owner to`root`
- file mode to 4755 -> it is run by the owner of the file (i.e. `root`) and with full permissions

## Problem

There has been a change in the CI environment which has affected our browser tests. From ~2025-01-16 onwards all browser test started to fail with this error:

```
Electron stderr: [2074:0120/103701.092234:FATAL:setuid_sandbox_host.cc(158)]
The SUID sandbox helper binary was found, but is not configured correctly.
Rather than run without sandboxing I'm aborting now. You need to make sure that
/home/runner/work/network/network/node_modules/electron/dist/chrome-sandbox
is owned by root and has mode 4755.
```

## Root cause

The root cause is that the GitHub has upgraded CI runner to use Ubuntu v24.04 instead of v22.04 (when `ubuntu-latest` is used, like we do). Our failures started during the rollout period of that change: https://github.com/actions/runner-images/issues/10636

The [9b426ef](https://github.com/streamr-dev/network/pull/2980/commits/9b426ef218bdac9e29e84f2611348e24c5eac5be) in this PR confirms the root cause: when we didn't have the fix in place, the the tests passed when `ubuntu-22.04` was used, and failed when `ubuntu-24.02` was used.

### Relevant resources

- GitHub's Ubuntu 24.04 image: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md
- the code which prints the error (in Chromium): https://chromium.googlesource.com/chromium/src/sandbox/+/refs/heads/main/linux/suid/client/setuid_sandbox_host.cc#161

### DIFF
--- a/.github/workflows/close-stale-issue-and-pr.yaml
+++ b/.github/workflows/close-stale-issue-and-pr.yaml
@@ -7,7 +7,7 @@ on:
 jobs:
   stale:
     name: Close stale issues and PRs
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/stale@v9
         with:
--- a/.github/workflows/close-stale-issue-and-pr.yaml
+++ b/.github/workflows/close-stale-issue-and-pr.yaml
@@ -7,7 +7,7 @@ on:
 jobs:
   stale:
     name: Close stale issues and PRs
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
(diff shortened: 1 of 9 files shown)


juslesan (approved):

harbu (approved):
Nice work :+1:

PR #2981, opened 2025-01-21
deps [NET-1401]: update to Node.js 22

harbu:
## Summary

Update Node.js to active LTS version 22. As per custom we maintain backwards compatibility (somewhat) to previous LTS 20.

In practice this means that
- We prefer v22 and use it for the Docker image and _.nvmrc_
- We still support v20 therefore we have it set for Node v20 types and tsconfig base ("lowest common denominator")

### Future improvements
- Update dependencies and replace `request` package with something else in autocertifier packages to get rid of `punycode` transitive dependency and the deprecation warning. (See NET-1409 and NET-1404 for more information)


packages/cli-tools/test/utils.ts
```diff
@@ -18,7 +18,8 @@ export const runCommand = async (commandLine: string, opts?: StartCommandOptions
 }
 
 export async function* startCommand(commandLine: string, opts?: StartCommandOptions): AsyncGenerator<string> {
-    const args: string[] = ['dist/bin/streamr.js']
+    // TODO: --no-deprecation needed to get around deprecation warning for "punycode" in Node.js 22, remove when warning has gone away
```
teogeb: link to https://linear.app/streamr/issue/NET-1409/get-rid-of-punycode-transitive-dependency-in-network-monorepo

teogeb (commented):
Maybe the PR description would be even more explicit about there current approach:
- we prefer v22, therefore we use v22 for the Docker image and `.nvmrc`
- we require v20, therefore we use Node v20 types and `tsconfig` base

packages/cli-tools/test/utils.ts
harbu:
```suggestion
    // TODO: --no-deprecation needed to get around deprecation warning for "punycode" in Node.js 22, remove when warning has gone away (NET-1409)
```

teogeb (approved):

PR #2984, opened 2025-01-21
refactor(test-utils): Unify `generateWalletWithGasAndTokens()` and `fetchPrivateKeyWithGas()`

teogeb:
Unified the two helper functions to have same method signature and renamed the functions.

Now there is a `createTestWallet(opts)` function, and a `createTestPrivateKey(opts)` function. The functionality is in `createTestWallet(opts)`, and `createTestPrivateKey(opts)` is just simple helper which we can use when we only need the private key.

Both functions have opts parameter which defines whether we the wallet / private key should have gas and/or tokens associated with it.

## Future improvements

- Maybe we could replace usages of `fastWallet() `and `fastPrivate()` with these functions
- We don't usually need the wallet to have a provider connected to it: this can be optimized in the future, if needed.


jtakalai (approved):
probably reconsider all names that talk about "gas".

(fetchPrivateKeyWithGas and generateWalletWithGasAndTokens are replaced, that's good)

`gas: true` maybe `gasTokens: true`.

Maybe not part of this ticket, let's make another one for namings, perhaps as part of documentation tickets.

PR #2985, opened 2025-01-22
feat(cli-tools): [NET-1405] Add internal `token-mint` command

teogeb:
Added an internal command for minting test tokens for given address. It can also transfer gas to the address.

Works only in `dev2` environment.

Usage:
```
/streamr internal token-mint 0x1234567890123456789012345678901234567890 123.4 5.678
```


jtakalai (approved):

jtakalai (commented):
probably rename gasAmount -> nativeTokenAmount

PR #2993, opened 2025-01-28
test: [NET-1420] Use `createTestWallet()` instead of `fastWallet()`

teogeb:
Also use `createTestPrivateKey()` instead of `fastPrivateKey()`. This way we have cleaner test API.

## Refactoring

As the new methods are async, it is no longer possible to create test wallets / private keys outside of the `describe` blocks. Therefore some wallet / private key setups were moved e.g. to `beforeAll()`.

The `createClient()` and `createTestClient()` utils were simplified: the `privateKey` parameter is now optional. Also we can omit the `auth` section from the client options if just a random private key is needed (the client will autogenerate it in that case).


jtakalai (approved):
TIL StreamrClient will generate random key instead of erroring if auth is left out. I guess it makes sense; if you do read-only stuff then random auth is just as good as no auth.

PR #2994, opened 2025-01-28
feat(network): NET-1419 Configurable propagation buffer sizes

juslesan:
## Summary

Make it possible to configure the size of the propagation buffers in the network.

### DIFF
--- a/packages/trackerless-network/src/logic/createContentDeliveryLayerNode.ts
+++ b/packages/trackerless-network/src/logic/createContentDeliveryLayerNode.ts
@@ -21,6 +21,7 @@ type ContentDeliveryLayerNodeOptions = MarkOptional<StrictContentDeliveryLayerNo
         minPropagationTargets?: number
         acceptProxyConnections?: boolean
         neighborUpdateInterval?: number
+        maxPropagationBufferSize?: number
     }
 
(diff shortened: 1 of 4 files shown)


packages/trackerless-network/src/logic/createContentDeliveryLayerNode.ts
```diff
@@ -34,6 +35,7 @@ const createConfigWithDefaults = (options: ContentDeliveryLayerNodeOptions): Str
     const minPropagationTargets = options.minPropagationTargets ?? 2
     const acceptProxyConnections = options.acceptProxyConnections ?? false
     const neighborUpdateInterval = options.neighborUpdateInterval ?? 10000
+    const maxPropagationBufferSize = options.maxPropagationBufferSize ?? 150
```
harbu: The `150` here and in file ProxyClient.ts should be extracted as a shared constant. Same applies for `minPropagationTargets` and the value `2`. Alternatively, leave the default value up to `Propagation.ts` to decide?

juslesan: Yeah I made it internal now. I'll open a separate PR about moving the rest of the default configs out of here similarly.
 
It might make sense to move some of these defaults to a higher layer (sdk maybe?) and some of these can probably just be hard coded constants. But I'll go over them case by case

harbu (approved):
lgtm!

ptesavol (approved):
Looks OK

PR #2997, opened 2025-01-29
refactor(network): Default configs as constants for layer2

juslesan:
## Summary

Added constants for default configurations for Layer2 classes. 

## Changes

- Random node view count is now configured with the nodeViewSize config option

## Future improvements

- Do the same for the discovery layer #3000 
- Remove remaining magic numbers around ControlLayerNode configuration (Could also clean up on the DHT side).
- Could maybe pass some of the defaults from a higher layer (for example the sdk). This will require some thought as some of the configs could be wanted on a per stream partition basis and others for all streams.
- Rename all fields with `listSize` to ie `listNodeCount `

### DIFF
--- a/packages/trackerless-network/src/logic/ContentDeliveryLayerNode.ts
+++ b/packages/trackerless-network/src/logic/ContentDeliveryLayerNode.ts
@@ -65,7 +65,9 @@ export interface StrictContentDeliveryLayerNodeOptions {
     rpcRequestTimeout?: number
 }
 
-const RANDOM_NODE_VIEW_SIZE = 20
+export const DEFAULT_NODE_VIEW_SIZE = 20
+export const DEFAULT_NEIGHBOR_TARGET_COUNT = 4
+export const DEFAULT_ACCEPT_PROXY_CONNECTIONS = false
 
 const logger = new Logger(module)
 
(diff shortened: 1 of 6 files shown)


harbu (approved):
lgtm

packages/trackerless-network/src/logic/ContentDeliveryLayerNode.ts
```diff
@@ -65,7 +65,9 @@ export interface StrictContentDeliveryLayerNodeOptions {
     rpcRequestTimeout?: number
 }
 
-const RANDOM_NODE_VIEW_SIZE = 20
+export const DEFAULT_NODE_VIEW_SIZE = 20
```
teogeb, line 68: https://github.com/streamr-dev/network/blob/main/developer-docs/index.md#naming-counts

```suggestion
export const DEFAULT_NODE_VIEW_CONTACT_COUNT = 20
```

Could be part of a separate PR (in that case, add a bullet point to the "Future improvements" section)

packages/trackerless-network/src/logic/ContentDeliveryLayerNode.ts
```diff
@@ -284,7 +286,7 @@ export class ContentDeliveryLayerNode extends EventEmitter<Events> {
         if (this.isStopped()) {
             return
         }
-        const randomContacts = this.options.discoveryLayerNode.getRandomContacts(RANDOM_NODE_VIEW_SIZE)
+        const randomContacts = this.options.discoveryLayerNode.getRandomContacts(this.options.nodeViewSize)
```
teogeb, line 289: This is a functionality change? Maybe worth mentioning in the PR description.

juslesan: It is a slight functionality change in that the node view sizes are usually the same across all structures.

PR #3000, opened 2025-01-30
refactor(network): Default configuration constants for DiscoveryLayerNode

juslesan:
## Summary

Added constants for default configuration of the DiscoveryLayerNode.

## Future Improvements

Add default configuration configs to the ControlLayerNode and DhtNode

### DIFF
--- a/packages/trackerless-network/src/logic/DiscoveryLayerNode.ts
+++ b/packages/trackerless-network/src/logic/DiscoveryLayerNode.ts
@@ -1,4 +1,13 @@
-import { DhtAddress, PeerDescriptor, RingContacts } from '@streamr/dht'
+import { DhtAddress, PeerDescriptor, RingContacts, ServiceID } from '@streamr/dht'
+import { StreamPartID } from '@streamr/utils'
+
+export const DEFAULT_DISCOVERY_LAYER_KBUCKET_SIZE = 4
+export const DEFAULT_DISCOVERY_LAYER_JOIN_TIMEOUT = 20000
+export const DEFAULT_DISCOVERY_LAYER_NEIGHBOR_PING_LIMIT = 16
+export const DEFAULT_DISCOVERY_LAYER_PERIODICALLY_PING_NEIGHBORS = true
+export const DEFAULT_DISCOVERY_LAYER_PERIODICALLY_PING_RING_CONTACTS = true
+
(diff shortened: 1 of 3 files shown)


packages/trackerless-network/src/logic/DiscoveryLayerNode.ts
```diff
@@ -1,4 +1,13 @@
-import { DhtAddress, PeerDescriptor, RingContacts } from '@streamr/dht'
+import { DhtAddress, PeerDescriptor, RingContacts, ServiceID } from '@streamr/dht'
+import { StreamPartID } from '@streamr/utils'
+
+export const DEFAULT_DISCOVERY_LAYER_KBUCKET_SIZE = 4
```
teogeb, line 4: https://github.com/streamr-dev/network/blob/main/developer-docs/index.md#naming-counts

```suggestion
export const DEFAULT_DISCOVERY_LAYER_KBUCKET_NODE_COUNT = 4
```

Could be part of a separate PR (in that case, add a bullet point to the "Future improvements" section)

At some point we refactored `PeerManager` so that it at least partially hides the `KBucket` terminology, and uses higher level concepts instead (e.g. neighbors, nearby contacts, connections). Does this config option control any of those higher level concepts, or should we keep this name?

packages/trackerless-network/src/logic/DiscoveryLayerNode.ts
```diff
@@ -1,4 +1,13 @@
...
+export const DEFAULT_DISCOVERY_LAYER_JOIN_TIMEOUT = 20000
+export const DEFAULT_DISCOVERY_LAYER_NEIGHBOR_PING_LIMIT = 16
```
teogeb, line 6: What this limits? E.g. the count of neighbors to ping, or interval of pings, or timeout of ping responses?

juslesan: It limits pings if the number of neighbors is above it. Maybe it could be called ie `DEFAULT_DISCOVERY_LAYER_NEIGHBOR_COUNT_PING_LIMIT`?

packages/trackerless-network/src/logic/DiscoveryLayerNode.ts
juslesan: This is used specifically to control the sizes of the k-buckets that the neighbors are placed in. So it's specifically used to tweak the behaviour of kademlia's k-buckets.  Simply calling it neighbor-something might be a bit challenging because of that.

PR #3002, opened 2025-02-06
docs: Operator expectations section

fonty1:
## Summary

A brief guide on being a good node operator. Here we explicitly mention a worst case, 24 hour requirement to update their fleet nodes if faced with a critical situation.


harbu (approved):
lgtm

PR #3006, opened 2025-02-17
fix(node) [NET-1426]: Fix raising flag on empty inspection array

harbu:
## Summary

Fixes issue in which a flag would be raised when the `inspectOverTime` function returns an empty result array. Added a check _not_ to raise flag if the array is empty. This could happen due to JSON RPC timing out causing the inspection process to abort before a single inspection was performed. Due to JSON RPC timeouts being common, this results in lots of false flags.

- Do we need a better list of JSON RPCs for Polygon?
- Is the `FallbackProvider` working correctly?
- Technical debt: write unit tests for operator stuff (`inspectOverTime` and `inspectRandomNode`)


teogeb (approved):

harbu:
Forgot changelog entry...

PR #3017, opened 2025-02-20
chore(dht): Use `URLSearchParams` to parse search params

mondoreale:
## Summary

I switch to `URLSearchParams` for query param parsing and kill a dependency (`querystring`).

`URLSearchParams` is
- Widely supported by browsers (97%+).
- Available in node (global scope) since v15, stable since v10.

`querystring` is
- Deprecated.
- Doesn't seem to work well in browsers (breaks browser tests if not treated properly).

## Changes

- Ditch `querystring`
- Use `URLSearchParams` instead

- n/a

### DIFF
--- a/packages/dht/src/connection/websocket/WebsocketServerConnector.ts
+++ b/packages/dht/src/connection/websocket/WebsocketServerConnector.ts
@@ -11,7 +11,6 @@ import { ipv4ToNumber, Logger, wait } from '@streamr/utils'
 import { attachConnectivityRequestHandler, DISABLE_CONNECTIVITY_PROBE } from '../connectivityRequestHandler'
 import { WebsocketServerConnection } from './WebsocketServerConnection'
 import { ConnectionType, IConnection } from '../IConnection'
-import queryString from 'querystring'
 import { isMaybeSupportedProtocolVersion, LOCAL_PROTOCOL_VERSION } from '../../helpers/version'
 import { shuffle } from 'lodash'
 import { sendConnectivityRequest } from '../connectivityChecker'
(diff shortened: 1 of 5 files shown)


ptesavol (approved):
LGTM

packages/dht/src/connection/websocket/WebsocketServerConnector.ts
```diff
@@ -75,7 +74,8 @@ export class WebsocketServerConnector {
         if (!this.abortController.signal.aborted && this.websocketServer) {
             this.websocketServer.on('connected', (connection: IConnection) => {
                 const serverSocket = connection as unknown as WebsocketServerConnection
-                const query = queryString.parse(serverSocket.resourceURL.query as string ?? '')
+                const urlParams = new URLSearchParams(serverSocket.resourceURL.query as string ?? '')
+                const query = Object.fromEntries(urlParams.entries())
```
harbu: Does this code path get exercised in an existing test? The change looks innocent enough but just wondering if we had coverage for this.

mondoreale: Good question. Took it "as-is" from https://github.com/streamr-dev/network/commit/704b4e9d3682f45fdb58a87e4cbdcc2fb9ee9918. @ptesavol do you know by any chance?

teogeb: We could get the action directly from the `urlParams`:
```
const action = urlParams.get('action') as (Action | undefined)
```

teogeb (approved):

mondoreale:
Getting this one in. I need it for further work in other projects.

PR #3018, opened 2025-02-20
chore: Selective `lodash` imports

mondoreale:
## Summary

I make all imports from `lodash` selective and precise. It's beneficial for our Webpack bundles.

streamr-sdk.web.js **before** (stat size 9.43MB)

image

streamr-sdk.web.js **after** (stat size 8.96MB)

image

## Changes

- `import { foo } from 'lodash'` → `import foo from 'lodash/foo'`

- n/a


harbu (approved):
Looks good to me, I think @teogeb has some opinion about the use of `lodash` so maybe wait for his input before landing.

Anyways one point here. Sooner or later someone will inadvertently import `lodash` directly and then the whole library is included in the bundle again. Would it make sense to create an eslint rule about this, apparently it is possible, see: https://stackoverflow.com/questions/69621518/custom-eslint-no-restricted-imports-for-lodash-vs-lodash-fp.

teogeb (approved):
Awesome! 😄

PR #3019, opened 2025-02-20
fix: NET-1428: Ensure proper ESM/CJS entry point in the "browser" export

mondoreale:
## Summary

The aim here is to improve DX for those who want to use the SDK in their apps (webpack, rollup, etc.).

It moves the static SDK bundle (streamr-sdk.web.js) to the "script" export and populates the "browser" export with proper browser-specific modifications & shimming.

- [x] Fix browser tests 😱 
- [x] Can we make the **browser test runner** smarter? Lots of remapping in there, still. 😬 If I remove them it breaks viciously.
  💭 Not really. Maybe with much more effort.
- [x] Can this thing work in a **Nextjs** project out-of-the-box(ish)?
  💭 Yes, although with some extra effort.
- [x] Docs: Explain the **package.json rewriting** strategy (why it's there / what does it do).

Goes really well with #3018 and #3017 (both merged now).

⚠️ This may break frontend apps. The static build was an all-in-one package (no polyfills needed). Now it is a host project's responsibility to apply proper polyfilling. See "Limitations and future improvements" for details.

---

Available as [`v102.2.0-rc.1`](https://www.npmjs.com/package/@streamr/sdk/v/102.2.0-rc.1) 🏷️

```
npm uninstall @streamr/sdk && npm i @streamr/sdk@internal
```

## Changes

- Define proper "browser" exports.
- Add a postbuild script to DHT – rewrites its package.json to fix pathnames in `dist/package.json`.
- SDK package gets published from its root folder.
- Strip SDK's postbuild script from any package.json related logic. All it takes care of now are the db migrations.
- Simplify the `release.sh` script and make it a bit more consistent. It's a good script.

- Devs still need to polyfill for their browser-targetted bundles (os, fs, stream, etc.). For instance,
  - [node-polyfill-webpack-plugin](https://github.com/Richienb/node-polyfill-webpack-plugin) is a good one for Webpack apps.
  - [vite-plugin-node-polyfills](https://www.npmjs.com/package/vite-plugin-node-polyfills) works perfectly for Vite apps.
- SDK and DHT use `files` field in their package.jsons to define the contents of the npm package. Other packages use .npmignore. Using the `files` field is a preferred way and we should migrate other packages soon, too.
- DHT package contains 2 package.json files in the npm package (package.json and dist/package.json). We have to get rid of the later one. It's copied over by `tsc` because one of the files (Handshaker.ts) references it.


eslint.config.mjs
```diff
@@ -14,7 +14,8 @@ export default [
         ignores: [
             '**/generated/**',
             '**/dist/**',
-            '**/node_modules/**'
+            '**/node_modules/**',
+            '**/scripts/rewrite-package.ts'
```
harbu: Any harm in linting rewrite-package.ts?

packages/autocertifier-server/package.json
```diff
@@ -13,6 +13,9 @@
   },
   "main": "dist/src/exports.js",
   "types": "dist/src/exports.d.ts",
+  "browser": {
+    "dist/src/exports.js": false
+  },
```
harbu, line 18: What's the idea behind setting this (and the rest) to false?

mondoreale:
Here's one interesting fact
https://github.com/streamr-dev/hub/pull/1981#issuecomment-2687986542

I'm starting to wonder that we may want an explicit browser export (`@streamr/sdk/browser`) so that others who use it don't have to do [jest hacks like I did there](https://github.com/streamr-dev/hub/pull/1981/commits/3d1f67ad2d654aab894896e02aca7053472e36b6).

packages/browser-test-runner/src/createWebpackConfig.ts
```diff
@@ -26,29 +34,17 @@ export const createWebpackConfig = (
...
-                    'fs': false,
-                    'module': false,
-                    'net': false,
-                    'timers': require.resolve('timers-browserify'),
-                    'os': false,
-                    'zlib': require.resolve('browserify-zlib'),
-                    'tls': false
-                }
+                alias,
+                fallback,
```
harbu, line 47: I believe you can remove `timers-browserify` from package.json now.

(Curiously,  `browserify-zlib` seems to be already missing from package.json)

packages/dht/package.json
```diff
@@ -7,12 +7,18 @@
...
   },
-  "main": "dist/src/exports.js",
-  "types": "dist/src/exports.d.ts",
+  "main": "./dist/src/exports.js",
+  "types": "./dist/src/exports.d.ts",
+  "browser": {
+    "./dist/src/connection/webrtc/NodeWebrtcConnection.js": "./dist/src/connection/webrtc/BrowserWebrtcConnection.js",
+    "./dist/src/connection/websocket/NodeWebsocketClientConnection.js": "./dist/src/connection/websocket/BrowserWebsocketClientConnection.js",
+    "./dist/src/helpers/browser/isBrowserEnvironment.js": "./dist/src/helpers/browser/isBrowserEnvironment_override.js"
+  },
```
harbu, line 16: Neat trick! :+1:

packages/sdk/src/StreamrClient.ts
```diff
@@ -1,3 +1,4 @@
+import 'timers'
```
harbu, line 5: Intentional change?

packages/sdk/webpack.config.js
```diff
@@ -95,42 +95,13 @@ module.exports = (env, argv) => {
             globalObject: 'globalThis',
         },
         resolve: {
```
harbu, line 97: We can probably remove `readable-stream` from package.json?

release.sh
```diff
@@ -68,9 +67,9 @@ npm publish --access public --tag $NPM_TAG
 cd ../../..
 
 cd packages/node || exit
```
harbu, line 69: Should the `dht` package be released from dist folder here in the release.sh (similar to sdk)?

harbu (changes requested):
Firstly, great stuff. This is a good direction to go towards modernizing our exports and all.

Two additional general points:

- The PR brings in `tsx` but we've been using `ts-node` in the project before. I got the impression they do more-or-less the same thing? If so, we should stick to one. Are there benefits to `tsx` over `ts-node` that would merit changing to it?

- I didn't find the reasoning behind re-writing the package.json into dist and publishing under dist approach. Is there a need to do this or is this more of a best practice you advocate for?

packages/sdk/scripts/rewrite-package.ts
```diff
@@ -0,0 +1,54 @@
...
+ * This script moves the package.json file into the dist folder while adjusting
+ * relative pathnames for exports. Since the compiled output is inside "dist",
+ * but the original package.json references paths as if it's in the root, we
+ * need to strip "dist" from all export fields (main, types, …).
+ *
+ * This ensures that consumers of the package import the correct files
+ * without referencing "dist" in their paths, maintaining proper module resolution.
+ */
+
+import pkg from "../package.json" assert { type: "json" }
```
teogeb: Would it make sense to unify the two rewrite-package.ts scripts and move that unified script to somewhere where both packages can access it? E.g. a new `scripts` directory in monorepo root.

packages/browser-test-runner/src/createWebpackConfig.ts
```diff
@@ -26,29 +34,17 @@ export const createWebpackConfig = (
                 ],
             },
             plugins: [
-                new NodePolyfillPlugin(),
-                new webpack.ProvidePlugin({
-                    process: 'process/browser'
+                new NodePolyfillPlugin({
```
teogeb, line 37: Question: Is the `process` polyfill which is now provided by `NodePolyfillPlugin` equal (or better) compared to the previous polyfill which was provided by the `webpack.ProvidePlugin`?

teogeb (changes requested):
Great improvements 👍 

As developers now need to add polyfills, we should update the documentation (in monorepo's `docs` folder) to reflect this change.

Let's also add to a CHANGELOG entry (with **BREAKING CHANGE:** highlight).

packages/autocertifier-server/package.json
mondoreale: The reasoning behind this goes as follows:
- We assume certain packages are node-only and the browser shouldn't touch them at all.
- Pointing their main exports to `false` (disabling them for browser) would then tell (break) if the browser makes any attempts to load them despite the above assumption.

Long story short, it's sort of a sanity check.

packages/browser-test-runner/src/createWebpackConfig.ts
mondoreale: NodePolyfillPlugin uses the same [`process`](https://github.com/defunctzombie/node-process) library (see [`node-stdlib-browser`](https://www.npmjs.com/package/node-stdlib-browser) for a comprehensive list).

packages/browser-test-runner/src/createWebpackConfig.ts
mondoreale: `browserify-zlib` was never there, yeah. The monorepo ✨ magic.

I removed `timers-browserify` now also.

packages/sdk/src/StreamrClient.ts
mondoreale: It is.

`timers-browserify` (polyfilling `timers`) patches global objects to provide missing timer functions (`setImmediate`, …).

I've left a comment in the code.

eslint.config.mjs
mondoreale: No harm at all. It just didn't work right away. I've been getting this error:

```
0:0  error  Parsing error: /Users/…/network/packages/dht/scripts/rewrite-package.ts was not found by the project service. Consider either including it in the tsconfig.json or including it in allowDefaultProject
```

This confusing message doesn't precisely indicate which tsconfig we're talking about. Now I know it's the tsconfig.jest.json.

I also thought it wants to pack my script into the dist folder on build (don't want that). Luckily now I see it doesn't do it.

mondoreale (commented):
Thanks for taking the time to look through it.

I'll address the remaining points in a follow-up session.

Cheers!

release.sh
mondoreale: It is now.

packages/sdk/scripts/rewrite-package.ts
mondoreale: I wouldn't press on it at this point. 2 scripts are totally manageable atm.

This could be a separate task: identify more shared scripts and have a `scripts` package then.

I'll leave it be for now.

mondoreale:
@harbu, about `tsx`

- Works out of the box for ESM scripts (import/export), but untested on CJS.
- Ignores tsconfig.json, so project-specific settings (module resolver, target, etc.) don’t matter.

I'd use it for standalone scripts. Why stop?

- Switching to ts-node broke the script (no import.meta, no JSON import assertions, etc.). It might work with the right options, but...
- ESLint complained about missing ESM features, too.

To satisfy ESLint and stick to ts-node, the easiest fix was compiling to proper CJS. It works really well without extra deps which is cool.

mondoreale:
@harbu, and on the rewriting of package.json:

With relative pathnames in "browser" (especially) we have to alter dist/package.json before publishing. It is crucial for bundler resolvers. I believe it was the reason for rewriting SDK's package.json originally.

Not sure why the SDK is published from dist, but I’m not against it. The code’s there, so maybe it makes sense. That said, a single package.json per package feels cleaner. Publishing outside dist means dealing with two: one at the root, one in dist. Efficiency matters.

mondoreale:
@teogeb @harbu, guys, have another look.

- I've updated docs. 
- I've mentioned the breaking nature of the change in the changelog, too.

harbu (approved):

harbu (approved):

teogeb (approved):

mondoreale:
Last minute hiccups.

- Had to drop rollup's usage tips and update Gatsby's (see docs). I pointed to a bad export previously. Getting them up and running wasn't so easy (esp. Gatsby). Next.js much smoother.
- Getting _GitHub Actions has encountered an internal error when running your job_. I'll ignore it.

PR #3020, opened 2025-02-20
chore(sdk) [NET-1394]: add new storage node cluster address

harbu:
## Summary

- Add new storage node cluster address constant `STREAMR_STORAGE_NODE_ADDRESS`
- Deprecate existing constant `STREAMR_STORAGE_NODE_GERMANY` 
- Adjust docs to reflect this change

### DIFF
--- a/packages/sdk/src/Config.ts
+++ b/packages/sdk/src/Config.ts
@@ -442,7 +442,7 @@ export type StrictStreamrClientConfig = MarkOptional<Required<StreamrClientConfi
  */
 export const STREAMR_STORAGE_NODE_GERMANY = '0x31546eEA76F2B2b3C5cC06B1c93601dc35c9D916'
 
-export const STREAMR_STORAGE_DEFAULT = '0x9dc08ff97f5c156181ec6a0b13fc3946454e529a'
+export const STREAMR_STORAGE_DEFAULT = '0x9dc08ff97f5c156181ec6a0b13fc3946454e529a' as HexString
 
 export const createStrictConfig = (input: StreamrClientConfig = {}): StrictStreamrClientConfig => {
     // TODO is it good to cloneDeep the input object as it may have object references (e.g. auth.ethereum)?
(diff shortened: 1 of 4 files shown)


mondoreale (approved):

packages/sdk/src/Config.ts
```diff
@@ -437,8 +437,13 @@ export type StrictStreamrClientConfig = MarkOptional<Required<StreamrClientConfi
     _timeouts: Exclude<DeepRequired<StreamrClientConfig['_timeouts']>, undefined>
 }
 
+/**
+ * @deprecated use {@link STREAMR_STORAGE_DEFAULT} instead
+ */
 export const STREAMR_STORAGE_NODE_GERMANY = '0x31546eEA76F2B2b3C5cC06B1c93601dc35c9D916'
 
+export const STREAMR_STORAGE_DEFAULT = '0x9dc08ff97f5c156181ec6a0b13fc3946454e529a'
```
teogeb: If we want to be very explicit, this could be named to `STREAMR_STORAGE_NODE_ADDRESS` and/or type this as `HexString`. Then it would be trivial that this is a value that can be given as an argument to `addStreamToStorageNode()`.

harbu: I'll add the `HexString` type. 

About the naming, it is a bit complicated. When we refer to a storage node, we don't actually refer to a single storage node (most of the time) but a cluster of storage nodes. The correct name would therefore be something like `STREAMR_STORAGE_NODE_CLUSTER_DEFAULT`. However, our internal naming convention currently refers to the address as `storageNodeAddress` so keeping in line with that it may be wise to stick with the `NODE` for now. Thoughts?

teogeb: As you mentioned, end-users don't know about the clustering. In SDK's nomenclature the entity to which we add/remove streams in order to get it stored is a _storage node_. This is an id to such entity, and therefore we should refer it as storage node.

In practice the entity can be a single node or a cluster of nodes. As long as a real cluster (i.e  more than one node) is not required, the current name is quite fine.

In summary: in my opinion we should now keep the current consistent term (`STREAMR_STORAGE_NODE_ADDRESS`). We can consider renaming the terminology in a follow-up PR, if needed.

harbu: Addressed here https://github.com/streamr-dev/network/pull/3020/commits/6a651e8b2593179462a91606685e91196b319e67

teogeb (approved):
Maybe would make sense to drop the term "cluster" also in CHANGELOG (lines 15 and 21)

PR #3036, opened 2025-03-04
chore(sdk)!: remove lit-protocol integration

harbu:
## Summary

Remove lit-protocol integration for offline key storing and fetching.

We may revisit this feature in the future, but right now as it stands, it doesn't work when enabled. (I tested it out by-hand and can confirm it doesn't work, most likely due to us running an ancient client of their protocol).

Better to remove it for now as it should contribute to a smaller bundle size.

- We may re-implement this offline key storing / fetching feature in the future (with or without lit-protocol), but for now we shall remove it as fixing the issue is not trivial as the whole approach of the lit-protocol library and protocol has changed quite a lot.


teogeb (approved):
Great that we can simplify this 👍

fonty1 (approved):

PR #3040, opened 2025-03-10
feat(network): new `doNotBufferWhileConnecting` transport flag.

juslesan:
## Summary

Added a transport option to avoid buffering while connecting. This is used in the trackerless-network during propagation when only real-time data is wanted.

## Changes

- Added new transport flag `doNotPropagateWhileConnecting`
- Added new configuration option to the trackerless-network to enable the feature.

- Might make sense to add a new configuration option ie. `fireAndForgetMode`. This would disable the use of all buffers during propagation if only real-time data is wanted.
- Make such configurations per stream instead of global.


ptesavol (approved):
LGTM

PR #3053, opened 2025-03-18
feat(DHT): NET-1436: Cut off connections to all application versions before v102.0.0

juslesan:
## Summary

Cut off connections to all application versions before v102.0.0

## Changes

- Renamed handshake error `UNSUPPORTED_PROTOCOL_VERSION` -> `UNSUPPORTED_VERSION`. Version incompatibilities are handled similarly between protocol and application versions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27359,6 +27359,7 @@
                 "lodash": "^4.17.21",
                 "lru-cache": "^11.0.2",
                 "node-datachannel": "^0.26.0",
+                "semver": "7.7.1",
                 "uuid": "^11.1.0",
                 "websocket": "^1.0.34",
                 "ws": "^8.18.1"
@@ -27369,6 +27370,7 @@
                 "@types/heap": "^0.2.34",
                 "@types/k-bucket": "^5.0.1",
                 "@types/lodash": "^4.17.16",
+                "@types/semver": "^7.5.8",
                 "@types/websocket": "^1.0.10",
                 "@types/ws": "^8.18.0",
                 "jest-leak-detector": "^27.3.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -27359,7 +27359,7 @@
                 "lodash": "^4.17.21",
                 "lru-cache": "^11.0.2",
                 "node-datachannel": "^0.26.0",
-                "semver": "7.7.1",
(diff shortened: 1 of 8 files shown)


teogeb (changes requested):
Could we have two separate enums: `UNSUPPORTED_PROTOCOL_VERSION` (i.e. the current) and new `UNSUPPORTED_APPLICATION_VERSION`? It would give us more detailed information why a handshake fails.

juslesan:
> Could we have two separate enums: `UNSUPPORTED_PROTOCOL_VERSION` (i.e. the current) and new `UNSUPPORTED_APPLICATION_VERSION`? It would give us more detailed information why a handshake fails.

We could but older versions would not support the new failing UNSUPPORTED_APPLICATION_VERSION error. The behavior on how to handle unsupported version is the same for both cases so there is no requirement to add the new field to break backwards compatibility @teogeb

teogeb:
> > Could we have two separate enums: `UNSUPPORTED_PROTOCOL_VERSION` (i.e. the current) and new `UNSUPPORTED_APPLICATION_VERSION`? It would give us more detailed information why a handshake fails.
> 
> We could but older versions would not support the new failing UNSUPPORTED_APPLICATION_VERSION error. The behavior on how to handle unsupported version is the same for both cases so there is no requirement to add the new field to break backwards compatibility @teogeb

What would happen in practice if a new client would send an error code `3` (the new `UNSUPPORTED_APPLICATION_VERSION`) which old nodes don't know about?

juslesan:
> > > Could we have two separate enums: `UNSUPPORTED_PROTOCOL_VERSION` (i.e. the current) and new `UNSUPPORTED_APPLICATION_VERSION`? It would give us more detailed information why a handshake fails.
> > 
> > 
> > We could but older versions would not support the new failing UNSUPPORTED_APPLICATION_VERSION error. The behavior on how to handle unsupported version is the same for both cases so there is no requirement to add the new field to break backwards compatibility @teogeb
> 
> What would happen in practice if a new client would send an error code `3` (the new `UNSUPPORTED_APPLICATION_VERSION`) which old nodes don't know about?

Don't know for sure but I would expect that the connection will eventually be cleaned up after a timeout. But there could be some undefined behavior that leads to problems

PR #3055, opened 2025-03-21
refactor(DHT): executeRecursiveOperation helper method in DhtNode

juslesan:
## Summary

added new executeRecursiveOperation helper method to the network

## Future improvements

- The `deleteDataFromDht` and `findClosestNodesFromDht` should use the function as well.
- Could make the creator parameter in storeDataToDht clearer. There should be no need to provide it for the function. The External API could simply be passed the storeManagers storeData method. Currently it is used in the public API for testing but it should be reserved for internal use only.

### DIFF
--- a/packages/dht/src/dht/DhtNode.ts
+++ b/packages/dht/src/dht/DhtNode.ts
@@ -550,11 +550,10 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
     }
 
     public async storeDataToDht(key: DhtAddress, data: Any, creator?: DhtAddress): Promise<PeerDescriptor[]> {
-        const connectedEntryPoints = this.getConnectedEntryPoints()
-        if (this.peerDiscovery!.isJoinOngoing() && connectedEntryPoints.length > 0) {
-            return this.storeDataToDhtViaPeer(key, data, sample(connectedEntryPoints)!)
-        }
-        return this.storeManager!.storeDataToDht(key, data, creator ?? this.getNodeId())
+        return this.executeRecursiveOperation(
+            () => this.storeManager!.storeDataToDht(key, data, creator ?? this.getNodeId()),
+            (connectedEntryPoint) => this.storeDataToDhtViaPeer(key, data, connectedEntryPoint)
+        )
     }
 
     public async storeDataToDhtViaPeer(key: DhtAddress, data: Any, peer: PeerDescriptor): Promise<PeerDescriptor[]> {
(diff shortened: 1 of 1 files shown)


packages/dht/src/dht/DhtNode.ts
```diff
@@ -550,11 +550,10 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
...
 
     public async storeDataToDht(key: DhtAddress, data: Any, creator?: DhtAddress): Promise<PeerDescriptor[]> {
-        const connectedEntryPoints = this.getConnectedEntryPoints()
-        if (this.peerDiscovery!.isJoinOngoing() && connectedEntryPoints.length > 0) {
-            return this.storeDataToDhtViaPeer(key, data, sample(connectedEntryPoints)!)
-        }
-        return this.storeManager!.storeDataToDht(key, data, creator ?? this.getNodeId())
+        return this.executeRecursiveOperation(
+            () => this.storeManager!.storeDataToDht(key, data, creator ?? this.getNodeId()),
+            (connectedEntryPoint) => this.storeDataToDhtViaPeer(key, data, connectedEntryPoint)
```
teogeb, line 555: The creator argument is not used in `this.storeDataToDhtViaPeer(key, data, connectedEntryPoint)`? Maybe we should remove that parameter from the public API?

packages/dht/src/dht/DhtNode.ts
```diff
@@ -568,12 +567,13 @@ export class DhtNode extends EventEmitter<Events> implements ITransport {
...
-        const connectedEntryPoints = this.getConnectedEntryPoints()
-        if (this.peerDiscovery!.isJoinOngoing() && connectedEntryPoints.length > 0) {
-            return this.fetchDataFromDhtViaPeer(key, sample(connectedEntryPoints)!)
-        }
-        const result = await this.recursiveOperationManager!.execute(key, RecursiveOperation.FETCH_DATA)
-        return result.dataEntries ?? []  // TODO is this fallback needed?
+        return this.executeRecursiveOperation(
+            async () => {
+                const result = await this.recursiveOperationManager!.execute(key, RecursiveOperation.FETCH_DATA)
+                return result.dataEntries ?? []
```
teogeb: Is the fallback needed? (as the comment was removed)

teogeb (changes requested):

packages/dht/src/dht/DhtNode.ts
juslesan: Added this to future improvements

teogeb (approved):

PR #3060, opened 2025-04-01
feat: [NET-1445] quantum resistant key exchange using ML-KEM

hpihkala:
## Summary

This PR introduces the ability to use (and to require the use of) `ML-KEM` for key exchange instead of `RSA`. The former is considered quantum resistant, while the latter is not. This PR does not change the defaults, but introduces a new config option to enable the use of the new quantum resistant algorithm.

Implementing the new algorithm itself was straight forward, but certain refactoring was necessary to use it. Most of the line changes in this PR are actually related to that refactoring instead of the core feature itself.

**Backwards compatibility**

The change is backwards compatible with the default settings, ie. the change is not breaking as such. However, the new feature can obviously only be used if both the publisher and subscriber have the new version. 

Backwards compatibility with the previous version was tested manually, and the results are as expected - ie. key exchange works with default (non-quantum-secure) settings, but fails if `requireQuantumResistantKeyExchange` is `true`, as the other side running the old version doesn't support the new algorithm.

## Changes

- Implement encryption and decryption using `ML-KEM` in `EncryptionUtil`
- Refactor interfaces in `EncryptionUtil` to be unified across different algorithms
- Add a new `StreamrClient` config option under the `encryption` block: `requireQuantumResistantKeyExchange: <boolean>` (default: false)
- Update `SubscriberKeyExchange` and `PublisherKeyExchange` to read that config option and act accordingly 
- Add to `GroupKeyRequest` and `GroupKeyResponse` protobuf messages an `encryptionType` field, the value of which is an `AsymmetricEncryptionType` enum. For backwards compatibility, the absence of this field is interpreted to imply RSA.
- Refactor away the technical debt of `OldGroupKeyRequest` and `OldGroupKeyResponse` vs. `NewGroupKeyRequest` and `NewGroupKeyResponse` mess and get rid of `GroupKeyRequestTranslator`. Now, the "new" group key classes (generated from protobuf) are used everywhere, and the old hand-rolled ones have been deleted.
- Add unit test cases for `EncryptionUtil` and integration test cases for `SubscriberKeyExchange` and `PublisherKeyExchange`
- Add a section to the end-to-end encryption page in the docs, describing the new feature
- Add a `--quantum` flag to the CLI tool to flip on the new switch

- Refactoring away the `OldStreamMessage` / `NewStreamMessage` / `StreamMessageTranslator` debt is out of scope, the old vs. new classes issue was only cleaned up in relation to the group key classes, as only they were relevant to the feature.
- In the future, encryption preferences could come from stream metadata by default, instead of explicit local config

- [x] Is this a breaking change? If it is, be clear in summary. **-> (No)**


packages/sdk/src/encryption/EncryptionUtil.ts
```diff
@@ -1,28 +1,146 @@
...
+     * Generic utility functions
+     */
+    static async encryptForPublicKey(plaintext: Uint8Array, publicKey: Uint8Array, type: AsymmetricEncryptionType): Promise<Buffer> {
+        if (type === AsymmetricEncryptionType.ML_KEM) {
+            return this.encryptWithMLKEMPublicKey(plaintext, publicKey)
+        }
+        if (type === AsymmetricEncryptionType.RSA) {
+            return this.encryptWithRSAPublicKey(plaintext, publicKey)
+        }
+        throw new Error(`Unexpected GroupKeyEncryptionType: ${type}`)
```
teogeb: Maybe could reword the error message as these are generic utility functions (can be used to encrypt/decrypt non-groupkey data in the future)

```suggestion
        throw new Error(`Unexpected encryption type: ${type}`)
```

packages/sdk/test/integration/PublisherKeyExchange.test.ts
```diff
@@ -67,63 +89,100 @@ describe('PublisherKeyExchange', () => {
...
+                    requireQuantumResistantKeyExchange: true,
+                }
+            })
+    
+            // The invalid request just prints a debug level error on the publisher.
+            // Since the error is swallowed, it's difficult to assert the error here.
+            await triggerGroupKeyRequest(publisherClient, stream)
+    
+            // Due to the above, rely on time - not foolproof
+            await new Promise((resolve) => setTimeout(resolve, 3000))
```
teogeb: Could use wait from `@streamr/utils`

```suggestion
            await wait(3000)
```

packages/sdk/src/encryption/SubscriberKeyExchange.ts
```diff
@@ -31,7 +30,7 @@ const MAX_PENDING_REQUEST_COUNT = 50000 // just some limit, we can tweak the num
 @scoped(Lifecycle.ContainerScoped)
 export class SubscriberKeyExchange {
 
-    private rsaKeyPair?: RSAKeyPair
+    private keyPair?: RSAKeyPair | MLKEMKeyPair
```
teogeb:
```suggestion
    private keyPair?: KeyExchangeKeyPair
```

packages/sdk/src/exports.ts
```diff
@@ -68,7 +68,6 @@ export type {
 // These are currently exported because NetworkNodeStub uses methods which operate on StreamMessage.
 // If we remove that semi-public class we can maybe remove these exports.
 export type { UserID } from '@streamr/utils'
-export type { EncryptedGroupKey } from './protocol/EncryptedGroupKey'
```
teogeb, line 71: The `EncrypedGroupKey` type is part of the semi-public API (as it is used as a field type of `StreamMessage#newGroupKey`). Therefore we should keep the export.

i.e. change the deleted line to:
`export type { EncryptedGroupKey } from '@streamr/trackerless-network'`

End-users could of course import the type directly from `@streamr/trackerless-network`, but we re-export these kind of types here so that the `sdk`'s API is self-contained.  (e.g. `DhtAddress` and `ProxyDirection`  are also re-exported like that.)

teogeb (commented):
Looks good! 👍 We should keep the `EncryptedGroupKey` type in the `exports.ts`, otherwise all ok.

packages/sdk/src/utils/crossPlatformCrypto.ts
```diff
@@ -0,0 +1,10 @@
+import crypto from 'crypto'
+
+export function getSubtle(): crypto.webcrypto.SubtleCrypto {
+    const subtle = typeof window !== 'undefined' ? window?.crypto?.subtle : crypto.webcrypto.subtle
+    if (!subtle) {
+        const url = 'https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto'
+        throw new Error(`SubtleCrypto not supported. This feature is available only in secure contexts (HTTPS) & Node 16+. ${url}`)
+    }
+    return subtle
+}
```
jtakalai, line 10: Is this part of ML-KEM pr?

jtakalai: I see, it's a DRY refactor (subtle needed now in EncryptionUtil in addition to RSAKeyPair, hence moved into a separate file)

teogeb (approved):

PR #3071, opened 2025-04-16
feat(DHT): [NET-938] WebRTCbuffer threshold high low handling

juslesan:
## Summary

Added buffer high/low handling for webrtc connections. This ensures that messages are sent more reliably when throughput is high.


ptesavol (approved):
Looking good, let's merge this and make a release ASAP so we can try whether it resolves the issue in the other project.

PR #3074, opened 2025-04-17
feat: [NET-1446] Quantum resistant identities and signatures

hpihkala:
## Summary

The goal of this PR is to add support for quantum secure cryptographic identities and signatures, as well as to make it easier to add more identity/signature types in the future. The PR includes some refactoring to create a new `Identity` abstraction which replaces the old `Authentication` interface.

Most of the line changes are due to renaming things + docs changes. The actual code changes in this PR are actually not that massive.

## Changes

**More significant changes:**
- Refactor old `Authentication` interface replaced by new `Identity` framework
- Added support for ML-DSA-87 identity and signing
- As an alternative to configuring keys, allow an `Identity` instance to be passed to `StreamrClient` as `auth.identity`
- Add `requireQuantumResistantSignatures` and `requireQuantumResistantEncryption` config options as well as related checks
- Add `--public-key` and `--key-type` options to CLI tool
- **Breaking change**: In CLI tool, rename `streamr wallet` to `streamr keys` (the only command there was `streamr wallet whoami`, which is now `streamr keys whoami`. Add a new command `streamr keys generate`.
- **Breaking change**: Removed `StreamrClient.generateEthereumAccount()`, the new way is `EthereumKeyPairIdentity.generate()` (can be done with other key/identity types as well).
- **Breaking change**: The string values in `Message.signatureType` now correspond with the `KeyType` values. This means the previously output value `SECP256K1` is now `ECDSA_SECP256K1_EVM`.
- Large-ish docs changes

**Additional minor changes:**
- Removed some technical debt from `StreamMessageTranslator` by removing `oldContentType` vs. `newContentType`, `oldEncryptionType` vs. `newEncryptionType`. The network-level enums are used directly now.
- Give more accurate names to `SignatureType`s, paving the road for more types to be added later.
- Namespace the exports in `signingUtils.ts` by algorithm type, eg. `ECDSA_SECP256K1_EVM` and `ML_DSA_87`
- Remove `address` as an acceptable parameter to key config
- Add a central place `identityConfig.ts` for wiring together identities, `SignatureType`s, and verification methods
- Remove repetition / enumeration of SignatureTypes from `Message.ts` and `SignatureValidator.ts`
- bugfix: Fix regex used for validating hex strings given to config

- The client config key `auth` was kept as-is for backwards compatibility, although renaming it to `identity` would probably make sense at some point.
- I didn't attempt to solve all overlap or inaccuracies in the docs (there are some). I mainly focused on the Identity article, the articles under the Security section, and any references to them.

- [x] Is this a breaking change? If it is, be clear in summary. **(no major breaking changes, but some exported supporting classes have changed)**


docs/docs/streamr-network/security/end-to-end-encryption.md
```diff
@@ -36,17 +41,63 @@ const streamr = new StreamrClient({
...
+});
+```
+
+## Pre-agreed keys
+
+If you don't want to exchange the keys via the network, you can pre-configure symmetric keys on both publishers and subscribers like this:
+
+```ts
+// Generates a new AES-256 key (32 random bytes)
+const key = new GroupKey('key-id', crypto.randomBytes(32))
```
teogeb: The class is exported as `EncryptionKey` (https://github.com/streamr-dev/network/blob/main/packages/sdk/src/encryption/LocalGroupKeyStore.ts#L11-L12)

```suggestion
const key = new EncryptionKey('key-id', crypto.randomBytes(32))
```

packages/sdk/src/Message.ts
```diff
@@ -1,8 +1,24 @@
...
+        )
+    ),
+    // These special ones need to be added manually
+    [SignatureType.ECDSA_SECP256K1_LEGACY]: 'ECDSA_SECP256K1_LEGACY',
+    [SignatureType.ERC_1271]: 'ERC_1271',
+}
 
 /**
- * Represents a message in the Streamr Network.
+ * Represents a message in the Streamr Network. This is an application-facing class, whereas StreamMessage is considered internal.
```
teogeb: Could move this comment to the `StreamMessage` class. There it would be available for those users who care about the fact that there are actually two different message structures. Typical users may just find this separation confusing as it is not relevant for them.

```suggestion
 * Represents a message in the Streamr Network. 
```

packages/cli-tools/bin/streamr-keys-generate.ts
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env node
+import '../src/logLevel'
+import { KeyPairIdentity } from '@streamr/sdk/dist/types/src/identity/KeyPairIdentity'
+import { identityConfig, validKeyTypeValues } from '@streamr/sdk'
+import { binaryToHex } from '@streamr/utils'
+import { createCommand, Options } from '../src/command'
+
+createCommand()
+    .description('generate a public/private key pair based on the --key-type option')
+    .option('--key-type [key-type]', `one of: [${validKeyTypeValues.join(', ')}]`)
```
teogeb: In the documentation we state that enums are in this form: (e.g. `quantum-security.md`)
```
streamr keys generate --key-type ml-dsa-87
```

This is a good format and in line with other cli-tool usage (e.g. the `--env` flag).

But currently the implementation requires the enums to be uppercase form (e.g. `ML_DSA_87`). It would be good to change it to allow only the lowercase-dash format.

(In the SDK/cli-tools we use two kinds of enums which partially contributes to this problem. TypeScript enums are always in the classic uppercase form. We've also started also to use string literal "enums", and those are in the lowercase form as that is the typical convention for those. Ideally we'd unify the style by refactoring all "enums" to be string literals.)

packages/cli-tools/bin/streamr-keys-generate.ts
```diff
@@ -0,0 +1,25 @@
...
+    .action(async (options: Options) => {
+        if (!options.keyType) {
```
teogeb: Could use `requiredOption` definition for commander to specify the required flag. That way the flag is mentioned in the help text of the command and we can implement the validation logic easily:
https://github.com/streamr-dev/network/blob/1025c2386fcd6f9a1cde970597770dc53efab84b/packages/cli-tools/src/command.ts#L41-L42

Note that there we use `option`, but same as this is a required flag we can us `requiredOption`.

packages/cli-tools/bin/streamr-keys-generate.ts
```diff
@@ -0,0 +1,25 @@
...
+            console.error(`Error: Please provide --key-type [one of: ${validKeyTypeValues.join(', ')}]`)
+        } else {
+            const config = identityConfig[options.keyType]
+            if (!config) {
+                console.error(`Error: Invalid key type. Must be one of: ${validKeyTypeValues.join(', ')}.`)
+            }
+            const identity = config.generate() as KeyPairIdentity
+            console.info(`Public key: ${await identity.getUserIdString()}`)
```
teogeb: We could output the data as JSON so that it is easy to read programatically. See e.g. `stream create`, and `internal operator-create` commands:
https://github.com/streamr-dev/network/blob/aa0869c1bccf893d44ea4e836acefe5996e42227/packages/cli-tools/bin/streamr-internal-operator-create.ts#L32

packages/cli-tools/src/client.ts
```diff
@@ -1,19 +1,31 @@
...
-    const authenticationOptions: StreamrClientConfig = 
-        (commandOptions.privateKey !== undefined) ? { auth: { privateKey: commandOptions.privateKey } } : {}
+
+    const keyPairConfig: KeyPairIdentityConfig | undefined = 
+        (commandOptions.privateKey) ? { privateKey: commandOptions.privateKey } : undefined
+
+    if (keyPairConfig) {
+        keyPairConfig.publicKey = commandOptions.publicKey
+        keyPairConfig.keyType = commandOptions.keyType
+    }
```
teogeb: Could we set all these 3 values (`privateKey`, `publicKey` and `keyType`) at line 11? Or do we need both the ternary and the statement?

teogeb (commented):
First round of comments (docs and cli-tools)

CHANGELOG.md
```diff
@@ -12,7 +12,9 @@ Changes before Tatum release are not documented in this file.
 
 #### Added
 
-- Add support for quantum secure key exchange using ML-KEM (https://github.com/streamr-dev/network/pull/3060)
+- Add support for quantum resistant key exchange using ML-KEM (https://github.com/streamr-dev/network/pull/3060)
+- Add support for quantum resistant signatures using ML-DSA (https://github.com/streamr-dev/network/pull/3074)
+- CLI tool allows generating key pairs with `streamr keys generate --key-type [...]`
```
teogeb: Move the CLI tools feature into the `@streamr/cli-tools` section (at line 59), and add github PR link for that item

packages/utils/src/signingUtils.ts
```diff
@@ -1,65 +1,92 @@
...
 
 const SIGN_MAGIC = '\u0019Ethereum Signed Message:\n'
 const keccak = new Keccak(256)
 
+export interface KeyPair {
+    publicKey: Uint8Array
+    privateKey: Uint8Array
+}
+
+export interface SigningUtil {
```
teogeb, line 17: Would it make sense to convert this to abstract class with a `getInstance`(type: 'md-dsa-87' | 'ecdsa_secp256k1_evm') method?

Alternatively we could rename the implementations e.g. to `MdDsa87SigningUtil` and `EcdsaSecp256k1EvmSigningUtil`.

packages/utils/src/signingUtils.ts
```diff
@@ -1,65 +1,92 @@
...
- * to ether.js v5.5.0.
- *
- * See test/benchmark/SigningUtils.ts and the original PR:
- * https://github.com/streamr-dev/streamr-client-protocol-js/pull/35
+ * EVM compatible signing scheme using keccak hash, magic bytes, and secp256k1 curve.
  */
+export const ECDSA_SECP256K1_EVM: SigningUtil = {
+    keccakHash(message: Uint8Array, useEthereumMagic: boolean = true): Buffer {
+        keccak.reset()
+        keccak.update(useEthereumMagic ? Buffer.concat([Buffer.from(SIGN_MAGIC + message.length), message]) : Buffer.from(message))
```
teogeb: This magic string applies only to this signer, could rename the constant e.g. to ECDSA_SECP256K1_EVM_SIGN_MAGIC

packages/cli-tools/bin/streamr.ts
```diff
@@ -9,6 +9,6 @@ program
     .command('stream', 'stream subcommands')
     .command('storage-node', 'storage node subcommands')
     .command('mock-data', 'mock data subcommands')
-    .command('wallet', 'wallet subcommands')
+    .command('keys', 'subcommands for keys and wallets')
```
teogeb: Could rename this to singular form. That way it would be unified with other commands.

To me the sub commands in singular sounds all ok:
- `streamr key generate ...` generates multiple keys, but one key pair
- `streamr key whoami` displays one key (the public key)

teogeb (commented):
Second round of comments (utils, cli-tools, changelog)

packages/sdk/src/Config.ts
```diff
@@ -9,22 +9,33 @@ import validate from './generated/validateConfig'
...
+import { Identity } from './identity/Identity'
+import { KeyType } from './identity/identityConfig'
 
-export interface ProviderAuthConfig {
-    /**
-     * The {@link https://docs.ethers.org/v6/api/providers/#Eip1193Provider Eip1193Provider} type is from the `ethers` library.
-     */
+/**
+ * For passing in an Ethereum provider (= wallet) for signing
+ */
```
teogeb: Could keep the link to the `ethers.js` documentation so that we don't have undocumented types in the API.

Typically the instance passed here is a MetaMask's global `window.ethereum` object.

```suggestion
/**
 * For passing in an Ethereum provider (e.g. MetaMask) for signing
 *
 * The {@link https://docs.ethers.org/v6/api/providers/#Eip1193Provider Eip1193Provider} type is from the `ethers` library.
 */
```

packages/sdk/src/Config.ts
```diff
@@ -357,13 +368,29 @@ export interface StreamrClientConfig {
...
+        /**
+         * Default: false. If true, configuring a non-quantum-resistant identity key pair will produce an error,
+         * and subscribed messages will only be accepted if they are signed using a quantum resistant algorithm.
+         */
+        requireQuantumResistantSignatures?: boolean
+
+        /**
+         * Default: false. If true, only data encrypted using quantum resistant algorithms will be accepted,
+         * and data using other algorithms as well as unencrypted data will be rejected. Note that also public (unencrypted)
+         * data will be rejected in this case.
```
teogeb: Why we'd want to reject public data? To me it would sound logical that _if_ encryption is used, this requires it to be quantum resistant.

packages/sdk/src/Message.ts
```diff
@@ -1,8 +1,24 @@
 import { HexString, StreamID } from '@streamr/utils'
-import { SignatureType, StreamMessage } from './protocol/StreamMessage'
+import { StreamMessage } from './protocol/StreamMessage'
+import { SignatureType } from '@streamr/trackerless-network'
+import { identityConfig, KeyType } from './identity/identityConfig'
+
+// Lookup structure for converting SignatureType to KeyType string
+export type MessageSignatureType = KeyType | 'ECDSA_SECP256K1_LEGACY' | 'ERC_1271'
```
teogeb, line 7: It is good to use string literal instead of TypeScript enum. For string literals the prefer lowercase format. The literals here weren't in that format before this PR, but maybe good to change now as this is anyway a breaking change.

```suggestion
export type MessageSignatureType = KeyType | 'ecdsaSecp256k1Legacy' | 'erc1271'
```

Same for literals in `KeyType`. 

Example of preferred usage of string literals in SDK: https://github.com/streamr-dev/network/blob/1025c2386fcd6f9a1cde970597770dc53efab84b/packages/sdk/src/Config.ts#L243-L244

packages/sdk/src/identity/Identity.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+export type SignerWithProvider = AbstractSigner<Provider>
+
+/**
+ * Identity represents a digital identity, specified by cryptographic keys,
+ * and capable of signing. You can configure new Identities in identityConfig.ts.
+ */
+export abstract class Identity {
+    abstract getUserIdString(): Promise<UserID>
+    abstract getUserIdBytes(): Promise<UserIDRaw>
```
teogeb: Could match the method names to their return types. Great to have both, and one without the suffix is a good default for typical use cases. 

```suggestion
    abstract getUserId(): Promise<UserID>
    abstract getUserIdRaw(): Promise<UserIDRaw>
```

packages/sdk/src/identity/EthereumKeyPairIdentity.ts
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    // eslint-disable-next-line class-methods-use-this
+    getSignatureType(): SignatureType {
+        return SignatureType.ECDSA_SECP256K1_EVM
+    }
+
+    // eslint-disable-next-line class-methods-use-this
+    getExpectedPublicKeyLength(): number {
+        // Address, not actual public key
```
teogeb, line 37: Is there some specific need that this returns the length of the address, not the length of the public key (i.e. 64)?

packages/sdk/src/identity/MLDSAKeyPairIdentity.ts
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * An identity that uses a quantum-resistant ML-DSA-87 key pair to sign messages.
+ */
+export class MLDSAKeyPairIdentity extends KeyPairIdentity {
+    assertKeyPairIsValid(): void {
+        // Validity of key pair is tested by signing and validating something
+        const payload = Buffer.from('data-to-sign')
+        const signature = ML_DSA_87.createSignature(payload, this.privateKey)
+        if (!ML_DSA_87.verifySignature(this.publicKey, payload, signature)) {
+            throw new Error(`The given publicKey and privateKey don't seem to match!`)
```
teogeb:
```suggestion
            throw new Error(`The given publicKey and privateKey don't match!`)
```

packages/sdk/src/identity/identityConfig.ts
```diff
@@ -0,0 +1,77 @@
...
+ * 
+ * How to configure new Identity types:
+ * 1. Add a new SignatureType entry to NetworkRpc.proto in network package
+ * 2. Add the needed SigningUtil to signingUtils.ts
+ * 3. Create the Identity implementation itself (eg. extend KeyPairIdentity)
+ * 4. Wire everything together below
+ */
+export const validKeyTypeValues = ['ECDSA_SECP256K1_EVM', 'ML_DSA_87'] as const
+export const identityConfig: Record<KeyType, {
+    fromConfig: (config: Pick<StrictStreamrClientConfig, 'auth'>) => Identity
```
teogeb: Remove the internal method from public API

```suggestion
    /** @internal */
    fromConfig: (config: Pick<StrictStreamrClientConfig, 'auth'>) => Identity
```

packages/sdk/src/identity/identityConfig.ts
```diff
@@ -0,0 +1,77 @@
...
+ * This is where config keyTypes are connected to Identity implementations.
+ * 
+ * How to configure new Identity types:
+ * 1. Add a new SignatureType entry to NetworkRpc.proto in network package
+ * 2. Add the needed SigningUtil to signingUtils.ts
+ * 3. Create the Identity implementation itself (eg. extend KeyPairIdentity)
+ * 4. Wire everything together below
+ */
+export const validKeyTypeValues = ['ECDSA_SECP256K1_EVM', 'ML_DSA_87'] as const
+export const identityConfig: Record<KeyType, {
```
teogeb: As this is a more a class-like structure than a variable, it would make sense to capitalize the identifier. But then it would collide with `IdentityConfig` type. Maybe could rename either this of the type?

packages/sdk/src/exports.ts
```diff
@@ -44,15 +46,22 @@ export {
     DEFAULT_ENVIRONMENT_ID,
     EntryPointDiscovery
 } from './Config'
+export { validKeyTypeValues, KeyType, DEFAULT_KEY_TYPE, identityConfig } from './identity/identityConfig'
```
teogeb: Do we need to export `identityConfig` in the public API? As identities are typically accessed via explicit classes  (e.g. `EthereumKeyPairIdentity.generate()`), maybe this is not needed.

teogeb (commented):
Third round of comments (most of the sdk)

packages/sdk/src/identity/KeyPairIdentity.ts
```diff
@@ -0,0 +1,68 @@
+import { toUserId, UserID, UserIDRaw } from '@streamr/utils'
+import { RpcProviderSource } from '../RpcProviderSource'
+import { Identity, SignerWithProvider } from './Identity'
+import { KeyPairIdentityConfig, StrictStreamrClientConfig } from '../Config'
+
+export abstract class KeyPairIdentity extends Identity {
+    publicKeyString: UserID
+    publicKey: UserIDRaw
+    privateKey: Uint8Array
+
```
teogeb:
```suggestion
    private readonly publicKeyString: UserID
    private readonly publicKey: UserIDRaw
    private readonly privateKey: Uint8Array

```

packages/cli-tools/bin/streamr-keys-whoami.ts
```diff
@@ -7,5 +7,5 @@ import { createClientCommand } from '../src/command'
 createClientCommand(async (client: StreamrClient) => {
     console.info(await client.getUserId())
 })
-    .description('displays your public address')
+    .description('displays your public key')
```
teogeb:
```suggestion
    .description('displays your user id')
```

packages/cli-tools/bin/streamr.ts
teogeb: Also another approach would be to rename this to `identity`, i.e.: 
- `streamr identity generate`
- `streamr identity whoami`

packages/cli-tools/bin/streamr-keys-generate.ts
teogeb: Currently a `KeyPairIdentity` consists of `userId` + `privateKey`, not `publicKey` + `privateKey`. As `userId` and publicKey are not the same in `Ethereum`, it is good to keep this seperation.

But whenever we output a `userId` it would be good to label it as user ID, not public key.
```suggestion
            console.info(`User ID: ${await identity.getUserIdString()}`)
```

packages/sdk/src/identity/KeyPairIdentity.ts
teogeb: Also good to rename `publicKeys` -> `userId`

packages/sdk/src/identity/EthereumKeyPairIdentity.ts
```diff
@@ -0,0 +1,65 @@
...
+import { RpcProviderSource } from '../RpcProviderSource'
+import { SignatureType } from '@streamr/trackerless-network'
+import { KeyPairIdentityConfig, StrictStreamrClientConfig } from '../Config'
+import { KeyPairIdentity } from './KeyPairIdentity'
+
+/**
+ * An Identity that derives an Ethereum address from a secp256k1 private key
+ * and uses that as the UserID instead of the actual public key.
+ */
+export class EthereumKeyPairIdentity extends KeyPairIdentity {
```
teogeb, line 13: Would it make sense to rename this to `EthereumIdentity`? That renaming should maybe go sync with the subclass, but maybe we could consider also renaming that class (as it doesn't contain `publicKey`+`privateKey` but `userId`+`privateKey`) 

Not sure what would be a good name for the base class.`StaticIdentity`, maybe?

packages/sdk/src/signature/SignatureValidator.ts
```diff
@@ -32,27 +39,36 @@ export class SignatureValidator {
...
+                toUserIdRaw(streamMessage.getPublisherId()),
+                createSignaturePayload(streamMessage),
+                streamMessage.signature
+            )
+        }
+
+        // Special handling: different payload computation, same SigningUtil
+        if (streamMessage.signatureType === SignatureType.ECDSA_SECP256K1_LEGACY) {
+            return ECDSA_SECP256K1_EVM.verifySignature(
+                // publisherId is hex encoded address string
```
teogeb:
```suggestion
                // publisherId is hex encoded Ethereum address string
```

packages/sdk/src/subscribe/messagePipeline.ts
```diff
@@ -51,6 +52,19 @@ export const createMessagePipeline = (opts: MessagePipelineOptions): PushPipelin
     const messageStream = new PushPipeline<StreamMessage, StreamMessage>
     const msgChainUtil = new MsgChainUtil(async (msg) => {
         await validateStreamMessage(msg, opts.streamRegistry, opts.signatureValidator)
+
+        if (!isCompliantEncryptionType(msg.encryptionType, opts.config)) {
+            throw new Error(`A message in stream ${
```
teogeb: Could throw a `StreamrClientError`. We can add a new error code there if needed, e.g. `UNACCEPTABLE_MESSAGE`.

(There we have again some non-unified way to use string enums names: the type is a string literal but keeps old enum naming style, partially for backwards compatibility reasons.)

packages/sdk/src/utils/encryptionCompliance.ts
```diff
@@ -0,0 +1,58 @@
...
+import { RSAKeyPair } from '../encryption/RSAKeyPair'
+import { MLKEMKeyPair } from '../encryption/MLKEMKeyPair'
+import { KeyExchangeKeyPair } from '../encryption/KeyExchangeKeyPair'
+
+const quantumResistantSignatureTypes = new Set<SignatureType>([SignatureType.ML_DSA_87])
+const quantumResistantAsymmetricEncryptionTypes = new Set<AsymmetricEncryptionType>([AsymmetricEncryptionType.ML_KEM])
+const quantumResistantEncryptionTypes = new Set<EncryptionType>([EncryptionType.AES])
+
+export function assertCompliantIdentity(identity: Identity, config: Pick<StreamrClientConfig, 'encryption'>): void {
+    const isQuantumSecure = isQuantumResistantIdentity(identity)
```
teogeb: Could use word quantum _resistant_ in place of quantum _secure_ so that we have only one name for that concept

packages/sdk/src/utils/encryptionCompliance.ts
```diff
@@ -0,0 +1,58 @@
...
+export function isCompliantEncryptionType(encryptionType: EncryptionType, config: Pick<StreamrClientConfig, 'encryption'>): boolean {
+    const isQuantumSecure = quantumResistantEncryptionTypes.has(encryptionType)
+    const quantumSecurityIsRequired = config.encryption?.requireQuantumResistantEncryption
+
+    return isQuantumSecure || !quantumSecurityIsRequired
+}
+
+export function isCompliantSignatureType(signatureType: SignatureType, config: Pick<StreamrClientConfig, 'encryption'>): boolean {
+    const isQuantumSecure = quantumResistantSignatureTypes.has(signatureType)
+    const quantumSecurityIsRequired = config.encryption?.requireQuantumResistantEncryption
```
teogeb: Should use this config option?

```suggestion
    const quantumSecurityIsRequired = config.encryption?.requireQuantumResistantSignatures
```

packages/sdk/test/integration/quantum-policy.test.ts
```diff
@@ -0,0 +1,170 @@
...
+            const testMsg = Msg()
+            await quantumPublisher.publish(stream.id, testMsg)
+            const received = []
+            for await (const msg of sub) {
+                received.push(msg.content)
+                if (received.length === 1) {
+                    break
+                }
+            }
+            expect(received).toEqual([testMsg])
```
teogeb: It is possible to collect the messages easily with the `collect` helper function:

```suggestion
            const received = await collect(sub, 1)
            expect(received.map((m) => m.content)).toEqual([testMsg])
```

packages/sdk/test/integration/quantum-policy.test.ts
```diff
@@ -0,0 +1,170 @@
...
+        })
+
+        // Need to use done callback instead of async/await because we use the error listener
+        it('fails if requirements are violated', (done) => {
+            const testMsg = Msg()
+
+            sub.on('error', (err: Error) => {
+                expect(err.message).toContain('signature')
```
teogeb: When we start to use `StreamrClientError` instances (see another comment), we can assert this with `toEqualStreamrClientError` jest custom matcher. E.g. 

```ts
expect(err).toEqualStreamrClientError({ message: '...', code: '...' })
```

packages/sdk/test/integration/quantum-policy.test.ts
```diff
@@ -0,0 +1,170 @@
...
+                done()
+            })
+
+            nonQuantumClient.publish(stream.id, testMsg)
```
teogeb: If the test can have assertions at the and of the test it is maybe bit more readable. In this case it doesn't matter too much, but we've previously refactored some larger tests to use that kind of style:

```suggestion
            const errorPromise = new Promise<Error>((resolve) => {
                sub.on('error', resolve)
            })
            nonQuantumClient.publish(stream.id, testMsg)
            const err = await errorPromise
            expect(err.message).toContain('signature')
```

packages/sdk/test/integration/quantum-policy.test.ts
```diff
@@ -0,0 +1,170 @@
...
+    afterEach(async () => {
+        await environment.destroy()
+    })
+
+    describe('pubsub under strict quantum settings', () => {
+
+        let sub: Subscription
+
+        beforeEach(async () => {
+            // Clients for pub/sub (can't make transactions)
```
teogeb: Didn't quite catch what kind of transactions are referred here? Why these client would try to do some transactions for pubsub?

packages/sdk/test/unit/validateStreamMessage2.test.ts
```diff
@@ -1,60 +1,60 @@
...
 
 describe('Validator2', () => {
 
     let getStreamMetadata: (streamId: string) => Promise<StreamMetadata>
     let isPublisher: (userId: UserID, streamId: string) => Promise<boolean>
     let isSubscriber: (userId: UserID, streamId: string) => Promise<boolean>
-    let publisherAuthentication: Authentication
-    let subscriberAuthentication: Authentication
+    let publisherAuthentication: Identity
+    let subscriberAuthentication: Identity
```
teogeb:
```suggestion
    let publisherIdentity: Identity
    let subscriberIdentity: Identity
```

teogeb (changes requested):
Rest of the comments about sdk.

Great improvements in this PR, with excellent readability 👍

docs/docs/streamr-network/security/end-to-end-encryption.md
hpihkala: Good catch - this is copy-pasted content from the deleted file `key-encryption.md`. I'll fix it.

packages/cli-tools/bin/streamr-keys-generate.ts
hpihkala: My thinking was that all of the following names and labels should have alignment:
- The names of the SignatureTypes (`SignatureType.ML_DSA_87`)
- The names of the SigningUtils (for example `import { ML_DSA_87 } from '@streamr/utils'`)
- The names of the `keyType` configuration options (for example `keyType: 'ML_DSA_87'`)
- The string value of `Message.signatureType` (for example `ML_DSA_87`)

Note that I'm not making any case re whether these should be upper case or lower case, my point is they should be aligned. Happy to make them lower case instead if preferred? Or are there some arguments why they should not be unified?

packages/cli-tools/bin/streamr-keys-generate.ts
hpihkala: Ok I see how to use the `formEnumArgValueDescription` and `createFnParseEnum` utilities, and I understand that `requireOption` removes the need to check in the action handler 👍 

However, I didn't understand how to avoid repeating the option definition in `command.ts` vs. `streamr-keys-generate.ts`. The `--key-type` should be optional for most commands, however required for `streamr keys generate`. Now it's like this, LMK if you see a better way:

In `command.ts`:

```
.option('--key-type [key-type]', `type of public/private key (${formEnumArgValueDescription(validKeyTypeValues, DEFAULT_KEY_TYPE)})`, createFnParseEnum('key-type', validKeyTypeValues))
```

In `streamr-keys-generate.ts`:

```
.requiredOption('--key-type [key-type]', `type of public/private key (${formEnumArgValueDescription(validKeyTypeValues)})`, createFnParseEnum('key-type', validKeyTypeValues))
```

Since this is a required option, I made the second argument (the default value) to `formEnumArgValueDescription` optional to avoid printing a default value in the help text since the option is required.

Works nicely now, examples:

```
% node dist/bin/streamr.js stream help subscribe
Usage: streamr-stream-subscribe [options] <streamId>

subscribe to a stream, prints JSON messages to stdout line-by-line

Options:
  -V, --version                output the version number
  --private-key <key>          a private key to establish your identity
  --key-type [key-type]        type of public/private key (one of: "ECDSA_SECP256K1_EVM", "ML_DSA_87", default: "ECDSA_SECP256K1_EVM")
  --public-key [public-key]    a public key - required by some key types
  --config <file>              read connection and authentication settings from a config file
  --env <environmentId>        use pre-defined environment (one of: "polygon", "polygonAmoy", "dev2", default: "polygon")
  --quantum                    require quantum resistant key exchange and signature algorithms to be used
  -p, --partition [partition]  partition (default: 0)
  -d, --disable-ordering       disable ordering of messages by OrderingUtil (default: false)
  -r, --raw                    subscribe raw (default: false)
  -m, --with-metadata          print each message with its metadata included (default: false)
  -h, --help                   display help for command
```

```
% node dist/bin/streamr.js keys generate
error: required option '--key-type [key-type]' not specified

Usage: streamr-keys-generate [options]

generate a public/private key pair based on the --key-type option

Options:
  -V, --version          output the version number
  --key-type [key-type]  type of public/private key (one of: "ECDSA_SECP256K1_EVM", "ML_DSA_87")
  -h, --help             display help for command
```

```
% node dist/bin/streamr.js keys generate --key-type foo
key-type must be one of: "ECDSA_SECP256K1_EVM", "ML_DSA_87"
```

packages/cli-tools/bin/streamr-keys-generate.ts
hpihkala: Ok fixed. The JSON format is a little harder to read in case of long keys, but arguably they are hard to read in any case, so I'd say it's not significantly worse.
Screenshot 2025-04-28 at 15 45 24

packages/utils/src/signingUtils.ts
hpihkala: Yeah, naming things is hard... Valid or not, these were the reasons why I didn't pick either of your suggested approaches:
- Abstract class: all the methods are static, so taking an instance-driven approach felt a bit awkward
- Naming: wanted to keep unified naming w/ regard to the other items listed in [this comment](https://github.com/streamr-dev/network/pull/3074#discussion_r2063649048). I don't think there's a naming pattern that would 'feel' right across enums, class names, and string values in config, but so far I prioritized alignment over naming convention. We could have a chat about the naming to see if some ideas come up.

packages/cli-tools/bin/streamr.ts
hpihkala: I like the `identity` option, let's go with that. The problem I have with `key` is that identities are *key pairs*, so the plural more correct than the singular.

packages/sdk/src/Config.ts
hpihkala: My issue here was that the expected object has nothing to do with `ethers` - we only borrow the type definition from ethers. Any type definition that complies with the [spec](https://eips.ethereum.org/EIPS/eip-1193) is equally good. But I'll update the comment to clarify what is expected while keeping the ethers reference.

packages/sdk/src/Config.ts
hpihkala: Yeah I was thinking about this as well, but it needs another round of thinking. 

Originally my thought was this: I was considering potential users of the quantum encryption, for example defense forces. They would want 100% certainty that any and all data traffic is encrypted, and encrypted with quantum resistant methods. That's what this config option currently does.

However, it doesn't make too much sense to prevent **subscribing** to public data despite this config option. Maybe it should only prevent **publishing** unencrypted data? That would at least prevent accidents and nasty stuff like this:
- Attacker compromises chain RPC and makes a stream appear as public
- Publishers start publishing unencrypted data because the stream appears to be public
- Attacker now gains access to the data

The config option would prevent this from happening.

hpihkala: I changed it so that subscribing to public data is ok despite the setting. It makes more sense now. Publishing public data is still prevented, I wouldn't expect that to work if somebody sets `requireQuantumResistantEncryption`.

packages/sdk/src/Message.ts
hpihkala: Deferring this to the discussion about which is better, unified naming or following convention

packages/sdk/src/identity/Identity.ts
hpihkala: That's fine yeah. Perhaps we could even reconsider the names `UserID` and `UserIdRaw` at some point, but that's way out of scope for this PR.

packages/sdk/src/identity/EthereumKeyPairIdentity.ts
hpihkala: This was actually removed in the second PR and the validation made cleaner, so ignoring this here

packages/sdk/src/identity/identityConfig.ts
hpihkala: I'll rename this to `IdentityMapping`

packages/sdk/src/exports.ts
hpihkala: It's needed by the CLI tool, so technically needs to be exported. But I'll mark it with `/** @internal */`

packages/sdk/src/identity/KeyPairIdentity.ts
hpihkala: `private readonly` added.

I think it gets pretty confusing if we use the `userId` name here instead of `publicKey`. This is a `KeyPairIdentity`, and the terminology "public key" and "private/secret key" is very established, while nobody outside of Streamr knows `userId` is. 

With that in mind, I feel we should rather rename `UserID` to `PublicKeyHex` and `UserIDRaw` to `PublicKey`, but maybe out of scope here?

packages/cli-tools/bin/streamr-keys-generate.ts
hpihkala: Somewhat disagree as mentioned above. First of all, in Ethereum terminology, there is no such thing as user ID. There is an 'address', which is derived from the public key by hashing it. For all intents and purposes, the address serves the same function as a public key, it's just slightly compressed.

A key pair consists of a public key and a private key, regardless of whether we (or Ethereum) give weird names like `address` or `userId` to them.

Especially in the bigger picture where Streamr supports many different types of key pairs (all of which call the public key the public key, except the Ethereum one), I would prefer to use the established and more common terminology.

packages/sdk/src/identity/KeyPairIdentity.ts
hpihkala: Ahh actually, at least in this PR they can't be made `private` because they are needed in the subclasses, and they are needed *sync* for the validation that happens in constructor. (`getUserId()` is async).

The validation part changed in the other PR, so I'll make a note there to check back on this.

packages/sdk/src/utils/encryptionCompliance.ts
hpihkala: Yep noticed this mistake too

packages/sdk/test/integration/quantum-policy.test.ts
hpihkala: Very handy! 👍

packages/sdk/test/integration/quantum-policy.test.ts
hpihkala: There's another client instance in this test called `nonQuantumClient` that uses an `EthereumKeyPairIdentity` which can sign transactions to create streams and set permissions. I'll clarify the comment(s).

packages/sdk/src/identity/KeyPairIdentity.ts
hpihkala: Checked the other PR and made these fields `protected`. The change is in the other PR. Naming challenge remains, let's discuss

packages/cli-tools/bin/streamr-identity-generate.ts
```diff
@@ -0,0 +1,24 @@
...
+
+createCommand()
+    .description('generate a public/private key pair based on the --key-type option')
+    .requiredOption('--key-type [key-type]', `type of public/private key (${formEnumArgValueDescription(validKeyTypeValues)})`, 
+        createFnParseEnum('key-type', validKeyTypeValues))
+    .action(async (options: Options) => {
+        const config = IdentityMapping[options.keyType!] // required option
+        if (!config) {
+            console.error(`Error: Invalid key type. Must be one of: ${validKeyTypeValues.join(', ')}.`)
+        }
```
teogeb: This is not needed as `createFnParseEnum()` does the validation:

https://github.com/streamr-dev/network/blob/a4c6a0c3882b1fd0e8f4afaca2f6054bdb4117ff/packages/cli-tools/src/common.ts#L32-L33

```suggestion
```

packages/sdk/src/publish/MessageFactory.ts
```diff
@@ -98,6 +100,9 @@ export class MessageFactory {
         const messageId = new MessageID(this.streamId, partition, msgRef.timestamp, msgRef.sequenceNumber, publisherId, msgChainId)
 
         const encryptionType = (await this.streamRegistry.hasPublicSubscribePermission(this.streamId)) ? EncryptionType.NONE : EncryptionType.AES
+        if (!isCompliantEncryptionType(encryptionType, this.config)) {
+            throw new Error(`Publishing to stream ${this.streamId} was prevented because configuration requires encryption!`)
```
teogeb: This could throw `StreamrClientError`.

(Actually all cases where we throw `new Error(`) could be converted to use `StreamrClientError`. Maybe most important for these kinds of user-facing errors as the error code helps the communication.)

packages/sdk/src/utils/encryptionCompliance.ts
```diff
@@ -0,0 +1,58 @@
...
+const quantumResistantSignatureTypes = new Set<SignatureType>([SignatureType.ML_DSA_87])
+const quantumResistantAsymmetricEncryptionTypes = new Set<AsymmetricEncryptionType>([AsymmetricEncryptionType.ML_KEM])
+const quantumResistantEncryptionTypes = new Set<EncryptionType>([EncryptionType.AES])
+
+export function assertCompliantIdentity(identity: Identity, config: Pick<StreamrClientConfig, 'encryption'>): void {
+    const isQuantumResistant = isQuantumResistantIdentity(identity)
+    const quantumResistanceIsRequired = config.encryption?.requireQuantumResistantSignatures
+
+    if (quantumResistanceIsRequired && !isQuantumResistant) {
+        throw new Error(`Quantum resistant signatures are required, but the configured key type doesn't enable quantum resistance!`)
```
teogeb: This could throw `StreamrClientError`

teogeb (changes requested):
Some more comments

packages/cli-tools/bin/streamr-keys-generate.ts
hpihkala: We discussed this with Teo and decided to go with this for now

packages/sdk/src/identity/KeyPairIdentity.ts
hpihkala: Chatted about this, ok as is

packages/utils/src/signingUtils.ts
hpihkala: @teogeb I will do the `getInstance` refactor in the other PR - it's easier there because I changed the interface in that PR + added the new type

packages/cli-tools/bin/streamr-keys-generate.ts
hpihkala: We talked about this and agreed that `UserID` is on a higher abstraction level compared to public key. `KeyPairIdentity` - consisting of a public key and private key - is defined as an `Identity` that uses the public key as the user ID. Therefore it's exactly right to:
- On the `Identity` level, use the highest abstraction level (define getters like `getUserId()`)
- On the `KeyPairIdentity` level, internally it's clearest to use `this.publicKey` and `this.privateKey` since those concepts are closest to the definition of what a `KeyPairIdentity` is, and `KeyPairIdentity` uses signing functions etc. that use the public key / private key terminology, so it's best to maintain naming correspondence with them.

teogeb (approved):
👍

PR #3077, opened 2025-04-22
feat(cli-tools): Add internal `operator-create` command

teogeb:
Added an internal command for creating operator contracts. Supports only `dev2` environment.

## Usage

```
streamr internal operator-create --cut 12.3 --redundancy-factor 2 --node-addresses  '...,...' --env dev2 --private-key ...
```

## Other changes

Small refactoring to `operatorContractUtils` so that `StreamrClient#client.getSigner()` is supported as `deployer`.

## Future improvements

Could promote this to be non-internal command. Currently `operatorContractUtils#deployOperatorContract()` always uses `TEST_CHAIN_CONFIG`, but we could pass e.g. environment ID as a parameter to support also other environments.

### DIFF
--- a/packages/sdk/src/contracts/operatorContractUtils.ts
+++ b/packages/sdk/src/contracts/operatorContractUtils.ts
@@ -66,7 +66,7 @@ export async function setupOperatorContract(
  * @hidden
  */
 export interface DeployOperatorContractOpts {
-    deployer: Wallet
+    deployer: SignerWithProvider
     operatorsCutPercentage?: number
     metadata?: string
     operatorTokenName?: string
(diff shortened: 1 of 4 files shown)


packages/sdk/src/contracts/operatorContractUtils.ts
```diff
@@ -80,7 +80,7 @@ export async function deployOperatorContract(opts: DeployOperatorContractOpts):
     logger.debug('Deploying OperatorContract')
     const abi = OperatorFactoryArtifact
     const operatorFactory = new Contract(TEST_CHAIN_CONFIG.contracts.OperatorFactory, abi, opts.deployer) as unknown as OperatorFactoryContract
-    const contractAddress = await operatorFactory.operators(opts.deployer.address)
+    const contractAddress = await operatorFactory.operators(await opts.deployer.getAddress())
```
jtakalai, line 83: getAddress is valid for Wallets too 👍

jtakalai (approved):

PR #3079, opened 2025-04-23
chore(sdk) [NET-1452]: Optimize ABI JSON files

teogeb:
Optimized SDK bundle size by converting contract ABI JSONs from `json` format to `full` format (https://docs.ethers.org/v6/api/abi/abi-coder/#FormatType). 

The full `format` is more compact than the `json` format. As we use named parameters in events (e.g. `deploySponsorshipContract()`) and structs (e.g. `convertChainPermissionsToStreamPermissions()`) we can't use the `minimal` format.

## Statistics

The `streamr-sdk.web.min.js` bundle decreased  67k in size.

## Conversion

These ABIs are converted from the full format ABIs by running this script:

```ts
import fs from 'fs'
import { Interface } from 'ethers'

const filePath = process.argv[2]
const rawAbi = JSON.parse(fs.readFileSync(filePath))
const iface = new Interface(rawAbi)
const formattedAbi = Object.values(iface.fragments).map(fragment => {
    // use full format so that we can use named event parameters
    return fragment.format('full')
})
fs.writeFileSync(filePath, JSON.stringify(formattedAbi, undefined, 4))
```


jtakalai (approved):
>Do we need error fragments?

Depends who this Contract object is for. Maybe if it's given out of SDK then it's good to have "everything" in. We can also parse better errors from them.

>Would it be possible to configure ether.js to utilize the error fragment information in order to provide us better error messages?

Sure

>Currently the contract errors typically contain the selector hash in the error messages (e.g. MinimumStake(uint256) = 0be9186a) but it is not automatically parsed by ethers.js?

Not sure, but can be manually done in error handling, though it might be flaky since RPCs vary in the format they answer with

PR #3081, opened 2025-04-23
refactor(sdk): Operator contract utils without token ABI

teogeb:
Refactored `delegate()` and `sponsor()` so that the token argument is not needed. The token address is now queried via a `Operator`/`Sponsorship` contract instance.

## Other changes

- removed obsolete `delegator` parameter from `transferAndCall()`
- simplified ABI definition in `getTestTokenContract()`
- added logging for `undelegate()`

## Future improvements

As we no longer need to pass the `token` argument for `delegate()` and `sponsor()` functions it should be possible to promote the related internal cli-tool command to non-internal commands.


jtakalai (approved):
>delegate doesn't need the data parameter, was ok to remove?

is ok to remove

PR #3083, opened 2025-04-24
feat(network): [NET-1451] Add option for Plum Tree like optimization of stream message propagation

juslesan:
## Summary

Added support for Plum Tree like optimization of stream traffic. With the mode enabled nodes propagate full messages only through the most optimal path. The path is constructed based on duplicate messages received by the node when it initially joins the stream. Nodes will propagate the `MessageID`-field of the Stream Messages to any nodes that are not on the optimal path. This allows reconstruction of the propagation tree in the presence of churn.

## Changes

- Added `StreamPartDeliveryOptions` for streams that make it possible to pass options of a per stream partition basis. This  structure is used to pass the option to use the plum tree optimization.

## TODO:

- More testing, unit tests


packages/trackerless-network/src/logic/ContentDeliveryRpcLocal.ts
```diff
@@ -29,10 +31,22 @@ export class ContentDeliveryRpcLocal implements IContentDeliveryRpc {
...
+        const previousNode = (context as DhtCallContext).incomingSourceDescriptor!
+        const previousNodeId = toNodeId(previousNode)
+        this.options.markForInspection(previousNodeId, message.messageId!)
+        if (this.options.plumTreeManager === undefined) {
+            if (this.options.markAndCheckDuplicate(message.messageId!, message.previousMessageRef)) {
+                this.options.broadcast(message, previousNodeId)
+            }
+        } else if (this.options.markAndCheckDuplicate(message.messageId!, message.previousMessageRef)) {
+            this.options.plumTreeManager.broadcast(message, previousNodeId)
+            await this.options.plumTreeManager.resumeNeighbor(
```
ptesavol: add comments to this branch and the else branch

packages/trackerless-network/src/logic/plumtree/PlumTreeManager.ts
```diff
@@ -0,0 +1,136 @@
...
+const logger = new Logger(module)
+
+interface Events {
+    message: (msg: StreamMessage, previousNode: DhtAddress) => void
+}
+
+export class PlumTreeManager extends EventEmitter<Events> {
+    private neighbors: NodeList
+    private localPeerDescriptor: PeerDescriptor
+    private localPausedNeighbors: PausedNeighbors = new PausedNeighbors()
```
ptesavol: add comments to clarify what are these localPausedNeighbors and remotePausedNeighbors

packages/trackerless-network/src/logic/plumtree/PlumTreeManager.ts
```diff
@@ -0,0 +1,136 @@
...
+    async sendBuffer(fromTimestamp: number, neighbor: PeerDescriptor): Promise<void> {
+        const remote = new ContentDeliveryRpcRemote(this.localPeerDescriptor, neighbor, this.rpcCommunicator, ContentDeliveryRpcClient)
+        const messages = this.latestMessages.filter((msg) => msg.messageId!.timestamp >= fromTimestamp)
+        await Promise.all(messages.map((msg) => remote.sendStreamMessage(msg)))
+    }
+
+    async onMetadata(msg: MessageID, previousNode: PeerDescriptor): Promise<void> {
+        // If the number of messages in the buffer is greater than 1, resume the sending neighbor
+        // This is done to avoid oscillation of the neighbors during propagation
+        if (this.latestMessages.filter((m) => m.messageId!.timestamp >= msg.timestamp).length > 1) {
```
ptesavol: implement working logic for this functionality

ptesavol (changes requested):
see comments

packages/trackerless-network/src/logic/plumtree/PlumTreeManager.ts
```diff
@@ -0,0 +1,157 @@
...
+interface Events {
+    message: (msg: StreamMessage, previousNode: DhtAddress) => void
+}
+
+export class PlumTreeManager extends EventEmitter<Events> {
+    private readonly neighbors: NodeList
+    private readonly localPeerDescriptor: PeerDescriptor
+    // These neighbors are paused by the local node
+    private readonly localPausedNeighbors: PausedNeighbors = new PausedNeighbors()
+    // These neighbors are paused by remote nodes, used to limit sending of pausing and resuming requests
```
ptesavol: maybe still would be better to rename the variables to neighborsThatWeHavePaused and neighborsThatHavePausedUs?

packages/trackerless-network/src/logic/plumtree/PlumTreeManager.ts
```diff
@@ -0,0 +1,157 @@
...
+        const latestMessageTimestamp = this.getLatestMessageTimestamp(msg.messageChainId)
+        if (latestMessageTimestamp < msg.timestamp) {
+            if (!this.metadataTimestampsAheadOfRealData.has(msg.messageChainId)) {
+                this.metadataTimestampsAheadOfRealData.set(msg.messageChainId, new Set())
+            }
+            this.metadataTimestampsAheadOfRealData.get(msg.messageChainId)!.add(msg.timestamp)
+            if (this.metadataTimestampsAheadOfRealData.get(msg.messageChainId)!.size > 1) {
+                console.log("resuming neighbor", latestMessageTimestamp, msg.timestamp, this.remotePausedNeighbors, this.localPausedNeighbors, this.neighbors.getAll().map((n) => toNodeId(n.getPeerDescriptor())))
+                await this.resumeNeighbor(previousNode, msg.messageChainId, this.getLatestMessageTimestamp(msg.messageChainId))
+                this.metadataTimestampsAheadOfRealData.get(msg.messageChainId)!.forEach((timestamp) => {
```
ptesavol, line 110: I think there is a possible race condition here. More metadatas might arrive while the resume is still in progress and we would call the resume again (maybe on the same neighbor, maybe on another). Should we add a resumeInProgress flag?

packages/trackerless-network/src/logic/plumtree/PlumTreeManager.ts
juslesan: We are using the remote / local pattern across the monorepo

packages/trackerless-network/src/logic/plumtree/PlumTreeManager.ts
juslesan: This is likely not required since any unnecessary resumed connections are quickly closed once duplicate messages arrive.

packages/trackerless-network/src/logic/plumtree/PlumTreeManager.ts
ptesavol: ok, in that case I'll modify the comments a little bit to make the situation clearer to me..I am still quite confused

ptesavol (approved):
This looks ok now

PR #3086, opened 2025-04-25
feat(node): [NET-1455] Autostaker plugin

teogeb:
## Summary

An experimental plugin that automatically manages sponsorship stakings. It periodically checks which sponsorships offer good payouts and stakes tokens to those sponsorships when possible (i.e. if there are unstaked tokens available or tokens can be unstaked from another sponsorship with a worse payout).

## Features

- Uses a payout-proportional strategy, see details in the comment of `payoutProportionalStrategy.ts`
- Simple error handling: if an error occurs during staking or unstaking, all subsequent actions in that run are skipped (this is acceptable because the actions will be retried in the next scheduled run)
- Uses an approximate scheduler
- Operators select different sponsorships if multiple sponsorships have the same `payoutPerSec`
- Filters out:
  - small transactions (the minimum transaction size is configurable via `minTransactionDataTokenAmount`)
  - sponsorships with too small payouts
  - sponsorships with a non-zero `minimumStakingPeriodSeconds`
  - sponsorships that require too many operators to join before payouts start (configurable via `maxAcceptableMinOperatorCount`)
  - sponsorships that have already reached the maximum number of operator stakes
  
## Authentication

As the plugin does transaction on behalf of the operator, it needs either operator's private key or a `CONTROLLER` role assigned to it so that it can act as a [staking agent](https://docs.streamr.network/streamr-network/network-roles/operators/#staking-agents). Granting the role is the preferred approach since controllers don't have permission to withdraw funds from the `Operator` contract

## Future Improvements

- Support sponsorships with `minimumStakingPeriodSeconds` by implementing logic to prevent premature unstaking, instead of filtering them out.
- Convert the smoke test to an integration test and run it in CI. The test is currently quite slow.
- Consider combining the rounding logic and the filtering of small transactions. A possible approach: 1) skip rounding, 2) filter out transactions that are too small, 3) balance the net change by removing or reducing the smallest stakes.  This would stake all available tokens but might do too many transactions overall.


packages/node/src/plugins/autostaker/AutostakerPlugin.ts
```diff
@@ -0,0 +1,87 @@
...
+                    }
+                `
+            }
+        })
+        const sponsorships: { id: string }[] = await collect(queryResult)
+        logger.info(`Available sponsorships: ${sponsorships.map((s) => s.id).join(',')}`)
+        if ((sponsorships.length) > 0 && (availableBalance >= STAKE_AMOUNT)) {
+            const targetSponsorship = sample(sponsorships)!
+            logger.info(`Stake ${formatEther(STAKE_AMOUNT)} to ${targetSponsorship.id}`)
+            await _operatorContractUtils.stake(
```
jtakalai: the "what" to do should be separated from the "when", like it is in the simulator

packages/node/src/plugins/autostaker/AutostakerPlugin.ts
```diff
@@ -0,0 +1,87 @@
...
+            .connect(new Wallet(this.pluginConfig.operatorOwnerPrivateKey, provider))
+        const stakedAmount = await operatorContract.totalStakedIntoSponsorshipsWei()
+        const availableBalance = (await operatorContract.valueWithoutEarnings()) - stakedAmount
+        logger.info(`Available balance: ${formatEther(availableBalance)} (staked=${formatEther(stakedAmount)})`)
+        // TODO is there a better way to get the client? Maybe we should add StreamrClient#getTheGraphClient()
+        // TODO what are good where consitions for the sponsorships query
+        // @ts-expect-error private
+        const queryResult = streamrClient.theGraphClient.queryEntities((lastId: string, pageSize: number) => {
+            return {
+                query: `
```
jtakalai, line 134: > How we should query the possible sponsorships, which we could stake to? Now we use projectedInsolvency_gt and spotAPY_gt fllters, but maybe these are not optimall.

`remainingWei_gt: "0"` is same as "the projectedInsolvency is in the past". It's a fine way to select only "active" sponsorships. Both should work the same.

There might not need to be hard cutoffs for other this, just order by `totalPayoutWeiPerSec` desc should be fine, take some reasonable maximum number, e.g. 100.

packages/node/bin/autostaker-manual-test.sh
```diff
@@ -0,0 +1,52 @@
+# Do not merge this manual test to main
+
+NODE_PRIVATE_KEY="1111111111111111111111111111111111111111111111111111111111111111"
+OWNER_PRIVATE_KEY="2222222222222222222222222222222222222222222222222222222222222222"
```
jtakalai: > We wouldn't like to require that operators provide their private key for this plugin. What other possibilities there are? The nodes can nowadays vote and flag behalf of the operator, so maybe they could also stake/unstake if that kind of permission could be added?

Nodes do node things; there's another role that can be used for staking/unstaking: CONTROLLER_ROLE. This is added by operator calling `operatorContract.grantRole("0x7b765e0e932d348852a6f810bfa1ab891e259123f02db8cdcde614c570223357", controllerAddress)` (where that hex mush can be read from `operatorContract.CONTROLLER_ROLE()`)

This is probably preferable as a key management mechanism, so that the autostaker would be run with a private key of this kind of "controller".

There's `revokeRole` for removing that role.

packages/node/src/plugins/autostaker/AutostakerPlugin.ts
jtakalai: i.e. put "what to do with the sponsorships from thegraph" into a separate strategies directory

packages/node/src/plugins/autostaker/AutostakerPlugin.ts
jtakalai: On the other hand, in tests, remainingWei isn't updated instantly; in a test, we can do a "zero sponsoring" to update the remainingWei (after it should be expired).

jtakalai: The current `projectedInsolvency_gt: ${Math.floor(Date.now() / 1000)}` is in other words totally fine, and perhaps it's easier to write the test as well. I'll mark this conversation resolved.

packages/node/src/plugins/autostaker/payoutProportionalStrategy.ts
```diff
@@ -0,0 +1,177 @@
...
+        }
+    }
+
+    const tooSmallAdjustments = adjustments.filter((a) => abs(a.difference) < operatorConfig.minTransactionAmount)
+    if (tooSmallAdjustments.length > 0) {
+        pull(adjustments, ...tooSmallAdjustments)
+        let netDifference = sum(tooSmallAdjustments.map((a) => a.difference))
+        while (netDifference < 0) {
+            // there are more stakings than unstakings: remove smallest of the stakings
+            const smallestStaking = minBy(adjustments.filter((a) => a.difference > 0), (a) => Number(a.difference))!
```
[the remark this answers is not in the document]
jtakalai: should be fine for comparator finding the smallest item

packages/node/src/plugins/autostaker/payoutProportionalStrategy.ts
```diff
@@ -0,0 +1,177 @@
...
+    )
+
+    const adjustments = [...targetStakes.keys()]
+        .map((sponsorshipId) => ({ 
+            sponsorshipId,
+            difference: targetStakes.get(sponsorshipId)! - (operatorState.stakes.get(sponsorshipId) ?? 0n)
+        }))
+        .filter(({ difference: difference }) => difference !== 0n)
+
+    // fix rounding errors by forcing the net staking to equal unstakedAmount: adjust the largest staking
```
jtakalai: I'd do the "tooSmallAdjustments" before this "force net staking to equal unstakedAmount", because there are two points to this forcing:
* that we never run out of tokens
* that all tokens are put to work

I see in the tooSmallAdjustments a safety to never adjust to remove more unstaking than staking, that's good and keeps the token balance on the correct side. Just then the comment "forcing the net staking to equal unstakedAmount" isn't accurate for the end-result.

I think it would make sense to just first filter out those too small adjustments, then do the forcing; that way, net staking will be same as unstakedAmount (i.e. all tokens are put to work) but also we shouldn't be able to run out of tokens.

packages/node/src/plugins/autostaker/payoutProportionalStrategy.ts
```diff
@@ -0,0 +1,177 @@
...
+    const netStakingAmount = sum(adjustments.map((a) => a.difference))
+    if (netStakingAmount !== operatorState.unstakedAmount && stakeableSponsorships.size > 0 && adjustments.length > 0) {
+        const largestDifference = maxBy(adjustments, (a) => Number(a.difference))!
```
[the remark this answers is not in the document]
jtakalai: probably fine for comparison

packages/node/src/plugins/autostaker/AutostakerPlugin.ts
```diff
@@ -0,0 +1,178 @@
...
+                this.pluginConfig.operatorContractAddress,
+                action.sponsorshipId,
+                action.amount
+            )
+        }
+    }
+
+    // eslint-disable-next-line class-methods-use-this
+    private async getStakeableSponsorships(streamrClient: StreamrClient): Promise<Map<SponsorshipID, SponsorshipConfig>> {
+        // TODO what are good where conditions for the sponsorships query so that we get all stakeable sponsorships
```
harbu: TODO still relevant?

packages/node/src/plugins/autostaker/AutostakerPlugin.ts
```diff
@@ -0,0 +1,178 @@
...
+                        ) {
+                            id
+                            totalPayoutWeiPerSec
+                        }
+                    }
+                `
+            }
+        })
+        const sponsorships = await collect(queryResult)
+        return new Map(sponsorships.map(
```
harbu: This was not in scope of the original ticket but:

I think we definitely need to add some simple handling for `minimumStakingPeriodSeconds` here. Either we (a) simply ignore / skip sponsorships that have set it to something greater than zero or (b) we have some maximum value we are prepared to handle (let's say 5 days) but the sponsorship shouldn't go insolvent within that period.

packages/node/src/plugins/autostaker/AutostakerPlugin.ts
```diff
@@ -0,0 +1,178 @@
...
+                unstakedAmount
+            },
+            operatorConfig: {
+                operatorContractAddress: this.pluginConfig.operatorContractAddress,
+                minTransactionAmount: parseEther(String(this.pluginConfig.minTransactionDataTokenAmount)),
+                maxSponsorshipCount: this.pluginConfig.maxSponsorshipCount
+            },
+            stakeableSponsorships,
+            environmentConfig: {
+                minStakePerSponsorship: 5000000000000000000000n  // TODO read from The Graph (network.minimumStakeWei)
```
harbu: Implement TODO

packages/node/src/plugins/autostaker/config.schema.json
```diff
@@ -0,0 +1,34 @@
...
+    "operatorContractAddress": {
+      "type": "string",
+      "description": "Operator contract Ethereum address",
+      "format": "ethereum-address"
+    },
+    "runIntervalInMs": {
+      "type": "integer",
+      "description": "The interval (in milliseconds) at which autostaking possibilities are analyzed and executed",
+      "minimum": 0,
+      "default": 30000
```
harbu: This should be maybe every 30min or smth?

packages/node/src/plugins/autostaker/AutostakerPlugin.ts
```diff
@@ -0,0 +1,178 @@
...
+            }
+        })
+        const signer = await streamrClient.getSigner()
+        for (const action of actions) {
```
harbu, line 116: It's critical that unstakes are executed before stakes. Haven't yet delved into `adjustStakes` function to see details of how the list is ordered. This could be somehow highlighted in a code comment so future maintainers do not accidentally break it.

packages/node/src/plugins/autostaker/AutostakerPlugin.ts
```diff
@@ -0,0 +1,178 @@
...
+        scheduleAtApproximateInterval(async () => {
+            try {
+                await this.runActions(streamrClient)
+            } catch (err) {
+                logger.warn('Error while running autostaker actions', { err })
+            }
+        }, this.pluginConfig.runIntervalInMs, 0.1, false, this.abortController.signal)
+    }
+
+    private async runActions(streamrClient: StreamrClient): Promise<void> {
```
harbu: In theory some `abortSignal` checks could be added between `await` invocations in the `runActions` body but I don't think in practice they bring a lot of value (and contribute to reduced code clarity).

packages/node/src/plugins/autostaker/AutostakerPlugin.ts
```diff
@@ -0,0 +1,178 @@
...
+                maxSponsorshipCount: this.pluginConfig.maxSponsorshipCount
+            },
+            stakeableSponsorships,
+            environmentConfig: {
+                minStakePerSponsorship: 5000000000000000000000n  // TODO read from The Graph (network.minimumStakeWei)
+            }
+        })
+        const signer = await streamrClient.getSigner()
+        for (const action of actions) {
+            logger.info(`Action: ${action.type} ${formatEther(action.amount)} ${action.sponsorshipId}`)
```
harbu: Logging guidelines would suggest something like:
```suggestion
            logger.info(`Execute action: ${action.type} ${formatEther(action.amount)} ${action.sponsorshipId}`)
```

packages/utils/src/sum.ts
```diff
@@ -0,0 +1,3 @@
+export function sum(values: bigint[]): bigint {
+    return values.reduce((acc, value) => acc + value, 0n)
+}
```
harbu, line 3: Too generic name imo, would prefer `sumBigintArray` or smth. I would also just keep the util in the autostaker plugin directory.

Doesn't lodash `sum` achieve the same?

harbu (commented):
I still need to read thru
- packages/node/src/plugins/autostaker/payoutProportionalStrategy.ts
- packages/node/test/smoke/autostaker.test.ts
- packages/node/test/unit/plugins/autostaker/payoutProportionalStrategy.test.ts

But here is first round of comments

jtakalai (approved):

packages/utils/src/sum.ts
teogeb: It seems to work in practice, at least partially:
```ts
const foo = sum([1, 2, 3, 10])
const bar = sum([1n, 2n, 3n, 4n])
console.log(foo, bar)
console.log(typeof foo) -> number
console.log(typeof bar) -> bigint
```

But the official types define this signature: `sum(collection: List<any> | null | undefined): number;`, so maybe best just to copy the `sum()` to `payoutProportionalStrategy.ts`.

harbu:
Two things that need solving still
- Handling of `minimumStakingPeriodSeconds`. I suggest we just ignore / skip sponsorships that have this set (in the spirit of MVP). Future improvement here is to add support for such sponsorships.
- Handling of `minOperatorCount` (or whatever the variable is named). I suggest there is a config setting (e.g. `maxAcceptableMinOperatorCount`) that controls what is the maximum acceptable number for this config for a sponsorship. Any sponsorship requiring a higher amount of operators is simply ignored / skipped.

packages/node/src/plugins/autostaker/payoutProportionalStrategy.ts
```diff
@@ -0,0 +1,181 @@
...
+        minStakePerSponsorship,
+        maxSponsorshipCount,
+        operatorContractAddress
+    )
+    const minStakePerSponsorshipSum = BigInt(selectedSponsorships.length) * minStakePerSponsorship
+    const payoutProportionalAmount = totalStakeableAmount - minStakePerSponsorshipSum
+    const payoutPerSecSum = sum(selectedSponsorships.map((id) => stakeableSponsorships.get(id)!.payoutPerSec))
+    const targetsForSelected: TargetStake[] = selectedSponsorships.map((id) => [
+        id,
+        minStakePerSponsorship + payoutProportionalAmount * stakeableSponsorships.get(id)!.payoutPerSec / payoutPerSecSum
```
harbu, line 119: Cursor brought this up.

Can `payoutPerSecSum` be zero? Division by zero could be an issue then. Generally sponsorships pay out something I suppose. But I wonder if in certain edge scenarios this could happen?

```ts
const payoutPerSecSum = sum(selectedSponsorships.map((id) => stakeableSponsorships.get(id)!.payoutPerSec))
const targetsForSelected: TargetStake[] = selectedSponsorships.map((id) => [
    id,
    payoutPerSecSum > 0n 
        ? minStakePerSponsorship + payoutProportionalAmount * stakeableSponsorships.get(id)!.payoutPerSec / payoutPerSecSum
        : minStakePerSponsorship
])
```

packages/node/src/plugins/autostaker/payoutProportionalStrategy.ts
```diff
@@ -0,0 +1,181 @@
...
+        }
+    }
+
+    const tooSmallAdjustments = adjustments.filter((a) => abs(a.difference) < operatorConfig.minTransactionAmount)
+    if (tooSmallAdjustments.length > 0) {
+        pull(adjustments, ...tooSmallAdjustments)
+        let netDifference = sum(tooSmallAdjustments.map((a) => a.difference))
+        while (netDifference < 0) {
+            // there are more stakings than unstakings: remove smallest of the stakings
+            const smallestStaking = minBy(adjustments.filter((a) => a.difference > 0), (a) => Number(a.difference))!
```
harbu: > Bug: The `minBy` function could return undefined if there are no positive adjustments left in the array. The non-null assertion ! would then cause a runtime error.

Cursor brought up this. However, I don't think this can happen. If `netDifference < 0` there must be at least one staking left?

packages/node/src/plugins/autostaker/payoutProportionalStrategy.ts
teogeb: Seems that it is currently ok for a sponsorship to not to pay out (i.e. `totalPayoutWeiPerSec` is zero). Added MIN_SPONSORSHIP_TOTAL_PAYOUT_PER_SECOND filter to avoid staking to such sponsorships.

It also avoids sponsorships, which have very small payout: e.g. the sponsorship has been already been paid out almost entirely but there is still some very small token amount left there.

packages/node/src/plugins/autostaker/payoutProportionalStrategy.ts
teogeb: There was an edge case in which there were only very small unstakes (due to sponsorship expirations). This threw an error in that situation if `minTransactionAmount` >= `minStakePerSponsorship`. Improved the logic so that it handles also that situation. The implementation is now more straightforward.

packages/node/src/plugins/autostaker/AutostakerPlugin.ts
teogeb: Not worth adding, just unnecessary complexity.

packages/node/src/plugins/autostaker/AutostakerPlugin.ts
teogeb: There is very explicit sorting in `payoutPropotionalStrategy.ts`:
```ts
(action) => ['unstake', 'stake'].indexOf(action.type)
```

But makes sense to add that to the function documentation as a requirement.

packages/node/src/plugins/autostaker/AutostakerPlugin.ts
harbu: We went with (a)

packages/node/src/plugins/autostaker/AutostakerPlugin.ts
```diff
@@ -142,13 +144,18 @@ export class AutostakerPlugin extends Plugin<AutostakerPluginConfig> {
...
                         }
                     }
                 `
             }
         })
         const sponsorships = await collect(queryResult)
-        return new Map(sponsorships.map(
+        const hasAcceptableOperatorCount = (item: SponsorshipQueryResultItem) => {
+            return (item.maxOperators === null) || (item.operatorCount < item.maxOperators)
+        }
```
harbu: We need to take into consideration whether we already have a stake, and if we do, the strict less-than should become less-than-or-equal.

packages/node/test/unit/plugins/autostaker/payoutProportionalStrategy.test.ts
```diff
@@ -286,5 +286,31 @@ describe('payoutProportionalStrategy', () => {
...
+                operatorState: { unstakedAmount: 0n, stakes: new Map([
+                    ['a', 10n],
+                    ['b', 1000n]
+                ]) },
+                operatorConfig: { minTransactionAmount: 50n, operatorContractAddress: '' },
+                stakeableSponsorships: new Map([]),
+                environmentConfig: { minStakePerSponsorship: 0n }
+            })).toEqual([
+                { type: 'unstake', sponsorshipId: 'b', amount: 1000n }
+            ])
```
harbu: We probably need to consider the following exceptional case:
- Sponsorship `S` is expired (i.e. it doesn't pay out anything anymore)
- But current operator has stake `n < minTransactionAmount` in sponsorship `S`

With current logic, we would not unstake because the transaction amount requirement is not fulfilled. However, what the user probably(?) wants is that we would unstake in this circumstance.

packages/node/src/plugins/autostaker/payoutProportionalStrategy.ts
teogeb: Added an item to "Future improvements"

harbu (changes requested):

packages/node/src/plugins/autostaker/AutostakerPlugin.ts
```diff
@@ -129,7 +129,10 @@ export class AutostakerPlugin extends Plugin<AutostakerPluginConfig> {
         }
     }
 
-    private async getStakeableSponsorships(streamrClient: StreamrClient): Promise<Map<SponsorshipID, SponsorshipConfig>> {
+    private async getStakeableSponsorships(
+        stakes: Map<SponsorshipID, WeiAmount>,
```
harbu, line 127: I would maybe give a more specific name e.g `currentStakes` or `myCurrentStakes`.

harbu (approved):

PR #3088, opened 2025-04-25
feat: [NET-1453] Support ECDSA with the secp256r1 curve

hpihkala:
## Changes

- Add support for ECDSA on `secp256r1` curve. Uses webcrypto (SubtleCrypto) for speed, as that's natively implemented in browsers.
- Had to refactor `createSignature` and `verifySignature` in `SigningUtil` interface to be `async`, because webcrypto (SubtleCrypto) only offers async methods.
- Offer an `assertValidKeyPair` utility function in all `SigningUtil`s
- The previous PR #3074 left some type safety holes in `signingUtils.ts`, they are fixed here.
- Unify `EthereumKeyPairIdentity` constructor with other `KeyPairIdentity` constructors. There's now a utility function `EthereumKeyPairIdentity.fromPrivateKey(string)` for creating it from a private key string.
- Moved `crossPlatformCrypto.ts` (utility to get `subtle`) from `sdk` package to `utils` package, as it was needed there too.

- [x] Is this a breaking change? If it is, be clear in summary. **-> NO**


hpihkala:
Note to self, based on the base PR: see if these can be made `private` in `KeyPairIdentity`:

```
private readonly publicKeyString: UserID
private readonly publicKey: UserIDRaw
private readonly privateKey: Uint8Array
```

EDIT - need sync access to `this.publicKey` and `this.privateKey` in subclasses, so set to `protected` instead. ✅

hpihkala:
Note to self: refactor `SigningUtil` interface to abstract class in this PR ✅

docs/docs/streamr-network/security/signing-and-verification.md
```diff
@@ -10,17 +10,15 @@ The type of cryptographic keys configured on the publisher determine the signatu
...
 
-The Streamr SDK is configured with cryptographic keys which determine the user's identity in the network. The type of keys used also determines the signature algorithm used. The Streamr SDK supports the following key/identity/signature schemes:
-
-- ECDSA with the `secp256k1` curve - commonly referred to in Streamr context as an 'Ethereum' identity, defined by a private key and an 'address' derived from the related public key. This is the default and most commonly used identity type, and the only one capable of signing transactions on EVM blockchains.
-- ML-DSA for quantum resistant signatures. Learn more about [Quantum Security](quantum-security.md).
+The Streamr SDK is configured with cryptographic keys which determine the user's identity in the network. The type of keys used also determines the signature algorithm used. The key/identity/signature schemes supported by Streamr are detailed in this article.
 
 The choice of keys and algorithms depends on the use case and whether the same identities are used outside of Streamr. The following table specifies the functionality available to each key type as well as the `keyType` and `Identity` class name required to use it:
```
teogeb, line 15: Should we use phrase `keyType` or `keyPairType`?

To me `keyType` sounds better as it is a short and descriptive name. As the constant names (`ECDSA_SECP256R1`, `ML_DSA_87`, ...) refer to algorithms, it is equally ok to say either that:
- the keys in the key pair are of that type
- the whole key pair is of that type

Either is fine. But shorter names are bit easier to use in the configs and in the command line, and bit more readable in the code.

If we keep the phrase `keyPairType`, then renaming should be done in these files (as the phrase `keyType` is used in place of `keyPairType`):
```
packages/cli-tools/bin/streamr-identity-generate.ts
packages/cli-tools/src/client.ts
packages/cli-tools/src/command.ts
packages/sdk/src/Config.ts
packages/sdk/src/identity/IdentityMapping.ts
docs/docs/streamr-network/security/quantum-security.md
docs/docs/streamr-network/security/signing-and-verification.md
```

packages/utils/src/signingUtils.ts
```diff
@@ -1,95 +1,271 @@
...
-import { areEqualBinaries } from './binaryUtils'
-import { UserIDRaw } from './UserID'
 import { randomBytes } from '@noble/post-quantum/utils'
+import { p256 } from '@noble/curves/p256'
+import { areEqualBinaries, binaryToHex } from './binaryUtils'
+import { UserIDRaw } from './UserID'
+import { getSubtle } from './crossPlatformCrypto'
+import { webcrypto } from 'crypto'
+
+export const SUPPORTED_KEY_PAIR_TYPES = [
```
teogeb: `SUPPORTED_` prefix would indicate that we have some specific unsupported key types

```suggestion
export const KEY_PAIR_TYPES = [
```

teogeb (commented):
First round of comments

docs/docs/streamr-network/security/signing-and-verification.md
hpihkala: I'm ok with `KeyType` - that's what is actually was before I changed it to worse :)

packages/utils/src/signingUtils.ts
hpihkala: If we rename `KeyPairType` to `KeyType` as per your suggestion, and since this is an array consisting of the different `KeyType` values, I guess the name of the array should then be `KEY_TYPES`?

packages/sdk/src/identity/IdentityMapping.ts
```diff
@@ -1,63 +1,68 @@
...
+ * 2. Add the needed SigningUtil to signingUtils.ts and tests to signingUtils.test.ts
+ * 3. Add the new SigningUtil to exports.ts in the utils package
+ * 3. Create the Identity implementation itself (eg. extend KeyPairIdentity) and tests for it
  * 4. Wire everything together below
  */
-export const validKeyTypeValues = ['ECDSA_SECP256K1_EVM', 'ML_DSA_87'] as const
-/** @internal */
-export const IdentityMapping: Record<KeyType, {
+export const IDENTITY_MAPPING: {
+    keyPairType: KeyType
```
teogeb:
```suggestion
    keyType: KeyType
```

packages/utils/src/signingUtils.ts
```diff
@@ -1,95 +1,284 @@
...
+        if (!ml_dsa87.verify(publicKey, payload, signature)) {
+            throw new Error(`The given ML-DSA public key and private key don't match!`)
+        }
+    }
+
+}
+
+// Declared at the bottom of the file because the classes need to be
+// declared first. TS makes sure all KeyPairTypes are present.
+const keyPairTypeToInstance: Record<KeyType, SigningUtil> = {
```
teogeb:
```suggestion
// declared first. TS makes sure all KeyTypes are present.
const keyTypeToInstance: Record<KeyType, SigningUtil> = {
```

packages/cli-tools/bin/streamr-identity-generate.ts
```diff
@@ -1,21 +1,19 @@
...
     .action(async (options: Options) => {
-        const config = IdentityMapping[options.keyType!] // required option
-        const identity = config.generate() as KeyPairIdentity
-        console.info(JSON.stringify({ 
-            publicKey: await identity.getUserId(), 
-            privateKey: binaryToHex(await identity.getPrivateKey()),
+        const signingUtil = SigningUtil.getInstance(options.keyType!) // required option
+        const keyPair = signingUtil.generateKeyPair()
+        console.info(JSON.stringify({
+            publicKey: toUserId(keyPair.publicKey),
```
teogeb: Same functionality, but more descriptive name

```suggestion
            publicKey: binaryToHex(keyPair.publicKey),
```

packages/sdk/src/identity/EthereumKeyPairIdentity.ts
```diff
@@ -1,50 +1,29 @@
...
 import { KeyPairIdentityConfig, StrictStreamrClientConfig } from '../Config'
 import { KeyPairIdentity } from './KeyPairIdentity'
 
+const signingUtil = new EcdsaSecp256k1Evm()
+
 /**
  * An Identity that derives an Ethereum address from a secp256k1 private key
  * and uses that as the UserID instead of the actual public key.
  */
 export class EthereumKeyPairIdentity extends KeyPairIdentity {
```
teogeb, line 15: Could annotate `fromConfig()` to be `@internal` as we don't need to expose that in the public API. Same for `MLDSAKeyPairIdentity` and `EthereumProviderIdentity`.

( The `ECDSAKeyPairIdentity` doesn't need the annotation as it isn't exported from `sdk` package.)

packages/sdk/src/identity/EthereumKeyPairIdentity.ts
```diff
@@ -55,11 +34,18 @@ export class EthereumKeyPairIdentity extends KeyPairIdentity {
     static fromConfig(config: Pick<StrictStreamrClientConfig, 'auth'>): EthereumKeyPairIdentity {
         const privateKey = (config.auth as KeyPairIdentityConfig).privateKey
         const address = (config.auth as KeyPairIdentityConfig).publicKey
-        return new EthereumKeyPairIdentity(privateKey, address)
+        return EthereumKeyPairIdentity.fromPrivateKey(privateKey, address)
+    }
+
+    static fromPrivateKey(privateKey: HexString, address?: HexString): EthereumKeyPairIdentity {
+        return new EthereumKeyPairIdentity(
+            hexToBinary(address ?? new Wallet(privateKey).address.toLowerCase()),
```
teogeb:
```suggestion
            hexToBinary(address ?? new Wallet(privateKey).address), 
```

packages/sdk/src/identity/IdentityMapping.ts
```diff
@@ -1,63 +1,68 @@
...
+    if (!IDENTITY_MAPPING.find((id) => id.keyPairType === keyType)) {
+        throw new Error(`keyType missing from IDENTITIES: ${keyType}`)
     }
 })
 
 /**
  * Creates an Identity instance based on what's in the StreamrClient config
  */
-export const createIdentityFromConfig = (config: Pick<StrictStreamrClientConfig, 'auth' | 'contracts'>): Identity => {
+export function createIdentityFromConfig(config: Pick<StrictStreamrClientConfig, 'auth' | 'contracts'>): Identity {
```
teogeb, line 56: Just stylistic change?

packages/utils/src/signingUtils.ts
```diff
@@ -1,95 +1,284 @@
...
 export interface KeyPair {
     publicKey: Uint8Array
     privateKey: Uint8Array
 }
 
-export interface SigningUtil {
-    createSignature: (payload: Uint8Array, privateKey: Uint8Array) => Uint8Array
-    verifySignature: (publicKey: UserIDRaw, payload: Uint8Array, signature: Uint8Array) => boolean
-    [key: string]: any // Allow additional properties
+export abstract class SigningUtil {
```
teogeb: As this is a class we could rename the file to `SigningUtil.ts`

packages/sdk/src/identity/ECDSAKeyPairIdentity.ts
```diff
@@ -0,0 +1,43 @@
...
+    async createMessageSignature(payload: Uint8Array): Promise<Uint8Array> {
+        // Cache the privateKey in JWK format for a performance optimization
+        this.cachedJWK ??= signingUtil.privateKeyToJWK(this.privateKey)
+        return signingUtil.createSignature(payload, this.cachedJWK)
+    }
+
+    static fromConfig(config: Pick<StrictStreamrClientConfig, 'auth'>): ECDSAKeyPairIdentity {
+        const keyPairConfig = KeyPairIdentity.getKeyPairFromConfig(config)
+        if (!keyPairConfig.publicKey) {
+            throw new Error(`ECDSA_SECP256R1 identity requires a publicKey to be given in the config!`)
```
teogeb: This could throw `StreamrClientError`

teogeb (commented):
👍

packages/sdk/src/identity/EthereumKeyPairIdentity.ts
hpihkala: `ECDSAKeyPairIdentity` is actually missing from exports, so good catch! I'll add it.

Is there any benefit to making `fromConfig` internal? I could imagine some use cases for it (for example, application reads a config file in their application and then pass it to `fromConfig`, kind of like we do as well). 

Also the idea is that the identity classes can provide additional `from*` utility methods, for example `EthereumKeyPairIdentity` provides a `fromPrivateKey` method. Would it be weird if some `from*` methods are marked as internal and some are not? I don't see much of a problem in them being public.

packages/sdk/src/identity/IdentityMapping.ts
hpihkala: Yeah - do we prefer one style over the other?

packages/sdk/src/identity/ECDSAKeyPairIdentity.ts
hpihkala: Actually it's possible to remove this requirement, as elliptic curves allow computing the public key from the private key. I'll do that. `EthereumKeyPairIdentity` works similarly, only the `privateKey` is required.

packages/sdk/src/identity/IdentityMapping.ts
teogeb: No preference, both ok

packages/sdk/src/identity/EthereumKeyPairIdentity.ts
teogeb: The object structure of the config parameter works well for our internal use case, as it matches the format used by our client's configuration. However, it's not ideal for external applications, since it includes "unnecessary" `auth`, `contracts`, and `ethereumNetwork` blocks. An external application would typically only want to pass something simple, such as
```
{
    ethereum: {
        ...
    }
    chainId
}
```

instead of 

```
{
    auth: {
        ethereum: {
            ...
        }
    };
    contracts: {
        ethereumNetwork: {
            chainId
        }
    }
}
```

Because the current format isn’t very suitable for external use, I suspect this method won’t be commonly used in practice. It’s easier to write a custom function that directly calls the `EthereumProviderIdentity` constructor.

To me it would make sense to provide multiple `from*` utility methods and hide the internal ones. This would simplify the API so that end-users only see methods that are actually relevant to them.

hpihkala: Ok agreed, will make the change

packages/sdk/src/identity/KeyPairIdentity.ts
```diff
@@ -3,40 +3,24 @@ import { RpcProviderSource } from '../RpcProviderSource'
 import { Identity, SignerWithProvider } from './Identity'
 import { KeyPairIdentityConfig, StrictStreamrClientConfig } from '../Config'
 
+/**
+ * KeyPairIdentity is an Identity that is defined by a public key and a private key.
+ * It uses the public key as the UserID.
+ */
 export abstract class KeyPairIdentity extends Identity {
```
teogeb, line 10: The `getKeyPairFromConfig()` could be annotated as internal

packages/sdk/src/exports.ts
```diff
@@ -46,15 +46,17 @@ export {
...
 export { GroupKey as EncryptionKey } from './encryption/GroupKey'
 export type { UpdateEncryptionKeyOptions } from './encryption/LocalGroupKeyStore'
 export type { StreamDefinition } from './types'
 export { formStorageNodeAssignmentStreamId, peerDescriptorTranslator } from './utils/utils'
 export { Identity, type SignerWithProvider } from './identity/Identity'
 export { EthereumKeyPairIdentity } from './identity/EthereumKeyPairIdentity'
-export { MLDSAKeyPairIdentity } from './identity/MLDSAKeyPairIdentity'
 export { EthereumProviderIdentity } from './identity/EthereumProviderIdentity'
+export { MLDSAKeyPairIdentity } from './identity/MLDSAKeyPairIdentity'
+export { ECDSAKeyPairIdentity } from './identity/ECDSAKeyPairIdentity'
```
teogeb, line 60: Could also 
`export { KeyPairIdentity } from './identity/KeyPairIdentity'`
so that the class is available in API docs (and usable in applications if someone needs)

teogeb (approved):

PR #3089, opened 2025-04-28
refactor(sdk): Use addresses instead of contract objects in `operatorContractUtils`

teogeb:
Refactored functions in `operatorContractUtils` so that we give contract address as an argument instead of `OperatorContract` object:
- `delegate()`
- `stake()`
- `unstake()`

Also refactored `SetupOperatorContractReturnType` so that it contains address instead of the contract object.

## Other changes

- unified argument names in cli-tool command. Now operator contract address is always named as `operatorContractAddress`
- using `unstake()` from contract utils instead of calling contract directly in some tests

### DIFF
--- a/packages/cli-tools/bin/streamr-internal-operator-stake.ts
+++ b/packages/cli-tools/bin/streamr-internal-operator-stake.ts
@@ -5,14 +5,14 @@ import { StreamrClient, _operatorContractUtils } from '@streamr/sdk'
 import { createClientCommand } from '../src/command'
 import { parseEther } from 'ethers'
 
-createClientCommand(async (client: StreamrClient, operatorContractAddress: string, sponsorshipAddress: string, dataTokenAmount: string) => {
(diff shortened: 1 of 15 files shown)


packages/cli-tools/bin/streamr-internal-operator-stake.ts
```diff
@@ -5,14 +5,14 @@ import { StreamrClient, _operatorContractUtils } from '@streamr/sdk'
...
-        operatorContract,
+        await client.getSigner(),
+        operatorAddress,
         sponsorshipAddress,
         parseEther(dataTokenAmount)
     )
 })
     .description('stake funds to a sponsorship')
-    .arguments('<operatorContractAddress> <sponsorshipAddress> <dataTokenAmount>')
+    .arguments('<operatorAddress> <sponsorshipAddress> <dataTokenAmount>')
```
jtakalai: operatorContractAddress is better, maybe operators themselves think their EOA is the "operator address"

jtakalai (approved):

PR #3092, opened 2025-04-28
test: move `setupOperatorContract()` to `test-utils`

teogeb:
Moved `setupOperatorContract()` utility from `sdk`'s `operatorContractUtils` to `test-utils`. It is a test utility which creates some test node wallets for an operator contract it deploys.

This move is a preparation for a PR in which we refactor `operatorContractUtils` to be environment-independent.

## Other changes

Removed unnecessary `createTestWallet` option in `checkOperatorValueBreach.test.ts`


packages/node/test/integration/plugins/operator/OperatorPlugin.test.ts
```diff
@@ -33,7 +32,7 @@ describe('OperatorPlugin', () => {
     beforeAll(async () => {
         const deployment = (await setupOperatorContract({
             nodeCount: 1,
-            createTestWallet
+            deployOperatorContract: _operatorContractUtils.deployOperatorContract
```
jtakalai, line 35: this indirection isn't really explained anywhere... why is deployOperatorContract an argument? It's passed in the same way everywhere, so it might as well not be passed at all.

jtakalai (approved):

packages/node/test/integration/plugins/operator/OperatorPlugin.test.ts
teogeb: That function is from `sdk` and test-utils can't have `sdk` as a dependency (as it has the dependency other way around: `test-utils` is a dependency of `sdk`)

It was explained in the original PR which added that indirection: https://github.com/streamr-dev/network/pull/2963

PR #3094, opened 2025-04-29
refactor(sdk): `environmentId` parameter for contract deploy utils

teogeb:
Added `environmentId` parameter for `deployOperatorContract()` and `deploySponsorshipContract()` functions. Created curried functions for testing.

## Other changes

- `StrictStreamrClientConfig` always contains value for `environment`
- renamed `setupOperatorContract()` test utility to `setupTestOperatorContract()` to unify the naming

## Future improvements

The internal `operator-create` and `sponsorship-create` cli-tool command now support all environments Therefore it should be possible to promote the commands to non-internal commands.


jtakalai (approved):
there could be a smoke test for creating a client e.g. in the live polygon network, and doing something that only involves reading? This would test that selecting something else than dev2 works

PR #3099, opened 2025-05-02
fix(network): Fixed function binding for event listener in PlumTreeManager

juslesan:
## Summary

Fixed function binding for event listener in PlumTreeManager


ptesavol (approved):
OK

PR #3107, opened 2025-05-07
feat(network): Limit number of paused connections in PlumTree

juslesan:
## Summary

Limit number of paused connections to 3 by default. 
Also if all connections are paused when a neighbor leaves, resume a neighbor to receive data faster.


ptesavol (commented):
looks ok, let's deploy this

ptesavol (approved):

PR #3110, opened 2025-05-08
bugfix: getting subtle in web workers

hpihkala:
`window` is not available inside web workers, and this was used for getting access to SubtleCrypto API in browsers.

Instead, use `self` - which in browser main thread is the same as `window`, but also grants access to `self.crypto.subtle` in web workers.

Also fixes an unrelated flaky test I happened to notice in the CI run.


hpihkala:
Merging without review as trivial

PR #3111, opened 2025-05-08
feat(sdk): add "peaq" environment

jtakalai:
## Summary

Add `peaq` option to `ENVIRONMENT_IDS` list

Config wizard wasn't updated. That tool is currently target primarily for operators, and therefore the current defaults of `polygon` and `polygonAmoy` are maybe optimal. We can add `peaq` as one of options later, if there is need for that.

## Other changes

Simplified `EnvironnmentId` type


packages/sdk/src/Config.ts
```diff
@@ -238,12 +238,12 @@ export interface EthereumNetworkConfig {
...
 // - note that there is no special handling for empty arrays in the applyConfig and therefore
-//   empty arrays will be applied as-is: we may want to remove "enthereum.rpcEndpoints" key 
+//   empty arrays will be applied as-is: we may want to remove "enthereum.rpcEndpoints" key
 //   from @streamr/config as the intention is to use system-defaults (e.g. Metamask defaults)
 //   in Ethereum network
-export type EnvironmentId = 'polygon' | 'polygonAmoy' | 'dev2'
+export type EnvironmentId = 'polygon' | 'polygonAmoy' | 'dev2' | 'peaq'
 
-export const ENVIRONMENT_IDS: EnvironmentId[] = ['polygon', 'polygonAmoy', 'dev2']
+export const ENVIRONMENT_IDS: EnvironmentId[] = ['polygon', 'polygonAmoy', 'dev2', 'peaq']
```
teogeb: Maybe could refactor this to:

```ts
export const ENVIRONMENT_IDS = ['polygon', 'polygonAmoy', 'peaq', 'dev2'] as const
export type EnvironmentId = typeof ENVIRONMENT_IDS[number]
```

packages/node/src/config/ConfigWizard.ts
```diff
@@ -21,7 +21,7 @@ import * as BrokerConfigSchema from './config.schema.json'
 
 const MIN_BALANCE = parseEther('0.1')
 
-type EnvironmentId = 'polygon' | 'polygonAmoy'
+type EnvironmentId = 'polygon' | 'polygonAmoy' | 'peaq'
```
teogeb: Which environments the `ConfigWizard` should support? If we want peaq, it should be added to the `getEnvironmentId()` function at line 249.

Maybe this `EnvironmentId` could use sdk's `EnvironmentId` as base type?

teogeb (changes requested):

teogeb (approved):

PR #3112, opened 2025-05-09
fix(network): Fix PlumTreePropagation test

juslesan:
## Summary

Reduced test flakiness and increased the number of messages in a single test run.

## Changes

- removed newly joining nodes from multiple publishers test
- calculate number of received messages separately in tests where multiple publish streaks happen
- Publish Messages in a tight loop in the `All nodes receive messages case`
- Increased the number of nodes in the test to 64


ptesavol (approved):
ok

PR #3114, opened 2025-05-12
refactor(network): Restructure trackerless-network directories

juslesan:
## Summary

Restructured network directories. No longer use the outdated style from the old network where most of the code is in a logic directory. Instead code is split in directories based on the layer it is used in. If code is multi-layer it is kept in the src directory.

## Future improvements

- Should code that is multilayer ie. NetworkStack, ContentDeliveryManager, StreamPartReconnect have their own directory?

- Rename the `plum-tree` directory to something else that describes the feature.


teogeb (changes requested):
Nice restructuring! 👍  It really makes the codebase a lot easier to follow

Some comments:
- the new directory names should be use the existing lowercase naming style:
  - `ContentDeliveryLayer` -> `content-delivery-layer`
  - `ControlLayer` -> `control-layer`
  - `DiscoveryLayer` -> `discovery-layer`
  - `plumtree` -> `plum-tree`
- as all other top-level directories are layer-specific, I'd remove the `NodeInfo` directory and move the files to the top level
- we could rename `inspect` -> `inspection` so that all directory names are nouns

Also, most of the directories are named by features (e.g. `inspect`, `propagation`), but `plumtree` is named by a protocol. Would it make sense to rename that e.g. to `message-dissemination`?

teogeb (approved):

PR #3135, opened 2025-05-28
feat(cli-tools): Add internal `operator-grant-controller-role` command

teogeb:
Added a new command for granting the operator controller role. This is currently an internal command, like the other operator-related commands. If needed, we can promote this to a public command later (no modifications needed).


jtakalai (approved):
Could add into the test that the controller can not do e.g. updateMetadata before getting the role, and after it can. So the reader can see the effect of granting the role.

PR #3141, opened 2025-06-02
fix(network): reduce the number of nodes in the test to fix flakyness

juslesan:
## Summary

Reduced the number of nodes in the `integration/PlumTreePropagtion.test.ts` to 8 fix flakyness  

## Future improvements

Investigate why the test fails with higher numbers of nodes. This could indicate a bug in the PlumTree protocol or the simulator

### DIFF
--- a/packages/trackerless-network/test/integration/PlumTreePropagation.test.ts
+++ b/packages/trackerless-network/test/integration/PlumTreePropagation.test.ts
@@ -13,7 +13,7 @@ describe('Propagation', () => {
     let contentDeliveryLayerNodes: ContentDeliveryLayerNode[]
     const STREAM_PART_ID = StreamPartIDUtils.parse('testingtesting#0')
     let totalReceived: number
-    const NUM_OF_NODES = 64
+    const NUM_OF_NODES = 8
     let simulator: Simulator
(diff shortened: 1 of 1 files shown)


packages/trackerless-network/test/integration/PlumTreePropagation.test.ts
```diff
@@ -110,7 +110,7 @@ describe('Propagation', () => {
             await until(() => totalReceived >= NUM_OF_NODES * i, 10000)
         }
         totalReceived = 0
-        const numberOfNewNodes = 32
+        const numberOfNewNodes = NUM_OF_NODES
```
teogeb: These are two different configurable values? I.e. should we have something like this:
```ts
const INITIAL_NODE_COUNT = 8
const NEW_NODE_COUNT = 8
```

teogeb (changes requested):

teogeb (approved):
> Investigate why the test fails with higher numbers of nodes. This could indicate a bug in the PlumTree protocol or the simulator.

It would be good to do this investigation soon after we merge this to `main`. Is there a Linear ticket about this?

PR #3143, opened 2025-06-04
feat(cli-tools): Add `--max-operator-count` flag to `sponsorship-create` command

teogeb:
**Note that this PR contains breaking changes to internal commands. But as there are no changes for public API, this is not a breaking change in semver sense.**

Added new flag to the internal `sponsorship-create` command.

Also removed support for the short flag `-c` which was an alias to `--min-operator-count` . Now neither of the operator count flags have a short alias.

## Related changes

SDK's internal `operatorContractUtils#deploySponsorshipContract()` now supports `maxOperatorCount` option. Some refactoring was needed there to support dynamic policy count (as we need to have additional policy if `maxOperatorCount`  is defined).

## Other changes

- explicit wording about the unit of token amount in two CLI tool commands
- small stylistic fix in `operatorContractUtils#deployOperatorContract()`


jtakalai (approved):

harbu (approved):
LGTM

~I'd say this is a breaking change in semver terms since flag `--c` goes away.~

PR #3145, opened 2025-06-04
refactor(sdk): Required `earningsPerSecond` field in `deploySponsorshipContract()`

teogeb:
Changed the `earningsPerSecond` to be a required field when calling `deploySponsorshipContract()`. It is very important configuration field for sponsorship deployment and therefore it makes sense that a value is always provided.

Before this PR only tests relied on the default value (1 data token per second). Maybe it is good to be explicit also in the tests, i.e. not move the default value to from `deploySponsorshipContract()` to `deployTestSponsorshipContract()` helper methods.


jtakalai (approved):

packages/node/test/integration/plugins/operator/OperatorPlugin.test.ts
```diff
@@ -112,7 +113,7 @@ describe('OperatorPlugin', () => {
         const stream = await createTestStream(client, module)
 
         const sponsorer = await createTestWallet({ gas: true, tokens: true })
-        const sponsorship = await deployTestSponsorshipContract({ streamId: stream.id, deployer: sponsorer })
+        const sponsorship = await deployTestSponsorshipContract({ streamId: stream.id, deployer: sponsorer, earningsPerSecond: parseEther('1') })
```
harbu:
```suggestion
        const sponsorship = await deployTestSponsorshipContract({ streamId: stream.id, deployer: sponsorer, earningsPerSecond: EARNINGS_PER_SECOND })
```

harbu (approved):

PR #3147, opened 2025-06-09
feat(network): [NET-1473] PlumTree can now be enabled from the sdk

juslesan:
## Summary

It is now possible to enable the PlumTree optimization from the SDK.

## Changes

- renamed `plumTreeOptimisation` to `plumtreeOptimization`

## Future Improvements

- Perhaps `NetworkNode#broadcast` should not automatically join streams. Instead when a message is published From the SDK for the first time `joinStreamPart` should be called separately.
 - Handle situation where user creates multiple subscriptions to the same stream part with mismatching delivery options. An error could be thrown, for example.
 - Rename files in `trackerless-network/src/content-delivery-layer/plum-tree` to be in line with the `plumtreeOptimization` naming convention #3150


packages/sdk/src/subscribe/Subscription.ts
```diff
@@ -37,13 +39,15 @@ export class Subscription extends MessageStream {
         streamPartId: StreamPartID,
         isRaw: boolean,
         erc1271ContractAddress: EthereumAddress | undefined,
+        deliveryOptions: StreamPartDeliveryOptions,
```
teogeb: As this is an optional configuration:

```suggestion
        deliveryOptions: StreamPartDeliveryOptions | undefined,
```

packages/sdk/src/subscribe/SubscriptionSession.ts
```diff
@@ -12,6 +12,10 @@ import { Subscription } from './Subscription'
  * A session contains one or more subscriptions to a single streamId + streamPartition pair.
  */
 
+const getAnyItemFromSet = <T>(set: Set<T>): T | undefined => {
+    return Array.from(set.values())[0]
```
teogeb:
```suggestion
    return set.values().next().value
```

packages/sdk/src/StreamrClient.ts
```diff
@@ -165,9 +167,10 @@ export class StreamrClient {
     async publish(
         streamDefinition: StreamDefinition,
         content: unknown,
-        metadata?: PublishMetadata
+        metadata?: PublishMetadata,
+        deliveryOptions?: StreamPartDeliveryOptions
```
teogeb, line 171: We should export this `StreamPartDeliveryOptions`  interface in `exports.ts` as it is part of our API.

If I remember correctly, we talked about the naming about a month ago and concluded that it could make sense to use a higher level term instead of `plumTreeOptimisation`? We didn't come up with any specific term at that time but maybe it would be good to revisit the topic?

If we'd chose to keep the low-level term, note that:
- the [protocol](https://www.dpss.inesc-id.pt/~ler/reports/srds07.pdf), defines the term as one word: _plumtree_ (not _plum tree_)
- we use US spelling in our code base so it should be _optimization_

teogeb (changes requested):
First round of comments. A changelog should be added.

packages/sdk/src/StreamrClient.ts
juslesan: Could rename plumtree to ie broadcast tree in a separate PR

harbu: No strong feelings here, my 2 cents:

I'd go with the specific algorithm, i.e. `plumTreeOptimization` or `plumtreeOptimization`. Another way to think about it is is of the useful effect for the user, e.g. `reduceRedundantPropagation` or `reduceDuplicateMessagePropagation` or something like that. Maybe I prefer the "specific algorithm" naming more, however.

packages/sdk/src/subscribe/SubscriptionSession.ts
```diff
@@ -102,7 +106,8 @@ export class SubscriptionSession {
     private async subscribe(): Promise<void> {
         this.node.addMessageListener(this.onMessageInput)
         if (!await this.node.isProxiedStreamPart(this.streamPartId)) {
-            await this.node.join(this.streamPartId)
+            const deliveryOptions = getAnyItemFromSet(this.subscriptions)!.deliveryOptions
+            await this.node.join(this.streamPartId, undefined, deliveryOptions)
```
harbu, line 110: I'd add to future improvements "Handle situation where user creates multiple subscriptions to the same stream part with mismatching delivery options. An error could be thrown, for example."

packages/sdk/src/StreamrClient.ts
harbu: Change from `optimisation` =>  `optimization` definitely worth implementing for consistency.

teogeb: Using `plumtreeOptimization` (with that spelling) is fine. Maybe we'll have another optimizations in the future, and in that case this is specific enough.

juslesan: Renamed to `plumtreeOptimization` also opened a new PR #3150 to rename appropriate files in the trackerless-network to match the convention

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ Changes before Tatum release are not documented in this file.
 - Add new storage node address `STREAMR_STORAGE_NODE_ADDRESS` (https://github.com/streamr-dev/network/pull/3020)
 - Add `peaq` environment (https://github.com/streamr-dev/network/pull/3111)
 - Add `iotex` environment (https://github.com/streamr-dev/network/pull/3142)
+- Add support for using the plumtree optimisation in stream partitions in the SDK (https://github.com/streamr-dev/network/pull/3147)
```
harbu: Something weird going on here or my github diff view is broken. This should be the only item in "unreleased" as v103 was pushed yesterday. Maybe merge gone wrong?

harbu (approved):

teogeb (changes requested):
The changelog item should be under "Unreleased" section, otherwise ok 👍

CHANGELOG.md
juslesan: @teogeb the changes to the changelist are under "unreleased" here

teogeb (approved):

PR #3150, opened 2025-06-11
refactor(network): rename PlumTree to Plumtree

juslesan:
## Summary

Renamed PlumTree to Plumtree to match naming of the configuration in #3147


harbu (approved):

packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -32,7 +32,7 @@ service NodeInfoRpc {
   rpc getInfo (NodeInfoRequest) returns (NodeInfoResponse);
 }
 
-service PlumTreeRpc {
+service PlumtreeRpc {
```
teogeb, line 35: Is this rename a breaking change in protocol level?

juslesan: Could be but as this feature is unreleased and StreamrTV is in alpha release at the moment it should not matter as much.

PR #3156, opened 2025-06-16
refactor(trackerless-network): unify naming and refactor selectParallelTargets

juslesan:
## Summary

No longer call nodes that are attempted as neighbors neighbors, instead call them targets. Added getExcludedIds helper function to remove duplicate code. Added clearer comments to explain the control flow. Fixed a bug where nodes were pushed to `excludedIds` too soon.


ptesavol (approved):
Looks OK

PR #3177, opened 2025-08-05
feat(network): Setting for max paused neighbors in Plumtree

juslesan:
## Summary

Added a setting that allows the user to limit the number of paused neighbors there are in Plumtree. This allows creating more stable plumtree topologies by setting the max to ie.2 This means that nodes should have 2 propagating neighbors. By doing this, there should be less of a change for nodes to drop packets as there is always a redundant connection.

## Other changes

- reduced flakyness of PlumtreePropagation test by setting the maxPausedNeighbors parameter to 2 (test passed 100 times in a row)


ptesavol (approved):
Looks OK

PR #3179, opened 2025-08-07
feat(network): No longer send pause messages if the maxNeighborPaused option is met

juslesan:
## Summary

No longer send pause messages if the maxPausedNeighbors condition is met. This limits the number of paused messages being sent and eliminates flickering of paused neighbors.

Fixes the flaky PlumTreePropagation test. Passed 350 times in row.

## Other changes

Refactored the PlumTreePropagation test


ptesavol (approved):
OK

PR #3191, opened 2025-08-27
feat(sdk, node) [NET-1486]: trigger autostaker on new sponsorhip created event

harbu:
## Summary

Autostaker now gets triggered whenever a new sponsorship is created (in addition to the interval based execution).

## Changes
- Add new `StreamrClient` event `sponsorshipCreated`
- New configuration option `sponsorshipFactoryChainAddress` must be supplied when configuring EVM manually (i.e. **not** using preset `env`).
- Add a mutex / lock that prevents more than one autostaking task of being executed concurrently.

- There could be other events of interest (e.g. sponsorship funded, or tokens delegated to operator) that would be worth triggering the autostaker analysis for. It would be have to looked into whether these events can be hooked up to somehow (do they trigger smart contract events? where are the contract events triggered?)
- Handling of expiring sponsorships. If we know a sponsorship ends in 15 minutes, it would be more efficient to trigger the autostaker analysis right after, instead of waiting a full hour (default interval). A bit of an edge case however.


harbu:
@teogeb PR not yet ready, but maybe you can have a quick look and comment on the approach?

packages/node/src/plugins/autostaker/AutostakerPlugin.ts
```diff
@@ -99,15 +99,26 @@ export class AutostakerPlugin extends Plugin<AutostakerPluginConfig> {
         await fleetState.start()
         await fleetState.waitUntilReady()
         const isLeader = await createIsLeaderFn(streamrClient, fleetState, logger)
-        scheduleAtApproximateInterval(async () => {
+        let alreadyRunning = false
```
teogeb:
```suggestion
        let running = false
```

packages/node/src/plugins/autostaker/AutostakerPlugin.ts
```diff
@@ -99,15 +99,26 @@ export class AutostakerPlugin extends Plugin<AutostakerPluginConfig> {
...
             } catch (err) {
                 logger.warn('Error while running autostaker actions', { err })
+            } finally {
+                // eslint-disable-next-line require-atomic-updates
+                alreadyRunning = false
             }
-        }, this.pluginConfig.runIntervalInMs, 0.1, false, this.abortController.signal)
+        }
+        scheduleAtApproximateInterval(triggerRun, this.pluginConfig.runIntervalInMs, 0.1, false, this.abortController.signal)
+        streamrClient.on('sponsorshipCreated', triggerRun)
```
teogeb, line 121: Could clean up the event listener in `AutostakerPlugin#stop()`

packages/sdk/test/end-to-end/sponsorship-created-event.test.ts
```diff
@@ -0,0 +1,51 @@
...
+            earningsPerSecond: parseEther('1')
+        })
+        const expectedSponsorshipAddress = toEthereumAddress(await sponsorship.getAddress())
+
+        const sponsorshipCreatedEvent: SponsorshipCreatedEvent[] = await listenerEvent
+
+        expect(sponsorshipCreatedEvent.length).toEqual(1)
+        expect(sponsorshipCreatedEvent[0].sponsorshipContractAddress).toEqual(expectedSponsorshipAddress)
+        expect(sponsorshipCreatedEvent[0].streamId).toBe(testStream.id)
+        expect(sponsorshipCreatedEvent[0].metadata).toBeDefined()
```
teogeb: Could deploy the sponsorship contract with metadata, so that we can test it here

harbu:
Fixed issues @teogeb , please have another look

packages/sdk/src/contracts/SponsorshipFactory.ts
```diff
@@ -0,0 +1,55 @@
...
+import { ConfigInjectionToken, StrictStreamrClientConfig } from '../Config'
+import { StreamrClientEventEmitter } from '../events'
+import { LoggerFactory } from '../utils/LoggerFactory'
+import { ChainEventPoller } from './ChainEventPoller'
+import { initContractEventGateway } from './contract'
+
+export interface SponsorshipCreatedEvent {
+    readonly sponsorshipContractAddress: EthereumAddress
+    readonly streamId: StreamID
+    readonly metadata: string
```
teogeb: This could be similar to `StreamMetadata`:

```suggestion
    readonly metadata: Record<string, unknown>
```

harbu: I'd put it future improvements now, since we don't rely on metadata atm

teogeb (approved):

PR #3213, opened 2025-10-03
fix: the Autocertifier-server to use the StreamrWebsocketPort when storing domains

juslesan:
## Summary

The autocertifier-server now uses the StreamrWebsocketPort as port that is stored to the database when creating or updating subdomains.


ptesavol (approved):
OK

PR #3215, opened 2025-10-06
feat(Autocertifier): [NET-1504] Autocertifier-server self cleanup

juslesan:
## Summary

When registering a new domain name from an IP, all previous domain names registered by the IP and port combinationare removed.


ptesavol (commented):
LGTM, the challenger will fail if somebody else answers to the websocket challenge, so it is safe to do this

PR #3218, opened 2025-10-10
fix(Autocertifier-server): Fix error when cleaning up subdomains

juslesan:
## Summary

Added `Promise.allSettled` instead of `Promise.all` to get rid of failing `createNewCertificateForSubdomain` requests


ptesavol (approved):
OK

PR #3224, opened 2025-10-24
refactor: Move process exit handlers from the trackerless-network to the node package

juslesan:
## Summary

Moved exit handlers from the `trackerless-network` to the `node` package. This ensures that the operator nodes will still gracefully leave the network if the process is exited. Also removed the on unload event listener in NetworkStack. Users of the SDK are being made aware of the change by updating the SDK's README to include a section about properly cleaning up the Network node when the StreamrClient is no longer needed.

## Other Changes 

- The `exit` case is no longer handled as we cannot do asynchronous operations there.

---

> [!NOTE]
> Move shutdown handling to the node CLI, remove global exit/unload handlers from trackerless-network, and document proper StreamrClient cleanup.
> 
> - **Node CLI (`packages/node/bin/streamr-node.ts`)**
>   - Add graceful shutdown handlers for `SIGINT`, `SIGTERM`, `SIGUSR1`, `SIGUSR2`.
>   - On `uncaughtException`/`unhandledRejection`, log fatal and stop broker before exit.
> - **Trackerless Network (`packages/trackerless-network/src/NetworkStack.ts`)**
>   - Remove global process/window exit handlers and instance tracking; simplify `stop()`.
> - **Docs**
>   - Add cleanup guidance to `docs/docs/usage/sdk/how-to-use.md` and `packages/sdk/README.md` to call `StreamrClient.destroy()` when done.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f2d88f0f5cb925fd43b48aa80f7440278bbbdd8e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/node/bin/streamr-node.ts
+++ b/packages/node/bin/streamr-node.ts
@@ -26,7 +26,7 @@ program
                 process.exit(exitCode)
             }
             
-            const exitEvents = ['SIGINT', 'SIGTERM', 'SIGUSR1', 'SIGUSR2', 'exit']
+            const exitEvents = ['SIGINT', 'SIGTERM', 'SIGUSR1', 'SIGUSR2']
             exitEvents.forEach((event) => {
                 process.on(event, () => shutdown(0))
             })
--- a/packages/node/bin/streamr-node.ts
+++ b/packages/node/bin/streamr-node.ts
@@ -1,10 +1,12 @@
 #!/usr/bin/env node
 import { program } from 'commander'
 import pkg from '../package.json'
-
+import { Logger } from '@streamr/utils'
 import { createBroker } from '../src/broker'
 import { readConfigAndMigrateIfNeeded } from '../src/config/migration'
 import { overrideConfigToEnvVarsIfGiven } from '../src/config/config'
+
+const logger = new Logger(module)
 
 program
     .version(pkg.version)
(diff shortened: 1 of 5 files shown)


packages/node/bin/streamr-node.ts
```diff
@@ -17,6 +19,28 @@ program
...
             overrideConfigToEnvVarsIfGiven(config)
             const broker = await createBroker(config)
+            
+            // Set up graceful shutdown handlers
+            const shutdown = async (exitCode: number) => {
+                await broker.stop()
+                process.exit(exitCode)
+            }
+            
+            const exitEvents = ['SIGINT', 'SIGTERM', 'SIGUSR1', 'SIGUSR2', 'exit']
```
teogeb: Is the exit event relevant? We can't do asynchronous operations there: https://nodejs.org/api/process.html#event-exit

teogeb (changes requested):
> Could be worth the effort to update the sdk's README to make sure that users are aware that they need to call StreamrClient#destroy always when shutting down the process.

Yes, let's add that

> For browser nodes the graceful leaving on unload event is still kept in the trackerless-network. Alternatively the on unload event handler could be removed as well. If this is done the users must be made aware that they need to call NetworkNode#stop manually when a browser window is closed.

The `StreamrClient#destroy()` does that, right? So when we have the documentation, we can remove this functionality, too.

juslesan:
So if I understood correctly do you @teogeb want to remove the clean up from the on unload events as well?

teogeb:
> So if I understood correctly do you @teogeb want to remove the clean up from the on unload events as well?

Yes

packages/sdk/README.md
```diff
@@ -118,6 +118,13 @@ streamr.resend(streamId, { last: 10 }, (msgs) => {
 });
 ```
 
+### Clean up
+After the StreamrClient is no longer used or the process is shutting down it is very important to call `destroy` on the StreamrClient. This ensures that the network node of the client will be shutdown gracefully.
```
teogeb:
```suggestion
After the `StreamrClient` is no longer used or the process is shutting down it is very important to call `destroy()` on the `StreamrClient`. This ensures that the network node of the client will be shutdown gracefully. 
```

teogeb: Worth mentioning also here? https://docs.streamr.network/usage/sdk/how-to-use

packages/node/bin/streamr-node.ts
```diff
@@ -17,6 +19,28 @@ program
...
             overrideConfigToEnvVarsIfGiven(config)
             const broker = await createBroker(config)
+            
+            // Set up graceful shutdown handlers
+            const shutdown = async (exitCode: number) => {
+                await broker.stop()
+                process.exit(exitCode)
+            }
+            
+            const exitEvents = ['SIGINT', 'SIGTERM', 'SIGUSR1', 'SIGUSR2']
```
teogeb, line 29: Could clarify in the PR description why `exit` event type is no longer handled?

packages/sdk/README.md
juslesan: Yes that would make sense. How do I update this page?

teogeb: It is this file in this repo: `docs/docs/usage/sdk/how-to-use.md`

PR #3227, opened 2025-11-06
Hotfixes for autostaker problems

hpihkala:
A set of small improvements to address problems currently affecting operators in the production network. All the changes are pretty "local": ideally, they could be generalized to apply to all transactions (ie. always retry transactions, always bump gas limit, etc), but implementing that in some clean and configurable way would take *much* more time. The scope of this PR is to fix the production issues as quickly as possible, therefore prioritizing the simplest solutions.

Changes:

- Bump gas limit on stake/unstake contract calls by 20% to defend against too low estimates in case contract change rapidly changes
- Retry actions in autostaker plugin
- Bump `@streamr/config` to remove bad RPC entry
- Guard against The Graph not being up to date when a new sponsorship event is seen on chain
- Broadcast autostaker stake/unstake transactions in parallel to speed up reaction time


packages/node/src/plugins/autostaker/AutostakerPlugin.ts
```diff
@@ -119,15 +120,49 @@ export class AutostakerPlugin extends Plugin<AutostakerPluginConfig> {
...
+                this.pluginConfig.operatorContractAddress,
+                action.sponsorshipId,
+                action.amount,
+                (tx) => resolve(tx) // resolve onSubmit
+            ).catch(reject)
+        })
+    }
+
+    // This will retry the transaction preflight checks, defends against various transient errors
+    private async submitActionWithRetry(action: Action, signer: SignerWithProvider,
```
teogeb: `maxRetries` and `retryDelay` are never used by caller, so we could just use values of 5 and 5000 at lines 157-158. Preferably via constants (e.g. `ACTION_SUBMIT_RETRY_COUNT` and   `ACTION_SUBMIT_RETRY_DELAY`)

packages/sdk/src/contracts/operatorContractUtils.ts
```diff
@@ -144,25 +144,47 @@ export const stake = async (
...
-    await (await contract.stake(sponsorshipContractAddress, amount)).wait()
+    amount: WeiAmount,
+    onSubmit: (tx: ContractTransactionResponse) => void = () => {},
+): Promise<ContractTransactionReceipt | null> => {
+    logger.debug('Stake', { amount: formatEther(amount), sponsorshipContractAddress })
+    const operatorContract = getOperatorContract(operatorContractAddress).connect(staker)
+    
+    // Gas limit needed for staking is a little unstable because others might 
+    // be staking at the same time so we bump the gas limit to be safe
+    const gasLimit = bumpGasLimit(await operatorContract.stake.estimateGas(sponsorshipContractAddress, amount))
```
teogeb: We have similar approach in `getEthersOverrides()`, but it `uses getProvider().getFeeData()` instead `estimateGas()` Would it make sense to use SDK's utility, or would that approach implement a slightly different functionality? 
https://github.com/streamr-dev/network/blob/1ad5ff123f4781bca450878e16674ad7c82b7d57/packages/sdk/src/ethereumUtils.ts#L13-L14

As bumping of gas limit is done for the AutoStaker user case, we could control that via a parameter. I.e. `highGasPrice: boolean`, which defaults to `false`, but AutoStaker calls it with `true`. That way the behavior doesn't change in other use cases, like in the CLI tools commands.

teogeb (commented):
Great improvements 👍

packages/sdk/src/contracts/operatorContractUtils.ts
hpihkala: Those are two different things (gas amount vs. gas price). That being said, having the possibility to optionally bump gas price via config would be useful and in line with our other contract-calling functionality, but unlike other code, the `operatorContractUtils` unfortunately isn't wired up to use the `ethersOverrides` system, so that would require medium-size refactoring which is probably beyond the scope of this PR.

packages/node/src/plugins/autostaker/AutostakerPlugin.ts
hpihkala: Sure, done

CHANGELOG.md
```diff
@@ -28,6 +30,12 @@ Changes before Tatum release are not documented in this file.
 
 #### Changed
 
+- Update default list of JSON RPC urls for Polygon (replace a dead one with a working one)
```
teogeb: Usually we haven't duplicated the info and we've just have the SDK.

e.g. like here
```
## [100.2.2] - 2024-04-03

### @streamr/sdk

#### Changed

- Update internal list of JSON RPC urls for Polygon
```

teogeb (approved):

PR #3242, opened 2025-11-21
deps: Bump `webpack` to v5.103.0

teogeb:
Upgraded webpack version and modified karma config to be compatible with the upgrade.

## Error

After the upgrade browser test runs, e.g. `npm run test-browser-unit `in the `sdk` package, failed like this:

```
Electron 34.2.0 (Node 20.18.2) ERROR
  Uncaught TypeError: __webpack_modules__[moduleId] is not a function
  at /tmp/_karma_webpack_665134/commons.js:3032:41
  
  TypeError: __webpack_modules__[moduleId] is not a function
      at __nested_webpack_require_24761__ (/tmp/_karma_webpack_665134/commons.js:3032:41)
      at ./src/utils.ts (/tmp/_karma_webpack_665134/commons.js:2661:16)
      at __nested_webpack_require_24761__ (/tmp/_karma_webpack_665134/commons.js:3032:41)
      at ./src/index.ts (/tmp/_karma_webpack_665134/commons.js:2430:14)
      at __nested_webpack_require_24761__ (/tmp/_karma_webpack_665134/commons.js:3032:41)
      at /tmp/_karma_webpack_665134/commons.js:3043:44
      at ../../node_modules/@jest/expect-utils/build/index.js (/tmp/_karma_webpack_665134/commons.js:3046:12)
      at __webpack_require__ (/tmp/_karma_webpack_665134/runtime.js:23:42)
      at /tmp/_karma_webpack_665134/commons.js:111126:20
      at /tmp/_karma_webpack_665134/commons.js:111366:3
``` 

## Changes to browser runner karma configuration

Webpack’s [recent change](https://github.com/webpack/webpack/releases/tag/v5.103.0) of `"Rename single nested __webpack_export__ and __webpack_require__ in already bundled code"` is likely causing this issue. Because our Karma test setup imports several Jest packages that are themselves pre-bundled, webpack’s re-bundling breaks their internal runtime. We fix this by declaring those Jest packages as externals in the Karma webpack config so they’re not re-bundled.

In this PR three dependencies have been defined as externals:
- @jest/expect-utils
- pretty-format
- jest-diff

Also added `nodeIntegration: true` definition to the webpack config so that Electron can use these external packages. Without this addition the run fails like this: `Uncaught ReferenceError: require is not defined`.

We may need externalize also `jest-matcher-utils` in the future as it is a similar dependency to these three packages. But as there is no error about that package, it hasn't been externalized in this PR

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20077,13 +20077,17 @@
             }
         },
         "node_modules/loader-runner": {
-            "version": "4.3.0",
-            "resolved": "https://registry.npmjs.org/loader-runner/-/loader-runner-4.3.0.tgz",
-            "integrity": "sha512-3R/1M+yS3j5ou80Me59j7F9IMs4PXs3VqRrm0TU3AbKPxlmpoY1TNscJV/oGJXo8qCatFGTfDbY6W6ipGOYXfg==",
+            "version": "4.3.1",
+            "resolved": "https://registry.npmjs.org/loader-runner/-/loader-runner-4.3.1.tgz",
+            "integrity": "sha512-IWqP2SCPhyVFTBtRcgMHdzlf9ul25NwaFx4wCEH/KjAXuuHY4yNjvPXsBokp8jCB936PyWRaPKUNh8NvylLp2Q==",
             "dev": true,
             "license": "MIT",
             "engines": {
                 "node": ">=6.11.5"
+            },
+            "funding": {
+                "type": "opencollective",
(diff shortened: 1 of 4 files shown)


mondoreale (approved):
LGTM.

mondoreale:
The diff I'm getting between SDK's bundle before and after is very minimal, which is good.

PR #3260, opened 2025-12-05
feat [NET-1639]: Bidirectional proxy connections

juslesan:
## Summary

Added support for opening bidirectional proxy connections to the network. These proxy connections can be used to publish and subscribe to messages in the network. The bidirectional connections are now the default behavior. This is an important feature for proxied nodes in the StreamrTV. Previously it was possible to only set unidirectional publish or subscribe proxy connections.  Support for opening publish or subscribe only proxy connections is kept in the network.


packages/trackerless-network/protos/NetworkRpc.proto
```diff
@@ -172,6 +172,7 @@ message CloseTemporaryConnection {
 enum ProxyDirection {
   PUBLISH = 0;
   SUBSCRIBE = 1;
+  BIDIRECTIONAL = 2;
```
teogeb: To me it sounds that direction is now an optional attribute of a proxy connection. I.e. it can ba restricted to _publish-only_, _subscribe-only_, or _none_. Therefore we could just make the field optional?

`NetworkRpc.proto`:
```protobuf
message ProxyConnectionRequest {
  optional ProxyDirection direction = 1;
  bytes userId = 2;
}
```

Methods, e.g. `StreamrClient#setProxies()`:
```ts
async setProxies(
    streamDefinition: StreamDefinition,
    nodes: NetworkPeerDescriptor[],
    direction?: ProxyDirection,
    connectionCount?: number
)
```

teogeb (changes requested):

packages/trackerless-network/protos/NetworkRpc.proto
hpihkala: Just commenting that regardless of the exact details of the interfaces, bidirectional should definitely be the default.

hpihkala (approved):
So this was basically a one-liner change (filter condition in `getSubscribers()`)? Nice! We should get this to Operators pretty quickly so that we can properly start testing using proxy connections in StreamrTV. Please wait for Teo's approval before merging.

teogeb (approved):

PR #3283, opened 2025-12-10
refactor: Unbind `Logger` from NodeJS `module` object (by type)

mondoreale:
This pull request updates the `Logger` utility to make its scoping more flexible and less dependent on NodeJS `module` objects.[^1] The changes allow the logger to be initialized with either a string or an object containing an `id`, making it easier to use in different contexts beyond just files or modules.

It is backward compatible.

### Changes

**Logger scope improvements:**

* Changed the `Logger` constructor to accept a `scope` parameter, which can be either a string or an object with an `id` property, instead of a `NodeJS.Module`. (`packages/utils/src/Logger.ts`)
* Updated the `Logger.createName` static method to handle the new `scope` type, extracting the identifier appropriately whether it's a string or an object. (`packages/utils/src/Logger.ts`)
* Introduced the `Scope` type alias for clarity and type safety. (`packages/utils/src/Logger.ts`)

### Future steps

* `module` is a Node-specific object and is not available in ES and browser making the logger hard to use in other environments without a polyfill. Next steps would be to stop initializing `Logger` with `module`, esp. in files targetting multiple environments – subject for a separate PR.

[^1]: Extracted (with changes) from #3279.

### DIFF
--- a/packages/utils/src/Logger.ts
+++ b/packages/utils/src/Logger.ts
@@ -72,6 +72,8 @@ function wrappedMethodCall(
     }
 }
 
+type Scope = string | { id: string }
+
 export class Logger {
     static NAME_LENGTH = 25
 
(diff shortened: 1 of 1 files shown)


packages/utils/src/Logger.ts
```diff
@@ -72,6 +72,8 @@ function wrappedMethodCall(
     }
 }
 
+type Scope = string | { id: string }
```
teogeb: Maybe we could keep using the term _module_? 
```suggestion
type LoggerModule = string | { id: string }
```

teogeb (changes requested):
Good generalization. Maybe we don't need to introduce a separate term? Typically we create separate loggers for each module (e.g. class or component) and therefore that term is quite descriptive.

teogeb (approved):

PR #3284, opened 2025-12-10
feat(sdk): Explicit encryption keys

juslesan:
## Summary

Make it possible to provide explicit encryption keys to the StreamrClient. These keys will be used in a given stream for all publishers as the encryption key.  Setting the explicit keys disables the group key exchange. Therefore, all participants in the stream need to agree on what key is being used for encryption and decryption to work properly.
 
## Changes

- Added new configuration field `keys` to the `encryption` section
- The `GroupKeyManager#fetchKey` returns explicitly defined keys if they are defined
- `Publisher` initializes the `GroupKeyQueue` with the explicit keys if they are defined


packages/sdk/src/encryption/GroupKeyManager.ts
```diff
@@ -9,12 +9,34 @@ import { uuid } from '../utils/uuid'
...
+): Promise<GroupKey | undefined> => {
+    if (config.keys !== undefined) {
+        for (const entry of Object.entries(config.keys)) {
+            if (await streamIdBuilder.toStreamID(entry[0]) === streamId) {
+                return new GroupKey(entry[1].id, Buffer.from(hexToBinary(entry[1].data)))
+            }
+        }
+    }
+    return undefined
+}
```
teogeb, line 33:
```suggestion
}

```

packages/sdk/src/config.schema.json
```diff
@@ -375,6 +375,21 @@
...
+                            "id": {
+                                "type": "string"
+                            },
+                            "data": {
+                                "type": "string"
+                            }
+                        },
+                        "required": ["id", "data"]
+                    }
                 }
```
teogeb, line 387: We could use `hex-string` as data type. 

Also maybe could improve readability by defining what kind of property keys are there in this map:

```
"keys": {
    "type": "object",
    "propertyNames": {
        "$ref": "#/definitions/streamIdOrPath"
    },
    "additionalProperties": {
        "$ref": "#/definitions/encryptionKey"
    }
}
```

and these under the `definitions` block:

```
"streamIdOrPath": {
    "type": "string"
},
"encryptionKey": {
    "type": "object",
    "properties": {
        "id": {
            "type": "string"
        },
        "data": {
            "type": "string",
            "format": "hex-string"
        }
    },
    "required": [
        "id",
        "data"
    ]
}
```

(Currently StreamID can be any string as we support also legacy format. In the future this streamIdOrPath can some strict a format pattern.)

packages/sdk/src/Config.ts
```diff
@@ -390,6 +390,11 @@ export interface StreamrClientConfig {
          * Note that subscribers will still accept unencrypted (public) data despite this setting.
          */
         requireQuantumResistantEncryption?: boolean
+
+        /**
+         * If this is defined only these encryption keys will be used. This will disable the Streamr key-exchange
```
teogeb:
```suggestion
         * If this is defined only these encryption keys will be used. This will disable the Streamr key-exchange.
```

packages/sdk/test/test-utils/utils.ts
```diff
@@ -219,9 +220,11 @@ export const createGroupKeyManager = async (
     groupKeyStore: LocalGroupKeyStore = mock<LocalGroupKeyStore>(),
     identity?: Identity 
 ): Promise<GroupKeyManager> => {
+    const usedIdentity = identity ?? await createRandomIdentity()
```
teogeb: Let's make `identity` as a required parameter

packages/sdk/test/unit/messagePipeline.test.ts
```diff
@@ -86,6 +87,7 @@ describe('messagePipeline', () => {
             groupKeyManager: new GroupKeyManager(
                 mock<SubscriberKeyExchange>(),
                 groupKeyStore,
+                new StreamIDBuilder(EthereumKeyPairIdentity.fromPrivateKey(publisher.privateKey)),
```
teogeb: Let's use local variable `const identity = ...` so that we can use it at lines `90` and `92`

packages/sdk/test/unit/GroupKeyManager.test.ts
```diff
@@ -29,6 +30,7 @@ describe('GroupKeyManager', () => {
         return new GroupKeyManager(
             subscriberKeyExchange,
             groupKeyStore,
+            new StreamIDBuilder(EthereumKeyPairIdentity.fromPrivateKey(wallet.privateKey)),
```
teogeb: Let's use local variable `const identity = ...` so that we can use it at lines 33 and 40

teogeb (changes requested):
Some small suggestions. Also please add a changelog item and use conventional commit title `feat(sdk): ...`

teogeb (approved):

PR #3286, opened 2025-12-11
refactor: Remove "browser" tsconfig jsons

mondoreale:
This pull request removes several `tsconfig.browser.json` files from multiple packages and updates the corresponding `tsconfig.karma.json` files to extend from the shared root configuration instead. This streamlines TypeScript configuration across packages, reducing redundancy and potential inconsistencies.

**TypeScript configuration cleanup:**

* Deleted `tsconfig.browser.json` files from the following packages to consolidate configuration into the shared root config: `autocertifier-server`, `cdn-location`, `geoip-location`, `proto-rpc`, `sdk`, `trackerless-network`, and `utils`.

**Test configuration updates:**

* Updated `tsconfig.karma.json` files in `dht`, `proto-rpc`, `trackerless-network`, and `utils` to extend from the shared root config (`../../tsconfig.browser.json`) instead of the deleted package-local config, and added necessary compiler options and includes directly.
* Ensured that all updated `tsconfig.karma.json` files explicitly specify `outDir`, `noImplicitOverride`, and the appropriate `include` paths.


packages/sdk/tsconfig.browser.json
```diff
@@ -1,24 +0,0 @@
-{
```
teogeb, line 1: Currently there is no `tsconfig.karma.json` in `sdk`. That file is there for all other packages which run browser tests (`dht`, `trackerless-network,` `proto-rpc`, `utils`). Is it ok that the file is missing in `sdk`? If the file is not needed there, why it needed in these other packages?

(Maybe `sdk` is a special case, but asking for clarification as I don't remember the reason at the moment).

Not strictly related to this PR.

teogeb (commented):
The root level `tsconfig.browser.json` could be renamed to `tsconfig.karma.json` as it is only referenced from tsconfig.karma.json files?

packages/sdk/tsconfig.browser.json
mondoreale: The SDK does not use `createWebpackConfig` from `browser-test-runner` – that's what points to the tsconfig. The SDK has its very own custom webpack config setup (also for karma).

PR #3360, opened 2026-01-20
refactor(sdk): Resolve `Resends` circular dependency

mondoreale:
## Summary

Resolve circular dependency around `Resends` and clean up dependency injection in the subscription module.

## Changes

- Move `Resends` injection from `MessagePipelineFactory` to `Subscriber`, **breaking the circular dependency**.
- Remove unnecessary `delay()` wrappers from `Subscriber` and `Resends` constructors.
- Refactor `SubscriptionSession` constructor to use options object pattern for better extensibility.


mondoreale:
Ok., found a better way!

`tsyringe` is actually pretty cool.

Onwards!

